PR #80, opened 2021-07-13
adding environment for sms and totp mfa and updating test urls

eddiekeller:
*Issue #, if available:*

n/a

*Description of changes:*

Adding a new backend environment for the SMS + TOTP MFA, and the sample app using the old Authenticator to go with it. 

Also updating the URLs in the test files to account for the new folder structure in the examples package.


examples/next/pages/ui/components/authenticator/sign-in-totp-sms-mfa/index.tsx
```diff
@@ -0,0 +1,15 @@
+import { Authenticator } from "aws-amplify-react";
```
ericclemmons, line 1: Should we add `@aws-amplify/ui-react` (and remove this dependency) to check for parity?

PR #150, opened 2021-08-02
Fix type issue in jest env setup

zchenwei:
*Issue #, if available:*

The testing library setup is missing type declaration, which causes custom matcher no found error during Jest runtime if we want to use them. 

```
Property 'toBeDisabled' does not exist on type 'Matchers<any>'
```

*Description of changes:*

1. Adding `@types/testing-library__jest-dom` as dev dependency and configure it to be not hoisted(`ts-jest` failed to load it if it gets hoisted for some reasons). Otherwise, we will not be able to use the custom matchers provided [here](https://github.com/testing-library/jest-dom#usage), though we `import '@testing-library/jest-dom';`

2. Create a `jest.setup.ts` file to remove some redundant setup in each test file

### DIFF
--- a/packages/react/src/primitives/Badge/__tests__/Badge.test.tsx
+++ b/packages/react/src/primitives/Badge/__tests__/Badge.test.tsx
@@ -1,6 +1,4 @@
 import { render, screen } from '@testing-library/react';
-import '@testing-library/jest-dom';
-import '@testing-library/jest-dom/extend-expect';
 import { Badge } from '../Badge';
 import { ComponentClassNames } from '../../shared';
 
--- a/packages/react/src/primitives/Card/__test__/Card.test.tsx
+++ b/packages/react/src/primitives/Card/__test__/Card.test.tsx
@@ -1,7 +1,5 @@
 import { Card } from '../Card';
 import { render, screen } from '@testing-library/react';
-import '@testing-library/jest-dom';
-import '@testing-library/jest-dom/extend-expect';
 import { ComponentClassNames } from '../../shared';
 
 describe('Card component', () => {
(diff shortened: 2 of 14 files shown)


packages/react/package.json
```diff
@@ -24,10 +24,17 @@
     "classnames": "^2.3.1",
     "qrcode": "^1.4.4"
   },
+  "workspaces": {
+    "nohoist": [
+      "@types/testing-library__jest-dom",
+      "@types/testing-library__jest-dom/**"
+    ]
+  },
```
ericclemmons, line 32: Normally any `workspaces` declarations go on the root `package.json`.

I noticed these warnings are occurring on `yarn install` now:

> **warning** nohoist config is ignored in "@aws-amplify/ui-react" because it is not a private package. If you think nohoist should be allowed in public packages, please submit an issue for your use case.

So this doesn't seem to be doing anything.  Can you test removing this change?

zchenwei: I tested that. It looks like the type files cannot be loaded correctly in Jest env when they are on root.

ericclemmons: So yarn _is_ using this `nohoist` config despite these warnings saying `config is ignored in` on install?

```
warning nohoist config is ignored in "@aws-amplify/ui-react" because it is not a private package. If you think nohoist should be allowed in public packages, please submit an issue for your use case.
warning nohoist config is ignored in "@aws-amplify/ui-react" because it is not a private package. If you think nohoist should be allowed in public packages, please submit an issue for your use case.
warning nohoist config is ignored in "@aws-amplify/ui-react" because it is not a private package. If you think nohoist should be allowed in public packages, please submit an issue for your use case.
```

zchenwei: Yup, I think so. Honestly, I don't like this `no hoist` config but if we removed that, the reality is the custom jest matchers from `'@testing-library/jest-dom'` cannot get recognized in runtime though we can get type reference.

PR #180, opened 2021-08-12
feat(ui-core): auth machine refactor with actors

wlee221:
*Issue #, if available:*

## Description of Changes

This PR leverages [actor model](https://xstate.js.org/docs/guides/actors.html#spawning-actors) to split each auth flow into its own actor. Particularly, the auth states are organized into four main flows/actors: `signIn`, `signUp`, `signOut`, and `resetPassword`. 

## Structure 

The top-level `authMachine` is a parent machine that:
- Define possible transitions between each flow (e.g. signIn → signUp)
- Spawn actor with appropriate context. Pass the necessary data from actor to actor.
- transition to next auth state based on the result of the current actor.
- Unaware of any service call or form handling internals.

Here's what the refactored parent machine looks like:

image

Child actors are spawned under the main authenticator. Only one is active at a time and they:
- own one specific auth flow (signIn, signUp, signOut, or resetPassword).
- responsible for managing the associated form and making service calls.
- have a clear notion of being done, so that its parent can make transition based on the final result. 
- Deliver predictable data when they're done, so that its parent doesn't know about the actor internals.

<details>
 <summary>Sign In Actor Visualization</summary>

Screen Shot 2021-08-17 at 1 00 38 PM

</details>

#### Simplify top-level machine
This PR simplifies the top-level machine to have 6 states without any substates. Any nested logic is moved to child actors.

#### Spawn only necessary context and logic
We only spawn necessary data for that actor's auth flow. For example, `signUp` doesn't spawn `challengeName` as it's not required. 

#### Be explicit how context are shared
We define a clear contract for passing context, which I'll describe below. This strictly defines which context is for passing data vs just setting data.

#### Group auth states into bigger auth flows
With this refactor, we group smaller services into well defined auth flow, instead of being 1-1 with Auth API. This well defines what each of auth flow really does -- e.g. `signIn` returns a valid user, whether or not it went through `confirmSignIn` or `forceNewPassword`. Doing this encapsulates all of possible branched flows into one flow.

## Passing data between actors
To pass data from actor to actor, the first actor sends "done data" when it reaches its final state. For example: https://github.com/aws-amplify/amplify-ui/blob/9a300eaed9c3876d876c9336fd1ed582a5b40ba6/packages/core/src/machines/actors/auth/signUp.ts#L133-L143

- `authAttributes` is used to pass any auth attribute data to the next actor. Here, it sends username and password if it exists.
- `intent` hints the next auth flow and helps the new actor to be aware of the context and what it should do next.

#### Getting remoteError:
```tsx
const { remoteError } = state.context.actorRef.context
```
#### Getting authState
```tsx
state.context.actorRef?.getSnapshot().matches('confirmSignUp')
```

I wrote some helper functions as these are reused often.

### Requirements

<details>
 <summary>Checkboxes</summary>

**core machine changes**
- [x] Create `signInActor` that contains `signIn`, `confirmSignIn`, `setupTOTP`, `federatedSignIn`, and `forceNewPassword`.
- [x] Initialize `signIn` with necessary context
- [x] Create `signUpActor`. This will contain   `signUp` and `confirmSignUp`.
- [x] Initialize `signUp` that contains necessary context
- [x] If we go from `signIn` to `confirmSignUp` due to unconfirmed user, make sure `signUpActor` starts with `confirmSignUp`.
- [x] Pass necessary context when transitioning between actors
- [x] reflect #184 changes

**Reflect state machine changes to each frameworks**
- [x] Angular
- [x] React
- [x] Vue

### TODO goals after this PR:
- [ ] Standardize form interface. Right now only signUp has input validation, which can be used in other states too. This probably deserves its own PR though.
- [ ] Reuse any actions, services, or substate (e.g. federatedSignIn!)
</details>


wlee221:
Made all the framework refactors! Waiting for the test results to see if I missed anything...

wlee221:
Last thing I have to do is to get #184 changes to #180, and this should be good to go!

packages/core/src/machines/actors/auth/signIn.ts
```diff
@@ -0,0 +1,280 @@
...
+        states: {
+          edit: {
+            entry: sendUpdate(),
+            on: {
+              SUBMIT: 'submit',
+              CHANGE: { actions: 'handleInput' },
+              FEDERATED_SIGN_IN: 'federatedSignIn',
+            },
+          },
+          federatedSignIn: {
```
wlee221, line 37: Now, federatedSignIn is under signIn flow as its goal matches that of signIn actor. This means that there will be some dry code in signUp.federatedSignUp, but we can later reuse code with shared state definition + spread operator.

packages/core/src/machines/actors/auth/signIn.ts
```diff
@@ -0,0 +1,280 @@
...
+        data: (context) => ({
+          user: context.user,
+        }),
+      },
+      rejected: {
+        type: 'final',
+        data: (context) => ({
+          intent: 'confirmSignUp',
+          authAttributes: {
+            username: context.formValues.username,
```
wlee221, line 180: If we get rejected due to unconfirmed user, we send back data with intent that we should `confirmSignUp`.

packages/core/src/machines/actors/auth/signOut.ts
```diff
@@ -0,0 +1,27 @@
...
+    states: {
+      pending: {
+        invoke: {
+          src: 'signOut',
+          onDone: 'resolved',
+          onError: 'rejected',
+        },
+      },
+      resolved: { type: 'final' },
+      rejected: { type: 'final' },
```
wlee221, line 19: Not sure why `Auth.signUp` would fail... but in that case it'd transition to `#authMachine.idle` to check current user anyway and transition to appropriate state.

packages/core/src/machines/actors/auth/signUp.ts
```diff
@@ -0,0 +1,229 @@
...
+            initial: 'idle',
+            states: {
+              idle: {
+                entry: sendUpdate(),
+                on: {
+                  SUBMIT: 'validate',
+                  FEDERATED_SIGN_IN: 'federatedSignIn',
+                },
+              },
+              federatedSignIn: {
```
wlee221, line 58: This is probably better named `federatedSignUp`... but leaving this as now to keep PR focused.

packages/core/src/machines/actors/auth/signUp.ts
```diff
@@ -0,0 +1,229 @@
...
+                    target: 'resolved',
+                    actions: ['setUser', 'setCredentials'],
+                  },
+                  onError: {
+                    target: 'idle',
+                    actions: 'setRemoteError',
+                  },
+                },
+              },
+              resolved: { type: 'final', always: '#signUpActor.confirmSignUp' },
```
wlee221, line 94: Note: I'm having explicit resolved state here so that `setCredentials` runs before `formValues` are cleared.

packages/core/src/machines/actors/auth/signUp.ts
```diff
@@ -0,0 +1,229 @@
...
+      resolved: {
+        type: 'final',
+        data: (context) => {
+          const { username, password } = context.authAttributes;
+          const canAutoSignIn = !!(username && password);
+          return {
+            user: context.user,
+            intent: canAutoSignIn ? 'autoSignIn' : null,
+            authAttributes: { username, password },
+          };
```
wlee221, line 138: Hint the parent machine whether we can auto sign in from here.

packages/core/src/machines/authMachine.ts
```diff
@@ -0,0 +1,172 @@
...
+          },
+        ],
+      },
+      signIn: {
+        entry: 'spawnSignInActor',
+        exit: stopActor('signInActor'),
+        on: {
+          SIGN_UP: 'signUp',
+          RESET_PASSWORD: 'resetPassword',
+          'done.invoke.signInActor': [
```
wlee221, line 54: This is the xstate built-in emitted event when signInActor completes. Need to use this because actors are spawned and not invoked, thus doesn't have an explicit `onDone` handler.

packages/core/src/machines/authMachine.ts
```diff
@@ -0,0 +1,172 @@
...
+      authenticated: {
+        on: { SIGN_OUT: 'signOut' },
+      },
+    },
+    on: {
+      CHANGE: { actions: 'forwardToActor' },
+      SUBMIT: { actions: 'forwardToActor' },
+      FEDERATED_SIGN_IN: { actions: 'forwardToActor' },
+      RESEND: { actions: 'forwardToActor' },
+      SIGN_OUT: { actions: 'forwardToActor' },
```
wlee221, line 99: These are the events that are forwarded to the actors. We can do wildcard descriptor `*` here, but wanted to be explicit here to prevent communications from being too chatty.

packages/core/src/machines/authMachine.ts
```diff
@@ -0,0 +1,172 @@
...
+      }),
+      spawnSignInActor: assign({
+        actorRef: (_, event) => {
+          const actor = signInActor.withContext({
+            authAttributes: event.data?.authAttributes,
+            user: event.data?.user,
+            intent: event.data?.intent,
+            formValues: {},
+            validationError: {},
+          });
```
wlee221, line 124: This is how we pass required context to actors. Note that `event.data` is coming from actor `done` events.

packages/core/src/helpers/auth.ts
```diff
@@ -76,3 +80,19 @@ export const getConfiguredAliases = (context: AuthContext) => {
   const [primaryAlias, ...secondaryAliases] = aliases;
   return { primaryAlias, secondaryAliases };
 };
+
+/**
+ * Get the state of current actor. This is useful for checking which screen
+ * to render: e.g. `getActorState(state).matches('confirmSignUp.edit').
+ */
+export const getActorState = (state: AuthMachineState) => {
```
eddiekeller, line 88: nit - I know that adding return types to functions isn't "necessary", but since the given return type isn't explicitly clear to someone looking at this without prior knowledge of what they're working with, what do you think about adding it here? Not sure how easy/difficult that would be with Xstate typings.

packages/core/src/machines/actors/auth/resetPassword.ts
```diff
@@ -0,0 +1,115 @@
...
+  },
+  {
+    actions: {
+      setRemoteError: assign({
+        remoteError: (_, event) => event.data?.message || event.data,
+      }),
+      setUsername: assign({
+        username: (context) => context.formValues.username,
+      }),
+      handleInput: assign({
```
slaymance, line 90: What do you think about changing this action name to `handleChange`?

packages/core/src/helpers/auth.ts
```diff
@@ -42,7 +46,7 @@ export enum FederatedIdentityProviders {
  */
 export const getAliasInfoFromContext = (context: AuthContext) => {
   const loginMechanisms = context.config?.login_mechanisms ?? ['username'];
-  const error = context.validationError['username'];
+  const error = context.actorRef?.context?.validationError['username'];
```
slaymance, line 49: Question here: I think it could be useful to have a constant referencing the string `'username'`, but it's a little confusing since `validationError['username']` isn't really related to how we use `'username'` as the default `login_mechanisms`. Any thoughts on how we can make our use of `'username'` a bit clearer?

packages/core/src/machines/actors/auth/resetPassword.ts
```diff
@@ -0,0 +1,115 @@
...
+        formValues: (context, event) => {
+          const { name, value } = event.data;
+          return { ...context.formValues, [name]: value };
+        },
+      }),
+      clearFormValues: assign({ formValues: {} }),
+      clearError: assign({ remoteError: '' }),
+      clearUsername: assign({ username: undefined }),
```
eddiekeller, line 98: Looks like these actions are gonna be reused across all of the actors. Are we able to pull these out into their own `actions.ts` file?

packages/vue/src/components/confirm-sign-up.vue
```diff
@@ -5,10 +5,10 @@
         <base-heading>
           {{ confirmSignUpHeading }}
         </base-heading>
-        <base-field-set :disabled="state.matches('confirmSignUp.pending')">
+        <base-field-set :disabled="actorState.matches('confirmSignUp.pending')">
           <user-name-alias
             :userNameAlias="true"
-            :userName="state?.context?.formValues[primaryAlias]"
+            :userName="actorState?.context?.formValues[primaryAlias]"
```
ErikCH, line 11: The formValues no longer populated at this point. It needs to be retrieved from the `authAttributes`

```suggestion
            :userName="actorState?.context?.authAttributes?.username"
```

packages/vue/src/components/confirm-sign-up.vue
```diff
@@ -117,9 +118,11 @@ export default defineComponent({
...
       value: { context },
     } = state;
+    const actorState = computed(() => getActorState(state.value));
+    const actorContext = actorState?.value.context;
 
     let [primaryAlias] = useAliases(context?.config?.login_mechanisms);
-    if (!context?.formValues?.confirm_password) {
+    if (!actorContext.formValues?.confirm_password) {
       primaryAlias = 'username';
     }
```
ErikCH, line 127: I don't believe this needs to exist anymore since we don't need to look up the primaryAlias any longer since the value is always being stored in `authAttributes.username` in the signUp Actor

packages/vue/src/components/confirm-sign-up.vue
ErikCH: Is it sometimes `user?.username` Might need to make that check not sure

packages/core/src/types/authMachine.ts
```diff
@@ -26,12 +51,13 @@ export type AuthEventTypes =
...
-  | 'CONFIRM_SIGN_IN'
-  | 'INPUT'
   | 'CHANGE'
   | 'FEDERATED_SIGN_IN'
-  | 'RESET_PASSWORD';
+  | 'RESET_PASSWORD'
+  | 'done.invoke.signInActor'
+  | 'done.invoke.signUpActor'
+  | 'done.invoke.signOutActor'
+  | 'done.invoke.resetPasswordActor';
```
eddiekeller, line 60: nit - since these are events of a different "format" than the previous `AuthEventTypes`, can we make these into their own `type`? Something like `export type InvokeActorEventTypes`. We can then either extend from `AuthEventTypes`, or just change where these types are used to except either `AuthEventTypes` or `InvokeActorEventTypes`.

Just a thought. Not blocking over it.

packages/vue/src/components/confirm-sign-up.vue
```diff
@@ -143,13 +146,15 @@ export default defineComponent({
...
+      const { user, authAttributes } = actorContext;
+      const username = user?.username ?? authAttributes?.username;
       send({
         type: 'SUBMIT',
         //@ts-ignore
         data: {
           //@ts-ignore
           ...Object.fromEntries(formData),
-          username: context?.formValues[primaryAlias],
+          username: username,
```
ErikCH, line 157: You fixed it for submit but you missed updating it for `onLostCodeClicked`. That's still trying to pull from formValues where it should be sending over the `username` like on the submit method.

packages/core/src/types/authMachine.ts
```diff
@@ -63,7 +89,7 @@ export type AuthInputAttributes = Record<AuthInputNames, InputAttributes>;
 
 export interface AuthEvent {
   type: AuthEventTypes;
-  data?: any; // TODO: strongly type data for each AuthEventType
+  data?: any;
```
eddiekeller, line 92: I think we can avoid using `any`. I know we don't necessarily know everything that `data` could be, since it encompasses our form submissions, api return values, etc. But it will always at least be an "object", right? Like, `data` won't ever be `'a string'`, or a `number`. What do you think about about making it have a type of `Record<string, any>`? It still uses `any`, which I guess is sort of unavoidable in this case, but it is more specific at least.

packages/react/src/components/Authenticator/ConfirmSignIn/ConfirmSignIn.tsx
```diff
@@ -17,15 +21,16 @@ export const ConfirmSignIn = (): JSX.Element => {
   } = useAmplify(amplifyNamespace);
 
   const [state, send] = useAuth();
-  const isPending = state.matches('confirmSignIn.pending');
+  const actorState = getActorState(state);
```
eddiekeller, line 24: It looks like we aren't using the `state` from the `useAuth` now. Maybe we should change that variable to something like an underscore to signify that we intentionally aren't using it.

Question - does `useAuth` still provide a valid state for us to work with? And if it does, is the `getActorState` helper necessary?

packages/vue/src/types/index.ts
```diff
@@ -67,7 +70,8 @@ export interface ConfirmPasswordSetupReturnTypes {
   confirmText: ComputedRef<string>;
   onLostCodeClicked: () => void;
   state: Ref;
-  send: PayloadSender<EventObject>;
+  actorState: Ref;
```
ErikCH, line 73: This probably should be `ComputedRef` . If you wanted to be more specific you'd need to add in some type assertion by defining what it will return inside the computed property, otherwise it defaults to `ComputedRef<any>`. 

To fix this, you'd need to go back into `getActorState` and add a return type in there, then add the type in this file.

packages/core/src/machines/actors/auth/resetPassword.ts
```diff
@@ -0,0 +1,115 @@
...
+    },
+    services: {
+      async resetPassword(context) {
+        const { username } = context.formValues;
```
eddiekeller, line 102: `username` won't always be in the `context.formValues`. For instance, we call this method when we `RESEND` the code on the `ConfirmResetPassword` screen, but there is no `username` field on that screen, so it gets sent up as a `data` property on the `onClick` call.

packages/react/src/components/Authenticator/ResetPassword/ConfirmResetPassword.tsx
```diff
@@ -53,9 +62,6 @@ export const ConfirmResetPassword = (): JSX.Element => {
             onClick={() => {
               send({
                 type: 'RESEND',
-                data: {
-                  username: state.context.username,
-                },
```
eddiekeller, line 58: Why was this removed? There is no `username` field on this page, so the `username` won't exist in the `formValues` context. We need to be able to pass this up to properly submit the `RESEND` request.

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
```diff
@@ -100,8 +103,9 @@ SignUp.AliasControl = ({
   const {
     components: { Input, Label, Text, ErrorText },
   } = useAmplify('Authenticator.SignUp.Password');
-  const [{ context }] = useAuth();
-  const error = context.validationError[name];
+  const [state] = useAuth();
+  const { validationError } = getActorContext(state);
+  const error = validationError[name];
```
eddiekeller, line 108: Why is `useAuth()` sufficient for providing `state` here, but we need to use the helper method `getActorState` in other components?

eddiekeller (changes requested):
Good work, big changes. Left a couple comments. Haven't pulled down the changes to play with myself yet, so will do that next.

packages/vue/src/components/sign-up.vue
```diff
@@ -9,7 +9,7 @@
           {{ signUpButtonText }}
         </base-heading>
         <federated-sign-in></federated-sign-in>
-        <base-field-set :disabled="state.matches('signUp.submit')">
+        <base-field-set :disabled="actorState.matches('signUp.submit')">
```
ErikCH, line 12: Federated sign in buttons on the sign up are not working for Vue or React. I get this error. I believe the federatedSignIn service doesn't exist in sign up.

```suggestion
        // add this into signUp services or call the one in SignIn actor
        
      async federatedSignIn(_, event) {
        const { provider } = event.data;
        const result = await Auth.federatedSignIn({ provider });

        return result;
      },
```

`Warning: No service found for invocation 'federatedSignIn' in machine 'signUpActor'.`

packages/core/src/machines/actors/auth/signIn.ts
```diff
@@ -0,0 +1,276 @@
...
+      autoSignIn: {
+        invoke: {
+          src: 'signIn',
+          onDone: { actions: 'setUser', target: '#signInActor.resolved' },
+          onError: '#signInActor.signIn',
+        },
+      },
+      signIn: {
+        initial: 'edit',
+        exit: 'clearFormValues',
```
ErikCH, line 27: Does this work? Anytime I switch between Sign in and Sign Up I get this error.

`Warning: No implementation found for action type 'clearFormValues'`

ErikCH (changes requested):

packages/vue/src/components/confirm-sign-up.vue
wlee221: Ah, thanks for the catch! And you're right, it's `user.username` if coming directly from `signUp`. Making that adjustment along with yours.

wlee221: You're right, thanks!https://github.com/aws-amplify/amplify-ui/pull/180/commits/b422514114380de35fa7dfb1a86fa8343ba8053c

packages/vue/src/components/confirm-sign-up.vue
wlee221: Sorry, this is a typo in my end. SignUp still stores usernames as email/phone_number depending on `primaryAlias`. I'll update the code to reflect this.

wlee221: In future PRs, we should consider storing that value as `username` on signUp, then we can disambiguate it during submit.

wlee221: Wait nvm!! Scratch all I said. You're right, the value is always passed from `authAttributes.username`. Thanks @ErikCH 🎉

packages/vue/src/components/confirm-sign-up.vue
wlee221: Thanks for the catch -- https://github.com/aws-amplify/amplify-ui/pull/180/commits/417d0775068b9d66ccc2683abc9f9437fa56e6e6 to remove primaryAlias as well as cover onLostCodeClicked

packages/core/src/helpers/auth.ts
```diff
@@ -76,3 +80,19 @@ export const getConfiguredAliases = (context: AuthContext) => {
...
+  return state.context.actorRef?.getSnapshot();
+};
+
+/**
+ * Get the context of current actor. Useful for getting any nested context
+ * like remoteError.
+ */
+export const getActorContext = (state: AuthMachineState) => {
+  return getActorState(state)?.context;
+};
```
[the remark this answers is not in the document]
wlee221: Right, we need to chain `.actorRef?.getSnapshot()` because context like `remoteError` is nested under `state.context`. Top level machine only has these contexts: https://github.com/aws-amplify/amplify-ui/blob/322c001eb2412efb0af08aa6594f5b28965cce84/packages/core/src/types/authMachine.ts#L7-L13

packages/core/src/helpers/auth.ts
wlee221: That's fair. This one is very doable, and @ErikCH has provided similar feedback about typing this. I'll add it in!

The only annoyance is that each actor has different context so we have to union them: something like `SignInState | SignUpState | 'SignOutState' | ResetPWState`. Let me try this.

packages/core/src/machines/actors/auth/resetPassword.ts
wlee221: Yes! And I tried this for a bit too. Here's where xstate typing gets more finicky though because we have different context types for each actor and action types don't work so well with unions. 

We might need to use some `as any` or `as AssignAction<SignInContext, AuthEvent>`. Let's pull this into a different PR, as I want to get the structure laid down first. Trust me, I dislike dry code as much as you do 😆

packages/core/src/types/authMachine.ts
wlee221: I like this idea! Done here -- https://github.com/aws-amplify/amplify-ui/pull/180/commits/fbed3464311ca46b93e7b7bb498ce8e242486bad

packages/core/src/types/authMachine.ts
wlee221: Right, let me do that. Eventually, we can have even stricter types with type guards and what not. For now, let me put `Record<PropertyKey, any>`.

wlee221: https://github.com/aws-amplify/amplify-ui/pull/180/commits/aa50ba3a4c14dfb0f9d5acf3227fed5298be910f

packages/core/src/machines/actors/auth/resetPassword.ts
wlee221: Ah, that's a good suggestion. I personally like `handleInput` more because it indicates that it's the "input" that's being handled. I think `handleChange` isn't that clear about which change it's handling.  

Maybe `updateInput`, or `setInput` can work better?

packages/core/src/machines/actors/auth/signIn.ts
wlee221: Yes it works, but I found this behavior too. I think the problems are ran in arbitrary order, so 'clearFormValues' is removed before `signIn` machine exits out. Let me see if tweaking this to "resolved" state fixes this.

packages/core/src/helpers/auth.ts
wlee221: Ah, good catch here. And I believe this implementation is wrong -- "email" / "password" are not stored as "email" / "password" when on `signUp` screen. We just never had problems with this because `username` was never verified. 

Let's keep this as now, but let's consider making every username alias stored as "username", and we can disambiguate them elsewhere so we have consistent `formValues` interface.

packages/vue/src/components/sign-up.vue
wlee221: Thanks for the catch -- https://github.com/aws-amplify/amplify-ui/pull/180/commits/b18cafa6cb49f0dd0ce81d8fe9280a78e6f32842. Validated locally 👍

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
wlee221: Good question -- `getActorState` is mostly used for matching states, e.g. `actorState.matches('signUp.confirmSignUp')`. Otherwise, `getActorContext` is enough to get the contexts.

packages/react/src/components/Authenticator/ConfirmSignIn/ConfirmSignIn.tsx
wlee221: We don't need much from `useAuth` anymore, as we're mostly interested in child substates (like `signIn.confirmSignIn`), which are in actor. `useAuth` is now primarily used to get the nested context, which are at `state.context.actorRef.getSnapshot().context` which my helper function simplifies a bit.

I like underscore though! Let me do that cross-platform.

wlee221: - React: https://github.com/aws-amplify/amplify-ui/pull/180/commits/1960cab3bebefa929318d419e0eecf5764bda544
- Vue: https://github.com/aws-amplify/amplify-ui/pull/180/commits/dfeba88777421970baeeef7809b165969855b4a9


I think Angular is good because I set those context at the same place.

packages/core/src/machines/actors/auth/resetPassword.ts
wlee221: Now that I think about it, just `INPUT` events makes the most sense 😄  Let me try that on a future pr, but also open to discussions.

packages/core/src/machines/actors/auth/resetPassword.ts
wlee221: Good catch -- I'll do `context.formValues?.username ?? context.username`.

packages/react/src/components/Authenticator/ResetPassword/ConfirmResetPassword.tsx
wlee221: We don't need to send `username` anywhere because it's persisted in `context.username` anyway. Instead of preparing data here, we're directly using context in this service call: https://github.com/aws-amplify/amplify-ui/blob/f63cfeb4f6bf6fe8fa95929c5397d84703e922ef/packages/core/src/machines/actors/auth/resetPassword.ts#L107-L108

packages/vue/src/types/index.ts
wlee221: Done! Thanks for the explanation.

packages/core/src/helpers/auth.ts
wlee221: Done here: 
https://github.com/aws-amplify/amplify-ui/blob/8030d30c0cc32bf72db49e6d46284e5412836abc/packages/core/src/types/authMachine.ts#L51-L59

and return type is specified now: https://github.com/aws-amplify/amplify-ui/blob/8030d30c0cc32bf72db49e6d46284e5412836abc/packages/core/src/helpers/auth.ts#L91-L101

packages/core/src/machines/actors/auth/resetPassword.ts
wlee221: Done here https://github.com/aws-amplify/amplify-ui/pull/180/commits/d67bb44bafa8d2f7e7396cb096eeb1180eed0e9a

wlee221:
I think all the comments are addressed! (or scoped out for a future PR). Sorry about spamming your emails 😄

packages/core/src/types/authMachine.ts
```diff
@@ -1,37 +1,83 @@
...
+
+export type SignInState = State<SignInContext, AuthEvent>;
+export type SignUpState = State<SignUpContext, AuthEvent>;
+export type SignOutState = State<SignOutContext, AuthEvent>;
+export type ResetPasswordState = State<ResetPasswordContext, AuthEvent>;
+export type AuthActorContext =
+  | SignInContext
+  | SignUpContext
+  | SignOutContext
+  | ResetPasswordContext;
```
eddiekeller, line 59:
```suggestion
export type AuthActorContext = ActorContextWithForms | SignOutContext;
```

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
```diff
@@ -71,7 +83,8 @@ export function SignIn() {
           {isPending ? <>Signing in&hellip;</> : <>Sign In</>}
         </Button>
       </Footer>
-      <Box data-amplify-error>{state.event.data?.message}</Box>
+      <Box data-amplify-error>{actorState.context.remoteError}</Box>
+      <ErrorText amplifyNamespace={amplifyNamespace} />
```
eddiekeller, line 87: We don't need both of these. The `ErrorText` component will look in the `state.context` for the `remoteError`, so if we keep both the `Box` and `ErrorText` here, we will be showing the same error twice.

I think instead, you'll just need to edit the `ErrorText` component to use the new `actorState` method.

packages/react/src/components/Authenticator/index.tsx
```diff
@@ -44,24 +45,24 @@ export function Authenticator({
           switch (true) {
             case state.matches('idle'):
```
eddiekeller, line 46: Should this one also be switched to `actorState`?

eddiekeller (changes requested):

packages/core/src/types/authMachine.ts
eddiekeller: `ActorContextWithForms` contains the `SignUp/SignIn/ResetPassword` contexts, so we probably don't need to list them twice. Unless you are worried about `ActorContextWithForms` possibly growing and getting out of sync with this `AuthActorContext` (though from the naming, I don't think it will?).

eddiekeller: (this is a "nit", btw. not blocking if you don't like the idea)

eddiekeller:
Bugs I have found playing with the React environments - 

* On the `ConfirmSignIn` page, hitting `Back to Sign In` button doesn't do anything.
* Can we make it so that the XState inspector doesn't open in a new tab every time we initiate a connection with `localhost:3000`?
* On the `SignUp` page, when going through a Federated sign up for the first time, I got an app crash an error saying `username is undefined`. It actually did still reroute over to Facebook properly for the sign in, and after redirecting back to the app, I was logged in fine. Haven't been able to reproduce it again. Would be a weird experience for customers though. Probably not something we have to be *too* concerned with now, but let's keep our eyes open for it.

Otherwise, I tested the `ConfirmSignIn`, `SignUp`, `ConfirmSignUp`, `ResetPassword`, `SignIn`, and `FederatedSignIn/Up` flows. All of them seem to work good. The only thing I didn't test is `ForceNewPassword`, because that one is a pain in the butt to set up and I figure if all the others work, that one probably does too? Or I'm being lazy?

Anyway, if we can fix the two main points above and the new review I added earlier, then I'll get this approved!

ErikCH:
Quick follow up from @eddiekeller comment. On Vue the ConfirmSignIn back to sign in was working. However, the Force New Password `Have an account Sign In` button was not working. 

Federated sign in, logging in, logging out, confirm signing up worked.

ericclemmons (approved):
Do you have visualizations for the individual machines as well? It seems like those are more complex than the parent visualization is letting on.

I don't have any objections to merging this – it's a huge amount of work, and any changes I propose will be the result of _actual usage_ and abstracting over things like `actorState?.matches()`

Great work @wlee221 – I can tell you put a **ton** of thought, effort, and iterations into this. 👏

packages/core/src/machines/authMachine.ts
```diff
@@ -0,0 +1,172 @@
...
+      clearUser: assign({
+        user: undefined,
+      }),
+      setAuthConfig: assign({
+        config(_, event) {
+          return event.data.auth;
+        },
+      }),
+      spawnSignInActor: assign({
+        actorRef: (_, event) => {
```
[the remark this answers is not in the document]
ericclemmons: I'm curious:

**What are the pros/cons to spawning all machines when spawning the `authenticatorMachine`?**

```
const authenticatorMachine = createMachine({
  ...

  context: {
    // Optional, in case it's valuable to know the current screen (https://xstate.js.org/docs/tutorials/reddit.html#spawning-subreddit-actors)

    // Screens are already created (or as part of the `init` flow) and `idle`
	screen: signInMachine,
    signIn: signInMachine,
    signOut: signOutMachine,
  }
})
```

One thing that seems obvious is initializing `context` between screens: saving `username` on the Sign In screen and putting it on the `Sign Up` `context`.

packages/core/src/types/authMachine.ts
wlee221: No, I like this! Comming it in.

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
wlee221: Right, that's my bad. I did already edit `ErrorText` component, so I'll just remove `Box`.
```suggestion
      <ErrorText amplifyNamespace={amplifyNamespace} />
```

packages/react/src/components/Authenticator/index.tsx
wlee221: No, and here's an inconvenience I find about this refactor: `idle` is not part of any actor, so it's caught in the parent machine. I'm considering having better helper function that abstracts this actor layer (or can this be improved high level?)

something like this:
```tsx
enum AuthUIState {
  SignIn: 'signIn',
  SignUp: 'signUp',
  // ...
}

const getAuthUIState = (State machineState): AuthUIState => {
  // returns AuthUIState.SignIn, AuthUIState.SignUp, ...
}
```

wlee221:
re: @eddiekeller,
> On the ConfirmSignIn page, hitting Back to Sign In button doesn't do anything.
Thanks for the catch, `SIGN_IN` event was forwarded. Fixed in https://github.com/aws-amplify/amplify-ui/pull/180/commits/a768043632487d141cb26f7e61c3c0607b07a7da
> Can we make it so that the XState inspector doesn't open in a new tab every time we initiate a connection with localhost:3000?
Totally! I had it enabled for local dev help, but I'll remove it pre-merge.
> On the SignUp page, when going through a Federated sign up for the first time, I got an app crash an error saying username is undefined. 
The problem was that `authAttributes` context didn't get a default value. Fixed in https://github.com/aws-amplify/amplify-ui/pull/180/commits/9ed3dc616e7f6c20082bee59be117cd4f021223d

eddiekeller:
Thanks for making the changes @wlee221 . I'm good with merging now. This PR is really big and future feature work will depend on getting this in place. If there are little bugs, we can fix em up as we go. But my surface level testing on the React side looked good. Dropping an approval bomb on this one.

eddiekeller (dismissed):

wlee221:
Ah, I think "Dismiss stale pull request approvals when new commits are pushed" option is on... I'll let you know all when it's ready for final approval. I think we should consider disabling this option.

Anyhow, I'll get this pr done by 3pm pdt as future feature prs depend on this one.

wlee221:
Re: @ErikCH,

> On Vue the ConfirmSignIn back to sign in button was not working either. Also the force New Password Have an account Sign In button does not work.

This is fixed in https://github.com/aws-amplify/amplify-ui/pull/180/commits/a768043632487d141cb26f7e61c3c0607b07a7da.

> The QR code did not load. 

Thanks for the catch! I missed `setupTOTP` onMounted event. Fixed here: https://github.com/aws-amplify/amplify-ui/pull/180/commits/5420401a58cabc6ebfc6df820c3778a84157ce06

packages/core/src/machines/actors/auth/resetPassword.ts
```diff
@@ -0,0 +1,115 @@
...
+
+import { AuthEvent, ResetPasswordContext } from '../../../types';
+import { Auth } from 'aws-amplify';
+
+export const resetPasswordActor = createMachine<
+  ResetPasswordContext,
+  AuthEvent
+>(
+  {
+    id: 'resetPasswordActor',
```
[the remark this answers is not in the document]
wlee221: That's a good point. To be very honest, I just suffixed `actor` to the ids because I preferred

`signInActor.signIn` vs `signIn.signIn` to distinguish substate names from the id.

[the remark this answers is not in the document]
wlee221: I agree with naming the variable name `resetPasswordMachine`. For id, let me think for a minute. Maybe `signIn.signIn` is okay... or `signInMachine.signIn`?

packages/core/src/machines/actors/auth/resetPassword.ts
```diff
@@ -0,0 +1,115 @@
...
+    initial: 'resetPassword',
```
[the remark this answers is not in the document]
wlee221: I removed it in favor of `withContext` in the parent machine: https://github.com/aws-amplify/amplify-ui/blob/317c7cefb8b8e76e0620c7bdbac5ec006f808d1b/packages/core/src/machines/authMachine.ts#L141-L148.

Though I see the point of adding it in `resetPasswordMachine` too for readability and initialization purposes. I'll descope it out of this PR because it's already too big, but I'll add this to my TODO list.

packages/core/src/machines/authMachine.ts
```diff
@@ -0,0 +1,172 @@
...
+    // url: 'https://statecharts.io/inspect', // (default)
+    iframe: false, // open in new window
+  });
+}
+
+export const authMachine = createMachine<AuthContext, AuthEvent>(
+  {
+    id: 'auth',
+    initial: 'idle',
+    context: {},
```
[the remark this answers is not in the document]
wlee221: Let's do it!
```suggestion
    context: {
      user: undefined,
      config: undefined,
      actorRef: undefined,
    },
```

packages/core/src/machines/actors/auth/resetPassword.ts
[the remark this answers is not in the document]
wlee221: Let's descope this to another PR, because it's mostly naming things. But I do agree!

packages/core/src/machines/authMachine.ts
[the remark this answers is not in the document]
wlee221: That's a great idea!! We can do what they did and still have `actorRef` active so that the references can be the same. This should also save us headache from starting and stopping actors.

One implementation detail I see is having to pass context after it's initialized, which is not a big deal but we would need some updateContext actions.

packages/core/src/helpers/auth.ts
```diff
@@ -76,3 +83,19 @@ export const getConfiguredAliases = (context: AuthContext) => {
...
   return { primaryAlias, secondaryAliases };
 };
+
+/**
+ * Get the state of current actor. This is useful for checking which screen
+ * to render: e.g. `getActorState(state).matches('confirmSignUp.edit').
+ */
+export const getActorState = (state: AuthMachineState): AuthActorState => {
+  return state.context.actorRef?.getSnapshot();
+};
```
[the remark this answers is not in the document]
wlee221: I agree. I had some idea here to Eddie's comment:

> I'm considering having better helper function that abstracts this actor layer (or can this be improved high level?)
> 
> something like this:
> 
> ```tsx
> enum AuthUIState {
>   SignIn: 'signIn',
>   SignUp: 'signUp',
>   // ...
> }
> 
> const getAuthUIState = (State machineState): AuthUIState => {
>   // returns AuthUIState.SignIn, AuthUIState.SignUp, ...
> }
> ```

packages/core/src/machines/authMachine.ts
wlee221: That's a fair point. Good thing is that there's no handler for `FEDERATED_SIGN_IN` in `resetPassword` flow, nor are there buttons in `resetPassword` screens able to trigger that event. But I see your point -- maybe this is worth moving those `forwardToActor` to each individual state. Adding to my TODOs.

wlee221:
Re: @ericclemmons,

Ah, my screenshot got buried in my long, long thread. Here's the `signIn` machine, which is the most complicated among the four:

image

(I had to rotate my monitor 270 degrees to take this screenshot 😆 ) I'll add this to the description too.

eddiekeller (approved):

wlee221:
This is ready for final approval. Thank you all so much for testing them thoroughly across frameworks as well -- I'm glad we caught so many errors pre-merge. I knew I must have missed something when I was making changes across the packages, but not this much 😅

packages/core/src/helpers/auth.ts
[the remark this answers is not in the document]
ericclemmons: Agreed.  For launch, we'll prioritize not the implementation, but _how_ our view layers rely on state & context – that's what customers will be using as well.

ericclemmons (approved):
Let's give this a go! Great job @wlee221 !

ErikCH (approved):
Looks great! Good job

wlee221:
🚢 For sake of tracking this, here's a list of TODOs:

- Share any reused actions 
- Better name for `handleInput` or `CHANGE` event
- Look into “Warning: No implementation found for action type 'clearFormValues' “
- Rename `[authFlow]Actor` to `[authFlow]Machine`, as well as their ids.
- Define Initial context for each machine
- Spawn all machines when we spawn `authenticatorMachine`
- Facade for `state.matches`
- Specify target for `forwardToActor`

Most of them (except facade) are short implementation details that we can improve incrementally. I promise I'll keep those PRs short!

PR #183, opened 2021-08-12
Vendor prefix

jacoblogan:
This pull request will use autoprefixer to add vendor prefixes to both the static styling and and inline styling for amplify theming and primitive styling.

The render-time inline styling all runs through the view component style prop, this includes both the developer defined style props passed into primitives through the style prop and also the amplify styling props such as `color`, `fontWeight`, etc.  So by wrapping the style object we compute with the autoprefixer function, we will catch almost all instances of inline styling.  Some css in js libraries will avoid this by creating class names at run time, but most css in js libraries also handle vendor prefixes.


packages/react/src/primitives/View/View.tsx
```diff
@@ -1,6 +1,10 @@
 import React from 'react';
 import { convertStylePropsToStyleObj, getNonStyleProps } from '../shared/utils';
 import { ViewProps } from '../types/index';
+import postcssJs from 'postcss-js';
+import autoprefixer from 'autoprefixer';
+
+const prefixer = postcssJs.sync([autoprefixer]);
```
zchenwei, line 7: I feel like it makes more sense to put the logic inside a util function and then import that function to use here. What do you think?

zchenwei: Or maybe could wrap it outside `convertStylePropsToStyleObj`

zchenwei (changes requested):
Looks good! Just a small comment

packages/react/src/primitives/shared/utils.ts
```diff
@@ -3,6 +3,10 @@ import {
   ComponentPropToStyleProp,
   ViewProps,
 } from '../types/index';
+import postcssJs from 'postcss-js';
```
reesscot, line 6: nit: mind moving external imports to the top of the file?

reesscot (dismissed):
LGTM!

zchenwei (approved):
LGTM!

reesscot (approved):

jaminmelville:
Hey @jacoblogan I'm trying to find out why my bundled JS is so large. It looks like the `autoprefixer` package is contributing significantly, as it has a dependency on `caniuse-lite` which is 1.44MB. 
I'm guessing that this may be an issue for me as I haven't found any related issues in github. 
Can you shed any light?
Thanks!

PR #187, opened 2021-08-13
Styling alternatives

jacoblogan:
The goal of this pull request is to add documentation on how amplify styling and theming interacts with the common CSS In JS libraries.  The documentation should explain how amplify components can be used in conjunction with these libraries and the expected behavior of conflicting styling between the two.


docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+---
+
+import Block from 'amplify-docs/src/components/Block';
+import BlockSwitcher from 'amplify-docs/src/components/BlockSwitcher';
+
+One of the goals of Amplify is to be a good citizen of the environment it exists in. As such Amplify works hard to not interfere with other tools that are being used.
+
+## CSS In JS
+
+There are many CSS In JS libraries out there that give a number of benefits. Amplify does not use a CSS In JS library to implement its theming system but operates well with the major CSS In JS libraries out there. Below are a few of the CSS In JS libraries out there and a description of how they interact with Amplify's theming system.
```
dbanksdesign, line 12: 'CSS in JS libraries out there' twice in succession, we can probably change the second sentence to just say 'Below are a few libraries with descriptions of how they work with Amplify's theming system.'

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+import { View } from '@aws-amplify/ui-react';
+
+const StyledView = styled(View)`
+  color: blue;
+`;
+
+<StyledView color="red">Using Styling props</StyledView>;
+```
+
+In the example above, the color of the view will be set to `red` because of the Amplify styling prop. [Here]() is a list of all Amplify styling props.
```
dbanksdesign, line 56: TODO for me: create that page that 'Here' links to

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+
+##### Amplify variation props
+
+These are Amplify props that are added to certain Amplify components to change the look and/or behavior of the Amplify component. Examples of these variation props are `size` or `variation`.
```
dbanksdesign, line 60: We could make this a bit tighter:

'These props change the look and/or behavior of certain Amplify components. Examples include `size` and `variation`.'

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+import { Button } from '@aws-amplify/ui-react';
+
+const StyledButton = styled(Button)`
+  color: blue
+`;
+
+<StyledButton variation="primary'>Primary Button</StyledButton>
+```
+
+In the example above, the color of the button will be set to `white`, because the primary variation of the button has a color set to `white` and it takes precedence over the Styled Components styling. A list of Amplify variations can be found on [here]().
```
dbanksdesign, line 74: Interesting, is this always the case or is it just because with variations our selector has a higher precedence? If you used a button without a variation would the color be blue?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+```
+
+```CSS
+/* External Style Sheet */
+:root {
+  --amplify-components-button-color: red;
+}
+```
+
+In the example above, the button color will be set to `blue` because the Styled Component will take precedence over the Amplify CSS variable.
```
dbanksdesign, line 124: Similar question to above, is this true for all CSS variable overrides?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+
+</Block>
+
+<Block name="JSS">
+[JSS](https://github.com/cssinjs/jss/tree/master/packages/react-jss) is a lib for generating Style Sheets with Javascript.
```
dbanksdesign, line 129: nit: we should probably be consistent with 'library' and 'lib'

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+  --amplify-components-button-color: red;
+}
+```
+
+In the example above, the button color will be set to `blue` because the JSS styling will take precedence over the Amplify CSS variable.
+
+</Block>
+
+<Block name="Emotion">
+[Emotion](https://github.com/emotion-js/emotion) is a performant and flexible CSS-in-JS library. It allows you to style apps quickly with string or object styles.
```
dbanksdesign, line 264: We might want to be a bit more objective in these descriptions so it doesn't look like we are endorsing one library over another.

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+```jsx
+/** @jsxImportSource @emotion/react */
+import { jsx, css } from '@emotion/react';
+import { Text } from '@aws-amplify/ui-react';
+
+const styles = {
+  color: 'blue',
+};
+
+<Text css={css(styles)} color="red">
```
dbanksdesign, line 298: The `css` prop works?? Cool. 

cool

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+  This is my Emotion Text{' '}
```
dbanksdesign, line 299: Is `{' '}` something special here?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+    color: 'blue',
+  },
+});
+
+<Text className={css(styles.blue)} color="red">
+  This is an aphrodite Text component
+</Text>;
+```
+
+In the example above, the color of the view will be set to `blue` because of the Aphrodite styling will take precedence. [Here]() is a list of all Amplify styling props.
```
dbanksdesign, line 421: It looks like this is the only library where this use-case will be 'blue' instead of 'red', is that correct?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+    color: 'blue',
+  },
+});
+
+<Button className={css(styles.blue)} variation="primary">
+  Primary Button
+</Button>;
+```
+
+In the example above, the color of the button will be set to `blue`, because the Aphrodite styling will take precedence over the Amplify variation styling. A list of Amplify variations can be found on [here]().
```
dbanksdesign, line 442: Same here, all other libraries this would be 'white', is that expected?

dbanksdesign (commented):
I love this PR! Just a few minor language changes and questions for ya

next level shit

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: yes by default aphrodite adds !important to all of their styling, there is a config option to turn this off but I just have the default behavior covered here.

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: yes by default aphrodite adds !important to all of their styling, there is a config option to turn this off but I just have the default behavior covered here.

dbanksdesign: ohhh interesting.... should we make a note of this in the docs?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: Yes this is because our variation styling rules have a higher precedence, without the variation on the button the library styling would take precedence over ours and the button would be color blue.

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: This is true because of styling precedence, but unfortunately styling precedence isn't completely consistent because if they target the variables that are used in rules with a higher precedence then the variables will create changes that override the css in js changes. Some of these variables would be for variations or states (i.e. hover). I'm not sure what the best way to express this to the user would be.

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: Yeah probably, for the most part I just took the first sentance or two of their description in their readme file and that is what I used for the description. I'll rework this one to be more objective.

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: nope, pulling that out

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: yes we should probably make a note of this, what I didn't want to do is write two docs for Aphrodite where one is how this behaves by default and the other is how this behaves if you take off !important.  Should I just make a note that this is happening because of !important or should I also note that Aphrodite can be configured to take !important off in which case it behaves more like the others?

dbanksdesign: Yea I think just make a note about the behavior and how they can change that with the aphrodite config.

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
reesscot: Couldn't the customer use a data attribute selector to make it work, like this?
```
const StyledButton = styled(Button)`
  &[data-variation='primary']: {
    color: blue
   }
`;
```

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+```
+
+```CSS
+/* External Style Sheet */
+.my-styled-view {
+  color: red;
+}
+```
+
+In the example above, the color of the view will be `blue` because the stlyed component will take precedence over the classname CSS.
```
reesscot, line 99:
```suggestion
In the example above, the color of the view will be `blue` because the styled component will take precedence over the classname CSS.
```

reesscot: Can you add an explanation of why the styled component CSS will take precedence over the className CSS?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+
+##### CSS Root Variable Overrides
+
+Amplify theming provides many CSS variables to customize the look and feel of the entire application. These CSS variables can be used at a root level to override all instances of component styling.
```
reesscot, line 103:
```suggestion
Amplify theming provides many CSS variables to customize the look and feel of the entire application. These CSS variables can be used with the `:root` CSS psuedo-class to override all instances of component styling.
```

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
reesscot: This makes me wonder if we should try to have the same specificity for all of our custom rules... The attribute selector rules certainly complicate things.

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
```diff
@@ -0,0 +1,501 @@
...
+});
+
+const classes = useStyles();
+
+<Button className={classes.styledButton} variation="primary">
+  Primary Button
+</Button>;
+```
+
+In the example above, the color of the button will be set to `white`, because the primary variation of the button has a color set to `white` and it takes precedence over the JSS styling. A list of Amplify variations can be found on [here]().
```
reesscot, line 200: Does JSS support `&[data-attribute]` selectors?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
dbanksdesign: Yea I'm thinking that as well. I guess that means we have to go with BEM or BEM-like syntax then?

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: yes `&[data-attribute]` selectors are supported. In this guide I was aiming to just describe the default behavior and the experience I would expect developers would run into if they just use CSS in JS libraries without specifically trying to interact with amplify's styling

docs/src/pages/ui/theme/alternativeStyling/index.page.mdx
jacoblogan: yes the developer can use data attribute selectors to make their rules take precedence. In this guide I was trying to describe the experience that they would run into if they simply use a CSS in JS library without putting in extra effort to work well with amplify.  It seems like there are variations and work arounds for anything they might want to do, but I was trying to not branch off into too many tangents for this guide.

joebuono (approved):
LGTM! Great PR.

PR #194, opened 2021-08-18
Use common button click step and make it case insensitive

slaymance:
*Issue #, if available:* N/A

*Description of changes:* With the introduction of I18N strings in authenticator ([PR 189](https://github.com/aws-amplify/amplify-ui/pull/189)), we need the common "Click button" Cypress action to be case insensitive to function correctly with the new strings.

Furthermore, I'm skipping the "Sign up with valid email & password" test in `sign-up-with-email.feature`. We continuously get a failure due to `Exceeded daily email limit for the operation or the account. If a higher limit is required, please configure your user pool to use your own Amazon SES configuration for sending email.` This will be addressed as part of my audit of skipped e2e tests ([Asana](https://app.asana.com/0/1200781756632584/1200751914824795)), but it makes the most sense right now to skip so CI can continue to function correctly.

### DIFF
--- a/packages/e2e/cypress/integration/common/index.tsx
+++ b/packages/e2e/cypress/integration/common/index.tsx
@@ -16,7 +16,7 @@ And('I type a valid password {string}', (password: string) => {
 });
 
 And('I click the {string} button', (name: string) => {
-  cy.findByRole('button', { name }).click();
+  cy.findByRole('button', { name: new RegExp(name, 'i') }).click();
 });
 
 Then('I see {string}', (message: string) => {
--- a/packages/e2e/cypress/integration/common/index.tsx
+++ b/packages/e2e/cypress/integration/common/index.tsx
@@ -15,7 +15,7 @@ And('I type a valid password {string}', (password: string) => {
   cy.findByLabelText(/password/i).type(Cypress.env(password));
 });
 
-And('I click the {string} button', (name: string) => {
+When('I click the {string} button', (name: string) => {
   cy.findByRole('button', { name: new RegExp(name, 'i') }).click();
 });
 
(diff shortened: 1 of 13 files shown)


packages/e2e/cypress/integration/ui/components/authenticator/sign-up-with-email.feature
```diff
@@ -13,7 +13,7 @@ Feature: Sign Up with Email
     And I don't see "Username" as an input field
     And I don't see "Phone Number" as an input field
 
-@next @react @vue @angular
+@next @react @vue @angular @skip
```
slaymance, line 16: This test is leading us to `Exceeded daily email limit for the operation or the account. If a higher limit is required, please configure your user pool to use your own Amazon SES configuration for sending email.`

packages/e2e/cypress/integration/common/index.tsx
```diff
@@ -15,8 +15,8 @@ And('I type a valid password {string}', (password: string) => {
   cy.findByLabelText(/password/i).type(Cypress.env(password));
 });
 
-And('I click the {string} button', (name: string) => {
-  cy.findByRole('button', { name }).click();
+When('I click the {string} button', (name: string) => {
```
slaymance, line 18: This step definition can be used by any test, so I just updated this one to be case insensitive so all tests can use this.

ErikCH (approved):

packages/e2e/cypress/integration/ui/components/authenticator/sign-up-with-email.feature
ericclemmons: @eddiekeller Didn't you experience this as well & request an increase?

eddiekeller: @ericclemmons The workaround is to switch to using our own email in Amazon SES rather than relying on Cognito to send out emails for us. Not sure if that is what we want to use long term or not.

PR #195, opened 2021-08-18
yarn react build|dev

ericclemmons:
- [x] Replace #177 
- [x] Replace #186 
- [x] Use `tsup` for bundling – https://tsup.egoist.sh/
- [x] Output `@aws-amplify/ui-react/styles.css`
- [x] Inline all `@aws-amplify/*` dependencies (so it's a single bundle)
- [x] Remove Next.s compilation for `packages/*`
- [x] `yarn react dev` for development
- [x] `yarn react build` for production build
- [x] `yarn core build`
- [x] Watch changes in `@aws-amplify/ui-core` when doing `yarn react dev`

```
❯ yarn react dev
yarn run v1.22.10
$ yarn workspace @aws-amplify/ui-react dev
$ rimraf dist
$ tsup --watch src --watch ../core/src
 CLI  Building entry: src/index.tsx
 CLI  Using tsconfig: /Users/ecclemm/Projects/aws-amplify/amplify-ui/packages/react/tsconfig.json
 CLI  tsup v4.14.0
 CLI  Using tsup config: /Users/ecclemm/Projects/aws-amplify/amplify-ui/packages/react/tsup.config.cjs
 CLI  Running in watch mode
 CLI  Target: node12
 CJS  Build start
 ESM  Build start
 ESM  Build success in 6788ms
 CJS  Build success in 7019ms
 CLI  Watching for changes in "src" | "../core/src"
 CLI  Ignoring changes in "**/{.git,node_modules}/**" | "dist"
 DTS  Build start
 DTS  Build success in 28272ms
 CLI  Change detected: change ../core/src/authService.ts
 CJS  Build start
 ESM  Build start
 CJS  Build success in 1408ms
 ESM  Build success in 1407ms
```


packages/react/package.json
```diff
@@ -2,18 +2,27 @@
   "private": true,
   "name": "@aws-amplify/ui-react",
   "version": "0.0.1",
-  "main": "dist/index.js",
+  "type": "module",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
+  "exports": {
+    ".": "./dist/index.js",
+    "./styles.css": "./dist/index.css"
```
ericclemmons, line 10: `import "@aws-amplify/ui-react/styles.css"` correctly resolves to `dist/index.css`.

Neat, right?

packages/react/package.json
```diff
@@ -36,6 +46,8 @@
     ]
   },
   "devDependencies": {
+    "@aws-amplify/ui-core": "^0.0.1",
+    "@aws-amplify/ui-theme-base": "^0.0.1",
```
ericclemmons, line 50: Build tools like `microbundle`, `tsup`, etc. externalize `dependencies` and inline `devDependencies`.

This is helpful for using packages that we don't want to publish.

packages/react/patches/tsup+4.14.0.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/tsup/dist/index.js b/node_modules/tsup/dist/index.js
+index 1fd3ea9..8f84cf0 100644
+--- a/node_modules/tsup/dist/index.js
++++ b/node_modules/tsup/dist/index.js
+@@ -13633,6 +13633,7 @@ async function runEsbuild(options, { format, css }) {
+       jsxFactory: options.jsxFactory,
+       jsxFragment: options.jsxFragment,
+       sourcemap: options.sourcemap,
++      inject: ['./src/react-shim.js'],
```
ericclemmons, line 9: https://esbuild.github.io/content-types/#auto-import-for-jsx

packages/react/src/index.tsx
```diff
@@ -1,3 +1,5 @@
+import './styles.css';
```
ericclemmons, line 1: This is really nice – building (via PostCSS) styles as part of a single package & process.

packages/theme-base/package.json
```diff
@@ -22,8 +22,5 @@
     "rimraf": "^3.0.2",
     "sass": "^1.35.2",
     "style-dictionary": "^3.0.1"
-  },
-  "dependencies": {
-    "@fontsource/inter": "^4.5.0"
```
ericclemmons, line 27: The bundler had issues with WOFF files inside of this dependency.

Admittedly, this is weird to me as part of a library.  Because fonts require asset files (e.g. `.woff`, `.ttf`, etc.), these have to be bundled by the client app.

Normally, I see styles only specify `font-family` and leave it to the consumer to determine **where** the fonts are located:

- https://material-ui.com/customization/typography/#self-hosted-fonts
- https://tailwindui.com/documentation#optional-add-the-inter-font-family
- https://create-react-app.dev/docs/adding-images-fonts-and-files/
- https://github.com/csstools/postcss-font-magician

Because fonts can be very expensive, I think we should revisit #165 and try to make bundle-size as much of an opt-in as possible, like https://tailwindui.com/documentation#optional-add-the-inter-font-family.

If a customer doesn't want it, `font-family` will naturally go to the next available font.

If a customer wants to load it in a blocking way, they can use `<link>`.

If a customer wants to load it in a non-blocking way, they can do what Next.js does:
> https://nextjs.org/docs/basic-features/font-optimization

packages/react/package.json
dbanksdesign: neat

packages/react/package.json
dbanksdesign: Broader question: do we want to publish ui-theme-base? I noticed we publish @aws-amplify/ui which is kind of similar to what ui-theme-base is now. I don't have a strong opinion, just wanted to bring it up. Should we rename ui-theme-base to just ui?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "install": "yarn run theme-base build:all",
+    "install": "yarn run theme-base build:all && yarn core build && yarn react build",
```
ericclemmons, line 4: Going to replace this with `turbo`

dbanksdesign, line 4: ohh I like, now I can just run install and then start up the docs or example next server

package.json
```diff
@@ -37,7 +37,8 @@
   },
   "dependencies": {
     "patch-package": "^6.4.7",
-    "postinstall-postinstall": "^2.1.0"
+    "postinstall-postinstall": "^2.1.0",
+    "tsup": "^4.14.0"
```
ericclemmons, line 41: `tsup` checked all the boxes that `vite` did, + `.d.ts` output.

`Nx` requires a ton of convention, which is **great**, but more than I wanted to change.

`microbundle` is slow because of rollup.

packages/core/package.json
```diff
@@ -21,5 +23,8 @@
     "@types/lodash": "^4.14.170",
     "@types/webpack-env": "^1.16.2",
     "tslib": "^2.3.0"
+  },
+  "peerDependencies": {
+    "aws-amplify": "3.x.x || 4.x.x"
```
ericclemmons, line 28: Otherwise our `amazon-cognito-identity-js` and `aws-amplify` dependencies get bundled!

packages/react/package.json
```diff
@@ -21,15 +28,18 @@
     "test:unit:watch": "jest --watch"
   },
   "dependencies": {
+    "@aws-amplify/ui-core": "^0.0.1",
+    "@aws-amplify/ui-theme-base": "^0.0.1",
```
ericclemmons, line 32: I'll move theme-base into core in another PR. We won't lose tree-shaking or anything, but it'll be 1 less bundle for customers to keep in sync.

packages/react/package.json
```diff
@@ -21,15 +28,18 @@
...
     "classnames": "^2.3.1",
     "postcss-js": "^3.0.3",
     "qrcode": "^1.4.4"
   },
   "peerDependencies": {
-    "@aws-amplify/ui-core": "^0.0.1",
-    "@aws-amplify/ui-theme-base": "^0.0.1"
+    "aws-amplify": "3.x.x || 4.x.x",
+    "react": ">= 16.8.0",
+    "react-dom": ">= 16.8.0"
```
ericclemmons, line 42: This way we don't bundle dependencies.

packages/react/src/index.tsx
ericclemmons: Somehow the SCSS stuff just worked?? I need to check the docs to be sure...

packages/react/src/globals.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * This shim exists for `esbuild` to `inject` into every .tsx file.
+ * Otherwise, `import * as React from 'react'` would need to be added manually
+ *
+ * See: https://esbuild.github.io/content-types/#auto-import-for-jsx
+ * See: https://github.com/egoist/tsup/issues/390
+ */
+
+import * as React from 'react';
+export { React };
```
ericclemmons, line 10: Vite does this automatically. I had to patch `tsup` to avoid changing EVERY `.tsx` file.

packages/react/tsconfig.json
```diff
@@ -1,14 +1,16 @@
 {
   "compilerOptions": {
+    "moduleResolution": "node",
+    "baseUrl": ".",
```
ericclemmons, line 4: I thought this was the default, but `tsup` & `vite` weren't picking up on `node_modules` until I added this.

dbanksdesign (approved):
Overall, I love it. A couple small questions, nothing blocking.

packages/react/tsconfig.json
```diff
@@ -1,14 +1,16 @@
...
     "declaration": true,
     "esModuleInterop": true,
     "jsx": "react-jsx",
     "rootDir": "./src",
     "outDir": "./dist",
     "skipLibCheck": true,
     "sourceMap": true,
-    "module": "commonjs",
-    "target": "es5"
+    "module": "esnext",
```
ericclemmons, line 12: So, TIL : https://stackoverflow.com/questions/41993811/understanding-target-and-module-in-tsconfig

`module` basically means `module.exports` or `import/export` or `import()`:
> https://github.com/microsoft/TypeScript/issues/24082

packages/react/tsconfig.json
```diff
@@ -1,14 +1,16 @@
...
+    "target": "es2015"
```
ericclemmons, line 13: `target` means the language features:
> https://stackoverflow.com/questions/41993811/understanding-target-and-module-in-tsconfig

Targeting evergreen browsers, this could even be `esnext` but I figure we'll standup more examples (CRA, Nuxt, etc.) to determine what that impact is.

But the less TS has to transpile, the faster our builds will be.

packages/react/tsup.config.cjs
```diff
@@ -0,0 +1,10 @@
+/**
+ * @type {import("tsup").Options}
+ */
+module.exports = {
```
ericclemmons, line 4: I did it this way for type safety vs. `tsup src/index.tsx --format cjs,esm ...`

patches/tsup+4.14.0.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/tsup/dist/index.js b/node_modules/tsup/dist/index.js
+index 1fd3ea9..8f84cf0 100644
+--- a/node_modules/tsup/dist/index.js
++++ b/node_modules/tsup/dist/index.js
+@@ -13633,6 +13633,7 @@ async function runEsbuild(options, { format, css }) {
+       jsxFactory: options.jsxFactory,
+       jsxFragment: options.jsxFragment,
+       sourcemap: options.sourcemap,
++      inject: ['./src/globals.js'],
```
ericclemmons, line 9: https://github.com/egoist/tsup/issues/390

packages/core/package.json
```diff
@@ -2,15 +2,17 @@
...
+  "module": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
-    "dist/*.{js,ts}",
-    "dist/**/*.{js,ts}"
+    "dist"
   ],
   "scripts": {
-    "build": "tsc",
+    "prebuild": "rimraf dist",
```
wlee221, line 13: Should we have a separate `yarn clean` script here? I find explicit clean scripts when I'm looking at module/build issues.

```suggestion
    "clean": "rimraf dist",
    "prebuild": "yarn clean",
```

patches/tsup+4.14.0.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/tsup/dist/index.js b/node_modules/tsup/dist/index.js
```
wlee221, line 1: TIL https://www.npmjs.com/package/patch-package

wlee221 (approved):
This is looking great! Thanks @ericclemmons. Just got some suggestions/questions.

CONTRIBUTING.md
```diff
@@ -19,7 +19,17 @@
 Internally, this content is served by a single, Next.js [optional catch all route](https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes):
 [`docs/src/pages/[[...slugs]].tsx`](docs/src/pages/[[...slugs]].tsx).
 
-## Next/React Development
+## `@aws-amplify/core`
+
+- `yarn core build` to rebuild for production.
```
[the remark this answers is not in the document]
ericclemmons: `yarn core dev` is the answer for local development (including watching).

(I also learned that the build tool may not always be the dev tool!)

package.json
ericclemmons: Exactly right @dbanksdesign – I want the majority of the packages/examples/docs/etc. to "just work" with `git clone` and `yarn install`.

The only exception is really `environments/`, and that's because they require an AWS account.

In the future, we can even avoid that by including `aws-exports.js` – we would only need to monitor those resources for usage/abuse.

packages/core/package.json
ericclemmons: You know, I went back-and-forth on this.  What I noticed was that it was tough to trace when something was being used and where. (`build:icons` is an example of this)

Having `clean` explicitly defined kinda felt like I was promoting it to a public API (e.g. `yarn react clean`) vs. having only `build|dev` being the public APIs.

I hope the indirection from `build => prebuild => clean` is clearer with `build => prebuild` 🙏 

(I don't have a strong opinion – I've done NPM scripts every which way since [per-env](https://github.com/ericclemmons/per-env))

packages/react/src/index.tsx
ericclemmons: Good question @dbanksdesign!  No, `import '@aws-amplify/ui-react/styles.css'` is still required.

What `import` does here is allow the bundler (`tsup`, `microbundle`, whatever) automatically output both JS & CSS with a single process. (Compared to having to watch CSS **and** JS separately)

This is consistent with how, say, an application is bundled with webpack: even though your app has JS, CSS, SVGs, PNGs, whatever, the bundler identifies all dependencies and outputs them based on your configuration.

ericclemmons (commented):
Responding to comments, but somehow it's a review? ¯\_(ツ)_/¯

packages/core/package.json
dbanksdesign: Is there a use-case to run _only_ a clean? I've realized whenever I'm doing a clean it's because I want to blow up the build artifacts and rebuild them, so just being part of `prebuild` makes sense to me. But if there are use cases for "clean and not build" I'm for it.

packages/react/package.json
ericclemmons: No, any cross-platform/framework dependencies (like the state machine, design tokens, or even specs (we ran into an issue with that!)) ideally lives in `@aws-amplify/ui-core`.

But you're right to ask about [`@aws-amplify/ui`](https://github.com/aws-amplify/amplify-js/tree/main/packages/amplify-ui).

`ui-core` would be a new package, but I don't think it makes sense to have.  `ui-core` is kinda our equivalent of `ui`, so you're right – maybe we rename & merge `ui-core` and `theme-base` into `ui`.

The only thing that I found weird is `aws-amplify` installs it by default:
https://github.com/aws-amplify/amplify-js/blob/da8ef238928fdcad7ab307350ffff8448eec5b26/packages/aws-amplify/package.json#L47

But I have no idea why, or see the need for it.

The more I type it out, the more it makes sense for `@aws-amplify/ui` to contain our core/shared assets and `@aws-amplify/ui-*` be framework-specific stuff.  `core` just seemed to be common convention in the ecosystem (e.g. `@babel/core`).

Let's queue this up for another cleanup task.

packages/core/package.json
ericclemmons: Great question. I don't have one.  Your use-case models my experience as well:

- "Build for production", which means a clean slate so we don't re-publish old artifacts.
- "Dev" _should not_ remove clean because I found out it breaks hot-reload when those references disappear.

I suppose there could be a reason to clean and do dev, but IME `yarn install` is basically our starting point for "setup and build everything, and I'll chose what part I'm working on".

PR #197, opened 2021-08-19
sorted translation keys and run all react strings through i18n

eddiekeller:
*Issue #, if available:*

n/a

*Description of changes:*

Sorted all translation strings alphabetically for better readability.
Wrapped all React strings in `I18n` translator.


packages/core/src/i18n/translations.ts
```diff
@@ -13,164 +13,163 @@
 
 export const dict = {
   de: {
-    'Sign In': 'Anmelden',
-    'Sign Up': 'Registrieren',
-    'Sign Out': 'Abmelden',
-    'Sign in to your account': 'Melden Sie sich mit Ihrem Account an',
-    Username: 'Benutzername',
```
ericclemmons, line 20: How'd you do this?  Did we want to leverage https://www.npmjs.com/package/prettier-plugin-sort-json to do this automatically?

packages/core/src/i18n/translations.ts
```diff
@@ -13,164 +13,163 @@
...
+    'Phone Number': 'Número de teléfono',
+    'Resend a Code': 'Reenviar un código',
+    'Resend Code': 'Mandar codigo otra vez',
+    'Send Code': 'Enviar código',
+    'Sign In': 'Iniciar sesíon',
+    'Sign in to your account': 'Iniciar sesíon',
+    'Sign Out': 'Desconectar',
+    'Sign Up': 'Regístrase',
+    Skip: 'Omitir',
+    Submit: 'Enviar',
```
ericclemmons, line 166: What do you think about `consistent` for prettier?
https://prettier.io/docs/en/options.html#quote-props

ericclemmons (approved):
This looks good @eddiekeller !

I had some questions about making these stylistic improvements automated. What do you think?

packages/core/src/i18n/translations.ts
eddiekeller: I just used a VS code extension. Sort JS Objects or something. Super simple.

packages/core/src/i18n/translations.ts
eddiekeller: Yeah, I like “consistent.” Let’s update our prettier to use that. I’ll make that change in a coming PR. Good suggestion!

PR #201, opened 2021-08-20
Select Primitive

zchenwei:
*Description of changes:*

This PR add a Select primitive and sharable tokens for form fields.


docs/src/pages/ui/primitives/select/index.page.mdx
```diff
@@ -0,0 +1,329 @@
...
+
+## Demo
+
+<Example>
+  <SelectDemo />
+</Example>
+
+## Usage
+
+Import the Badge primitive and styles.
```
jacoblogan, line 19: Import the Select primitive and styles.

docs/src/pages/ui/primitives/select/index.page.mdx
```diff
@@ -0,0 +1,329 @@
...
+:root {
+  --amplify-components-select-background-color: yellow;
+}
+/* OR */
+.amplify-select {
+  background-color: yellow;
+}
+```
+
+To replace the Badge styling, unset it:
```
jacoblogan, line 216: To replace the Select styling, unset it:

packages/react/src/primitives/types/select.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type SelectSize = 'small' | 'large';
+
+export type SelectVariation = 'filled' | 'flushed';
+
+export interface SelectProps
+  extends BaseComponentProps,
+    BaseStyleProps,
+    AriaProps {
+  name?: string;
```
reesscot, line 14: nit: would you please alphabetize these?

packages/react/src/primitives/Select/__tests__/Select.test.tsx
```diff
@@ -0,0 +1,203 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import { Select } from '../Select';
+import { IconExpandMore } from '../../Icon';
+import { ComponentClassNames } from '../../shared';
+
+describe('Select primitive test suite', () => {
```
reesscot, line 8: Can you please add a test that shows this select component can be [controlled](https://reactjs.org/docs/forms.html#the-select-tag)?

packages/react/src/primitives/Select/__tests__/Select.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+    expect(optionTwo.selected).toBeFalsy();
+    expect(optionThree.selected).toBeFalsy();
+
+    userEvent.selectOptions(select, '3');
+    expect(optionOne.selected).toBeFalsy();
+    expect(optionTwo.selected).toBeFalsy();
+    expect(optionThree.selected).toBeFalsy();
+  });
+
+  it('should be able to focus with tab if not be disabled', async () => {
```
reesscot, line 152: Same comment above about this already being default browser behavior.

packages/react/src/primitives/Select/__tests__/Select.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const select = (await screen.findByTestId(
+      'test-select'
+    )) as HTMLSelectElement;
+
+    expect(document.body).toHaveFocus();
+    userEvent.tab();
+    expect(select).toHaveFocus();
+  });
+
+  it('should not be able to focus with tab if be disabled', async () => {
```
reesscot, line 169: Same comment above about this already being default browser behavior.

packages/react/src/primitives/Select/__tests__/Select.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+    expect(optionTwo.selected).toBeTruthy();
+    expect(optionThree.selected).toBeFalsy();
+
+    userEvent.selectOptions(select, '3');
+    expect(optionOne.selected).toBeFalsy();
+    expect(optionTwo.selected).toBeFalsy();
+    expect(optionThree.selected).toBeTruthy();
+  });
+
+  it('should not be able to select an option if disabled', async () => {
```
reesscot, line 104: Alternatively you could just test that the the disabled attribute is passed through correctly to the select element. We probably don't need to test this, since disabled selects not being interactive is default behavior of the browser, not functionality we are adding.

reesscot (changes requested):
This is looking great. 🙌 Had some thoughts below regarding built-in accessibility and a some stylistic tweaks.

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <View
+        as="select"
+        required={isRequired}
+        data-size={size}
+        data-variation={variation}
+        className={classNames(ComponentClassNames.Select, className)}
+        {...rest}
+      >
+        {placeholder && (
+          <option value="" disabled>
```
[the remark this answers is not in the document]
reesscot: I wonder if we should leave off the placeholder option and let customers pass `<option value="">Choose a State</option>` themselves. This would give them more control over what they would like the value of the placeholder to me (in case they wanted to do something like `<option value="DEFAULT">Customer placeholder prompt</option>`. I don't think it would be a big loss to not have a placeholder prop, and in fact make the component more flexible.

docs/src/components/SelectPropControls.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+
+import { Flex, Select, SelectProps } from '@aws-amplify/ui-react';
+
+import { FieldLabeler } from './FieldLabeler';
+
+export interface SelectPropContorlsProps extends SelectProps {
```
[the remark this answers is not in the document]
zchenwei: Good catch!

docs/src/components/SelectPropControls.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Flex justifyContent="space-between">
+        <FieldLabeler id="size">
+          <Select
+            name="size"
+            id="size"
+            defaultValue=""
+            onChange={(event) =>
+              setSize(event.target.value as SelectProps['size'])
+            }
+            width="8rem"
```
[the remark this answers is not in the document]
zchenwei: It wouldn't. I also try `width="fit-content'`, looks like we have to specify a decent width explicitly :( 

This is because we need to respect the icon to the right, otherwise it will overlap with the content.

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -0,0 +1,49 @@
...
+    isRequired,
+    ...rest
+  } = props;
+  return (
+    <View className="amplify-select-wrapper">
+      <View
+        as="select"
+        required={isRequired}
+        data-size={size}
+        data-variation={variation}
```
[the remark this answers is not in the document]
zchenwei: Sounds great!

packages/react/src/primitives/Select/Select.tsx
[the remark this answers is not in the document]
zchenwei: Yea, that's a good idea. I mean the `placeholder` is optional. They can totally pass a custom one themselves.

packages/react/src/primitives/Select/__tests__/Select.test.tsx
zchenwei: Yup! Will do

packages/react/src/primitives/Select/__tests__/Select.test.tsx
zchenwei: For sure, let me trace back and remove those browser default behavior tests in favor of controlled component tests

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import { Flex } from '../Flex';
+import { View } from '../View';
+import { IconExpandMore } from '../Icon';
+import { SelectProps } from '../types';
+import { ComponentClassNames } from '../shared';
+
+export const Select: React.FC<SelectProps> = (props) => {
```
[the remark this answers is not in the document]
zchenwei: I like this idea! I would like to build the `SelectField` in a separate PR with your `<Label>` component out of box to avoid squeezing too much in this one because things like testing, docs may also need to update as well

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {...rest}
+      >
+        {placeholder && (
+          <option value="" disabled>
+            {placeholder}
+          </option>
+        )}
+        {children}
+      </View>
+      <Flex
```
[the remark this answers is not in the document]
zchenwei: Yes, exactly. It will let the click get through

docs/src/components/SelectPropControls.tsx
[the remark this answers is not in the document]
reesscot: Hmm, this sounds like a problem a customer would run into that we should fix at the Select component? Could we use extra `padding-inline-end` to prevent the option text from going underneath the icon?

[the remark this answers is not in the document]
zchenwei: Humm, i think `padding-inline-end` will have no difference with `padding`? To fully address it, I think we should use percentage for `padding` so that the padding for field could be adjusted according to the `width`. And the icon could use the same percentage for its `right`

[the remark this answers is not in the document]
zchenwei: Okay, it works. It will override the `padding-right` for select.

docs/src/pages/ui/primitives/select/index.page.mdx
```diff
@@ -110,7 +110,7 @@ import '@aws-amplify/ui-react/styles.css';
 
 ### Variation
 
-Select comes in 3 variants, `filled`, `flushed` and none(default). Pass the `variation` prop and set it to either of these values.
+Select comes in 2 variants, `flushed` and none(default). Pass the `variation` prop and set it to either of these values.
```
reesscot, line 113: I think we decided in the meeting recently to change `flushed` to `quiet` to be more semantic.

packages/theme-base/src/css/component/_field.scss
```diff
@@ -29,6 +29,16 @@
     line-height: var(--amplify-components-field-large-line-height);
   }
 
+  &[data-variation='flushed'] {
```
reesscot, line 32: Thanks for moving this! I'm moving to my file now too

packages/react/src/primitives/Select/__tests__/Select.test.tsx
```diff
@@ -101,14 +127,9 @@ describe('Select primitive test suite', () => {
...
-  it('should not be able to select an option if disabled', async () => {
+  it('should be able to select an option as an uncontrolled component', async () => {
     render(
-      <Select
-        testId={testId}
-        placeholder={placeholder}
-        defaultValue=""
-        isDisabled
-      >
+      <Select testId={testId} placeholder={placeholder} defaultValue="">
```
reesscot, line 132: For the uncontrolled component we should confirm that passing `defaultValue="2"` actually sets that value instead of the placeholder. Should we add a test variation that sets the defaultValue to 2?

docs/src/pages/ui/primitives/select/index.page.mdx
```diff
@@ -145,17 +139,51 @@ import '@aws-amplify/ui-react/styles.css';
...
+
+// use the placeholder out of box
+<Select defaultValue="" placeholder="Please select a fruit">
+  <option value="apple">Apple</option>
+  <option value="banana">Banana</option>
+  <option value="orange">Orange</option>
+</Select>;
+
+// or pass a custom placeholder
+<Select defaultValue="">
```
reesscot, line 170: Would anything be selected here since the `defaultValue`is not DEFAULT?

docs/src/pages/ui/primitives/select/index.page.mdx
```diff
@@ -145,17 +139,51 @@ import '@aws-amplify/ui-react/styles.css';
...
+
+```jsx
+import { Select } from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+
+import { Select } from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+
+// use the placeholder out of box
+<Select defaultValue="" placeholder="Please select a fruit">
```
reesscot, line 163: Is it necessary to always specify the defaultValue prop? I think if you leave it off it will still be an uncontrolled component, but I'm not sure if the placeholder will be selected

packages/theme-base/src/tokens/components/select.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+  width: { value: '100%' },
+  transitionProperty: { value: 'background-color' },
+  transitionDuration: { value: '{time.transition.medium.value}' },
+  //variation
+  flushed: {
```
reesscot, line 6: Shouldn't these move to field.js?

packages/theme-base/src/tokens/components/select.js
```diff
@@ -0,0 +1,28 @@
...
+    borderTop: { value: 'none' },
+    borderLeft: { value: 'none' },
+    borderRight: { value: 'none' },
```
reesscot, line 9: We should try to use logical border properties (borderInlineStart/End etc) instead of borderLeft/borderRight etc

packages/theme-base/src/tokens/components/select.js
```diff
@@ -0,0 +1,28 @@
...
+    borderRadius: { value: '0' },
+    _focus: {
+      boxShadow: { value: '0 4px 2px -2px hsl(220, 50%, 50%)' },
```
reesscot, line 12: Please see my branch for focus border color and box shadow variables

docs/src/pages/ui/primitives/select/index.page.mdx
zchenwei: I forgot to update the value here but the first option will always be selected if no matching value found

docs/src/pages/ui/primitives/select/index.page.mdx
zchenwei: The `defaultValue` is used by uncontrolled component to initialize I think. For controlled component, it should always use `value`, the `defaultValue` doesn't work. If we leave it off, then it will select the first option by default

packages/react/src/primitives/Select/__tests__/Select.test.tsx
zchenwei: I don't think it is necessary? I think the `defaultValue` is a behavior of React select for uncontrolled component to initialize. At least, I don't think `defaultValue` will work when `value` present at the same time.

packages/theme-base/src/tokens/components/select.js
zchenwei: I am curious how to use the `borderInlineStart/End ` in this case? Looks like there is no `border-inline-top`? :(

packages/theme-base/src/tokens/components/select.js
zchenwei: The `boxShadow` for quiet may be a slightly different but  I will refer the color there

PR #188, opened 2021-08-17
Migrate Amplify JS UI Packages To Amplify UI Repo

ErikCH:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200398964162505/1200435140994537/f)

*Description of changes:*
Move all UI related packages over from [Amplify JS](https://github.com/aws-amplify/amplify-js) repo.

- [x] Move over all UI packages from Amplify JS with git history
- [ ] Verify all packages can build, and tests pass
- [ ] Add build scripts that can be ran from root of workspace
- [x] Combine @aws-amplify/ui-react (legacy and new Authenticator) into one package

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #204, opened 2021-08-23
build(angular): prepare angular package for deployment

wlee221:
*Issue #, if available:*

*Description of changes:* This pr makes minor changes to prepare angular package for deployment. 

- [x] Add explicit `@aws-amplify/ui-angular` dependency to `example-angular`
- [x] Build `@aws-amplify/ui-angular` in production mode
- [x] Distribute `theme.css` from `dist/ui-angular`.
- [x] Declare workspace for actual `@aws-amplify/ui-angular` package, which is at subdirectory `packages/angular/projects/ui-angular`. Details in comments
- [x] Various typos/build fixes I found on the way

 
## Angular dev structure

```bash
packages/angular/ 
├── projects/ui-angular/ # where the actual `@aws-amplify/ui-angular` library lives
│   ├── src/ # contains src files for actual library code.
│   └── package.json # name: `@aws-amplify/ui-angular`, private: false
│                    # contains dependencies that library needs (e.g. xstate)
│
├── dist/ui-angular/
│   └── bundles/
│   └── esm2015/
│   └── fesm2015/
│   └── lib/
│   └── package.json # same as the above, but defines extra fields like
│                    # main, module, esm2015, typings, sideEffects.
│ 
└── package.json # name: 'amplify-ui-angular-mono`, private: true
                 # contains base dependencies for developing angular packages.
                 # This is only for infrastructure purpose! Shouldn't be
                 # published or be referenced by other dependencies.
```


examples/angular/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "private": true,
   "dependencies": {
+    "@aws-amplify/ui-angular": "0.0.1",
```
wlee221, line 16: Adding explicit dependency to `@aws-amplify/ui-angular` here. Will be bumped to "2.0.1" on changesets pr later this week.

examples/angular/src/pages/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.component.ts
```diff
@@ -2,19 +2,20 @@ import { Component, NgZone, OnInit } from '@angular/core';
...
-});
 
 @Component({
   selector: 'sign-up-with-phone',
   templateUrl: 'sign-up-with-phone.component.html',
 })
 export class SignUpWithPhoneComponent implements OnInit {
-  constructor() {}
+  constructor() {
+    Amplify.configure({
```
wlee221, line 12: Typo: this configure needs to be inside constructor to ensure `Amplify` is only configured when this component renders.

examples/angular/tsconfig.app.json
```diff
@@ -6,5 +6,5 @@
     "types": []
   },
   "files": ["src/main.ts", "src/polyfills.ts"],
-  "include": ["src/**/*.d.ts", "../../packages/core"]
```
wlee221, line 9: "@aws-amplify/ui-core" no longer needs to be explicitly included now that it's getting built with `tsup` 👍

package.json
```diff
@@ -27,7 +25,8 @@
     "angular-docs",
     "environments",
     "examples/*",
-    "packages/*"
+    "packages/*",
+    "packages/angular/projects/ui-angular"
```
wlee221, line 29: This subdirectory, not `packages/angular`, contains the actual library code we want to publish. `packages/angular` is more of Angular workspace root to provide infrastructure for Angular development. Relevant docs: [link](https://angular.io/guide/creating-libraries#getting-started).

packages/angular/package.json
```diff
@@ -41,13 +40,15 @@
...
   "workspaces": {
     "nohoist": [
       "@angular-devkit/build-angular"
     ]
   },
-  "typings": "./dist/ui-angular/index.d.ts"
+  "typings": "./dist/ui-angular/index.d.ts",
+  "publishConfig": {
+    "directory": "dist/ui-angular"
+  }
```
wlee221, line 53: Angular packager builds the package to publish in `dist/ui-angular` and not at its root. This is quite an anti-pattern to how most monorepo deployment tools publish, and you can find related conversation in https://github.com/lerna/lerna/issues/1282. 

Fortunately both lerna and changesets support subdirectory publish: https://github.com/atlassian/changesets/pull/428, though I haven't tested this out yet.

examples/angular/tsconfig.json
```diff
@@ -15,8 +15,7 @@
     "lib": ["es2018", "dom"],
     "esModuleInterop": true,
     "paths": {
-      "@aws-amplify/ui-angular": ["../../packages/angular/dist/ui-angular/*"],
-      "@aws-amplify/ui-core": ["../../packages/core/*"],
+      "@aws-amplify/ui-angular": ["../../packages/angular/dist/ui-angular"],
```
wlee221, line 18: Due to how angular outputs file to this subdirectory, this has to be explicitly stated through paths (else it'd go to the source files).

packages/angular/projects/ui-angular/src/index.ts
```diff
@@ -7,5 +7,6 @@ export * from './lib/common/types';
 
 export * from './lib/components';
 export * from './lib/primitives';
-export * from './lib/services';
+export * from './lib/services/authenticator-context.service';
+export * from './lib/services/state-machine.service';
```
wlee221, line 11: TIL: for production builds, services should be explicitly imported instead of index.ts else it outputs an circular dependency. Related stackoverflow -- weird behavior, but this fixed production build issues.

examples/angular/src/pages/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.component.ts
ErikCH: Yeah, I noticed this too, on a couple of examples where I had to move it into the constructor. I'm not sure why

examples/angular/tsconfig.json
ErikCH: Have you tried setting the package.json `main` property to the dist folder output file? 

IIRC if you set your `package.json` to the correct `main` property, then the `@aws-amplify/ui-angular` import will default to that path not the src folder (even apps in the yarn workspace). So you wouldn't have too explicitly set this in the `tsconfig` file.

wlee221: Yeah, just tried it now! You were right, that correctly routes the imports to `dist`. But then it was having trouble locating "@aws-amplify/ui-angular/theme.css", because it's not part of the `main` file. 

This is frustrating because `dist/ui-angular` has different `package.json` from `lib/ui-angular`, with its own "main", "module", etc entries so workspaces aren't working as intended. If they were in the same directory, redirecting files through workspaces would've been much easier... But I think we need to use `path` so that it finds the correct directory to reference its package.json.

We can discuss this offline as needed.

wlee221:
Looking over the failing angular test. Looks like it's having trouble registering both legacy and current angular components in production build.

examples/angular/tsconfig.json
ErikCH: OK!

ErikCH (approved):

examples/angular/src/pages/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.component.ts
ericclemmons: My understanding is, because it's a single-page app that's executing all the imported code, `Amplify.configure` would be called as a side-effect of each component, creating tons of conflicts.

Instead, having it within `constructor` at least makes this "refresh-friendly" so that we configure it _once_ per component (assuming 1 component is rendered at a time)

examples/angular/src/pages/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.component.ts
```diff
@@ -2,19 +2,20 @@ import { Component, NgZone, OnInit } from '@angular/core';
...
+      ...awsExports,
+      auth: {
+        login_mechanisms: ['phone_number'],
```
ericclemmons, line 15: I learned yesterday that this structure looks more like:

```js
/* eslint-disable */
// WARNING: DO NOT EDIT. This file is automatically generated by AWS Amplify. It will be overwritten.

const awsmobile = {
    "aws_project_region": "us-east-1",
    "aws_cognito_identity_pool_id": "us-east-1:ea9c8067-5fdc-4203-b2e7-555960e0df2d",
    "aws_cognito_region": "us-east-1",
    "aws_user_pools_id": "us-east-1_FVBg0Ieqr",
    "aws_user_pools_web_client_id": "2p6k4ofki50lg0196r12sniij9",
    "oauth": {},
    "aws_cognito_login_mechanism": [],
    "aws_cognito_signup_attributes": [
        "PHONE_NUMBER",
        "EMAIL",
        "PREFERRED_USERNAME"
    ],
    "aws_cognito_mfa_configuration": "OFF",
    "aws_cognito_mfa_types": [
        "SMS"
    ],
    "aws_cognito_password_protection_settings": {
        "passwordPolicyMinLength": 8,
        "passwordPolicyCharacters": [
            "REQUIRES_LOWERCASE",
            "REQUIRES_NUMBERS",
            "REQUIRES_SYMBOLS",
            "REQUIRES_UPPERCASE"
        ]
    }
};


export default awsmobile;
```

Nothing to change, but just an FYI for when we start supporting it.

packages/angular/projects/ui-angular/ng-package.json
```diff
@@ -1,13 +1,20 @@
...
     "umdModuleIds": {
       "xstate": "xstate",
       "aws-amplify": "aws-amplify",
-      "@aws-amplify/core": "@aws-amplify/core",
-      "@aws-amplify/ui-core": "@aws-amplify/ui-core"
+      "@aws-amplify/core": "@aws-amplify/core"
     }
-  }
+  },
+  "allowedNonPeerDependencies": [
```
ericclemmons, line 13: What does this mean or do?  (It looks like we have to restate our dependencies here. What happens if we forget one?)

packages/angular/package.json
ericclemmons: An alternative that we can try is instead doing:

```json
"publishConfig": {
  "main": "dist/ui-angular/main.js",
  "module": "dist/ui-angular/module.js",
  ...
}
```

packages/angular/projects/ui-angular/src/index.ts
ericclemmons: I agree.  I think we should also have a clean-up task to remove all `export * from` statements and only have very explicit exports.

It's more verbose, but it's an O(n) cost.

packages/angular/projects/ui-angular/src/index.ts
```diff
@@ -7,6 +7,19 @@ export * from './lib/common/types';
...
+  AmplifyS3Text,
+  AmplifyS3TextPicker,
+  AmplifyS3Image,
+  AmplifyPicker,
+  AmplifyPhotoPicker,
+  AmplifyChatbot,
+  AmplifyToast,
+  AmplifyInput,
+  AmplifyButton,
+} from './lib/proxies';
```
ericclemmons, line 25: Is this the same code that @ErikCH has for #188 ?

package.json
ericclemmons: What's the primary difference between how Angular works in this repo vs. https://github.com/aws-amplify/amplify-js/tree/main/packages/aws-amplify-angular?

examples/angular/tsconfig.json
ericclemmons: This is how I got `@aws-amplify/ui-react/styles.css` to resolve correctly:

https://github.com/aws-amplify/amplify-ui/blob/47255d081883cd42c8b11c6198c9175139aa8323/packages/react/package.json#L8-L11

examples/angular/src/pages/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.component.ts
slaymance: This is interesting. @ErikCH you noticed this strictly for the Angular examples? I almost kind of like this because I could imagine a way to have a common frontend implementation where we dynamically pass in aws-exports for different environments as needed by e2e tests. Just spitballing, I don't know if that would actually work. It would clean up our examples a lot though since they're all mostly just a simple `Authenticator` component.

ericclemmons (commented):
Only some comments to aid my understanding & share some of my research as well.

This is tricky!  Angular seems to be different today than what I've seen in https://github.com/aws-amplify/amplify-js/tree/main/packages/aws-amplify-angular.

package.json
wlee221: 1) We're now using [`ng-packagr`](https://github.com/ng-packagr/ng-packagr) + `@angular/cli` to build packages. This outputs more bundles than we used to according to [Angular Package Format](https://github.com/ng-packagr/ng-packagr).

2) We used to build our packages somewhat [manually](https://github.com/aws-amplify/amplify-js/blob/main/packages/aws-amplify-angular/package.json#L10-L14), with explicit configurations. Now these are being done automatically with `ng build`.

3) On our legacy component, we declared most of Angular dependency as [`devDependency`](https://github.com/aws-amplify/amplify-js/blob/main/packages/aws-amplify-angular/package.json#L29-L68). Angular package in this repo does this by a mini workspace structure -- these angular infrastructure are defined in `packages/angular` root `package.json`.  The actual library code is defined in `packages/angular/projects/ui-angular` subdirectory. Its [`package.json`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/angular/projects/ui-angular/package.json#L5-L6) is free of these angular infrastructure because they are hoisted are at its parent.

wlee221: To visualize this, 
```bash
packages/angular/
├── projects/ui-angular/
│   ├── src/ # contains src files for actual library code.
│   └── package.json # name: `@aws-amplify/ui-angular`, private: false
│                        # contains dependencies that library needs (e.g. xstate)
│
└── package.json # name: 'amplify-ui-angular-mono`, private: true
                 # contains base dependencies for developing angular packages

packages/angular/package.json
wlee221: Right, the problem is that this is not the `package.json` we want to pack for publish as mentioned in the comment above. We are interested in publishing `dist/ui-angular/package.json`, which has the library-specific dependencies.

wlee221: I can do a quick demo what the structure looks like, since this structure is pretty tricky.

packages/angular/projects/ui-angular/src/index.ts
wlee221: +1, especially at our root/index.ts file.

examples/angular/src/pages/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.component.ts
wlee221: Cool!

packages/angular/projects/ui-angular/ng-package.json
wlee221: Angular recommends using `peerDependency` [whenever possible](https://github.com/ng-packagr/ng-packagr/blob/master/docs/dependencies.md#general-recommendation-use-peerdependencies-whenever-possible). To add non peer-dependencies, they require we add those to this array. If we forget to add one, `ng build` will complain and fail the build.

While I agree with their approach, dependencies like `xstate` should be `dependencies` and shouldn't need to be added by customers, which is why I added it here.

packages/angular/projects/ui-angular/src/index.ts
wlee221: Yeah, this just makes the exports more explicit because production build was complaining with just `export * from './lib/proxies'`.

examples/angular/tsconfig.json
wlee221: Ah, thanks! I'll add that to my package.json. 

But from dev side, this still has the same problem of divided subdirectories vs workspaces:

Reference to `@aws-amplify/angular` goes to `packages/angular/projects/ui-angular` as defined by workspaces, but this should really head to `packages/angular/dist/ui-angular` which has the auto-generated `package.json` that defines all the module declarations. This is why I see these structure as anti pattern, it works poorly against workspaces. But community has solved these nuances with `path` and `publishConfig.directory`

wlee221:
I'll handle the slow `ng start` problem in another pr, and see if I can make optimization cross-platform.

PR #211, opened 2021-08-24
chore(angular): use serve to serve built package

wlee221:
*Issue #, if available:*

*Description of changes:* This uses `serve` npm package to serve the built `next-example/dist`, so that `yarn build` does not build the same package twice.

Looked into using this in Vue/React as well, but their `yarn start` command were already using the `dist` directory as intended.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,7 +91,7 @@ jobs:
         run: yarn workspace ${{ matrix.framework }}-example build
 
       - name: Start ${{ matrix.framework }} example
-        run: yarn workspace ${{ matrix.framework }}-example start & npx wait-on -c waitOnConfig.json -t 120000 http-get://localhost:3000/ui/components/authenticator/sign-in-with-username
+        run: yarn workspace ${{ matrix.framework }}-example start & npx wait-on -c waitOnConfig.json -t 20000 http-get://localhost:3000/ui/components/authenticator/sign-in-with-username
 
       - name: Run E2E tests against ${{ matrix.framework }} example
         run: yarn workspace e2e test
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -91,7 +91,7 @@ jobs:
         run: yarn workspace ${{ matrix.framework }}-example build
 
       - name: Start ${{ matrix.framework }} example
-        run: yarn workspace ${{ matrix.framework }}-example start & npx wait-on -c waitOnConfig.json -t 120000 http-get://localhost:3000/ui/components/authenticator/sign-in-with-username
+        run: yarn workspace ${{ matrix.framework }}-example start & npx wait-on -c waitOnConfig.json -t 20000 http-get://localhost:3000/ui/components/authenticator/sign-in-with-username
```
wlee221, line 94: Since all the [framework]-example packages are directly consuming `dist`, we can shorten this to 20 seconds, or less. But I think 20 seconds is short enough to catch anything unexpected during serving without getting intermittent timeouts.

examples/angular/package.json
```diff
@@ -42,6 +42,7 @@
     "karma-jasmine": "~4.0.0",
     "karma-jasmine-html-reporter": "~1.5.0",
     "protractor": "~7.0.0",
+    "serve": "^12.0.0",
```
wlee221, line 45: http-server also works, but I found `serve` to be a bit easier for SPA routed apps. http-server needed a few more flags than above to work.

slaymance (approved):
Thanks for this, William! Btw, this is over 13% performance improvement for our Angular e2e tests in CI. Pretty substantial!

PR #310, opened 2021-08-27
Vue: Add Way To Build Package

ErikCH:
*Issue #, if available:* [Asana](https://app.asana.com/0/0/1200663219572565/f)

*Description of changes:*
Created build for Vue and setup tests with Jest


packages/core/src/types/authMachine.ts
```diff
@@ -33,6 +33,7 @@ export interface SignUpContext {
   login_mechanisms?: string[];
   intent?: string;
   authAttributes?: Record<string, any>;
+  challengeName?: string;
 }
```
ErikCH, line 37: When I compiled my Vue, it had errors that the types weren't correct for some of these contexts. I had to add these so it would compile. Since they are all optional it should be OK.

package.json
```diff
@@ -32,7 +32,9 @@
     "@mdx-js/mdx/**/trim": "^0.0.3",
     "fs-extra": "^10.0.0",
     "next": "^11.0.1",
-    "ws": "^7.4.6"
+    "ws": "^7.4.6",
+    "jest": "^26.6.3",
+    "ts-jest": "^26.5.6"
```
ErikCH, line 37: Vue test utils doesn't work with Jest 27, so I had to lock this at 26. The React tests seems to still work with it, so this should be fine.

packages/vue/package.json
```diff
@@ -1,13 +1,46 @@
...
+    "dist"
+  ],
   "scripts": {
-    "build": "echo \"Skipped\"",
-    "test": "echo \"Skipped: no test specified\""
+    "prebuild": "rimraf dist",
+    "build": "vite build",
+    "dev": "vite build --watch",
+    "clean": "rimraf dist",
+    "test": "jest __tests__"
```
reesscot, line 20: Maybe add a test:watch command as well?

ErikCH: OK! I'll add the same ones as React

packages/vue/src/components/sign-up.vue
```diff
@@ -16,9 +16,6 @@
           <user-name-alias />
           <sign-up-password-control />
           <sign-up-confirm-password-control />
-          <base-box data-ui-error v-if="error">
-            {{ error }}
-          </base-box>
```
wlee221, line 21: Why was this removed?

packages/core/src/types/authMachine.ts
wlee221: This is probably on me! But I prefer we don't add `challengeName` to other types, as this context is only necessary in signIn. Where have you found these errors?

wlee221 (commented):
Thanks @ErikCH! Just got few nits, but changes look good to me otherwise.

packages/vue/src/components/sign-up.vue
ErikCH: This was derived from sold old code that was on line 150 for a watcher. With the new redesign this actually was broke, I believe. There is already an error tracked elsewhere on this page, so I don't think we need 2 of them.

wlee221 (approved):

packages/vue/src/components/sign-up.vue
ErikCH: For now, let's leave the `challengeName` in the TypeScript file. For some reason the plugin I'm using likes it. Let's set a follow up task to troubleshoot it for the future.

packages/core/src/types/authMachine.ts
wlee221: Resolving as discussed, we'll tackle this later.

PR #315, opened 2021-08-30
Password Field & Field Group

reesscot:
This is a Draft PR implementation of PasswordField and FieldGroup. It also makes changes to TextField to allow a component to be inserted to the right (logical "end") of the input field.

Showing Firefox password generation support:

Showing LastPass support for `current-password` autocomplete field:
Screen Shot 2021-09-01 at 10 18 40 AM

TODO:

- [x] Unit tests
- [x] Write docs


packages/react/src/primitives/PasswordField/PasswordField.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const PasswordField: React.FC<PasswordFieldProps> = ({
+  label,
+  className,
+  hideShowPassword = false,
+  ...rest
+}) => {
+  const [type, setType] = React.useState<TextFieldType>('password');
+
+  const showPasswordOnClick = React.useCallback(() => {
+    console.log('hello from showpassword click', type);
```
dbanksdesign, line 18: nit: remove?

packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import { ShowPasswordButtonProps } from '../types/';
+import { Button } from '../Button';
+import { ComponentClassNames } from '../shared/constants';
+import { IconVisibility } from '../Icon';
+import { Text } from '../Text';
+import classNames from 'classnames';
+
+export const ShowPasswordButton: React.FC<ShowPasswordButtonProps> = ({
+  onClick = () => {},
```
dbanksdesign, line 10: Should we use React's `MouseEventHandler` type here?

packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+}) => {
+  return (
+    <Button
+      className={classNames(
+        ComponentClassNames.FieldShowPassword,
+        ComponentClassNames.FieldGroupField
+      )}
+      onClick={onClick}
+    >
+      <Text className="sr-only">Show password</Text>
```
dbanksdesign, line 20: What is the benefit from having screen reader only text and an empty alt tag over having this in an alt tag?

packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <IconVisibility size="large" alt="" />
```
dbanksdesign, line 21: Should we swap icons when it is visible or invisible? `<IconVisibilityOff>`

packages/react/src/primitives/TextField/TextField.tsx
```diff
@@ -59,25 +62,29 @@ export const TextField: React.FC<TextFieldProps> = ({
...
+          onCut={onCut}
+          onInput={onInput}
+          onPaste={onPaste}
+          onSelect={onSelect}
+          size={size}
+          type={type}
+          value={value}
+          {...rest}
+        />
+        {inputEndComponent}
```
dbanksdesign, line 86: For balance sake, should we also add `inputStartComponent`? Also, would this work if `inputEndComponent` is actually multiple React nodes?

packages/react/src/primitives/TextField/TextField.tsx
```diff
@@ -26,6 +27,8 @@ export const TextField: React.FC<TextFieldProps> = ({
   justifyContent,
   label,
   labelHidden = false,
+  hideShowPassword = true,
```
dbanksdesign, line 30: Do we want to have this on the TextField rather than just the PasswordField?

dbanksdesign (commented):
Just a few questions/comments, overall looks good though!

packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
reesscot: Yes, I forgot to do this. Will update.

packages/react/src/primitives/TextField/TextField.tsx
reesscot: Yes, I think we should add inputStartComponent as well.

Do you have an example of where we would want more than one end component? I see that you can do something similar with [bootstrap inputgroups](https://getbootstrap.com/docs/4.0/components/input-group/#multiple-addons). Perhaps we could take a similar approach to support multiple nodes.

packages/react/src/primitives/TextField/TextField.tsx
reesscot: Yes, I must have forgotten to remove it. This one doesn't do anything anymore.

packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
reesscot: Good call. I think I should actually move "Show password" to the Button as an `aria-label`, and should toggle it between "hide password" and "show password"

reesscot: https://gomakethings.com/icon-accessibility-and-aria-label/

packages/react/src/primitives/PasswordField/PasswordField.tsx
reesscot: Done

packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
reesscot: I don't think this default value is needed, so I've removed it. The `onClick` prop is already typed as `MouseEventHandler`

packages/react/src/primitives/TextField/TextField.tsx
reesscot: Updated to add `inputStartComponent`, and moved logic into the `FieldGroup` component.

packages/theme-base/src/css/component/button.scss
```diff
@@ -3,6 +3,7 @@
  */
 
 .amplify-button {
+  display: inline-block;
```
reesscot, line 6: This is the default browser display for buttons, which was unset to inline by our `all: unset` rule.

packages/theme-base/src/css/component/text.scss
```diff
@@ -1,39 +1,47 @@
 .amplify-text {
   color: var(--amplify-components-text-color);
   display: block;
-  
+
+  @at-root b#{&},
```
reesscot, line 5: This is to fix the issue where `<Text as="span">` is being rendered as a block element. This results in a rule like:
```
b.amplify-text, em.amplify-text, etc... { display: inline; }
```

docs/src/pages/ui/primitives/passwordfield/demo.tsx
```diff
@@ -0,0 +1,257 @@
...
+};
+
+export const LoginFormExample = () => (
+  <Flex as="form" direction="column" gap="1rem">
+    <Heading level={3}>Login</Heading>
+    <TextField label="Username" name="username" autoComplete="username" />
+    <PasswordField
+      label="Password"
+      name="password"
+      autoComplete="current-password"
```
dbanksdesign, line 146: 👀

docs/src/pages/ui/primitives/passwordfield/demo.tsx
```diff
@@ -0,0 +1,257 @@
...
+      }
+      name="password"
+      isRequired={true}
+    />
+    <Button type="submit">Submit</Button>
+  </Flex>
+);
+
+export const PasswordFieldStyledPropsExample = () => (
+  <TextField
```
dbanksdesign, line 244: Should this be `<PasswordField`?

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -0,0 +1,345 @@
+import {
+  Flex,
+  PasswordField,
+  TextField,
+  Button,
+  Heading,
+} from '@aws-amplify/ui-react';
+import {
+  PasswordFieldDemo,
+  LoginFormExample,
```
dbanksdesign, line 10: Minor, but maybe a pattern we should start doing for the docs: should we split the demo code and examples? It might make the code a bit more organized?

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -0,0 +1,345 @@
...
+  SignUpFormExample,
+  ChangePasswordFormExample,
+  DescriptiveTextExample,
+  RequiredPasswordFieldExample,
+  PasswordFieldStyledPropsExample,
+} from './demo';
+import { Example } from '@/components/Example';
+
+The PasswordField primitive can be used allow users to input passwords.
```
dbanksdesign, line 19: Minor: 'The PasswordField primitive allows users to input passwords'. Maybe we could also add something about how there is the show/hide button and that it works with password managers too?

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+```jsx
+<PasswordField label="Password" name="password" hideShowPassword={true} />
+```
+
+<PasswordField label="Password" name="password" hideShowPassword={true} />
+
+### Autocomplete - supporting password managers
+
+Use the `autoComplete` prop to tell browser how to populate a password field. Options for `autoComplete` are `current-password` (default) and `new-password`.
```
dbanksdesign, line 52: TIL! Nice!

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+#### How to choose an `autoComplete` option
```
dbanksdesign, line 54: Nit: we might not need this header?

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+If customers are logging in, you don't need to do set `autoComplete` because `current-password` is already the default.
+If customers are signing up for a new account, set `autoComplete` to `new-password`.
+
+See [MDN for more information on autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete).
```
dbanksdesign, line 59: Love linking out to MDN for this stuff! We should do that more!

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -0,0 +1,345 @@
...
+</Flex>
+```
+
+<Example>
+  <LoginFormExample />
+</Example>
+
+_Change password example: `current-password` and `new-password`_
+
+```jsx
```
dbanksdesign, line 106: Just thinking out loud here... I wonder if there is a way we could keep the example code in the codeblock shown to users in sync with the actual example code? If we needed to make a change to this we would need to do it in 2 files. Not a huge deal or anything, just thoughts...

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -0,0 +1,345 @@
...
+</Flex>
+```
+
+<Example>
+  <ChangePasswordFormExample />
+</Example>
+
+### Accessibility / Label behavior
+
+The PasswordField primitive is accessible by default. A matching `label` HTML element is rendered with the `PasswordField` by default,
```
dbanksdesign, line 137: The first 2 sentences both say 'by default', maybe we can combine them or something?

dbanksdesign (approved):
A couple minor comments about documentation, but other than that looks great!

docs/src/pages/ui/primitives/passwordfield/demo.tsx
reesscot: Whoops, how did I leave that behind? 🤦‍♂️

docs/src/pages/ui/primitives/passwordfield/react.mdx
reesscot: That's a great idea, I'll move them to an `examples.ts` file

docs/src/pages/ui/primitives/passwordfield/react.mdx
reesscot: Good call, will update

docs/src/pages/ui/primitives/passwordfield/react.mdx
reesscot: Agreed, I'll remove it

docs/src/pages/ui/primitives/passwordfield/react.mdx
reesscot: Yeah, we need a good solution for this. I'll discuss with @slaymance , I think they solved this for the Authenticator docs.

docs/src/pages/ui/primitives/passwordfield/react.mdx
reesscot: Default default default, just trying to get my point across 🤣. Will update.

dbanksdesign (approved):
LGTM!

PR #317, opened 2021-08-31
SwitchField

jacoblogan:
SwitchField form primitive.  This pull request adds the SwitchField form primitive implementation, documentation, and tests.

switchField


reesscot:
Needs rebase/merge from `main` before getting reviewed.

packages/ui/src/theme/css/component/switchField.scss
```diff
@@ -0,0 +1,70 @@
...
+  transition-duration: var(
+    --amplify-components-switchfield-track-transition-duration
+  );
+  background: var(--amplify-components-switchfield-track-background);
+  &[data-checked] {
+    background: var(--amplify-components-switchfield-track-checked-background);
+  }
+  &[data-disabled] {
+    opacity: var(--amplify-components-switchfield-disabled-opacity);
+    cursor: default;
```
reesscot, line 42: Should this be `cursor: not-allowed`?

packages/ui/src/theme/css/component/switchField.scss
```diff
@@ -0,0 +1,70 @@
...
+    flex-direction: column-reverse;
+  }
+}
+
+.amplify-switch-track {
+  display: inline-flex;
+  justify-content: flex-start;
+  box-sizing: content-box;
+  cursor: pointer;
+  border-radius: 9999px;
```
reesscot, line 29: Should this be a CSS variable in case customers want to change it?

packages/ui/src/theme/css/component/switchField.scss
```diff
@@ -0,0 +1,70 @@
...
+  width: var(--amplify-components-switchfield-thumb-width);
+  height: var(--amplify-components-switchfield-thumb-width);
+  border-color: var(--amplify-components-switchfield-thumb-border-color);
+  overflow-wrap: break-word;
+  cursor: pointer;
+  &[data-checked] {
+    transform: var(--amplify-components-switchfield-thumb-checked-transform);
+  }
+  &[data-disabled] {
+    cursor: default;
```
reesscot, line 64: Should this be cursor: not-allowed?

packages/ui/src/theme/css/component/switchField.scss
```diff
@@ -0,0 +1,70 @@
+.amplify-switchfield {
```
reesscot, line 1: It feels like more of these should be CSS variables, but I'm unsure given how specific they are to our Switch design... But perhaps they still should be in case we want to override them at a theme level. Any thoughts @hvergara @dbanksdesign?

packages/react/src/primitives/types/switchField.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export type labelPositions = 'start' | 'end' | 'top' | 'bottom';
+export interface SwitchFieldProps
+  extends InputProps,
+    FieldProps,
+    BaseStyleProps {
+  /**
+   * If isChecked is provided, this will be a controlled SwitchField
+   */
+  isChecked?: boolean;
```
reesscot, line 15: nit: alphabetize

packages/ui/src/theme/css/component/switchField.scss
```diff
@@ -0,0 +1,70 @@
...
+  &[data-label-position='end'] {
+    flex-direction: row-reverse;
+  }
+  &[data-label-position='top'] {
+    flex-direction: column;
+  }
+  &[data-label-position='bottom'] {
+    flex-direction: column-reverse;
+  }
+}
```
reesscot, line 22: I think there needs to be more padding between the label and the switch itself. Not sure where to comment on that, just noticed it from the docs.
switch

reesscot (changes requested):
Looking good, just some questions below

packages/react/src/primitives/SwitchField/useSwitch.ts
```diff
@@ -0,0 +1,32 @@
+import React, { ChangeEvent, useCallback, useState } from 'react';
+
+export const useSwitch = (props) => {
+  const { onChange, isChecked, defaultChecked, isDisabled } = props;
+  const isControlled = typeof isChecked !== 'undefined';
+  const [isOn, setIsOn] = useState(isControlled ? isChecked : defaultChecked);
```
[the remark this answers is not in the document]
jacoblogan: We could switch this to isChecked but then we would have to rearrange the prop that is currently being used for isChecked (this is the prop that turns this into a controlled input and defines the state of the checkbox).  Because  the current usage of isChecked is being exposed to the user and the current implementation of isOn is hidden from the user I thought it would be better in this order. But I am open to switching them.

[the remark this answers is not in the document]
reesscot: Makes sense

examples/next/pages/_app.page.tsx
```diff
@@ -1,5 +1,5 @@
 import '@aws-amplify/ui-react/styles.css';
-import './examples/listings/styles.scss';
+// import './examples/listings/styles.scss';
```
reesscot, line 2: Can you please remove this change?

docs/src/pages/ui/primitives/switchfield/react.mdx
```diff
@@ -0,0 +1,290 @@
...
+
+The `SwitchField` form primitive is a toggleable input type with a checked (on) and unchecked (off) state.
+
+## Demo
+
+<SwitchDemo />
+
+## Usage
+
+The most basic usage simply includes a switchField component passing in a required label prop.
```
reesscot, line 13:
```suggestion
The most basic usage simply includes a SwitchField component passing in a required label prop.
```

packages/ui/src/theme/tokens/space.js
```diff
@@ -7,4 +7,14 @@ module.exports = {
   xl: { value: '2.0rem' },
   xxl: { value: '3.0rem' },
   xxxl: { value: '4.5rem' },
+  relative: {
```
reesscot, line 10: What's the intended use of these, they appear to have all the same values as the non-relative sizes?

reesscot (commented):
Looking great, just a few more questions and a few unanswered questions left

packages/ui/src/theme/tokens/space.js
jacoblogan: the difference is the units, `rem` vs `em` so that we can size relative to the parent instead of the root.

jacoblogan: just creating a set of sizes using `em` instead of `rem` for sizing based off of the parent instead of the root. These are being used for resizing the switchfield with a single rule on the parent element instead of placing a data-size attribute/rule on every piece that needs to be resized.

packages/ui/src/theme/css/component/switchField.scss
jacoblogan: I've added a few more CSS variables, but some of the basic default styles are still being set directly.  Maybe a pass can be made on these styles as well as other defaults for other primitives based on the design decision that is made.

packages/ui/src/theme/css/component/switchField.scss
jacoblogan: more padding has been added

packages/ui/src/theme/tokens/space.js
jacoblogan: added a comment to the code

reesscot (approved):
Great work! Let's get this shipped! 🛳

PR #319, opened 2021-08-31
SelectField Primitive

zchenwei:
*Description of changes:*

This PR adds the `SelectField` primitive in favor of accessibility which doesn't provide by default in #201


packages/react/src/primitives/TextField/__tests__/TextField.test.tsx
```diff
@@ -86,11 +86,19 @@ describe('TextField component', () => {
...
+          variation="quiet"
+        />
+      );
 
-      const field = await screen.findByRole('textbox');
-      expect(field.dataset['size']).toBe('small');
-      expect(field.dataset['variation']).toBe('quiet');
+      const textField = await screen.findByTestId('testField');
+      const input = await screen.findByRole('textbox');
+      expect(textField).toHaveAttribute('data-size', 'small');
```
zchenwei, line 100: I found a bug earlier in `TextField` that we don't pass `size` prop to the high-level `field`. We should test the sizing of the `field` as a whole, which includes label.

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -23,11 +24,16 @@ export const Select: React.FC<SelectProps> = (props) => {
...
   } = props;
   return (
-    <View className="amplify-select-wrapper">
+    <View className={ComponentClassNames.SelectWrapper}>
       <View
+        aria-invalid={hasError}
         as="select"
         value={value}
-        defaultValue={defaultValue}
+        defaultValue={
```
zchenwei, line 32: `value === undefined` is to get rid of controlled component. React will complain if we set both `defaultValue` and `value`

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -23,11 +24,16 @@ export const Select: React.FC<SelectProps> = (props) => {
...
+          value === undefined && defaultValue === undefined && placeholder
```
zchenwei, line 33: value === undefined is to get rid of controlled component. React will complain if we set both defaultValue and value

packages/theme-base/src/css/component/fieldMessages.scss
```diff
@@ -2,7 +2,8 @@
   font-style: var(--amplify-components-fieldmessages-description-font-style);
   font-size: var(--amplify-components-fieldmessages-description-font-size);
 }
-.amplify-field__error-message {
+// Increase specificity here to avoid color overridden by .amplify-text color
+.amplify-field .amplify-field__error-message {
```
zchenwei, line 6: The `color` will be overridden by the one set in `.amplify-text` if we don't increase specificity here

packages/react/src/primitives/TextField/TextField.tsx
```diff
@@ -45,7 +45,8 @@ export const TextField: React.FC<TextFieldProps> = ({
     <Flex
       alignContent={alignContent}
       alignItems={alignItems}
-      className={classNames(ComponentClassNames.TextField, className)}
+      className={classNames(ComponentClassNames.Field, className)}
+      data-size={size}
```
zchenwei, line 49: Here is the small bug. We need to apply sizing prop here, otherwise the styling defined for it in `field.scss` will not work

reesscot: Thanks for fixing this!

packages/theme-base/src/css/component/fieldMessages.scss
reesscot: We should be able to fix the specificity battle by importing the `text.scss` file at the top of the `theme.scss` file, rather than after the `fieldMessages.scss` file.

reesscot (changes requested):
This is looking great! Just a few suggestions below

packages/react/src/primitives/Select/Select.tsx
zchenwei: > Should we check for `null` and `undefined` using loose equality check? (`== null` will check for `null `and `undefined`)

React will take `null` as a valid value passed to a prop. It just indicates nothing. So we will got the same complain if we rule it out along with `undefined`. Also, it will warn when `value` is set to `null` and request us to instead use empty string or `undefined`.

packages/react/src/primitives/SelectField/SelectField.tsx
```diff
@@ -0,0 +1,78 @@
...
+      direction={direction}
+      gap={gap}
+      justifyContent={justifyContent}
+      testId={testId}
+      wrap={wrap}
+    >
+      <Label htmlFor={fieldId} visuallyHidden={labelHidden}>
+        {label}
+      </Label>
+      <Select
```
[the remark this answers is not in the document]
zchenwei: IMO, It looks a bit wired that there is a `FieldDescription` above a `Select` because there will always be a placeholder option there. But for consistency, I am good to include that

packages/react/src/primitives/SelectField/__tests__/SelectField.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+        <SelectField label={label} testId={testId} className={className}>
+          <option value="1">1</option>
+          <option value="2">2</option>
+          <option value="3">3</option>
+        </SelectField>
+      );
+
+      const selectField = await screen.findByTestId(testId);
+      expect(selectField).toHaveClass(className);
+      expect(selectField).toHaveClass(ComponentClassNames.Field);
```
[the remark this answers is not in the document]
zchenwei: Yes, this is the high-level field and we share the sizing styles through `Field`. Should I have another `SelectField` and copy all the styles from it? I expect customer would more likely to style the inner Select with `amplify-select` and that's how I document about the styling section as well

packages/react/src/primitives/TextField/TextField.tsx
```diff
@@ -45,7 +45,8 @@ export const TextField: React.FC<TextFieldProps> = ({
     <Flex
       alignContent={alignContent}
       alignItems={alignItems}
-      className={classNames(ComponentClassNames.TextField, className)}
+      className={classNames(ComponentClassNames.Field, className)}
```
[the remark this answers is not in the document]
zchenwei: Ah yea, this seems related to a previous question. That's what I am wondering actually, I think this is a good idea.

packages/theme-base/src/css/component/_fieldControl.scss
```diff
@@ -0,0 +1,88 @@
+%amplify-field-control {
+  box-sizing: border-box;
+  font-size: var(--amplify-components-field-control-font-size);
+  line-height: var(--amplify-components-field-control-line-height);
+  padding: var(--amplify-components-field-control-padding);
+  transition: all var(--amplify-components-field-control-transition-duration);
+  width: 100%;
```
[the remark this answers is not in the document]
zchenwei: I will update it

packages/theme-base/src/css/theme.scss
```diff
@@ -45,4 +46,3 @@ html {
 @import './component/rating.scss';
 @import './component/select.scss';
 @import './component/text.scss';
```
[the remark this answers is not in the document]
zchenwei: Good catch!

packages/react/src/primitives/SelectField/__tests__/SelectField.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+          <option value="3">3</option>
+        </SelectField>
+      );
+      const selectField = await screen.findByTestId(testId);
+      expectFlexStyleProps(selectField);
+    });
+  });
+
+  describe('Label', () => {
+    it('should render expected field classname', async () => {
```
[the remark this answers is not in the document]
zchenwei: Sounds great!

packages/react/src/primitives/Select/Select.tsx
reesscot: Yeah, that's a good point. if it's set to null then the customer passed it in, versus undefined means the prop wasn't given. 👍

packages/react/src/primitives/SelectField/SelectField.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { FieldErrorMessage } from '../Field';
+import { Flex } from '../Flex';
+import { Select } from '../Select';
+import { Label } from '../Label';
+import { SelectFieldProps } from '../types/selectField';
+
+export const SelectField: React.FC<SelectFieldProps> = (props) => {
+  const {
+    alignContent,
+    alignItems,
```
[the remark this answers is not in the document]
zchenwei: Cool! And that reminds me the `autoComplete` isn't passed in `Input` int the TextField. I will fix that as well

joebuono (approved):
LGTM!

docs/src/pages/ui/primitives/selectfield/react.mdx
```diff
@@ -281,6 +281,70 @@ import '@aws-amplify/ui-react/styles.css';
...
+  >
+    <option value="apple">Apple</option>
+    <option value="banana">Banana</option>
+    <option value="orange">Orange</option>
+  </SelectField>
+</Example>
+
+### State
+
+`isDisabled` is the only available state for SelectField. A disabled field will be not be focusable not mutable and will not be submitted with form data.
```
reesscot, line 317: nit: could mention that if customers want a readonly field that's still submitted with the form, they can make all the options besides the selected option be disabled:

```
<option disabled></option>
<option value="default"></option>
<option disabled></option>
```
ref: https://stackoverflow.com/a/23428851

packages/react/src/primitives/SelectField/SelectField.tsx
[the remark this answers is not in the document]
reesscot: Oh yeah, I found that too and fixed in PasswordField PR :)

reesscot (approved):
LGTM! 🛳

docs/src/pages/ui/primitives/selectfield/react.mdx
zchenwei: I saw that as well when trying to find out if `select` has a `readonly` state!! Then I think it makes sense that `select` doesn't has that support because the use case is wired. A read-only select could totally be an `input` 🤣

PR #314, opened 2021-08-30
Rename @aws-amplify/ui-core to @aws-amplify/ui

ericclemmons:
*Issue #, if available:* https://app.asana.com/0/1200334286823427/1200857187579948/f

*Description of changes:*

- [x] Rename `@aws-amplify/ui-core` to `@aws-amplify/ui`
- [x] Migrate `theme-base` into `ui`
- [ ] ~~Remove `scripts/generateCSS.js` in favor of output of build process from `src/index.ts`~~

	So this is totally doable, but can be done in separate PR.  `tsup`, `esbuild`, `microbuild` & others all support `import "./styles.css"` and automatically running that through PostCSS. So that means we can build CSS & JS in 1 step with watch-mode support!	

- [x] Split `theme.css` to just be CSS variables & `styles.css` to include all primitive styles.
- [x] Rename `dict` export to `translations` for clarity/discoverability on what they are, and to be similar to https://docs.amplify.aws/ui/customization/translations/q/framework/react/. (Not using capital `T` translations because it's not equivalent to the enums used before)

### Directory Structure

- Note `dist:
	- Includes CJS `.js` and ESM `.mjs`
	- CSS Variables are `theme.css`
	- Theme + all component styles are in `styles.css`

```console
❯ tree packages/ui -L 3 -I node_modules
packages/ui
├── CHANGELOG.md
├── README.md
├── dist
│   ├── index.d.ts
│   ├── index.js
│   ├── index.js.map
│   ├── index.mjs
│   ├── index.mjs.map
│   ├── styles.css
│   └── theme.css
├── package.json
├── scripts
│   └── generateCSS.js
├── sd.config.js
├── src
│   ├── authService.ts
│   ├── globals.js
│   ├── helpers
│   │   ├── auth.ts
│   │   └── index.ts
│   ├── i18n
│   │   ├── index.ts
│   │   └── translations.ts
│   ├── index.ts
│   ├── machines
│   │   ├── actions
│   │   ├── actors
│   │   ├── authMachine.ts
│   │   └── index.ts
│   ├── theme
│   │   ├── css
│   │   ├── index.ts
│   │   ├── theme.ts
│   │   └── tokens
│   ├── types
│   │   ├── authMachine.ts
│   │   ├── index.ts
│   │   └── validator.ts
│   └── validators
│       └── index.ts
├── tsconfig.json
└── tsup.config.js
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #320, opened 2021-08-31
Reusable actions

eddiekeller:
*Issue #, if available:*

n/a

*Description of changes:*

Move Xstate actions into a separate file for easy reuse across actors.

### DIFF
--- a/packages/core/src/machines/actors/auth/signIn.ts
+++ b/packages/core/src/machines/actors/auth/signIn.ts
@@ -1,8 +1,24 @@
-import { createMachine, assign, sendUpdate } from 'xstate';
 import { get, isEmpty } from 'lodash';
+import { createMachine, sendUpdate } from 'xstate';
 
-import { AuthEvent, AuthChallengeNames, SignInContext } from '../../../types';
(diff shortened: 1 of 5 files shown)


packages/core/src/machines/actors/auth/signIn.ts
```diff
@@ -265,40 +287,19 @@ export const signInActor = createMachine<SignInContext, AuthEvent>(
...
+      clearFormValues,
+      clearUnverifiedAttributes,
+      handleInput,
+      setChallengeName,
+      setConfirmResetPasswordIntent,
+      setConfirmSignUpIntent,
+      setRemoteError,
+      setUnverifiedAttributes,
+      setUser,
+      setUsernameAuthAttributes,
```
ericclemmons, line 302: 😍

ericclemmons (approved):
Looks good!  #314 includes a test to ensure we're not re-exporting too much for `ui-react`, which we'll want to adopt for just `ui` – just want to keep in mind that we don't accidentally increase our exported API surface area 🙏

PR #316, opened 2021-08-30
build(vue): declaration file type conflict

wlee221:
*Issue #, if available:*

*Description of changes:* Following up https://github.com/aws-amplify/amplify-ui/pull/310#discussion_r698049403. Opened this pr for mainly ci checks and keeping track of progress.

Error we get is: 
```
[rpt2] /home/runner/work/amplify-ui/amplify-ui/packages/vue/src/components/confirm-sign-in.vue?vue&type=script&lang.ts(37,52): semantic error TS2339: Property 'challengeName' does not exist on type 'AuthActorContext'.
```

although we assert `ActorContext` to be `SignInContext` here:https://github.com/aws-amplify/amplify-ui/blob/64ef5f95e3d125b82ec8970f78b5b12f8b6213cf/packages/vue/src/components/confirm-sign-in.vue#L93-L95

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #321, opened 2021-08-31
Vue: Verify User Components

ErikCH:
*Issue #, if available:* [Asana](https://app.asana.com/0/0/1200689363526604/f)

*Description of changes:*
Add page if user is confirmed but not verified


examples/vue/package.json
```diff
@@ -21,6 +21,6 @@
     "typescript": "^4.3.2",
     "vite": "^2.4.2",
     "vite-plugin-voie": "^0.7.3",
-    "vue-tsc": "^0.0.24"
+    "vue-tsc": "^0.3.0"
```
wlee221, line 24: Just curious, what is this package used for?

packages/core/src/types/authMachine.ts
```diff
@@ -30,15 +30,18 @@ export interface SignInContext extends BaseFormContext {
 
 export interface SignUpContext extends BaseFormContext {
   login_mechanisms?: string[];
+  unverifiedAttributes?: Record<string, string>;
 }
 
 export interface ResetPasswordContext extends BaseFormContext {
   username?: string;
+  unverifiedAttributes?: Record<string, string>;
```
wlee221, line 38: Ah, so this is the same problem as https://github.com/aws-amplify/amplify-ui/pull/310#discussion_r698049403. I dived into this a little bit yesterday and found some clue I wrote in https://github.com/aws-amplify/amplify-ui/pull/316#issuecomment-909509741. tldr, Vue has problem inferring computed property types. (Not sure if it's a problem with Vue or rtp2...)
 
Ideally we would want to fix this error, but I'm good to descope this to another pr. We can revisit this later after @eddiekeller's state machine improvements.

packages/vue/src/components/confirm-verify-user.vue
```diff
@@ -0,0 +1,145 @@
...
+    BaseWrapper,
+  },
+  inheritAttrs: false,
+  setup(
+    _,
+    { emit, attrs }: SetupEventContext
+  ): ConfirmVerifyUserSetupReturnTypes {
+    const { state, send } = useAuth();
+    const actorState: ComputedRef<SignInState> = computed(
+      () => getActorState(state.value) as SignInState
```
wlee221, line 93: Looking at this, `getActorState` really needs to be explicit about its return type -- it's making us use type assertion all over the places, and it can fix the problem above 😅

packages/vue/src/components/verify-user.vue
```diff
@@ -0,0 +1,154 @@
...
+      });
+    };
+
+    const onSkipClicked = (): void => {
+      if (attrs?.onSkipClicked) {
+        emit('skipClicked');
+      } else {
+        send({
+          type: 'SKIP',
+        });
```
wlee221, line 137: fyi, you can do 
```suggestion
        send('SKIP');
```
to be a little bit more concise. But this is totally stylistic and is up to you :D

wlee221 (approved):
LGTM ! Just got some comments, but vue functionality looks clean 🙏

examples/vue/package.json
ErikCH: It's used to check vue's typescript files

packages/core/src/types/authMachine.ts
ErikCH: @wlee221 Thanks for doing the research! Yes, that must be what's happening, weird.

packages/vue/src/components/verify-user.vue
ErikCH: Sure I'll update it

PR #323, opened 2021-09-01
Don't start authService as a side-effect

ericclemmons:
*Description of changes:*

Importing `@aws-amplify/ui-react` automatically starts the `authService` in development (because there's no tree-shaking):

```console
TypeError: Cannot read property 'auth' of undefined

      at config (node_modules/@aws-amplify/ui-core/dist/index.cjs:1:14895)
      at node_modules/xstate/lib/utils.js:371:61
          at Array.reduce (<anonymous>)
      at Object.updateContext (node_modules/xstate/lib/utils.js:353:48)
      at Object.resolveActions (node_modules/xstate/lib/actions.js:412:53)
      at StateNode.resolveTransition (node_modules/xstate/lib/StateNode.js:895:36)
      at StateNode.transition (node_modules/xstate/lib/StateNode.js:831:17)
      at node_modules/xstate/lib/interpreter.js:705:28
      at Object.provide (node_modules/xstate/lib/serviceScope.js:12:16)
      at Interpreter.nextState (node_modules/xstate/lib/interpreter.js:704:34)
      at node_modules/xstate/lib/interpreter.js:107:31
      at Scheduler.process (node_modules/xstate/lib/scheduler.js:69:7)
      at Scheduler.schedule (node_modules/xstate/lib/scheduler.js:48:10)
      at Interpreter.send (node_modules/xstate/lib/interpreter.js:103:23)
      at actor.id (node_modules/xstate/lib/interpreter.js:1000:15)
```

More importantly, **there should not be side-effects from importing**. The state machines should **only** be started upon render.


eddiekeller (approved):

hvergara (approved):

dbanksdesign (approved):
LGTM! :shipit:

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,6 +1,10 @@
-import { useActor } from '@xstate/vue';
-import { authService } from '@aws-amplify/ui';
+import { authMachine } from '@aws-amplify/ui';
+import { useActor, useInterpret } from '@xstate/vue';
 
 export const useAuth = () => {
-  return useActor(authService);
+  const service = useInterpret(authMachine, {
+    devTools: process.env.NODE_ENV === 'development',
```
reesscot, line 6: Will `process` be defined in the browser?

ericclemmons: This is normal convention for bundlers since webpack2 to do dead-code elimination by setting `process.env.NODE_ENV = "production"` and evaluating this to `devTools: 'production' === 'development'`, **but I assume this is true for Vite as well**.

Let me check...

Yes, it should work:
https://github.com/vitejs/vite/commit/d7ceabe92efe1c523ee05a6941e3f5042733b2bf

If not, we can remove this.  But I'm still on the fence if this should be a default value in `useAuth` vs. for development we use `<AmplifyProvider services={authService}>` to override it 🤔  We'll need to test out the DX some more – that visualization can be essential to validating flows.

wlee221 (approved):

packages/vue/src/composables/useAuth.ts
ericclemmons: @ErikCH Ahhh, I think the issue here is that this is **repeatedly** creating the `authMachine` & returning a service – **so it's not a single machine**.

For React, the `Authenticator` renders once and creates the machine there, passing down the service via context:

https://github.com/aws-amplify/amplify-ui/blob/e9be592f172d6ac2613dd76123d5c5de47357b85/packages/react/src/components/Authenticator/index.tsx#L15-L23

What's the best way to do this with Vue?

ericclemmons: Should https://github.com/aws-amplify/amplify-ui/blob/main/packages/vue/src/components/authenticator.vue `provide` the service and each sub-screen `inject` it rather than `useAuth`?

ErikCH: Provide and inject could work, but it's a little messy trying to provide the service at the authenticator level. I decided instead to use `ref` which creates a reactive variable, that is stored in memory.

ErikCH: This should fix Vue from failing to run.

ErikCH (approved):

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,6 +1,14 @@
+import { authMachine } from '@aws-amplify/ui';
 import { useActor } from '@xstate/vue';
-import { authService } from '@aws-amplify/ui';
+import { interpret } from 'xstate';
+import { ref } from 'vue';
+
+const service = ref(
+  interpret(authMachine, {
+    devTools: process.env.NODE_ENV === 'development',
+  }).start()
```
ericclemmons, line 9: I think this will still be a problem because the service is being `start`'d the moment it's imported, rather than created & started _when the Authenticator is rendered_.

Otherwise, it's still a side-effect from an `import` and can break when tree-shaking and be unintended behavior for anyone _not_ using an Authenticator.

ericclemmons: For comparison, I removed the `authService` for exactly this reason:
https://github.com/aws-amplify/amplify-ui/pull/323/files#diff-06592eff0be23980380da9854605df47ca1355d6d9b5df6fdd9a35aabeb35335L6

ErikCH: There might be a way to model React's Context with Vue Provide/Inject, but it will cause `useAuth` to no longer work outside of Vue internally. I've pushed up another possible solution.

PR #328, opened 2021-09-01
Added in internationalization

ErikCH:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200881947262804/1200795341338117/f)

*Description of changes:*
Add I18n to labels


eddiekeller (approved):

packages/vue/src/components/confirm-reset-password.vue
```diff
@@ -124,15 +126,17 @@ export default defineComponent({
       getActorState(state.value)
     ) as ComputedRef<ResetPasswordState>;
     // Computed Properties
-    const backSignInText = computed(() => BACK_SIGN_IN_TEXT);
-    const lostYourCodeText = computed(() => LOST_YOUR_CODE_TEXT);
-    const resendCodeText = computed(() => RESEND_CODE_TEXT);
-    const confirmationCodeText = computed(() => CONFIRMATION_CODE_TEXT);
-    const confirmResetPasswordHeading = computed(
-      () => CONFIRM_RESET_PASSWORD_HEADING
+    const backSignInText = computed(() => I18n.get(BACK_SIGN_IN_TEXT));
```
ericclemmons, line 129: Should we remove the usage of `BACK_SIGN_IN_TEXT` from `DefaultTexts.ts` and instead use `Back to Sign In`? Otherwise this won't actually translate anything yet, right? 🤔 

https://github.com/aws-amplify/amplify-ui/blob/4f3dfd0df2e4bbaaff48a1dc9b12a4dd67c84433/packages/ui/src/i18n/translations.ts#L20

ericclemmons (commented):
The `I18n` adoption looks good, but I'm wondering if this will work with the definitions that now exist in https://github.com/aws-amplify/amplify-ui/blob/4f3dfd0df2e4bbaaff48a1dc9b12a4dd67c84433/packages/ui/src/i18n/translations.ts.

packages/vue/src/components/confirm-reset-password.vue
ErikCH: @ericclemmons The default texts shows 
`export const BACK_SIGN_IN_TEXT = 'Back to Sign In';`
So that should translate correctly, since that text matches ` 'Back to Sign In': 'Zurück zur Anmeldung', `

ErikCH: I did go through and update a handful of `BACK_SIGN_IN_TEXT` to match the translations files

ericclemmons: Oooh, I see.  So it's a bit of indirection here!  So I guess we have 2 paths in the future:

1. Directly use the `keyof translations[lang]` like `Back to Sign In` or
2. Have constants like `BACK_SIGN_IN_TEXT` (https://github.com/aws-amplify/amplify-js/blob/46ee5dd91c61f49bad4da8286b2f97c737d96631/packages/amplify-ui-components/src/common/Translations.ts#L4) who's value matches something in `translations`.

I think we're good here, but we'll need to revisit I18n in a way that creates a better contract for ourselves & customers to customize:
> https://docs.amplify.aws/ui/customization/translations/q/framework/react/

PR #329, opened 2021-09-02
Add Country Code selector for React SignUp

slaymance:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200881947262804/1200864094580680)

*Description of changes:* Adds a country code selector (default to US phone number `+1`) which will be prepended to end-users' phone number upon sign up. This `select` tag only applies to the `SignUp` screen where `phone_number` is a required field.


packages/ui/src/machines/authMachine.ts
```diff
@@ -130,7 +130,7 @@ export const authMachine = createMachine<AuthContext, AuthEvent>(
           const actor = signUpActor.withContext({
             authAttributes: event.data?.authAttributes ?? {},
             intent: event.data?.intent,
-            formValues: {},
+            formValues: { country_code: '+1' },
```
slaymance, line 133: This is where I stuck the default `country_code` state. Would love some feedback on the best way to handle this.

packages/e2e/cypress/integration/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.steps.ts
```diff
@@ -17,6 +17,7 @@ Then("I don't see {string} as an input field", (name: string) => {
 });
 
 When('I type the phone number {string}', (phone: string) => {
+  cy.findByRole('combobox', { name: /country code/i }).select('+1');
```
ericclemmons, line 20: This may belong in a separate `When I select the country code '+1'` command because we'll also test that `+1` is the default and that `8008675309` works without explicitly selecting the country code.

packages/react/src/components/common/country-dial-codes.tsx
```diff
@@ -0,0 +1,208 @@
+export const countryDialCodes = [
```
ericclemmons, line 1: I assume this came from the previous implementation?

packages/ui/src/machines/actors/auth/signUp.ts
```diff
@@ -188,18 +188,20 @@ export const signUpActor = createMachine<SignUpContext, AuthEvent>(
...
+            `${formValues.country_code}${formValues.phone_number}`.replace(
+              /[^A-Z0-9+]/gi,
+              ''
+            );
         }
 
         const username = formValues[primaryAlias];
         delete formValues[primaryAlias];
         delete formValues.confirm_password; // confirm_password field should not be sent to Cognito
+        delete formValues.country_code;
```
ericclemmons, line 204: FYI, mutating this `formValues` object via `delete` is a sign we're assuming they're nearly 1-to-1 with `attribtues`.  That's not correct.

`attributes` (on line 209 below) should be created _from_ `formValues` (and any other data sources).

For example, what happens when signUp fails?  Will `formValues` still have the original `formValues.country_code`?

packages/ui/src/machines/authMachine.ts
ericclemmons: Honestly, this makes sense to me since we're defaulting the value for the view (vs. the view having the default itself on every platform).

ericclemmons (approved):
I don't think it warrants blocking, but wanted to call out a couple risks that I want to get your take on:

1. Automatically selecting the country code in e2e tests vs. explicitly.
2. `delete` and use of `formValues` as `attributes`.

packages/react/src/components/common/country-dial-codes.tsx
ErikCH: Could this be moved to `ui` so Angular and Vue can grab it too, without having to copy and paste it into our own implementation?

ericclemmons: Yes, we should.  Apparently everyone's re-using this handy list:

https://github.com/uber/baseweb/blob/8255ff344606e4f702af13bd479e5a1e1d0e395d/src/phone-input/constants.js#L18-L321

From: https://github.com/jackocnr/intl-tel-input/blob/master/src/js/data.js

Frankly, I think this is actually a better list since it'll allow us to to key off of the `navigator.language` property by default.

packages/ui/src/machines/actors/auth/signUp.ts
slaymance: It's a bit cheeky since `formValues` is created by spreading the remaining values from `context` after pulling out `password`. So it already is a (shallow) copy of the actual form values passed from the UI. We could be more explicit about what values we're pulling from `formValues`, though this is a bit less scalable. What do you think @ericclemmons?

packages/react/src/components/common/country-dial-codes.tsx
slaymance: I just leaned on how `amplify-js` defined country dial codes: https://github.com/aws-amplify/amplify-js/blob/df95ea3724eb6406f64b03f25086cd3e8644cb5f/packages/aws-amplify-react/src/Auth/common/country-dial-codes.tsx

I like the idea of using that list and building our dial codes from that. I'll make a change.

packages/e2e/cypress/integration/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.steps.ts
slaymance: Yeah that makes sense. Any thoughts on where the `'+1'` comes from in the feature file? Would we just hard code it?

ErikCH (approved):

packages/ui/src/machines/actors/auth/signUp.ts
ericclemmons: Ah, I see.  Yeah, I think I'm biased towards explicitness on knowing what the valid `attributes` even are.  IME, Forms aren't 1-to-1 with data models (Country code is a good example of this, as is confirm password).

I think I'm looking for an answer on _how do we know what valid attributes should even be sent_?  If we accept 16 different custom attributes, won't we end up doing something like `const attributes = pick(formValues, signUpAttributes)`?

Nothing to change here today, but just a callout that mutating an object can pose a risk and that we'll likely require a different solution when we support all custom attributes.

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
```diff
@@ -118,25 +118,39 @@ SignUp.AliasControl = ({
...
+          name="country_code"
+          defaultValue={formValues.country_code}
+        >
+          {countryDialCodes.map((dialCode) => (
+            <option key={dialCode} value={dialCode}>
+              {dialCode}
+            </option>
+          ))}
+        </select>
+      )}
```
eddiekeller, line 144: We should try using the SelectField primitive that we have instead.

https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/SelectField/SelectField.tsx

eddiekeller (changes requested):

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
slaymance: Good call, this will also be implemented as part of [PhoneNumberField](https://app.asana.com/0/1200881947262804/1200683483662286) primitive. I had mostly just implemented core functionality in this PR.

PR #336, opened 2021-09-03
feat: adding initial styling for the sign in page

eddiekeller:
*Issue #, if available:*

n/a

*Description of changes:*

These are not final style changes. There are questions that we need the product team to answer regarding styling. Additionally, the mocks make use of tabs, a primitive that we don't have ready quite yet (that will replace the out-of-place "create account" section), we still need social provider icons, etc. However! Progress! It looks at least usable now.

Updating the sign in page to use the new primitives (FieldGroup, Flex, PasswordField, etc).
Adding styling based on mocks - https://www.figma.com/proto/74U1PE5KhnRslVJbvLxeKA/Duplicate-Me?node-id=539%3A32964&scaling=min-zoom&page-id=539%3A23088

Screenshot - 
Screen Shot 2021-09-03 at 3 09 35 PM


packages/react/src/components/Authenticator/styles.css
```diff
@@ -63,3 +63,46 @@
 [data-amplify-authenticator] :disabled {
   @apply opacity-50;
 }
+
+:root {
+  /* primary button colors */
+  --amplify-components-button-primary-background-color: #7256ff;
+  --amplify-components-button-primary-border-color: #7256ff;
```
ericclemmons, line 70: @dbanksdesign This change would mean that these default values would be overridden for all components for anyone importing the Authenticator.

Should we scope these to `[data-amplify-authenticator]` or re-evaluate the default values of these variables?

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignInButtons/FederatedSignInButton.tsx
```diff
@@ -28,5 +31,20 @@ export const FederatedSignInButton = (
     });
   };
 
-  return <Button onClick={handleClick}>{text}</Button>;
+  // TODO - get Google and Amazon icons
+  let iconComponent;
+  if (icon === 'facebook') {
+    iconComponent = <IconFacebook />;
+  }
```
ericclemmons, line 38: We can steal these SVGs rather than expecting `IconFacebook` to be exported as a primitive:

https://github.com/aws-amplify/amplify-js/blob/d815f8bb825163fcb5d6f713b235b1421e08b1a6/packages/amplify-ui-components/src/components/amplify-icon/icons.tsx#L31-L68

(I need to write a proposal/PR for our re-exporting of Material UI icons)

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignInButtons/FederatedSignInButton.tsx
```diff
@@ -1,15 +1,18 @@
 import { useAmplify, useAuth } from '../../../../hooks';
 import { FederatedIdentityProviders } from '@aws-amplify/ui';
 
+import { IconFacebook } from '../../../../primitives';
+
 export interface FederatedSignInButtonProps {
+  icon?: 'facebook' | 'google' | 'amazon';
```
ericclemmons, line 7: I think we'll be revisiting this implementation and how it's exposed for customers to customize via slots (Vue/Angular) or provider (React/RN), so start thinking about what abstraction makes the most sense for customers to replace/inject OAuth/SAML providers (e.g. Auth0, Discord, GitHub, Twitter) in the UI.

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
```diff
@@ -52,43 +51,60 @@ export function SignIn() {
...
-        <Button onClick={() => send({ type: 'SIGN_UP' })} type="button">
-          {I18n.get('Create account')}
+        <Button
+          borderRadius={0}
+          isDisabled={isPending}
+          isFullWidth={true}
+          type="submit"
+          variation="primary"
+          isLoading={isPending}
+          loadingText={I18n.get('Signing in')}
```
ericclemmons, line 74: This will need a translation I'm guessing...

ericclemmons (commented):
Thanks for getting this looking right @eddiekeller!

The only blocker I have is the usage of `:root` for CSS variables – should those be our new default, or scoped to `[data-amplify-authenticator]`?

(To unblock for merging, I recommend the latter. We need a kitchen-sink, comprehensive layout of primitives to finalize the right default values IMO)

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignInButtons/FederatedSignInButton.tsx
eddiekeller: Nice, sounds good.

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignInButtons/FederatedSignInButton.tsx
eddiekeller: Sure, this is just to get us started for now.

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
eddiekeller: Yep!

packages/react/src/components/Authenticator/styles.css
eddiekeller: Yeah, let me know what would be the correct way to go. I went with `:root` because that's what the primitives docs say - https://ui.docs.amplify.aws/ui/primitives/button .

eddiekeller: Updated in new commit to use `data-amplify-authenticator`

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignInButtons/FederatedSignInButton.tsx
eddiekeller: Removing this for now until we pull in the others.

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
```diff
@@ -52,43 +51,61 @@ export function SignIn() {
...
+          variation="link"
+          size="small"
+          fontWeight="normal"
+        >
+          {I18n.get('Forgot your password? ')}
         </Button>
-      </Footer>
+
+        <Footer>
+          <Text fontWeight="normal">{I18n.get('No account? ')}</Text>
```
dbanksdesign, line 92: I don't think you need `fontWeight="normal"` it should be normal by default?

packages/react/src/components/Authenticator/shared/UserNameAlias.tsx
```diff
@@ -14,17 +14,19 @@ export function UserNameAlias(props: UserNameAliasProps) {
...
-      <Text>{I18n.get(label)}</Text>
-      <Input
-        onChange={handleInputChange}
-        name="username"
-        required
-        type={type}
-      />
-      <ErrorText>{error}</ErrorText>
-    </Label>
+    <TextField
```
dbanksdesign, line 20: yay

dbanksdesign (commented):
Looks great! The main stuff is we just need to align on the default styling. I think the Authenticator should just use the default style of the primitives for consistency. I'll work to make sure the mocks and primitives are in line.

packages/react/src/components/Authenticator/styles.css
```diff
@@ -63,3 +63,46 @@
...
+  --amplify-components-button-link-loading-color: #8e7bfa;
+
+  /* default button colors */
+  --amplify-components-button-color: #7256ff;
+  --amplify-components-button-border-color: #7256ff;
+
+  --amplify-components-button-hover-color: white;
+  --amplify-components-button-hover-background-color: #8e7bfa;
+  --amplify-components-button-hover-border-color: #7256ff;
+}
```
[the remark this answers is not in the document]
ericclemmons: Totally agree @dbanksdesign.  The way I've solved this in the past was to have a kitchen-sink example (dashboards are usually great for these!) so we can see that all default styles look good holistically.

What do you think?

ericclemmons (approved):
This looks like it'll unblock some more screens, so I'm good to go!

One of our main goals will be **for the Authenticator to use primitives without additional CSS styles**. (Specifying a few props is totally expected)

packages/react/src/components/Authenticator/styles.css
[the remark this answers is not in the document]
dbanksdesign: Yup I love it! Maybe a default docs page like /ui/primitives has a kitchen sink example?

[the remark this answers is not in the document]
eddiekeller: So for now then, I will remove these custom CSS overrides. For my own curiosity - will the color scheme be the purple as seen in the mocks? Or the teal color that the primitives currently use?

packages/react/src/components/Authenticator/styles.css
```diff
@@ -63,3 +63,46 @@
...
+
+.federated-sign-in-container {
+  margin-top: 1rem;
+}
+
+.federated-sign-in-button {
+  border-color: black;
+  border-radius: 0;
+  color: black;
+}
```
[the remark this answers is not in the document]
eddiekeller: It can be! I'm down for that.

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
eddiekeller: Yep, good catch!

packages/react/src/components/Authenticator/styles.css
```diff
@@ -63,3 +63,46 @@
...
+  --amplify-components-button-color: #7256ff;
+  --amplify-components-button-border-color: #7256ff;
+
+  --amplify-components-button-hover-color: white;
+  --amplify-components-button-hover-background-color: #8e7bfa;
+  --amplify-components-button-hover-border-color: #7256ff;
+}
+
+.federated-sign-in-container {
+  margin-top: 1rem;
```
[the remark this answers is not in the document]
eddiekeller: Cool, works for me.

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignInButtons/FederatedSignInButton.tsx
```diff
@@ -28,5 +31,20 @@ export const FederatedSignInButton = (
...
+  let iconComponent;
+  if (icon === 'facebook') {
+    iconComponent = <IconFacebook />;
+  }
+
+  return (
+    <Button
+      onClick={handleClick}
+      className="federated-sign-in-button"
+      fontWeight="normal"
```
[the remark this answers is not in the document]
eddiekeller: I'd prefer for the default style to have a `normal` font weight. If I leave this off, the default style is pretty bold.

PR #340, opened 2021-09-07
Audit of feature tests

slaymance:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200881947262804/1200751914824795)

*Description of changes:* This PR adds a couple of `.feature` to capture features, as well as adds comments for skipped tests regarding what's needed to have them functioning. A list of features which need to be addressed:

1. **Sign up features**: All sign up features are disabled due to SES limits for our Cognito environments
2. **Force change password features**: All features regarding force change password are skipped due to expiration of temporary passwords used by test users with the `FORCE_CHANGE_PASSWORD` Cognito status
3. **SMS MFA**: There's some cleanup needed on the Angular feature
4. **TOTP MFA**: Angular TOTP isn't currently functioning. Invalid TOTP code feature needs clean up across frameworks
5. **Verify user**: We run into backend limits when sending verification codes to test users and we don't have a way to receive those codes. Angular verify is currently not functioning.


eddiekeller:
Thanks for doing this audit, @slaymance . Before looking at the code, just a couple notes to add to your overview description.

---

> SMS MFA expired code (with resend code): This feature has not yet been implemented

If I remember correctly, there isn't an API to resend a SMS MFA code. The old authenticator didn't support this either.

> TOTP MFA expired code: This feature is not yet implemented (?)

We don't send anything with TOTP, as the code is generated inside of the user's chosen third-party auth application.

> SMS MFA full flow: A full e2e flow for SMS MFA is not currently implemented as we need some way of receiving a confirmation code

I *think* the only way we can do this is by using something like Twilio. However, at that point, we are introducing more libraries and weirdness into our tests that aren't actually part of the user experience. An alternative could be to not care about getting the correct SMS code and instead submit something random, and make sure the API call works as expected (meaning that we see an error message pop up on the screen from Cognito).

If we do it like this, we can rest easy that *our part* (the UI) of the authenticator is doing what it is supposed to do - namely, calling the Auth API. The only thing we miss out on testing is if the *value* is being submitted properly, but if that were the case, user tickets might be the only solution. Unless we do decide to use Twilio for it.

> Sign in federated full flow: We don't currently have a way to test third-party authentication

And I don't think we should. We don't want to test Google/Facebook/Amazon itself.

packages/e2e/features/ui/components/authenticator/sign-in-sms-mfa.feature
```diff
@@ -7,16 +7,25 @@ Feature: Sign In with SMS MFA
...
     Then I see "User does not exist"
+
+  # This feature is not yet implemented
+  @skip
+  Scenario: Expired SMS code
+    When I type a valid pone number "VALID_PHONE_NUMBER"
+    And I type a valid password "VALID_PASSWORD"
+    And I click the "Sign In" button
+    And I enter my confirmation code after some delay
+    Then I see 'Resend Code'
```
eddiekeller, line 31: Not possible with current APIs, AFAIK

packages/e2e/features/ui/components/authenticator/sign-in-totp-mfa.feature
```diff
@@ -7,23 +7,32 @@ Feature: Sign In with TOTP MFA
...
     Then I will be redirected to the setup mfa page
+
+  # This feature is not yet implemented
+  @skip
+  Scenario: Expired TOTP code
+    When I type a valid pone number "VALID_PHONE_NUMBER"
+    And I type a valid password "VALID_PASSWORD"
+    And I click the "Sign In" button
+    And I enter my confirmation code after some delay
+    Then I see 'Resend Code'
```
eddiekeller, line 38: This won't ever be possible, since we don't send TOTP codes ourselves to begin with. They are generated inside a third-party app, like google authenticator.

eddiekeller (changes requested):

ericclemmons:
When we refer to "Resend Code", are we talking about the presentation logic or the behavior logic of actually clicking it?

https://github.com/aws-amplify/amplify-js/blob/836c94aca1676fcd1684f9fcae9587a2da14945e/packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx#L227-L241

eddiekeller:
@ericclemmons That `resendConfirmCode` you linked looks like it is just for the `ConfirmSignUp` flow. I don't think there is an API for resending the MFA SMS code.

packages/e2e/features/ui/components/authenticator/sign-in-force-new-password.feature
```diff
@@ -7,14 +7,14 @@ Feature: Sign In with Force New Password flow
   Background:
     Given I'm running the example "ui/components/authenticator/sign-in-with-phone"
 
-  @react @vue @skip
+  @next @react @vue @skip
```
ericclemmons, line 10: What do you think about adding a `//` comment like we would a TODO to know what's needed for this to work?

```suggestion
  // The `FORCE_CHANGE_NUMBER` env variable isn't in the correct state to test this
  @skip @angular @next @react @vue
```

packages/e2e/features/ui/components/authenticator/sign-in-sms-mfa.feature
ericclemmons: I thought we already had Resend Code on this screen (without the need to enter a confirmation code)

https://github.com/aws-amplify/amplify-js/blob/836c94aca1676fcd1684f9fcae9587a2da14945e/packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx#L227-L241

packages/e2e/features/ui/components/authenticator/sign-in-totp-mfa.feature
ericclemmons: For TOTP, this would be more of a test like:

```
And I enter an invalid code
Then I see 'Some error message saying it is not valid'
```

Right?

packages/e2e/features/ui/components/authenticator/sign-in-with-email.feature
```diff
@@ -32,6 +32,8 @@ Feature: Sign In with Email
     And I click the "Sign In" button
     Then I see "Sign out"
 
+  # FORCE_CHANGE_PASSWORD tests are skipped as the temporary passwords used for these
```
ericclemmons, line 35: What state do these go into when they expire?

packages/e2e/features/ui/components/authenticator/sign-in-with-multi-alias.feature
```diff
@@ -10,25 +10,25 @@ Feature: Sign In with Multiple Aliases
   Background:
     Given I'm at the sign in page
 
-  @react
+  @next @react
```
ericclemmons, line 13: These aren't Angular or Vue because, I assume, it hasn't been implemented there yet?

packages/e2e/features/ui/components/authenticator/sign-up-federated.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Sign Up with Federation
+
+  Background:
+    Given I'm running the example "ui/components/authenticator/sign-in-federated"
+    When I click "Create account"
+
+  @angular @next @react @vue
+  Scenario: Sign up using OAUTH
+		Then I see the "Google" sign in button
```
ericclemmons, line 9: Weird how this did tabs instead of soft-tab spaces, isn't it?

packages/e2e/features/ui/components/authenticator/sign-up-with-phone.feature
```diff
@@ -7,13 +7,14 @@ Feature: Sign Up with Phone
...
 
-@next @react @vue @angular
+  @angular @next @react @vue
   Scenario: Login mechanism set to "phone_number"
     Then I see "Phone Number" as an input field
     And I don't see "Username" as an input field
     And I don't see "Email" as an input field
 
-@next @react @vue @skip
+  # Sign up tests skipped due to SES limits
```
ericclemmons, line 16: Do we have an answer for this?  I assume it's the same thing as customers:
https://docs.aws.amazon.com/ses/latest/dg/manage-sending-quotas-request-increase.html

Which means, should this be documented for them too?

ericclemmons (commented):
This is a really good starting point for our QA session. I have a few Qs as we get closer to that day.

packages/e2e/features/ui/components/authenticator/sign-in-sms-mfa.feature
eddiekeller: Different thing. That's resending a confirm sign up code from Cognito, not an SMS MFA code.

ericclemmons: Ahh thanks! All these flows are blurring together in my mind 😰

packages/e2e/features/ui/components/authenticator/sign-in-with-multi-alias.feature
slaymance: I actually addressed this in my improved credentials pr. The functionality is there for both frameworks, we just needed to set up a proper example that uses the multi-alias backend.

packages/e2e/features/ui/components/authenticator/sign-up-federated.feature
slaymance: Yeah e2 task: style rules for gherkin.

packages/e2e/features/ui/components/authenticator/sign-in-totp-mfa.feature
slaymance: Yeah that's kind of what I had gleaned from the corresponding ticket. Just testing for a general message saying the improper code was entered. I can update this scenario.

slaymance:
> When we refer to "Resend Code", are we talking about the presentation logic or the behavior logic of actually clicking it?
> 
> https://github.com/aws-amplify/amplify-js/blob/836c94aca1676fcd1684f9fcae9587a2da14945e/packages/amplify-ui-components/src/components/amplify-confirm-sign-up/amplify-confirm-sign-up.tsx#L227-L241

> @ericclemmons That `resendConfirmCode` you linked looks like it is just for the `ConfirmSignUp` flow. I don't think there is an API for resending the MFA SMS code.

And we do have that captured in the `confirm-sign-up` tests (though they're skipped for now). I had just misread the ticket for `End user is notified that MFA-SMS has expired after entering a code` so I'll update it to just check that we notify the user that the code is expired/invalid.

packages/e2e/features/ui/components/authenticator/sign-in-with-email.feature
slaymance: They still have a `FORCE_CHANGE_PASSWORD` status in Cognito, and we receive the `remoteError` "Temporary password has expired and must be reset by an administrator."

PR #342, opened 2021-09-07
initial styling for SignUp

eddiekeller:
*Issue #, if available:*

n/a

*Description of changes:*

Initial styling for the `SignUp` page. The "Sign In" button will be removed once the `Tabs` component is ready.

Old sign up page - 
Screen Shot 2021-09-07 at 3 43 22 PM

New sign up page -
Screen Shot 2021-09-07 at 3 44 41 PM

### DIFF
--- a/packages/e2e/cypress/integration/ui/components/authenticator/sign-up-with-email/sign-up-with-email.steps.ts
+++ b/packages/e2e/cypress/integration/ui/components/authenticator/sign-up-with-email/sign-up-with-email.steps.ts
@@ -23,11 +23,11 @@ When('I type the email {string}', (email: string) => {
 });
 
 And('I type the password {string}', (password: string) => {
-  cy.findByLabelText(/^password$/i).type(Cypress.env(password));
+  cy.findByPlaceholderText(/^password$/i).type(Cypress.env(password));
 });
 
 And('I confirm the password {string}', (password: string) => {
-  cy.findByLabelText(/confirm password/i).type(Cypress.env(password));
+  cy.findByPlaceholderText(/confirm password/i).type(Cypress.env(password));
 });
 
 Then('I see {string}', (message: string) => {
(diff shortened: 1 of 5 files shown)


packages/e2e/cypress/integration/ui/components/authenticator/sign-up-with-email/sign-up-with-email.steps.ts
```diff
@@ -23,11 +23,11 @@ When('I type the email {string}', (email: string) => {
 });
 
 And('I type the password {string}', (password: string) => {
-  cy.findByLabelText(/^password$/i).type(Cypress.env(password));
+  cy.findByPlaceholderText(/^password$/i).type(Cypress.env(password));
```
ericclemmons, line 26: Cypress/testing-lib couldn't find the `label` field even though it's visually hidden?

https://github.com/aws-amplify/amplify-ui/blob/0e4fea4fdc3257193b20e4dcc1a1e6a080de2d97/packages/react/src/primitives/Label/Label.tsx#L15-L25

I guess that's by design?

eddiekeller: Yeah, seems that way. I had to make the same changes to the SignIn screen. I guess Cypress is only looking for what it is visually on the page?

reesscot: I found that testing-lib can't find it even when it's visible as well (I got the same error in the unit tests). There appears to be some bug in the library or possibly JSDOM, but I couldn't track it down

ericclemmons: @reesscot Did that testing include the fix to `arria-labelledby`?

From Slack today:

I think the problem is aria-labelledby using the input’s `id` instead of a label `id`:

```html
<input class="amplify-input amplify-field-group__control" id="amplify-field-0108a3625c5f" aria-invalid="false" required="" type="text" aria-labelledby="amplify-field-0108a3625c5f" name="username" placeholder="Fake placeholder" data-amplify-usernamealias="true" value="">
```

> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

slaymance (approved):
Nice, this cleans up `SignUp` a lot!

PR #344, opened 2021-09-07
fix: fixing Alert display styling and remove duplicate scss import

zchenwei:
*Issue #, if available:*

`Alert` is using `Flex` layout. Styling it with `display: block` would override what the layout container does, causing  the close `x`  down to next line

Screen Shot 2021-09-07 at 4 00 59 PM

*Description of changes:*

1. Remove `display: block` in styles
2. Remove duplicate SCSS import

Screen Shot 2021-09-07 at 4 01 25 PM


dbanksdesign (approved):
LGTM! :shipit:

reesscot (approved):
Thanks for fixing!

joebuono (approved):
Thanks for catching this!

PR #346, opened 2021-09-08
Tabs component

joebuono:
*Issue #, if available:*

*Description of changes:*

This PR is ready to go! Adds `Tabs` and `TabItem` primitives that can be used like so:

```jsx
<Tabs>
  <TabItem title="Tab 1">
    Tab 1 content
  </TabItem>
  <TabItem title="Tab 2">
    Tab 2 content
  </TabItem>
</Tabs>
```

Also, this fixes a docs issue where demo CSS overrides were being overridden by the base styling because NextJS doesn't honor import order in JS imports, but does in CSS imports. 

https://user-images.githubusercontent.com/321279/133197969-a9adac31-a06e-469c-ad93-32abbc16a744.mp4


docs/src/components/TabsPropControls.tsx
```diff
@@ -0,0 +1,75 @@
...
+  setJustifyContent,
+}) => {
+  return (
+    <DemoBox primitiveName="Tabs">
+      <FieldLabeler id="defaultTab">
+        <select
+          name="defaultTab"
+          id="defaultTab"
+          value={defaultTab}
+          onChange={(event) =>
```
joebuono, line 33: For some reason, the `defaultTab` prop is not getting applied in the demo, but works in the Usage section of the docs.

docs/src/pages/ui/primitives/tabs/demo.tsx
```diff
@@ -0,0 +1,41 @@
...
+    Cannot click
+  </TabItem>,
+];
+
+export const TabsDemo = () => {
+  const tabsProps = useTabsProps({
+    defaultTab: 0,
+    grow: '',
+    justifyContent: 'flex-start',
+    children: demoChildren.map((child) => child),
```
joebuono, line 22: This was done to avoid a TypeScript error. Perhaps there's a more elegant way to set the children for the demo.

packages/react/package.json
```diff
@@ -27,14 +27,15 @@
...
+    "@radix-ui/react-tabs": "0.0.16",
     "@xstate/react": "^1.4.0",
     "autoprefixer": "^10.3.1",
     "classnames": "^2.3.1",
     "deepmerge": "^4.2.2",
     "nanoid": "^3.1.25",
     "postcss-js": "^3.0.3",
-    "qrcode": "^1.4.4",
-    "@aws-amplify/ui-react-v1": "npm:@aws-amplify/ui-react@1.2.9"
+    "@aws-amplify/ui-react-v1": "npm:@aws-amplify/ui-react@1.2.9",
```
joebuono, line 37: My original PR somehow omitted this package. Eric C told me that it needs to be included.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import {
+  Root,
+  List,
+  Trigger as RadixTab,
+  Content as Panel,
```
joebuono, line 6: Perhaps this renaming isn't necessary. It just helped me in thinking through how to use Radix.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+} from '@radix-ui/react-tabs';
+import { TabsProps, TabItemProps } from '../types';
+import { Flex } from '../Flex';
+import { ComponentClassNames } from '../shared/constants';
+import classNames from 'classnames';
+import { convertStylePropsToStyleObj, prefixer } from '../shared/utils';
+
+const isTabsType = (child: any): child is React.Component<TabItemProps> => {
```
joebuono, line 14: Maybe this function should be in the shared `util` file?

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+  return (
+    typeof child === 'object' &&
+    child.hasOwnProperty('props') &&
+    child.props.title != null &&
+    child.props.children != null
+  );
+};
+export interface Tabs {
```
joebuono, line 22: Although TypeScript doesn't complain, I don't think the interface and the component should both be named `Tabs`.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+  gap = '0',
+  grow,
+  justifyContent,
+  wrap,
+  ...rest
+}) => {
+  const tabs = React.Children.map(children, (child) => {
+    if (!isTabsType(child)) {
+      console.warn(
+        'Amplify UI: <Tabs> component only accepts <TabItem> as children.'
```
joebuono, line 43: I'm not sure what exactly our error messages should look like.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+  className,
+  defaultTab = 0,
+  direction,
+  gap = '0',
+  grow,
+  justifyContent,
+  wrap,
+  ...rest
+}) => {
+  const tabs = React.Children.map(children, (child) => {
```
joebuono, line 40: TypeScript wouldn't let me simply `map` over the `children` directly, so I used this as a workaround.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+    if (!isTabsType(child)) {
+      console.warn(
+        'Amplify UI: <Tabs> component only accepts <TabItem> as children.'
+      );
+      return null;
+    }
+
+    return child.props;
```
joebuono, line 48: The reason I returned the entire `child.props` object is because I wanted to allow the customer to pass down an arbitrary number of style props to the tabs.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  return (
+    <Root defaultValue={`${defaultTab}`}>
```
joebuono, line 52: Radix accepts the `defaultValue` as a `string`.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <List aria-label={ariaLabel}>
```
joebuono, line 53: Maybe `aria-label` should be required?

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+          wrap={wrap}
+          {...rest}
+        >
+          {tabs.map(({ className, isDisabled, title, ...rest }, index) => (
+            <RadixTab
+              className={classNames(ComponentClassNames.TabItems, className)}
+              data-grow={grow}
+              disabled={isDisabled}
+              key={index}
+              style={prefixer(convertStylePropsToStyleObj(rest, {}))}
```
joebuono, line 70: I first tried using the `View` primitive to take care of this, but I ran into issues with the style props getting passed to the `RadixTab`. There's probably a better way to pass style props to each tab.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+      {tabs.map((tab, index) => (
+        <Panel key={index} value={`${index}`}>
+          {tab.children}
+        </Panel>
+      ))}
+    </Root>
+  );
+};
+
+export const TabItem: React.FC<TabItemProps> = () => <></>;
```
joebuono, line 87: The `TabItem` item is essentially a shell component used as syntactic sugar. An example from React Spectrum inspired me to implement it this way, which I feel is more visually intuitive and offers a better DX. However, there are tradeoffs in terms of styling and customization.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import {
+  Root,
+  List,
+  Trigger as RadixTab,
+  Content as Panel,
+} from '@radix-ui/react-tabs';
```
joebuono, line 7: The main reasons for using [Radix](https://www.radix-ui.com/docs/primitives/components/tabs) is because it adheres to the Tabs WAI-ARIA design pattern and is also a small package size (3.35 kB).

packages/react/src/primitives/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { render, screen } from '@testing-library/react';
+import { Tabs, TabItem } from '../Tabs';
+import { ComponentClassNames } from '../../shared';
+import { ComponentPropsToStylePropsMap } from '../../types';
+import { kebabCase } from 'lodash';
+
+describe('Tabs: ', () => {
```
joebuono, line 7: Regretfully, I've yet to write tests for Tabs. Here are some ideas:

Tabs should:
- only accept `TabItem` components as children
- initially display the content panel of a specific tab based on the `defaultTab` prop, which is an index that defaults to zero (the first tab)
- apply different `className`s at the `Tabs` level and the `TabItem` level
- take up remaining space differently depending on whether `equal` or `relative` was passed to the `grow` prop
- apply any `FlexStyleProps`

TabItems should:
- accept a string or custom component (e.g., `<Text/>`) as the `title` prop
- render certain tabs unclickable based on the `isDisabled` prop

Is there a way to test Accessibility? Radix should be taking care of it though.

packages/react/src/primitives/types/tabs.ts
```diff
@@ -0,0 +1,39 @@
...
+    FlexStyleProps {
+  /**
+   * The Tabs component only accepts TabItem components as children.
+   */
+  children: React.ReactElement | React.ReactElement[];
+
+  /**
+   * Change which Tab content is initially displayed. Pass in the index of the Tab you wish to show. The default is index 0 (the first tab).
+   */
+  defaultTab?: number;
```
joebuono, line 21: Perhaps `defaultIndex` is a better name?

packages/ui/src/theme/tokens/components/tabs.js
```diff
@@ -0,0 +1,36 @@
+module.exports = {
```
joebuono, line 1: All of these styles are my best guesses. Definitely needs review from @dbanksdesign.

packages/ui/src/theme/css/component/tabs.scss
```diff
@@ -0,0 +1,66 @@
...
+  &[data-disabled] {
+    background-color: var(
+      --amplify-components-tabs-item-disabled-background-color
+    );
+    border-color: var(--amplify-components-tabs-item-disabled-border-color);
+    color: var(--amplify-components-tabs-item-disabled-color);
+    cursor: not-allowed;
+  }
+
+  &[data-grow='equal'] {
```
joebuono, line 59: Perhaps this is a bit hacky, but it does make each tab take up equal space.

packages/ui/src/theme/css/component/tabs.scss
```diff
@@ -0,0 +1,66 @@
+.amplify-tabs {
+  background-color: var(--amplify-components-tabs-background-color);
+  box-shadow: var(--amplify-components-tabs-box-shadow);
+  margin-bottom: var(--amplify-components-tabs-margin-bottom);
+}
+
+.amplify-tabs-item {
```
joebuono, line 7: I wanted to name this class `.amplify-tab-item`, but then we wouldn't be able to include it in the same `.js` file as `.amplify-tabs` for Style Dictionary.

packages/react/src/primitives/Tabs/Tabs.tsx
reesscot: Makes sense to me

packages/react/src/primitives/Tabs/Tabs.tsx
reesscot: In most cases I would agree, but for function interfaces, I think it's logical for them to have the same name

packages/react/src/primitives/types/tabs.ts
dbanksdesign: +1 to `defaultIndex`

docs/src/components/TabsPropControls.tsx
reesscot: I think that's because once the tabs have rendered, that prop is no longer acted upon.

packages/react/src/primitives/Tabs/Tabs.tsx
dbanksdesign: LGTM

packages/ui/src/theme/tokens/components/tabs.js
dbanksdesign: I went through and updated the styles, only a few minor tweaks but overall great job @joebuono !

packages/react/src/primitives/Tabs/Tabs.tsx
dbanksdesign: I think if it is not used in any other file, it is fine to have it in this file only.

docs/src/components/TabsPropControls.tsx
dbanksdesign: Updated the demo to use `currentIndex` and `onChange` so the demo controls work!

docs/src/pages/ui/primitives/tabs/demo.tsx
dbanksdesign: I wasn't getting this Typescript error, removed this map and everything works!

docs/src/components/Example.tsx
```diff
@@ -1,6 +1,8 @@
-export function Example({ children }) {
+export function Example({ children, className = '' }) {
```
dbanksdesign, line 1: @ericclemmons added this so we can scope CSS styles to an example to make the actual example code more closely reflect the actual code.

docs/src/pages/_app.page.tsx
```diff
@@ -1,7 +1,5 @@
 import Head from 'next/head';
-import '../styles/styles.css';
-import '@aws-amplify/ui-react/styles.css';
-import '../styles/primitivesStyles.css';
+import '../styles/index.css';
```
dbanksdesign, line 2: @ericclemmons NextJS was including these styles in incorrect order, breaking the docs example CSS. Changed this to a single CSS import and handled ordering of other imports there.

docs/src/pages/ui/primitives/tabs/demo.tsx
```diff
@@ -0,0 +1,52 @@
...
+          indicatorPosition={tabsProps.indicatorPosition}
+        >
+          {tabsProps.children}
+        </Tabs>
+      </Example>
+    </Flex>
+  );
+};
+
+export const ControlledTabExample = () => {
```
dbanksdesign, line 40: Controlled example 💪

docs/src/pages/ui/primitives/tabs/index.page.mdx
```diff
@@ -0,0 +1,7 @@
+---
+title: Tabs
+---
+
+import { Fragment } from '@/components/Fragment';
+
+<Fragment>{({ platform }) => import(`./${platform}.mdx`)}</Fragment>
```
dbanksdesign, line 7: TODO: will do a cleanup of all primitives docs pages to use multiple fragments rather than a single one.

docs/src/styles/styles.css
```diff
@@ -1,6 +1,3 @@
-@import 'prism-theme-night-owl';
```
dbanksdesign, line 1: Moved these imports to index.css, which will only import files.

docs/src/styles/tabsStyles.css
```diff
@@ -0,0 +1,44 @@
...
+  background-color: var(--amplify-colors-brand-primary-90);
+}
+
+.custom-tab-item[data-state='active'] {
+  color: var(--amplify-colors-font-inverse);
+  border-color: transparent;
+  background-color: var(--amplify-colors-brand-primary-100);
+}
+
+.pseudo-demo .amplify-tabs-item:hover {
```
dbanksdesign, line 24: Using example classnames and descendent selectors to scope example CSS

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,102 @@
...
+      return null;
+    }
+
+    return child.props;
+  });
+
+  // mapping our props to Radix's props
+  // value (currentIndex) and defaultValue (defaultIndex) must be strings
+  // https://www.radix-ui.com/docs/primitives/components/tabs#api-reference
+  const rootProps = {
```
dbanksdesign, line 57: 👀 mapping our props to Radix's and only setting `value` if `currentIndex` is defined.

dbanksdesign (commented):
Ok let's get this looked at and get this in!

docs/src/components/Example.tsx
reesscot: Nice, this is great!

docs/src/pages/ui/primitives/tabs/index.page.mdx
reesscot: This will be especially useful if we can share primitive documentation text and swap out the code based on platform, like is done in Authenticator.

docs/src/components/TabsPropControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <FieldLabeler id="justifyContent">
+        <select
+          name="justifyContent"
+          id="justifyContent"
+          value={justifyContent}
+          onChange={(event) =>
+            setJustifyContent(event.target.value as TabsProps['justifyContent'])
+          }
+        >
+          <option value="flex-start">flex-start</option>
```
reesscot, line 88: TODO: (not in this PR) We should extract the Flex Styles dropdown into reusable components across our demos.

docs/src/pages/ui/primitives/tabs/index.page.mdx
dbanksdesign: Yup!

docs/src/components/TabsPropControls.tsx
dbanksdesign: Added an EE task for this!

eddiekeller (approved):

eddiekeller:
Thanks for working on this! Excited to get these primitive in! :) :)

reesscot (changes requested):
Looks great. Just had a few comments/questions on CSS import order and docs.

PR #347, opened 2021-09-08
Re-export AmplifyAuthenticator (legacy)

ericclemmons:
For customers that are incrementally adopting the new UI components (e.g. `View`, `Button`), re-export the `AmplifyAuthenticator` for use.

> Screen Shot 2021-09-08 at 12 37 41 PM
> – http://localhost:3000/ui/components/amplify-authenticator/sign-in-with-email

What crossed my mind was also re-exporting `withAuthenticator`, which would be backwards-compatible:
> https://github.com/aws-amplify/amplify-js/blob/main/packages/amplify-ui-react/src/withAuthenticator.tsx

But then about the new `Authenticator`'s `withAuthenticator`?  A couple options crossed my mind:

1. New `Authenticator` & `withAuthenticator` from `@aws-amplify/ui-react`
	a. Legacy re-exported as `AmplifyAuthenticator` & `withAmplifyAuthenticator`.
2. Only export new packages from `@aws-amplify/ui-react`
	a. Re-export all legacy packages from `@aws-amplify/ui-react/legacy`

My personal preference is 2️⃣ , because it avoids namespace collisions & the confusing `Amplify*` prefixes to component names.

---


examples/next/pages/ui/components/amplify-authenticator/sign-in-with-email/index.page.tsx
```diff
@@ -0,0 +1,15 @@
+import { AmplifyAuthenticator, translations } from '@aws-amplify/ui-react';
```
ericclemmons, line 1: This could become:

```suggestion
import { AmplifyAuthenticator } from '@aws-amplify/ui-react/legacy';
```

examples/next/pages/ui/components/amplify-authenticator/sign-in-with-email/index.page.tsx
```diff
@@ -0,0 +1,15 @@
...
+import awsExports from '@environments/auth-with-email/src/aws-exports';
+import { Amplify, I18n } from 'aws-amplify';
+
+Amplify.configure(awsExports);
+
+I18n.putVocabularies(translations);
```
ericclemmons, line 7: This should be removed, since the legacy translations aren't compatible with the new ones:

```suggestion
```

packages/react/src/index.tsx
```diff
@@ -11,6 +11,10 @@ export * as primitives from './primitives';
 export { theme, translations } from '@aws-amplify/ui';
 
 export {
+  /**
+   * @deprecated `AmplifyAuthenticator` is the v1 release. Prefer `Authenticator` for continued support.
+   */
+  AmplifyAuthenticator,
   AmplifyChatbot,
   AmplifyPhotoPicker,
   AmplifyPicker,
```
ericclemmons, line 20: All legacy packages could be re-exported from `@aws-amplify/ui-react/legacy`.

ericclemmons (commented):
I'm more & more interested in a `/legacy` subfolder across all frameworks, since we learned that re-exporting *can* work.

packages/react/src/index.tsx
eddiekeller: I'm in favor of exporting from an explicit `legacy` path.

examples/next/pages/ui/components/amplify-authenticator/sign-in-with-email/index.page.tsx
hvergara: what would be the actual benefit? Cleaner exports?

packages/react/src/index.tsx
hvergara: I also like the idea to keep it as separated as possible, but IMO that could break existing users that rely on AmplifyAuthenticator (like customers depending on `@aws-amplify/ui-react`: "latest"`.

Is there a export symbol name conflict that would force us to do it? Or tree-shaking considerations?

examples/next/pages/ui/components/amplify-authenticator/sign-in-with-email/index.page.tsx
ericclemmons: Yes, but mostly (per the description) allowing us to re-export the new `withAuthenticator` **and** the legacy `withAuthenticator`.

This would give you 1 package with both new & legacy to use interchangeably.

packages/react/src/index.tsx
ericclemmons: `@aws-amplify/ui-react@2` is already going to be a breaking change:

- `Authenticator` is new whereas the previous version had the prefix `AmplifyAuthenticator`.
- Plus, `withAuthenticator` isn't backwards-compatible (`usernameAlias` and `formFields` are gone).
- Neither are our CSS variables for theming.

For customers who use `latest` (which would be a bit rare, since npm & yarn both resolve to version number (plus lock files prevent this):

```console
$ npm install @aws-amplify/ui-react@latest --save
```

```json
  "dependencies": {
    "@aws-amplify/ui-react": "^1.2.14"
  }
```

ericclemmons: If we don't need to do this, then we don't have to.  This was something I stumbled upon since we're already re-exporting `AmplifyS3Image` for customers to use alongside the new Authenticator.

But I really like the idea of re-exporting under `/legacy` because it's near 0 cost to do so and we can continue creating net-new components/primitives without worrying about "cleaning up" our exports.

ErikCH: I like this `/legacy` . It will break in 2.0 but most people should expect that.

examples/next/pages/ui/components/amplify-authenticator/sign-in-with-email/index.page.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export default function AuthenticatorWithUsername() {
```
ErikCH, line 9: Should this be `AuthenticatorWithEmail` ?

packages/react/src/index.tsx
hvergara: thanks for providing so much information & context to understand the issue, I'm convinced now about using `/legacy` now. Are you going to include that change on this PR or will create a separate one?

examples/next/pages/ui/components/amplify-authenticator/sign-in-with-email/index.page.tsx
hvergara: got it, thanks!

packages/react/legacy/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "private": true,
+  "name": "@aws-amplify/ui-react-legacy",
+  "main": "../dist/legacy.js",
+  "module": "../dist/esm/legacy.js",
+  "types": "../dist/legacy.d.ts"
```
ericclemmons, line 6: This file is similar to https://github.com/aws-amplify/amplify-js/blob/82544a6c8e0c0ae30f988b066e7e250f49862f64/packages/amplify-ui-components/loader/package.json in that helps TypeScript find `@aws-amplify/ui-react/legacy`'s types, since `types` can only be a string, not an array.

But this also helps for CJS bundlers that don't support the `exports` field.

packages/react/package.json
```diff
@@ -5,6 +5,7 @@
   "module": "dist/esm/index.js",
   "exports": {
     ".": "./dist/esm/index.js",
+    "./legacy": "./dist/esm/legacy.js",
```
ericclemmons, line 8: Despite `legacy/package.json`, ESM treats `exports` as an allowlist.

packages/react/src/legacy.tsx
```diff
@@ -0,0 +1,14 @@
...
+  AmplifyChatbot,
+  AmplifyPhotoPicker,
+  AmplifyPicker,
+  AmplifyS3Album,
+  AmplifyS3Image,
+  AmplifyS3ImagePicker,
+  AmplifyS3Text,
+  AmplifyS3TextPicker,
+  /** @deprecated `withAuthenticator` from legacy. Prefer the latest `withAuthenticator` release for continued support. */
+  withAuthenticator,
```
ericclemmons, line 13: Technically all `Amplify*` components are deprecated, but without an alternative implementation they're technically supported.

packages/react/src/index.tsx
ericclemmons: Included & updated!

packages/react/package.json
ErikCH: Wow, I'eve never seen this, interesting

ErikCH (approved):
Looks good. Also glad we are testing all the imports 👍

wlee221 (approved):
LGTM, I agree with the second approach in your pr description as well.

PR #333, opened 2021-09-02
feat(core): Export send alises for common xstate events

wlee221:
*Issue #, if available:*

*Description of changes:* This PR exports facade functions that developers can use to send common auth events without having to know xstate internals.

Specifically, this adds `getSendEventAliases` that returns an object of "helper" facade functions that can be used by package consumers. React for example:

```tsx
export const useAuthenticator = () => {
  import { authService, getSendEventAliases } from '@aws-amplify/ui';

  const [state, send] = useActor(authService);
  const { user } = state.context;
  const helpers = getSendAliases(send);
  /**
    * helpers = {
    *   signIn: (data?: AuthEventData) => send({ type: "SIGN_IN", data }),
    *   resend: (data?: AuthEventData) => send({ type: "RESEND", data }),
    *   ...
    * }
    */
  return { state,  user, ...helpers };
}
```

Then customer can use that hook/service to use those helpers without having to know xstate internals. 

```tsx
import { Authenticator, useAuthenticator } from "@aws-amplify/ui-react"

export default function App() {
  const { state, user, signOut } = useAuthenticator()
  
  if (state !== 'authenticated') {
    return <Authenticator />
  }
  
  return (
    <>
      <h1>Welcome {user.username}!</h1>
      <button onClick={signOut}>Sign Out</button>
    </>
  )
}
```

For this PR, I'm just adding this helper function that each framework can share. Each framework expert can take this on to make public facing hooks 👍  

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #349, opened 2021-09-09
feat(Angular): expose relevant context from each template outlet

wlee221:
*Issue #, if available:*

*Description of changes:* This adds context to every template outlet that customers can use to get `helper` functions or relevant context like `user`.

This enables in-template use cases like 

```tsx
<amplify-authenticator>
  <ng-template amplifyOverride="authenticated" let-user="user">
    <h1>Welcome, {{ user.username }}</h1>
  </ng-template>
</amplify-authenticator>
```

where `let-user="user"` is coming from template outlet context.

### DIFF
--- a/examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
+++ b/examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
@@ -1,1 +1,10 @@
-<amplify-authenticator></amplify-authenticator>
+<amplify-authenticator>
+  <ng-template
+    amplifyOverride="authenticated"
+    let-user="user"
+    let-signOut="signOut"
+  >
+    <h2>Welcome, {{ user.username }}!</h2>
+    <button (click)="signOut($event)">Sign Out</button>
+  </ng-template>
+</amplify-authenticator>
--- a/examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
+++ b/examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
@@ -5,6 +5,6 @@
     let-signOut="signOut"
   >
     <h2>Welcome, {{ user.username }}!</h2>
-    <button (click)="signOut($event)">Sign Out</button>
+    <button (click)="signOut()">Sign Out</button>
   </ng-template>
 </amplify-authenticator>
--- a/packages/angular/projects/ui-angular/src/lib/components/amplify-authenticator/amplify-authenticator.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/amplify-authenticator/amplify-authenticator.component.ts
@@ -50,9 +50,9 @@ export class AmplifyAuthenticatorComponent implements AfterContentInit {
    * Class Functions
    */
   public get context() {
-    const user = this.stateMachine.user;
     const { signOut } = this.stateMachine.services;
-    return { user, signOut };
+    const user = this.stateMachine.user;
+    return { signOut, user };
   }
 
   public get actorState() {
--- a/packages/angular/projects/ui-angular/src/lib/components/amplify-authenticator/amplify-authenticator.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/amplify-authenticator/amplify-authenticator.component.ts
@@ -49,6 +49,12 @@ export class AmplifyAuthenticatorComponent implements AfterContentInit {
   /**
    * Class Functions
    */
+  public get context() {
+    const user = this.stateMachine.user;
+    const { signOut } = this.stateMachine.services;
+    return { user, signOut };
+  }
+
   public get actorState() {
     return getActorState(this.stateMachine.authState);
   }
@@ -67,11 +73,5 @@ export class AmplifyAuthenticatorComponent implements AfterContentInit {
     });
 
     return customComponents;
-  }
-
-  public get context() {
-    const user = this.stateMachine.user;
-    const { signOut } = this.stateMachine.services;
-    return { user, signOut };
   }
 }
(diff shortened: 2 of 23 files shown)


examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
```diff
@@ -1 +1,10 @@
...
+<amplify-authenticator>
+  <ng-template
+    amplifyOverride="authenticated"
+    let-user="user"
+    let-signOut="signOut"
+  >
+    <h2>Welcome, {{ user.username }}!</h2>
+    <button (click)="signOut($event)">Sign Out</button>
+  </ng-template>
+</amplify-authenticator>
```
wlee221, line 10: This is how you can do sign-out within template. Note that we don't need to import any services to do so! Service can be imported for more headless or advanced usage like `state.matches` (or whichever facade we will provide).

packages/angular/projects/ui-angular/src/lib/components/amplify-authenticator/amplify-authenticator.component.ts
```diff
@@ -54,6 +49,12 @@ export class AmplifyAuthenticatorComponent implements AfterContentInit {
   /**
    * Class Functions
    */
+  public get context() {
+    const { signOut } = this.stateMachine.services;
+    const user = this.stateMachine.user;
+    return { signOut, user };
+  }
```
wlee221, line 56: I used typescript getters to provide context, because 1) it reevaluates context on every request, which will pick up changes in `user` or other non-static contexts and 2) templates can simply write `context` instead of `context()`.

examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
ericclemmons: Does `$event` need to be passed, or is this best-practices because this is an `onClick` HTML event?

examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
```diff
@@ -1 +1,10 @@
-<amplify-authenticator></amplify-authenticator>
+<amplify-authenticator>
+  <ng-template
+    amplifyOverride="authenticated"
+    let-user="user"
+    let-signOut="signOut"
```
ericclemmons, line 5: In the future, if we have a react hook like `useAuthenticator()` that returns a `{ user, signOut }`, could this template become:

```html
<amplify-authenticator>
  <ng-template amplifyOverride="authenticated" let-context="authenticator">
    <h2>Welcome, {{ authenticator.user.username }}!</h2>
    <button (click)="authenticator.signOut($event)">Sign Out</button>
  </ng-template>
</amplify-authenticator>
```

I'm looking to understand if having N `let-*` attributes is common or not in Angular :)

packages/angular/projects/ui-angular/src/lib/components/amplify-authenticator/amplify-authenticator.component.ts
ericclemmons: In this case, do we want `const signOut = () => send('SIGN_OUT')` vs. calling the service directly?

Otherwise this is the equivalent of `Auth.signOut()` and the Authenticator won't know to respond?

ericclemmons: OOh, nevermind – this is from #333 using the Angular state machine service, not _XState's_ `.services` property 🤦

ericclemmons (approved):
Looking good!  I had to basically talk through some of these changes because `stateMachine.services` makes me think _XState_ `createMachine({ services })`, not the Angular-specific service that uses the new facades from #333.

examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
wlee221: This is fairly new, so this is not so common in the Angular wild. I could find `let-` usage in angular-bootstrap: https://ng-bootstrap.github.io/#/components/datepicker/overview#day-template.

Yup that's totally possible too! Or even: 

```html
<amplify-authenticator>
  <ng-template amplifyOverride="authenticated" let-authenticator>
    <h2>Welcome, {{ authenticator.user.username }}!</h2>
    <button (click)="authenticator.signOut($event)">Sign Out</button>
  </ng-template>
</amplify-authenticator>
```
This looks like a great idea actually! I'll consider this.

For more finer control, I still expect customers to use `AuthenticatorService` directly instead of `let-` so that they have programmatic control over these files. `let-` is more for convenience for fetching and using related objects directly. I can write a quick use case document after this pr.

examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
wlee221: Oops, nope! Unless you want to do `$event.preventDefault()`, which is not really necessary here. I'll remove it for simplicity.

examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
```diff
@@ -1 +1,10 @@
...
+  >
+    <h2>Welcome, {{ user.username }}!</h2>
+    <button (click)="signOut($event)">Sign Out</button>
```
wlee221, line 8:
```suggestion
    <button (click)="signOut()">Sign Out</button>
```

packages/angular/projects/ui-angular/src/lib/components/amplify-authenticator/amplify-authenticator.component.ts
wlee221: Right! These are aliased helpers ✨  I'll clarify the naming conventions next sprint.

examples/angular/src/pages/ui/components/authenticator/sign-up/sign-up.component.html
wlee221: I'll put this into our facade task after qa. But that above code snippet looks promising for in-template use cases.

PR #350, opened 2021-09-09
chore: ignore .npmrc

wlee221:
*Issue #, if available:*

*Description of changes:* This adds `.npmrc` to `.npmignore`. `.npmrc` is used in our deployment procedure to authenticate ourselves using npm token. I created this PR to be safe.


.gitignore
```diff
@@ -5,3 +5,4 @@
 node_modules
 aws-exports.js
 aws-exports.ts
+.npmrc
```
ericclemmons, line 8: Can you add a comment of why?  `.npmrc` files are valid (like `.yarnrc` here: https://github.com/aws-amplify/amplify-js/blob/main/.yarnrc) for defaults across teams.

In the future, if we want to customize this and say, "oops! this couldn't be ignored!", we won't remove it accidentally.

ericclemmons (approved):
Makes sense to me!  May be worth having a comment on why though, since `.npmrc` is common for team-based repos.

.gitignore
wlee221: Good point, I'll do that.
```suggestion
# .npmrc is used in our deployment procedure to login to npm. Ignored for security.
.npmrc
```

PR #351, opened 2021-09-09
build(angular): move publishConfig to ui-angular package.json

wlee221:
*Issue #, if available:*

*Description of changes:* This is a follow up from #204. `publishConfig` should be located where `@aws-amplify/ui-angular` `package.json` is, as this is what `changeset` will reference when we run `changeset publish`. Previously this was located in angular workspace root `amplify-ui-angular-mono`, which changeset ignores because it's private (as desired).
 

ericclemmons (approved):
Totally makes sense – we're having to fit within changeset's expectations.

PR #353, opened 2021-09-09
initial styling for ConfirmSignUp

eddiekeller:
*Issue #, if available:*

https://app.asana.com/0/1200881947262804/1200935010977983/f

*Description of changes:*

Initial styling for the `ConfirmSignUp` screen. Also made some tweaks to the `signIn` and `signUp` machines in order to make the UX a bit better and remove the extra username input from the `ConfirmSignUp` screen.

Screen Shot 2021-09-09 at 11 24 02 AM

### DIFF
--- a/packages/ui/src/machines/actors/auth/signUp.ts
+++ b/packages/ui/src/machines/actors/auth/signUp.ts
@@ -140,7 +140,6 @@ export const signUpActor = createMachine<SignUpContext, AuthEvent>(
       resolved: {
         type: 'final',
         data: (context) => {
-          console.log(context);
           const { username, password } = context.authAttributes;
           const canAutoSignIn = !!(username && password);
           return {
--- a/packages/ui/src/machines/actors/auth/signUp.ts
+++ b/packages/ui/src/machines/actors/auth/signUp.ts
@@ -1,3 +1,4 @@
+import { get } from 'lodash';
 import { createMachine, sendUpdate } from 'xstate';
 
 import { Auth } from 'aws-amplify';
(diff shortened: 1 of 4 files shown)


packages/ui/src/machines/actors/auth/signUp.ts
```diff
@@ -167,13 +168,19 @@ export const signUpActor = createMachine<SignUpContext, AuthEvent>(
...
     },
     services: {
-      async confirmSignUp(_, event) {
-        const { username, confirmation_code: code } = event.data;
+      async confirmSignUp(context, event) {
+        const { user, authAttributes } = context;
+        const { confirmation_code: code } = event.data;
+
+        const username =
+          get(user, 'username') || get(authAttributes, 'username');
```
ericclemmons, line 176: Can you share more about what's going on here? I can't recall where `user.username` and `authAttributes.username` both come from for this screen.

(It may be worth a code comment so we can understand why the precedence exists)

ericclemmons (approved):
Looks good!  I trust your implementation, but just had a question so I can understand the _why_ behind how we derive `username`.

packages/ui/src/machines/actors/auth/signUp.ts
eddiekeller: `user` and `authAttributes` get set in different places depending on which screen we are coming from - either `SignIn` or `SignUp`. Both screens can redirect to this `ConfirmSignUp` screen, so we need to check in a couple places for the `username` value. If the `user` value is set into the context (via `SignUp`), then we will use that, but if we are coming from `SignIn`, then we need to grab the value from the `authAttributes` instead.

Me using lodash's `get` is just preference here instead of dealing with nullish coalescense stuff.

PR #354, opened 2021-09-09
initial styling for ResetPassword flow

eddiekeller:
*Issue #, if available:*

https://app.asana.com/0/1200881947262804/1200935010977981/f

*Description of changes:*

Initial styling pass through for the reset password flow. This flow is actually not in the mocks for some reason, so I'm just following the example of other screens.

!!! Fixed the title string in the screenshots below so both should be the same now (`Reset your password`). Don't feel like taking a new screenshot though.

Screen Shot 2021-09-09 at 1 29 18 PM

Screen Shot 2021-09-09 at 1 29 25 PM

### DIFF
--- a/packages/react/src/components/Authenticator/ResetPassword/ResetPassword.tsx
+++ b/packages/react/src/components/Authenticator/ResetPassword/ResetPassword.tsx
@@ -7,7 +7,7 @@ import { ErrorText, TwoButtonSubmitFooter } from '../shared';
 export const ResetPassword = (): JSX.Element => {
   const amplifyNamespace = 'Authenticator.ResetPassword';
   const {
-    components: { Fieldset, Footer, Form, Heading, Input, Label, Text },
+    components: { FieldGroup, Flex, Form, Heading, TextField },
   } = useAmplify(amplifyNamespace);
 
   const [state, send] = useAuth();
@@ -29,6 +29,8 @@ export const ResetPassword = (): JSX.Element => {
     });
   };
 
+  const inputLabel = I18n.get('Enter your username');
+
   return (
     <Form
       data-amplify-authenticator-resetpassword=""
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Authenticator/ResetPassword/ResetPassword.tsx
```diff
@@ -29,6 +29,8 @@ export const ResetPassword = (): JSX.Element => {
     });
   };
 
+  const inputLabel = I18n.get('Enter your username');
```
ericclemmons, line 32: Should this have a `// TODO` to indicate we want this to say `Enter your email` or `Enter your username` or `Enter your phone number` based on the primary login mechanism?

packages/react/src/components/Authenticator/ResetPassword/ResetPassword.tsx
```diff
@@ -46,29 +48,30 @@ export const ResetPassword = (): JSX.Element => {
...
+      <Flex direction="column">
+        <Heading level={3}>{headerText}</Heading>
 
-      <Fieldset disabled={isPending}>
-        <Label data-amplify-resetpassword-label="">
-          <Text>{I18n.get('Username')}</Text>
-          <Input
+        <FieldGroup direction="column" disabled={isPending}>
+          <TextField
             autoComplete="username"
```
ericclemmons, line 56: This `autoComplete` value will depend on the primary login mechanism, right?

ericclemmons (approved):
@eddiekeller You're quick af! I just had a question about what the username usage will eventually turn into.

Looking good!

packages/react/src/components/Authenticator/ResetPassword/ResetPassword.tsx
eddiekeller: Eh. It'll come up in the QA session, right?

packages/react/src/components/Authenticator/ResetPassword/ResetPassword.tsx
eddiekeller: Mmm I suppose it will, yes. So when we update the string value, we can update this too

PR #358, opened 2021-09-10
Improve Test User Credentials

slaymance:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200881947262804/1200615823577439)

*Description of changes:* This PR simplifies test user credentials for e2e tests. Instead of having a different environment variable for each user with a given status, this PR instead will dynamically compute a test user's alias based on the `loginMechanism` and Cognito `status` specified in the test. It also does a few other things:

- Adds framework feature tags where applicable for better coverage
- Removes `@skip` tag in favor of leaving off feature tags. `@skip` is better used for local testing and shouldn't be committed.
- Removes `@react` tag in favor of `@next`
- Adds `auth-with-multi-alias` examples for Angular and Vue
- Standardizes a lot of step definitions for reuse


.github/workflows/tests.yml
```diff
@@ -96,32 +96,12 @@ jobs:
       - name: Run E2E tests against ${{ matrix.framework }} example
         run: yarn workspace e2e test
         env:
-          # Override on the default value in `cypress.json` with `@react and not @skip`
-          TAGS: '@${{ matrix.framework }} and not @skip'
```
slaymance, line 100: Let's move to using `@skip` for local testing only. If we want to merge tests that aren't run for a specific framework, then we should just leave that framework's tag off the test.

packages/e2e/features/ui/components/authenticator/reset-password.feature
```diff
@@ -7,16 +7,16 @@ Feature: Reset Password
   Background:
     Given I'm running the example "ui/components/authenticator/sign-in-totp-mfa"
 
-  @next @react @vue @angular
```
slaymance, line 10: We can remove `@react` since we use `@next` to run our e2e tests against our Next examples.

packages/e2e/cypress/support/commands.ts
```diff
@@ -0,0 +1,67 @@
...
+const countryCodeByStatus = (status: string) => {
+  switch (status) {
+    case 'CONFIRMED':
+      return '1';
+    case 'UNKNOWN':
+      return '2';
+    case 'UNCONFIRMED':
+      return '3';
+  }
+};
```
slaymance, line 45: I may revisit these helpers function. Maybe they should live in a separate `helpers.ts` file?

.github/workflows/tests.yml
wlee221: I like this! For running tests locally in dev, how would I set this environment variable?

`TAGS=@angular yarn e2e dev`?

packages/e2e/cypress/integration/common/index.tsx
```diff
@@ -1,19 +1,11 @@
...
-Given("I'm running the example {string}", (url) => {
-  cy.visit(url);
-});
-
-When('I type a valid username {string}', (username: string) => {
-  cy.findByRole('textbox', { name: /username/i }).type(Cypress.env(username));
-});
-
-And('I type a valid password {string}', (password: string) => {
-  cy.findByLabelText(/password/i).type(Cypress.env(password));
```
wlee221, line 16: I love seeing codes deleted ✨

packages/e2e/cypress/support/commands.ts
wlee221: I like the idea of having `helpers.ts` file that we can put constants in.

packages/e2e/cypress/support/commands.ts
```diff
@@ -0,0 +1,67 @@
...
+    const buildAlias = cond([
+      [eq('username'), constant(appendStatusToAlias(status))],
+      [
+        eq('email'),
+        constant(`${appendStatusToAlias(status)}@${Cypress.env('DOMAIN')}`),
+      ],
+      [
+        eq('phone number'),
+        constant(
+          `+${countryCodeByStatus(status)}${Cypress.env('PHONE_NUMBER')}`
```
wlee221, line 60: How are we selecting country code with this helper? My understanding was that country code was selected through a dropdown.

wlee221 (approved):
I just appreciate how this change only adds +18 line total. LGTM 🚢

.github/workflows/tests.yml
slaymance: You can do `yarn e2e dev -e "TAGS=@angular and not @skip"`. And to run the exact script CI runs you can do `yarn e2e test -e "TAGS=@angular"`

packages/e2e/cypress/support/commands.ts
slaymance: Basically I just arbitrarily assigned them like this:
```
+1 === 'CONFIRMED'
+2 === 'UKNOWN'
+3 === 'UNCONFIRMED'
```
and each once corresponds to a test user in a backend configured with `phone_number`. Not stoked about this solution, but it at least simplifies how many phone numbers we have as environment variables for now.

.github/workflows/tests.yml
ericclemmons: I agree that this is the right decision long-term.

`@focus` is what I use locally, but `@skip` is also a helpful tool in the toolbelt.

The only problem with "leave that framework's tag off the test" is that it means _we won't show content for unsupported use-cases_.

This is where the dilemma is:

1. `.feature`s describe supported use-cases
2. The `@react` framework tag indicates that this is _supported behavior for React_ (React & Next.js aren't always 1-to-1)
3. But, we recognize that some behavior **cannot be tested in CI**.

So it sounds like we may need an alternative solution here to differentiate those two – perhaps with an explicit `@skip-ci` tag?

packages/e2e/cypress/support/commands.ts
ericclemmons: If it's specific to Cypress, then having it near to `e2e/cypress` vs. `@aws-amplify/ui` is my only constraint.

But I don't see much value in moving these to yet another file since they're needed for these commands TBH, because there's not re-use.

packages/e2e/features/ui/components/authenticator/reset-password.feature
ericclemmons: As mentioned above, the docs' framework drop-down & content relies on these tags to know that `@react` and `@next` frameworks both support this feature:

https://github.com/aws-amplify/amplify-ui/blob/911c82c2c7e2985bd0b8a6d4cbbe92a2fde7b53d/docs/src/components/Feature.tsx#L76-L84

e2e tests aren't running any `react` examples:

https://github.com/aws-amplify/amplify-ui/blob/911c82c2c7e2985bd0b8a6d4cbbe92a2fde7b53d/.github/workflows/tests.yml#L55-L60

ericclemmons (commented):
The primary concern I have is the removal of `@react`, since the docs rely on features to display supported use-cases per-framework.

`@react` is supported for these cases still, as is `@next`. We just don't have a React-specific example (like Create React App) for e2e tests today since Next.js is _close enough_.

It does seem that there's a need to skip in CI when we run into technical issues – where something's supported & manually validated, but we don't have automated regression tests.

packages/e2e/features/ui/components/authenticator/reset-password.feature
slaymance: We can just update the docs to point to `@next` by default. I'm not sure I understand when we would want docs to show React supports a feature that Next doesn't?

.github/workflows/tests.yml
slaymance: I guess it leads to the question of: do we want to have `.feature` files that will never translate to a working test? And if not, how do we want to include those features in our docs?

ericclemmons: What's something that will "never translate to a working test"?

My take is we want to codify all behavioral expectations because this is the one source of truth for our behavior that's consistent between product, engineering, and customer expectations.

So if something can't be tested for a technical reason, then at least it's documented, tracked, & can be unblocked one-day.

packages/e2e/features/ui/components/authenticator/sign-up.feature
```diff
@@ -5,24 +5,26 @@ Feature: Sign Up
...
-    Then I see "Create a new account"
 
-  @next @react @vue @angular @skip
-  Scenario: Sign up with a new username & password
-    When I type a new username
-    And I type a new password
-    And I type a new confirm password
-    And I type the email "test@example.com"
-    And I type the phone number "+15558675309"
+  @angular @next @vue
```
ericclemmons, line 9: `@react` needs to be here because, otherwise, this feature isn't supported on this platform:

> Screen Shot 2021-09-10 at 4 20 05 PM
> – https://ui.docs.amplify.aws/ui/components/authenticator?platform=react#sign-up


> Screen Shot 2021-09-10 at 4 20 14 PM
> – https://ui.docs.amplify.aws/ui/components/authenticator?platform=next#sign-up

docs/src/components/Feature.tsx
```diff
@@ -58,7 +58,7 @@ function getGitHubUrlForExample(platform) {
 export function Feature({ name = required('Missing feature name') }) {
   const [source, setSource] = React.useState(null);
   const { pathname, query } = useRouter();
-  const { platform = 'react' } = query;
+  const { platform = 'next' } = query;
```
ericclemmons, line 61: `react` is still our most common framework and should be the default for when `?platform=...` isn't explicitly set:

> https://ui.docs.amplify.aws/ui/components/authenticator#sign-up

ericclemmons (commented):
Sharing some screenshots & examples where removing `@react` removes documentation & support.

PR #361, opened 2021-09-10
initial styling for ForceNewPassword

eddiekeller:
*Issue #, if available:*

https://app.asana.com/0/1200881947262804/1200952565427106/f

*Description of changes:*

Initial styling for the `ForceNewPassword` screen. This adds a new `confirm_password` field to the screen. As a result, I also modified the state machine to handle validating that the passwords match each other, identical to the `SignUp` screen (literally copy/pasted most of it).

This PR *might* break functionality in Vue and Angular on the `ForceNewPassword` screen due to changes in the state machine.

Screen Shot 2021-09-10 at 11 32 11 AM
Screen Shot 2021-09-10 at 11 33 03 AM
Screen Shot 2021-09-10 at 11 33 40 AM

### DIFF
--- a/packages/e2e/cypress/integration/ui/components/authenticator/sign-in-force-new-password/sign-in-force-new-password.steps.ts
+++ b/packages/e2e/cypress/integration/ui/components/authenticator/sign-in-force-new-password/sign-in-force-new-password.steps.ts
@@ -10,7 +10,11 @@ And('I type in the phone number {string}', (phone: string) => {
 });
 
 And('I type in the password {string}', (password: string) => {
-  cy.findByPlaceholderText(/password/i).type(Cypress.env(password));
+  cy.findByPlaceholderText(/^password$/i).type(Cypress.env(password));
+});
+
+And('I confirm the password {string}', (password: string) => {
+  cy.findByPlaceholderText(/^confirm_password$/i).type(Cypress.env(password));
 });
 
 Then('I should see the Force Change Password screen', () => {
--- a/packages/ui/src/machines/actors/auth/signIn.ts
+++ b/packages/ui/src/machines/actors/auth/signIn.ts
@@ -3,17 +3,20 @@ import { createMachine, sendUpdate } from 'xstate';
 
 import { Auth } from 'aws-amplify';
 
+import { passwordMatches, runValidators } from '@/validators';
 import {
   clearAttributeToVerify,
   clearChallengeName,
   clearError,
   clearFormValues,
   clearUnverifiedAttributes,
+  clearValidationError,
   handleInput,
   setChallengeName,
   setConfirmResetPasswordIntent,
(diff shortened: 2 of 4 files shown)


packages/e2e/cypress/integration/ui/components/authenticator/sign-in-force-new-password/sign-in-force-new-password.steps.ts
```diff
@@ -10,7 +10,11 @@ And('I type in the phone number {string}', (phone: string) => {
 });
 
 And('I type in the password {string}', (password: string) => {
-  cy.findByPlaceholderText(/password/i).type(Cypress.env(password));
+  cy.findByPlaceholderText(/^password$/i).type(Cypress.env(password));
+});
+
+And('I confirm the password {string}', (password: string) => {
+  cy.findByPlaceholderText(/^confirm_password$/i).type(Cypress.env(password));
```
slaymance, line 17: These tests are skipped but I think the regex for the placeholder selector should be `/^confirm password$/i`.

packages/ui/src/machines/actors/auth/signIn.ts
```diff
@@ -401,6 +451,11 @@ export const signInActor = createMachine<SignInContext, AuthEvent>(
 
         return result;
       },
+      async validateFields(context, _event) {
+        const { formValues } = context;
+        const validators = [passwordMatches]; // this can contain custom validators too
+        return runValidators(formValues, validators);
+      },
```
wlee221, line 458: Can we share this in `../../actions`?

wlee221 (approved):
All lgtm! We should consider abstracting "form with validation" interface so that we can reuse it across subcomponents. Could be its own actor, or just a dynamic object. But I'm digressing again, code changes here lgtm 🚢

packages/ui/src/machines/actors/auth/signIn.ts
eddiekeller: I wonder how that works with services? I'll take a look at this for a future PR!

slaymance (approved):
Left a small comment for the e2e test selector.

packages/ui/src/machines/actors/auth/signIn.ts
wlee221: Oh shoot, yeah this is services 😓  my bad. We should be able to share those anyways, but I agree and this is not imminent.

packages/e2e/cypress/integration/ui/components/authenticator/sign-in-force-new-password/sign-in-force-new-password.steps.ts
eddiekeller: Good point! It should be. I'll have another PR following this one pretty quickly so I'll make the change there.

PR #374, opened 2021-09-14
Vue sign in page new style

ErikCH:
*Issue #, if available:* [Asana](https://app.asana.com/0/0/1200953035425696/f)

*Description of changes:*
Added new styles to Vue

image


examples/vue/src/main.ts
```diff
@@ -1,7 +1,6 @@
-import { createApp } from "vue";
-import App from "./App.vue";
-import router from "./router";
+import { createApp } from 'vue';
+import App from './App.vue';
+import router from './router';
+import './styles.scss';
```
ErikCH, line 4: There are styles in the next example, that I copied over to the Vue example so the look and feel would match.

packages/vue/src/components/sign-in-password-control.vue
```diff
@@ -1,19 +1,82 @@
...
+        class="
+          amplify-button
+          amplify-field-group__control
+          amplify-field__show-password
+        "
+        data-fullwidth="false"
+        type="button"
+        @click="togglePasswordText"
+      >
+        <svg
```
ErikCH, line 31: I inlined the SVGS for now.

eddiekeller (approved):

examples/vue/src/main.ts
wlee221: Gotcha, I'll probably have to do this in Angular too... and hope Angular is cooperative with scss.

wlee221: Do you recall which primitives needed this scss file? I suppose this is filling in the gap of not having primitives on Angular/Vue.

wlee221 (approved):
✨  LGTM! I will for sure reference this when I get to Angular parity styles too 😛

examples/vue/src/main.ts
ErikCH: All the ones with the !import were used for sure. Also I added a font.

PR #378, opened 2021-09-14
feat(angular): Add confirm password in forceNewPassword screen

wlee221:
*Issue #, if available:*

*Description of changes:* This reflects changes from #361.

## MP4

https://user-images.githubusercontent.com/43682783/133348023-1f7dccce-54c8-482a-b9c1-c8e199170e0b.mp4

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.html
@@ -11,8 +11,14 @@
       <amplify-form-input
         name="password"
         type="password"
-        label="Change password"
-        autocomplete="password"
+        autocomplete="new-password"
+      ></amplify-form-input>
+
+      <amplify-form-input
+        name="confirm_password"
+        label="Confirm Password"
+        type="password"
+        autocomplete="new-password"
       ></amplify-form-input>
 
       <div data-amplify-footer>
--- a/packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.ts
@@ -59,7 +59,6 @@ export class AmplifyForceNewPasswordComponent
   }
 
   onStateUpdate(state: AuthMachineState): void {
-    console.log({ ...state });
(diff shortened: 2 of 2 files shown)


packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.ts
```diff
@@ -59,9 +59,14 @@ export class AmplifyForceNewPasswordComponent
   }
 
   onStateUpdate(state: AuthMachineState): void {
+    console.log({ ...state });
```
eddiekeller, line 62: nit - we can delete this

eddiekeller (approved):

packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.html
```diff
@@ -11,8 +11,14 @@
       <amplify-form-input
         name="password"
         type="password"
-        label="Change password"
-        autocomplete="password"
+        autocomplete="new-password"
```
eddiekeller, line 14: nit - do we need a label in this `input` as well?

packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.ts
wlee221: ;)

wlee221:
```suggestion
```

packages/angular/projects/ui-angular/src/lib/components/amplify-force-new-password/amplify-force-new-password.component.html
wlee221: No, I have it so that it infers label automatically from from the input name:

https://github.com/aws-amplify/amplify-ui/blob/911c82c2c7e2985bd0b8a6d4cbbe92a2fde7b53d/packages/ui/src/helpers/auth.ts#L38

This will be cleaned up with Angular i18n coming this week.

wlee221: We can provide a specific `label` value if we want to override the default label, such as "Confirm Password".

PR #381, opened 2021-09-15
Re-do of tabs PR

dbanksdesign:
*Issue #, if available:*

*Description of changes:* There were lots of merge conflicts from the tabs PR being open so long, so starting this one fresh. Addressed @reesscot's comments as well from #346 All the credit goes to @joebuono for doing most of the work on this PR

```jsx
<Tabs>
  <TabItem title="Tab 1">
    Tab 1 content
  </TabItem>
  <TabItem title="Tab 2">
    Tab 2 content
  </TabItem>
</Tabs>
```

Also, this fixes a docs issue where demo CSS overrides were being overridden by the base styling because NextJS doesn't honor import order in JS imports, but does in CSS imports and adds a className to the `Example` component in the docs so we can scope the CSS for examples where we show CSS overrides. 

https://user-images.githubusercontent.com/321279/133472312-0ccbeb95-316e-404f-a823-939fa5b66495.mp4


eddiekeller (approved):
Thanks for doing this! Looks good!

ericclemmons:
> Also, this fixes a docs issue where demo CSS overrides were being overridden by the base styling because NextJS doesn't honor import order in JS imports, but does in CSS imports and adds a className to the Example component in the docs so we can scope the CSS for examples where we show CSS overrides.

Can you annotate the PR with the specifics on these 2 points? These gotchas sound worthwhile.  If I understand correctly.

1. When doing `import "...css"`, it's best to have a single `import ".../styles.css"` in `_app.jsx` and ensure correct import order _within the `.css` file_.

    > https://github.com/vercel/next.js/discussions/27953

2. Some examples need to be specifically targeted, so `<Example className="something">` (or something like `<Example id="global-import">`) allows docs-specific styling to target _just that example_ vs. all examples.

ericclemmons (approved):
> Also, this fixes a docs issue where demo CSS overrides were being overridden by the base styling because NextJS doesn't honor import order in JS imports, but does in CSS imports and adds a className to the Example component in the docs so we can scope the CSS for examples where we show CSS overrides.

Can you annotate the PR with the specifics on these 2 points? These gotchas sound worthwhile.  If I understand correctly.

1. When doing `import "...css"`, it's best to have a single `import ".../styles.css"` in `_app.jsx` and ensure correct import order _within the `.css` file_.

    > https://github.com/vercel/next.js/discussions/27953

2. Some examples need to be specifically targeted, so `<Example className="something">` (or something like `<Example id="global-import">`) allows docs-specific styling to target _just that example_ vs. all examples.

docs/src/pages/_app.page.tsx
```diff
@@ -1,7 +1,5 @@
 import Head from 'next/head';
-import '../styles/styles.css';
-import '@aws-amplify/ui-react/styles.css';
-import '../styles/primitivesStyles.css';
+import '../styles/index.css';
```
dbanksdesign, line 2: Moving CSS imports to a single `index.css` file so that we can control the order of styles. At least in development NextJS was not importing these CSS files in the right order with the `primitiveStyles.css` being loaded before `@aws-amplify/ui-react/styles.css`, which blew up those CSS overrides for examples where we add a className to a primitive to style it.

docs/src/styles/index.css
```diff
@@ -1,3 +1,13 @@
+/* Docs styles */
```
dbanksdesign, line 1: This is the new single CSS file that the docs will load globally with the proper order of imports.

docs/src/components/Example.tsx
```diff
@@ -1,6 +1,8 @@
-export function Example({ children }) {
+export function Example({ children, className = '' }) {
```
dbanksdesign, line 1: Adding the ability to add a className to an Example so we can add scoped CSS for examples where we show overriding CSS for primitives without adding extra classNames or IDs to the primitive. For example, the code we show on the docs:

```css
.amplify-tabs {
  background-color: var(--amplify-colors-background-secondary);
}

.amplify-tabs-item {
}
```

Currently we mimic that behavior by either adding classNames or style props on the primitives in the example like [here](https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/pages/ui/primitives/alert/react.mdx):

```css
/* styles.css */
:root {
  --amplify-components-alert-background-color: yellow;
}
/* OR */
.amplify-alert {
  background-color: yellow;
}
```

```jsx
<Alert backgroundColor="yellow">Change the default Alert to yellow</Alert>
```

Being able to add a className to an example block will allow us to write example CSS that more closely resembles the actual CSS showed in the code block, like: 

Example code in docs:

```css
/* styles.css */
/* When a Tab is selected and displaying its corresponding content panel */
.amplify-tabs-item[data-state='active'] {
  background-color: var(--amplify-colors-brand-primary-20);
}

/* When a Tab is disabled */
.amplify-tabs-item[data-disabled] {
  background-color: var(--amplify-colors-background-tertiary);
}
```

Code in CSS

```css
.tabs-data-demo .amplify-tabs-item[data-state='active'] {
  background-color: var(--amplify-colors-brand-primary-20);
}

.tabs-data-demo .amplify-tabs-item[data-disabled] {
  background-color: var(--amplify-colors-background-tertiary);
}
```

PR #382, opened 2021-09-15
React: Country Code Select for Sign In

slaymance:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200953035425686/1200974176066572)

*Description of changes:* This PR updates the React `SignIn` component to include a country code selector when the alias is `phone_number`. It also does the following:

- Updates the React `UserNameAlias` to include the country code selector when the alias is `phone_number`
- Updates the React `SignUp` component to use the `UserNameAlias` to reduce code duplication (and set up an easy transition when `PhoneNumberField` primitive lands)
- Updates state machine to have a default `country_code` and prepend selected `country_code` to a user's phone number on `signIn`
- Updates various e2e tests to perform the country code selection step
- Skips signing in with phone number for Angular and Vue until country code selector is implemented for those frameworks

### Using default country code
default-country

### Selecting country code
select-country


packages/react/src/components/Authenticator/shared/UserNameAlias.tsx
```diff
@@ -1,33 +1,67 @@
...
 
+  const isPhoneAlias = type === 'tel';
+
+  useEffect(() => {
+    isPhoneAlias &&
+      send({
+        type: 'CHANGE',
+        data: { name: 'country_code', value: actorContext.country_code },
+      });
+  }, []);
```
slaymance, line 36: On init, update xstate with the default country code when the alias is a phone number.

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
```diff
@@ -114,12 +114,12 @@ export function SignUp() {
           )}
 
           {secondaryAliases
-            .filter((alias) => !includes(socialProviderLoginMechanisms, alias))
+            .filter((alias) => alias as UserNameAlias)
```
eddiekeller, line 117: I don't think this is going to properly filter.

https://www.typescriptlang.org/play?#code/C4TwDgpgBAggNgSwIYGcoF4oHICuKIBOAdkgLYRZQA+2EpSCclNWYAFgPZEUDcAsAChQkKAGUOAY2RwM2AGZIJEAEYcOAa2bYA5mu1wK1bGSQAvLln5Dw0AFIcERWfGRoa4qUjhXBErimAoJAICAC4oe0cAbQBdWSjcfGIyCgAaWnpGLHSsE3MibOx2LjSdPQMsGKs-IgCoOUZgQggAE1lgggA6BrgmggAKfoA3LwBKDAA+KBGZVFhEVFGfARqUDgNOuA5tfqwevtaoBDQAWhPsJd9-dYhN7f795paloA

packages/ui/src/helpers/auth.ts
```diff
@@ -48,12 +50,24 @@ export enum FederatedIdentityProviders {
...
   const loginMechanisms = context.config?.login_mechanisms ?? ['username'];
   const error = context.actorRef?.context?.validationError['username'];
 
+  if (alias as UserNameAlias) {
+    return {
+      label: authInputAttributes[alias].label,
+      type: authInputAttributes[alias].type,
+      error,
+    };
+  }
```
eddiekeller, line 69: Same here. I don't believe type casting can be used in place of filtering.

https://www.typescriptlang.org/play?#code/C4TwDgpgBAggNgSwIYGcoF4oHICuKIBOAdkgLYRZQA+2EpSCclNWYAFgPZEUDcAsAChQkKAGUOAY2RwM2AGZIJEAEYcOAa2bYA5mu1wK1bGSQAvLln5Dw0AFIcERWfGRoa4qUjhXBErimAoJAICAC4oe0cAbQBdWSjcfGIyCgAaWnpGLHSsE3MibOx2LjSdPQMsGKs-IgCoOUZgQggAE1lgggA6BrgmggAKfoA3LwBKDAA+KBGZVFhEVFGfARqUDgNOuA5tfqwevtaoBDQAWhPsJd9-dYhN7f795pbLgUE5DgIoftXAmagOORBELjADegigR0Bwy8QTQLkWUDBAghENWNzuOxmLwhAF9BDigA

packages/ui/src/machines/actors/auth/signIn.ts
```diff
@@ -390,7 +390,12 @@ export const signInActor = createMachine<SignInContext, AuthEvent>(
         const source = !!(context.intent && context.intent === 'autoSignIn')
           ? context.authAttributes
           : context.formValues;
-        const { username, password } = source;
+        let { username, password } = source;
+
+        if (source.country_code) {
```
eddiekeller, line 395: nit - add `country_code` to the destructuring you are doing here

packages/ui/src/machines/authMachine.ts
```diff
@@ -129,8 +132,9 @@ export const authMachine = createMachine<AuthContext, AuthEvent>(
         actorRef: (context, event) => {
           const actor = signUpActor.withContext({
             authAttributes: event.data?.authAttributes ?? {},
+            country_code: DEFAULT_COUNTRY_CODE,
             intent: event.data?.intent,
-            formValues: { country_code: '+1' },
+            formValues: {},
             validationError: {},
             login_mechanisms: context.config?.login_mechanisms,
```
eddiekeller, line 139: nit - would be nice to pull out the common default context values here into a separate object for reuse across the machine. Looks like we use most of these default values in a couple places.

packages/ui/src/types/authMachine.ts
```diff
@@ -101,7 +102,15 @@ export const userNameAliasArray = [
...
-export const socialProviderLoginMechanisms = ['amazon', 'google', 'facebook'];
+export const socialProviderLoginMechanisms = [
+  'amazon',
+  'google',
+  'facebook',
+] as const;
+
+export type LoginMechanism =
+  | UserNameAlias
+  | typeof socialProviderLoginMechanisms[number];
```
eddiekeller, line 113: nit - pull this into its own type, similar to `UserNameAlias` on line 103.

packages/vue/src/components/sign-up.vue
```diff
@@ -64,8 +64,9 @@ import {
   AuthInputAttributes,
   authInputAttributes,
   getActorState,
+  LoginMechanism,
```
eddiekeller, line 67: doesn't look like this value is being used

packages/vue/src/composables/useUtils.ts
```diff
@@ -1,13 +1,15 @@
...
+import { LoginMechanism, userNameAliasArray } from '@aws-amplify/ui';
 
+export const useAliases = (
+  login_mechanisms: LoginMechanism[]
+): LoginMechanism[] => {
   if (login_mechanisms) {
     if (login_mechanisms.length === 1 && login_mechanisms[0] === 'username') {
-      return defaultMechanisms;
+      return [...userNameAliasArray];
     } else {
```
eddiekeller, line 9: We don't need this block since the default return value is also `[...userNameAliasArray]`

eddiekeller (changes requested):

packages/e2e/features/ui/components/authenticator/sign-in-force-new-password.feature
```diff
@@ -8,13 +8,15 @@ Feature: Sign In with Force New Password flow
     Given I'm running the example "ui/components/authenticator/sign-in-with-phone"
 
   Scenario: Sign in using a valid phone number and password and user is in a FORCE_CHANGE_PASSWORD state
-    When I type my "phone number" with status "FORCE_CHANGE_PASSWORD"
+    When I select my country code with status "FORCE_CHANGE_PASSWORD"
```
ericclemmons, line 11: Why is status necessary for selecting a country code?

packages/ui/src/machines/authMachine.ts
slaymance: @eddiekeller By "default context values" do you mean the ones we hard code in a couple of the actor spawners? So what you suggest would look something like:
```javascript
const DEFAULT_CONTEXT_VALUES = {
  country_code: '+1',
  formValues: {},
  validationError: {},
};

...

      spawnSignInActor: assign({
        actorRef: (_, event) => {
          const actor = signInActor.withContext({
            authAttributes: event.data?.authAttributes,
            user: event.data?.user,
            intent: event.data?.intent,
            ...DEFAULT_CONTEXT_VALUES,
          });
          return spawn(actor, { name: 'signInActor' });
        },
      }),
```

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
slaymance: Good catch, I fixed it and improved the typing so it's clear these aliases aren't going to be our `socialProviderLoginMechanisms`.

packages/e2e/features/ui/components/authenticator/sign-in-force-new-password.feature
slaymance: Country code is my method of determining which status a user with a given phone number has in Cognito. Since I reduced our environment variables to a single "alias", we needed some way of programmatically signing in with a phone number that corresponds to a Cognito status in our e2e like we do for username and email.

packages/ui/src/machines/authMachine.ts
eddiekeller: Something like that. You don't have to do it for this PR. It was just a thought I had while reading through the code.

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
```diff
@@ -33,11 +35,12 @@ export function SignUp() {
...
   const isPending = actorState.matches('signUp.pending');
-  const { remoteError } = actorState.context;
   const { validationError } = getActorContext(_state) as SignUpContext;
 
-  const [primaryAlias, ...secondaryAliases] = _state.context.config
-    ?.login_mechanisms ?? ['username', 'email', 'phone_number'];
+  const [primaryAlias, ...secondaryAliases] =
+    _state.context.config?.login_mechanisms?.filter(
+      (alias: any): alias is UserNameAlias => userNameAliasArray.includes(alias)
+    ) ?? userNameAliasArray;
```
eddiekeller, line 43: Let's clean this up. `alias` shouldn't be `any` - either `string` or `LoginMechanism`, right? We also don't really need the `alias is UserNameAlias` inside of the filter there.

```
_state.context.config?.login_mechanisms?.filter(
  (alias: LoginMechanism) => userNameAliasArray.includes(alias)
) ?? userNameAliasArray;
```

packages/react/src/components/Authenticator/shared/UserNameAlias.tsx
```diff
@@ -1,33 +1,67 @@
...
+          {countryDialCodes.map((dialCode) => (
+            <option key={dialCode} value={dialCode}>
+              {dialCode}
+            </option>
+          ))}
+        </SelectField>
+      )}
+      <TextField
+        type={type}
+        name={alias ?? 'username'}
```
eddiekeller, line 56: Have you tested that SignIn/SignUp still works as expected with this change? Xstate looks specifically for the `username` value. I think changing the name here will cause errors.

Integ tests passing won't catch this since those don't actually perform the SignIn/signUp actions.

packages/ui/src/machines/actors/auth/signIn.ts
```diff
@@ -390,7 +390,10 @@ export const signInActor = createMachine<SignInContext, AuthEvent>(
         const source = !!(context.intent && context.intent === 'autoSignIn')
           ? context.authAttributes
           : context.formValues;
-        const { username, password } = source;
+        let { country_code, username, password } = source;
+
+        username = (country_code ?? '') + username;
```
eddiekeller, line 395: nit - would prefer to make the destructured variables as `const` and just make a new value for the `username` on line 395. Won't block over it, though.

eddiekeller (changes requested):

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
slaymance: Yeah the `any` typing for `alias` isn't ideal and I just did it to get around how TypeScript types the `.includes` method. I'm essentially running into this [problem](https://github.com/microsoft/TypeScript/issues/26255), where `LoginMechanism` isn't a valid search value in `userNameAliasArray` because it includes the social providers. I'm open to other suggestions here, I mostly just wanted to limit how often we're writing `['username', 'email', 'phone_number']`.

The `alias is UserNameAlias` is nice because we're saying that `primaryAlias` will be `UserNameAlias` and `secondaryAliases` will be `UserNameAlias[]` after the filter is complete.

packages/react/src/components/Authenticator/shared/UserNameAlias.tsx
slaymance: Only the `signIn` service looks for `username` value and the `SignIn` component doesn't use `alias` because there will only ever be one input field. The `signUp` service looks for a name specific to the alias (email, username, phone_number) so that's why I use the `alias` as the name when it's specified by the parent component.

eddiekeller: Ok cool, thanks for clarifying that for me!

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
eddiekeller: Hm ok. I'll approve for now since playing with TypeScript isn't worth blocking over, but this might be something to revisit moving forward then.

eddiekeller (approved):

packages/react/src/components/Authenticator/SignUp/SignUp.tsx
slaymance: I captured this as an E2 task https://app.asana.com/0/1200334286823427/1201001812784243. Each `alias` begins as being a `LoginMechanism`, so I think revisiting our typings can get us to removing the `any` type to get around this issue.

packages/react/src/components/Authenticator/shared/UserNameAlias.tsx
slaymance: It's a little tricky, it may be worth revisiting to make it clearer. Maybe doing something in the `SignIn` component where it's agnostic to the `name` prop, but always updates the state machine with that prop as `username`.

PR #386, opened 2021-09-15
adding tabs, radio, and federated icons

eddiekeller:
Items in PR
- [x] Add tabs to `SignIn`/`SignUp` pages
- [x] Add icons to federated social login buttons
- [x] Add styling and radio buttons to the `VerifyUser`/`ConfirmVerifyUser` pages
- [x] Update tests to reflect new `SignIn`/`SignUp` pages

*Issue #, if available:*

https://app.asana.com/0/1200953035425686/1200992085363886/f
https://app.asana.com/0/1200953035425686/1200992085363877/f
https://app.asana.com/0/1200953035425686/1200935010977984/f

*Description of changes:*

Adding Tabs to the `SignIn` and `SignUp` pages.
Adding styling to the `VerifyUser` and `ConfirmVerifyUser` pages.
Adding Icons to the federated social sign in buttons.

Small add - loaded in custom font for our example next/react apps to use the custom font shown in the mocks. Screenshots attached below.

Screen Shot 2021-09-15 at 2 54 05 PM
Screen Shot 2021-09-15 at 2 54 11 PM
Screen Shot 2021-09-15 at 2 54 24 PM
Screen Shot 2021-09-15 at 2 54 30 PM


packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
```diff
@@ -17,10 +18,9 @@ const generateRadioGroup = (
 
   for (const [key, value] of Object.entries(attributes)) {
     const radio = (
-      <label key={key}>
-        <input type="radio" name="unverifiedAttr" value={key} />
+      <Radio name="unverifiedAttr" value={key} key={key}>
```
reesscot, line 21: If there's a single attribute only (Email), would it make sense to use a checkbox instead? Seems odd to show a radio button for a single option as shown in the screenshot. IME, radio buttons are usually used for picking between at least two options.

reesscot:
A few styling questions:
* Could "Skip" be a secondary button instead of a link variation?
* Could reduce the padding between the panel and the tabs? Not sure if this is a Tabs primitive styling issue.

packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
eddiekeller: There can be multiple options. It's just that the example I'm using only has `email` as an attribute to verify.

eddiekeller:
> Could "Skip" be a secondary button instead of a link variation?

Mocks show it as a `link` variation. https://www.figma.com/proto/74U1PE5KhnRslVJbvLxeKA/Duplicate-Me?node-id=608%3A43717&scaling=min-zoom&page-id=539%3A23088

But it can be secondary if that's what we want.

> Could reduce the padding between the panel and the tabs? Not sure if this is a Tabs primitive styling issue.

I don't believe I am adding any padding there myself. I think it is likely coming from the Tabs, but I'll double check.

reesscot:
> @reesscot
> 
> > Could "Skip" be a secondary button instead of a link variation?
> 
> Mocks show it as a `link` variation. https://www.figma.com/proto/74U1PE5KhnRslVJbvLxeKA/Duplicate-Me?node-id=608%3A43717&scaling=min-zoom&page-id=539%3A23088
> 
> But it can be secondary if that's what we want.
> 
> > Could reduce the padding between the panel and the tabs? Not sure if this is a Tabs primitive styling issue.
> 
> I don't believe I am adding any padding there myself. I think it is likely coming from the Tabs, but I'll double check.
> 
> Edit - confirmed. `amplify-tabs` has a `margin-bottom` of `var(--amplify-components-tabs-margin-bottom);` on it, which in this case equals `16px`. Added an override on that class for now to get rid of the margin.

Just my two cents that secondary would look better, but I'm not not the designer.

@dbanksdesign Do you think we want a margin below the Tabs? Seems like this would be a common issue to want to control the margin/padding below the tabs. I think it would make sense to remove it from the Tabs primitive.

dbanksdesign:
Yea that makes sense to remove the tabs margin. Having the margin kinda breaks our thinking about not using margin in general

dbanksdesign:
👀  removing tab margin: #390 390

ErikCH (approved):
Looks good

PR #388, opened 2021-09-15
Added New Styles to Sign Up Page

ErikCH:
*Issue #, if available:* [Asana](https://app.asana.com/0/0/1200953035425697/f)

*Description of changes:*
Update styles to match React/Mock

Deferring adding tabs until next PR

image


examples/vue/src/styles.scss
```diff
@@ -15,6 +15,9 @@ body {
   padding: 0;
   margin: 0;
 }
+.amplify-icon[data-size='large'] {
+  height: 2rem !important;
```
ericclemmons, line 19: Were there issues with the icon that required this? Ideally we don't have any additional styles besides what the primitives provide.

packages/vue/src/components/authenticator.vue
```diff
@@ -321,7 +307,7 @@ const emit = defineEmits([
 ]);
 
 const s = useInterpret(authMachine, {
-  devTools: process.env.NODE_ENV === 'development',
+  devTools: process.env.NODE_ENV !== 'development',
```
ericclemmons, line 310: This shouldn't be enabled for production should it? What did you experience?

```suggestion
  devTools: process.env.NODE_ENV === 'development',
```

packages/vue/src/components/password-control.vue
```diff
@@ -60,18 +60,23 @@
...
+import { HIDE_PASSWORD, SHOW_PASSWORD } from '../defaults/DefaultTexts';
 import { I18n } from 'aws-amplify';
-const passwordLabel = computed(() => I18n.get(PASSWORD_LABEL));
+
 const showPassword = I18n.get(SHOW_PASSWORD);
 const hidePassword = I18n.get(HIDE_PASSWORD);
 const showHideType = ref('password');
 const showHideLabel = ref(showPassword);
+
+const random = Math.floor(Math.random() * 999999);
```
ericclemmons, line 72: What's `random` used for? Is `amplify-password-control-${name}` not sufficient?

packages/vue/src/components/styles.css
```diff
@@ -14,10 +14,6 @@
   @apply flex flex-row items-center space-x-1;
 }
 
-[data-amplify-wrapper] [data-amplify-fieldset] {
-  @apply grid gap-6;
-}
-
```
ericclemmons, line 20: Should we remove all of these `styles.css` and entirely defer to `@aws-amplify/ui/styles.css`?

ericclemmons (commented):
Looking good @ErikCH!  I had a few questions before I give it a thorough review.

examples/vue/src/styles.scss
ErikCH: The chevron on the drop down for the select box was smaller then it was in React. So I added this to make it look exact. And for some reason the styles in the example app, don't automatically override the styles in the library.

packages/vue/src/components/styles.css
ErikCH: I don't think so. It would break the look and feel of the Vue app. We have these styles in Angular and React too. 

Maybe eventually we'll need to get rid of these...

packages/vue/src/components/password-control.vue
ErikCH: The `password-control` component is re-used twice in the `sign-up` component. Therefore we need an unique string for the `for=` attribute label. It's pseudo random number, so it's not unique, but the chances of it repeating in the sign-up is rare.

packages/vue/src/components/authenticator.vue
ErikCH: Good catch, I'll fix this. I changed this because when you build locally it always thinks it's in production for some reason.

examples/vue/src/styles.scss
ErikCH: In the `amplify-ui/examples/next/pages/examples/listings/styles.scss` there is an overwrite to make the icon large-height to xxl 

`
  --amplify-components-icon-large-height: var(--amplify-font-sizes-xxl);
`

When I added this line into the Vue example styles sheet, it was getting overwritten by the package, so the chevron was smaller.

So added this line and just marked is as 2rem to match `--amplify-font-sizes-xxl` so it would look exact.

eddiekeller (approved):

wlee221 (approved):

examples/vue/src/styles.scss
ericclemmons: @dbanksdesign ☝️ Something to consider as a default styling so icons look good by default.

Shouldn't `rem` be our default of `em`, also?

dbanksdesign: `rem` is our default sizing unit: https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/tokens/fontSizes.js https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/tokens/components/icon.js

For changes like this in the future, we should propose those changes in the theme so we don't have style overrides in different places. I think bumping up the size of the icons makes sense in general so we should do it at the theme level. Could we open a new PR making that change in the theme and removing this override?

PR #391, opened 2021-09-16
chore(angular): Remove amplify-sign-up

wlee221:
*Issue #, if available:*

*Description of changes:*
- Remove `amplify-sign-out` in favor of template variables we support. 
- Update examples to use template variables. Customers can also do so through `AuthenticatorService` (useAuthenticator hook equivalent), but I'll document that in my public interfacing pr.


examples/angular/src/pages/ui/components/authenticator/auth-with-multi-alias/auth-with-multi-alias.component.html
```diff
@@ -1 +1,10 @@
-<amplify-authenticator></amplify-authenticator>
+<amplify-authenticator>
+  <ng-template
+    amplifyOverride="authenticated"
+    let-user="user"
+    let-signOut="signOut"
+  >
```
wlee221, line 6: These are render props of Angular world ✨

examples/angular/src/pages/ui/components/authenticator/sign-up-with-phone/sign-up-with-phone.component.html
```diff
@@ -1 +1,10 @@
-<amplify-authenticator></amplify-authenticator>
+<amplify-authenticator>
+  <ng-template
+    amplifyOverride="authenticated"
+    let-user="user"
+    let-signOut="signOut"
+  >
+    <h2>Welcome, {{ user.username }}!</h2>
+    <button (click)="signOut()">Sign Out</button>
+  </ng-template>
```
ErikCH, line 9: This looks good. I like the idea of exposing signOut and user. Maybe in our discussion today we can chat on what other things we need to expose through slots

ErikCH (approved):
LGTM

PR #392, opened 2021-09-17
PhoneNumberField primitive

slaymance:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200953035425686/1200683483662286)

*Description of changes:* This PR adds the `PhoneNumberField` primitive which includes a "country code" drop down selector and an input field for an end-user's phone number. The primitive is used by the `SignIn` and `SignUp` fields when a phone number is required. However, when multi-alias authentication is enabled at sign in, the `PhoneNumberField` primitive is not used.

This pull request doesn't export `PhoneNumberField` as an exported primitive. That step will come in a [follow-up pull request](https://github.com/aws-amplify/amplify-ui/pull/456), and this one just gets the primitive ready for use in the Authenticator.

### PhoneNumberField in SignIn
country-code-signin

### PhoneNumberField in SignUp
country-code-signup


reesscot:
Please add a `min-width` CSS rule to the country code select:
Screen Shot 2021-09-21 at 11 32 42 AM

packages/react/src/primitives/PhoneNumberField/__tests__/PhoneNumberField.test.tsx
```diff
@@ -0,0 +1,62 @@
+import { render, screen } from '@testing-library/react';
+
+import { PhoneNumberField } from '../PhoneNumberField';
+
+describe('PhoneNumberField primitive', () => {
+  it('should render a country code selector', async () => {
```
reesscot, line 6: Please add the following tests:
```
  it('should render classname for PhoneNumberField` async () => {
  it('should be able to set a size', async () => {
  it('should be able to set a variation', async () => {
  it('should fire onChange and onInput', async () => {
```
See:
https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/PasswordField/__tests__/PasswordField.test.tsx#L11

docs/src/pages/ui/primitives/phonenumberfield/react.mdx
```diff
@@ -0,0 +1,29 @@
...
+See https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands for more information.
+
+### Additional Usage
+
+`PhoneNumberField` is a custom primitive built on top of the `TextField` primitive. Refer to the [TextField primitive documentation](textfield#usage)
+for more information regarding additional usage.
+
+## Styling
+
+The `PhoneNumberField` uses the same styling rules as the `TextField` primitive. Refer to the [TextField primitive documentation](textfield#css-styling)
```
reesscot, line 28: Our documentation strategy for primitives is to lean towards being repetitive rather than linking to other primitives. We feel like users will want to know everything about a primitives from a single page. Please add examples of use of `size`, `variation`, and CSS Styling. I think we also need to record the `onChange` and `onInput` event handler use mentioned in previous comment and an example of using `defaultCountryCode` See https://quip-amazon.com/fOKyAeLgbNCa/Template-for-Primitive-Docs

packages/react/src/components/Authenticator/shared/UserNameAlias.tsx
```diff
@@ -31,37 +30,33 @@ export function UserNameAlias(props: UserNameAliasProps) {
...
-        placeholder={i18nLabel}
-        label={label}
-        labelHidden={true}
-        autoComplete="username"
-        errorMessage={error}
-        {...attrs}
-      />
-    </>
+  return isPhoneAlias ? (
+    <PhoneNumberField
```
reesscot, line 38: Can you to add `autoComplete="username"` here instead of in the primitive? I'm guessing this is the use case where users use their phone number as their username?

packages/react/src/primitives/types/phoneNumberField.ts
```diff
@@ -0,0 +1,6 @@
+import { TextFieldProps } from './textField';
+
+export interface PhoneNumberFieldProps extends TextFieldProps {
+  defaultCountryCode: string;
+  type?: 'tel';
```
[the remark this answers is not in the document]
reesscot: This is really where I wish we could use `Omit`, because we don't actually allow the customer to set the type on this primitive. This is a good solution for now.

reesscot (changes requested):
This is looking great! So exciting to see a `PhoneNumberField` composed with the `inputStartComponents` prop. 🎉🎉🎉 

Just had some implementation suggestions and questions below.

packages/react/src/primitives/types/phoneNumberField.ts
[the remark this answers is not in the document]
slaymance: @reesscot I agree, I originally wanted to prevent `type` from being passed as a prop to the primitive entirely. I kind of got backed into a corner though since we're ultimately extending `BaseComponentProps` which has a catch all `[key: string]: any;` prop. So this was what I came up with to prevent the wrong type from being passed.

Another idea is I could use `Omit` like you suggested, pull out the `type` prop when destructuring arguments to `PhoneNumberField`, then force `type="tel"` so it doesn't get overwritten by a custom `type` attribute. Neither are entirely elegant though, so I'll leave this unless you have a preference.

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -0,0 +1,48 @@
...
+  ...rest
+}) => {
+  return (
+    <TextField
+      inputStartComponents={
+        <SelectField
+          name="country_code"
+          label="country code"
+          labelHidden={true}
+          defaultValue={defaultCountryCode}
```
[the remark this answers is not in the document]
slaymance: Nice, good catch!

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -0,0 +1,48 @@
+import { countryDialCodes } from '@aws-amplify/ui';
+
+import { SelectField } from '../SelectField';
+import { TextField } from '../TextField';
+import { PhoneNumberFieldProps } from '../types';
+
+export const PhoneNumberField: React.FC<PhoneNumberFieldProps> = ({
+  autoComplete = 'username',
```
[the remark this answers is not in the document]
slaymance: Good callout, I was showing my biased Authenticator mindset :smile:

I'll update our usage docs to reflect this and guide users who want to use autocomplete for password managers.

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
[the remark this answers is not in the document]
slaymance: Check out packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx where I added this attribute.

packages/react/src/primitives/types/phoneNumberField.ts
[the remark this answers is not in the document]
slaymance: I wanted to do some cheekiness like 
```typescript
interface PhoneNumberFieldProps extends
  Omit<TextFieldProps, 'onChange'>,
  Pick<SelectFieldProps, 'onChange'> {
...
```
but then discovered that `Pick` and `Omit` utility types don't actually do deep inheritance. I'm sure you all have come across that and is why we aren't using those types often.

packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
```diff
@@ -0,0 +1,29 @@
+import { useMemo } from 'react';
```
zchenwei, line 1: nit: should we add some basic unit tests for this small component? No blocker though

zchenwei (approved):
LGTM!

packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
slaymance: Yeah I think that is a great idea. Let me address it in my follow-up PR so we get this component unit-tested before exporting.

ErikCH:
Looks good @slaymance !

I updated the Vue look and feel to match React

image

image

packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const CountryCodeSelect: React.FC<CountryCodeSelectProps> = (props) => {
+  const countryCodeOptions = useMemo(
+    () =>
+      countryDialCodes.map((dialCode) => (
+        <option key={dialCode} value={dialCode}>
+          {dialCode}
+        </option>
+      )),
+    countryDialCodes
```
reesscot, line 16: dependency array should just have the object reference in it, rather than all the codes.
```suggestion
    [countryDialCodes]
```

packages/react/src/primitives/PhoneNumberField/__tests__/PhoneNumberField.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { PhoneNumberField } from '../PhoneNumberField';
+import { PhoneNumberFieldProps } from '../../types';
+import { ComponentClassNames } from '../../shared/constants';
+
+describe('PhoneNumberField primitive', () => {
+  const setup = async ({
+    defaultCountryCode = '+1',
+    label = 'Phone Number',
+    ...rest
+  }: Partial<PhoneNumberFieldProps>) => {
```
reesscot, line 13: Why make all these props optional?

packages/react/src/primitives/types/phoneNumberField.ts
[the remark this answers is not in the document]
reesscot: They actually should work, but there's a problem with our BaseComponent type which makes Omit and Pick not work correclty. TLDR is that Omit does a union of all the keys that forces everything to `[key: string]: any`.

packages/react/src/primitives/PhoneNumberField/__tests__/PhoneNumberField.test.tsx
slaymance: I made the props optional so that I could lean on default parameters here instead of passing `defaultCountryCode` and `label` to every `setup` function invocation. I'm open to changing it if you think it makes this setup function unclear for the tests.

packages/react/src/primitives/types/phoneNumberField.ts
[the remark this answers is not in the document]
slaymance: Oh ok maybe that was the issue I was running into. It would be nice to be able to use these utility types to build out different primitives based on existing typings.

packages/ui/src/theme/css/component/phoneNumberField.scss
```diff
@@ -0,0 +1,13 @@
+.amplify-countrycodeselect {
+  // default styles
+  min-width: var(--amplify-components-phonenumberfield-min-width);
+  white-space: var(--amplify-components-phonenumberfield-white-space);
+  &[data-size='small'] {
+    min-width: var(--amplify-components-phonenumberfield-small-min-width);
+    white-space: var(--amplify-components-phonenumberfield-white-space);
```
reesscot, line 7: This doesn't change from default styles above, so we can remove it.
```suggestion
```

packages/ui/src/theme/css/component/phoneNumberField.scss
```diff
@@ -0,0 +1,13 @@
...
+  }
+  &[data-size='large'] {
+    min-width: var(--amplify-components-phonenumberfield-large-min-width);
+    white-space: var(--amplify-components-phonenumberfield-white-space);
```
reesscot, line 11:
```suggestion
```

reesscot (approved):

PR #394, opened 2021-09-17
Add @aws-amplify/ui-standalone package

ericclemmons:
*Description of changes:*

This was an attempt to use the `Authenticator` in a Vanilla JS context with `aws-amplify`, **but this is seriously blocked by https://github.com/aws-amplify/amplify-ui/issues/268**.

```js
import {Authenticator} from '@aws-amplify/ui-standalone/authenticator';

const container = document.querySelector<HTMLDivElement>('#app')!;

new Authenticator(container);
```

- [x] Creates `@aws-amplify/ui-standalone`
- [x] Use `@preact/compat` in place of `react` and `react-dom`

---


ericclemmons:
Also, `qrcode` isn't ESM/Browser friendly:
> Uncaught ReferenceError: Buffer is not defined

https://github.com/soldair/node-qrcode/issues/215

examples/js/src/main.ts
```diff
@@ -0,0 +1,11 @@
+import { Amplify } from 'aws-amplify';
+import { Authenticator } from '@aws-amplify/ui-standalone';
+import awsExports from '../../../environments/auth-with-username-no-attributes/src/aws-exports';
+
+import '@aws-amplify/ui-react/styles.css';
```
ayush987goyal, line 5: I think this should either be:
1. Style injected (even for the react package perhaps?)
2. Or be re-exported from the `ui-standalone` package. Since usage of `Authenticator` from the standalone package should not involve pulling in something from `ui-react`

packages/standalone/src/Authenticator/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+import ReactDOM from 'react-dom';
+
+export class Authenticator {
+  constructor(container: HTMLElement) {
+    ReactDOM.render(
+      <AuthenticatorComponent>{() => <h1>Howdy</h1>}</AuthenticatorComponent>,
+      container
+    );
+  }
+}
```
ayush987goyal, line 12: This is a great start! I am planning to make it more generic post this change gets merged.

examples/js/src/main.ts
ericclemmons: Re-export is the way we've been going 👍

packages/standalone/src/Authenticator/index.tsx
ericclemmons: I'm curious what the behavior should be render props for an imperative wrapper 🤔 

Maybe:

```js
new Authenticator(container).show().then(({ session, signOut, user }) => {
  alert(`Howdy ${user.username}`)
});
```

Honestly, the "Vanilla Authenticator" example is a bit contrived – I don't have much customer data on what the use-cases are for this specifically.

ayush987goyal: I think that's pretty much the correct analogy. Treating the return value from the imperative wrapper as a `ref` and acting accordingly would perhaps be the right usage.

PR #397, opened 2021-09-17
Vue: Add Styles For Confirm Sign Up - Reset Password - Add Tabs To sign in/up Page and Icons for Federated

ErikCH:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200953035425686/1200953035425684/f)

*Description of changes:*
Added styles for

- [x] Confirm Sign Up
- [x] Reset Password
- [x] Tabs To Sign Up and Sign In Pages
- [x] Federated Icons

image

image

image

image


packages/vue/src/components/authenticator.vue
```diff
@@ -407,4 +412,24 @@ const onConfirmVerifyUserSubmitI = (e: Event) => {
...
+      send({
+        type: 'SIGN_UP',
+      });
+    } else {
+      send({
+        type: 'SIGN_IN',
+      });
+    }
+  }
+);
```
ErikCH, line 434: This ties into the simple `base-two-tabs` component. I use this to toggle which pages to show.

packages/vue/src/components/reset-password.vue
```diff
@@ -1,28 +1,46 @@
...
         </base-field-set>
 
-        <base-footer>
+        <base-footer
+          class="amplify-flex"
+          style="flex-direction: column; align-items: unset"
+        >
           <template #footert="{ slotData }">
             <slot
               name="footer"
```
ErikCH, line 46: After every refactor I'm trying to save the slots I setup originally. I left this one here, in case someone wants to replace the footer on this page.

packages/vue/src/composables/useSelect.ts
```diff
@@ -0,0 +1,7 @@
+import { ref } from 'vue';
+
+// state
+
+const active = ref(false);
+
```
ErikCH, line 6: This maybe overkill, but I think it will be used later when vue adds real primitives. For now it's a simple reactive value that helps keep track of what tab is added.

packages/vue/src/components/authenticator.vue
```diff
@@ -326,6 +326,11 @@ const confirmResetPasswordComponent = ref(null);
 const verifyUserComponent = ref(null);
 const confirmVerifyUserComponent = ref(null);
 
+// computed
+
+const signInLabel = computed(() => I18n.get(CREATE_ACCOUNT_LABEL));
```
ericclemmons, line 331: What does `computed` do here, or how will it know if `I18n`'s value has changed?

`I18n.get` is a lookup, so is there a benefit from having it assigned to `signInLabel` vs. used directly in the view?

packages/vue/src/components/styles.css
```diff
@@ -111,3 +96,18 @@
...
+}
+
+.federated-sign-in-button-row {
+  position: relative;
+}
+
+.federated-sign-in-icon {
+  position: absolute;
+  left: 0;
+}
```
ericclemmons, line 113: I assume these values exist because there's not primitive CSS for this, and the React primitives would be writing out dynamic styles that you have to do by hand?

packages/vue/src/defaults/DefaultTexts.ts
```diff
@@ -42,3 +42,8 @@ export const CODE_TEXT = 'Code';
 export const PASSWORD_TEXT = 'Password';
 export const HIDE_PASSWORD = 'Hide password';
 export const SHOW_PASSWORD = 'Show password';
+export const SIGN_IN_GOOGLE = 'Sign in with Google';
```
ericclemmons, line 45: @wlee221 What's our latest answer for directly referencing our translations (and getting type safety) vs. maintaining an enum lookup?

ericclemmons (approved):
Nothing blocking, only nits! Thanks for the screenshots @ErikCH

packages/vue/src/components/authenticator.vue
```diff
@@ -1,5 +1,11 @@
 <template>
   <div v-bind="$attrs" data-amplify-authenticator>
+    <base-two-tabs
+      v-if="actorState?.matches('signIn') || actorState?.matches('signUp')"
+    >
+      <base-two-tab-item :label="createAccountLabel" :id="44472" />
+      <base-two-tab-item :label="signInLabel" :firstTab="true" :id="44471" />
```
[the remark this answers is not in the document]
ErikCH: The id are unique strings that will form the attributes on the base-two-tab-item.

For example `radix-id-${id}-8-trigger-0`

This is to match the primitives

packages/vue/src/components/authenticator.vue
ErikCH: The computed creates a reactive variable around the `I18n.get(CREATE_ACCOUNT_LABEL)` That way if they dynamically change the language, the computed property will get re-rendered with the correct value. 

I think this could work directly in the view, it probably will work with Vue's reactivity system, we would have to test it to see if dynamically changing the language updates correctly.  

This will work fine though

packages/vue/src/components/styles.css
ErikCH: These styles I pulled directly from the React styles that weren't in the primitives. They were slight adjustments to adding the icons to the federated sign in buttons.

packages/vue/src/defaults/DefaultTexts.ts
wlee221: Ideally we should use `DefaultTexts` from `@aws-amplify/ui`: https://github.com/aws-amplify/amplify-ui/blob/36ce52646f455e6a3f2d7e9befb4f808efc1061b/packages/ui/src/i18n/translations.ts#L27-L35

This will be the source of truth across the frameworks.

wlee221: If you wish to use to use the string literals directly, you can do so with `translate` helper I provided here: https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/i18n/translations.ts#L82-L84. This type is inferred from `DefaultTexts` above, so this is consistent with the above enum.

wlee221 (approved):
LGTM ✨

packages/vue/src/components/authenticator.vue
ericclemmons: I guess what I'm wondering is _how will Vue know when these values change_?

My understanding is that anything within `computed` relies on values defined within Vue's components or composables. 

Like, suppose this renders on the page and then you did a `setTimeout(..., 5000)` that did `I18n.put({ en: { CREATE_ACCOUNT_LABEL: 'updated' } })` (not exactly, but you get the idea) – I don't know how Vue would know to update this.

packages/vue/src/components/styles.css
ericclemmons: @dbanksdesign Should we go ahead & remove all of our proof-of-concept styles from React, fix the fallout within the primitive styles, and then re-apply to Angular & Vue?

packages/vue/src/components/authenticator.vue
ErikCH: Anytime any values in the computed property change, it gets recalculated. So if the `CREATE_ACCOUNT_LABEL` changes it gets updated. If the `I18n.get()` returns a different value, then it recalculates. It's also cached too.

packages/vue/src/defaults/DefaultTexts.ts
ErikCH: @wlee221 @ericclemmons Perhaps in a separate pull request I can start pulling the values from the ui translations file. The defaultTexts is just the source of truth for all Vue.js texts. But it would be nice to use the same one used in Angular and React.

packages/vue/src/components/styles.css
dbanksdesign: That's a great idea!

PR #398, opened 2021-09-17
Checkbox and CheckboxField Primitive

zchenwei:
*Description of changes:*

This PR will be adding `CheckboxField` and its unit `Checkbox`


packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,78 @@
+import classNames from 'classnames';
```
zchenwei, line 1: I was thinking if this is over engineering..But looks like we can just export the `Checkbox`, just follow the same patterns as how we build other fields.

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,78 @@
...
+import React from 'react';
+
+import { Checkbox } from '../Checkbox';
+import { FieldErrorMessage, FieldDescription } from '../Field';
+import { Flex } from '../Flex';
+import { Label } from '../Label';
+import { CheckBoxFieldProps } from '../types/checkboxField';
+import { ComponentClassNames } from '../shared';
+
+export const CheckboxField: React.FC<CheckBoxFieldProps> = ({
```
zchenwei, line 11: I was thinking if this level is necessary. Looks like `Checkbox` is good enough. Just follow the same patterns as how we build other fields.

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,79 @@
...
+      )}
+      data-size={size}
+      direction={direction}
+      gap={gap}
+      justifyContent={justifyContent}
+      testId={testId}
+      wrap={wrap}
+      {...rest}
+    >
+      <FieldDescription
```
zchenwei, line 55: I removed the `Label` here because `Checkbox` is label wrapped. Not sure if we should remove `FieldDescription` as well. It looks cumbersome for a checkbox anyway :)

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const CheckboxField: React.FC<CheckboxFieldProps> = ({
+  alignContent,
+  alignItems,
+  checked,
+  children,
+  className,
+  defaultChecked,
+  descriptiveText,
+  direction = 'column',
```
reesscot, line 18: This should be set in CSS, not as a render style prop.

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onChange,
+  testId,
+  size,
+  value,
+  wrap,
+  ...rest
+}) => {
+  return (
+    <Flex
+      alignContent={alignContent}
```
reesscot, line 40: Most of these Flex style properties are not going to have any affect on the primary layout of the `Checkbox` component, which seems confusing. What do you think of passing the Flex style props down to the Checkbox component instead? That seems like a more common use case for the flex style props. If customers really need to change the position of the Error Message, they could use CSS. What do you think?

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,79 @@
...
+        ComponentClassNames.CheckboxField,
+        className
+      )}
+      data-size={size}
+      direction={direction}
+      gap={gap}
+      justifyContent={justifyContent}
+      testId={testId}
+      wrap={wrap}
+      {...rest}
```
reesscot, line 53: I'm really feeling like all the styling props should go to the Checkbox primitive

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,79 @@
...
+        defaultChecked={defaultChecked}
+        hasError={hasError}
+        id={id}
+        isEmphasized={isEmphasized}
+        isDisabled={isDisabled}
+        isReadOnly={isReadOnly}
+        isRequired={isRequired}
+        name={name}
+        onChange={onChange}
+        testId={`${testId}-checkbox`}
```
reesscot, line 70: If undefined, this will be `undefined-checkbox` in the DOM.

docs/src/pages/ui/primitives/checkboxfield/react.mdx
```diff
@@ -0,0 +1,234 @@
...
+</CheckboxField>;
+```
+
+<Example>
+  <ControlledCheckbox />
+</Example>
+
+### Sizes
+
+Use the `size` prop to change the SelectField size. Available options are `small`, `large`, and none (default).
```
reesscot, line 58: Add examples for states such as isDisabled, isReadonly, hasError, etc?

docs/src/pages/ui/primitives/checkboxfield/react.mdx
```diff
@@ -0,0 +1,234 @@
...
+import { CheckboxField } from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+
+<CheckboxField name="subscribe" value="true" gap="1rem">
+  subscribe
+</CheckboxField>;
+```
+
+<Example>
+  <CheckboxField name="subscribe" value="true" gap="1rem">
```
reesscot, line 231: Maybe give an example of value that doesn't look like a boolean, since this really is a string?
```suggestion
  <CheckboxField name="subscribe" value="on" gap="1rem">
```

packages/react/src/primitives/Checkbox/useCheckbox.ts
```diff
@@ -0,0 +1,28 @@
...
+  onChangeProp: React.ChangeEventHandler
+) => {
+  const [dataChecked, setDataChecked] = useState(initialChecked);
+  const [dataFocus, setDataFocus] = useState(false);
+
+  const onChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setDataChecked(e.target.checked);
+
+    // this is the onChange function user passing in
+    if (onChangeProp) {
```
reesscot, line 14: Check for function type first:
```suggestion
    if (isFn(onChangeProp)) {
```

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,79 @@
...
+  gap,
+  hasError = false,
+  id,
+  isEmphasized,
+  isDisabled,
+  isReadOnly,
+  isRequired,
+  justifyContent,
+  label,
+  labelHidden = false,
```
reesscot, line 29: Is there a use case for even hiding a checkbox label? There's no other way to hint at what the checkbox is for. I think we should ignore this prop entirely, and remove it once we fix the `[key: any]: string` issue.

reesscot: Oh I see, it's used for the description. I guess it will go away when you remove that.

packages/react/src/primitives/types/checkbox.ts
```diff
@@ -0,0 +1,18 @@
+import { InputProps } from './input';
+
+export interface CheckboxProps extends InputProps {
+  /**
```
reesscot, line 4: Please make children required, since that's the label text.

packages/ui/src/theme/css/component/checkbox.scss
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.amplify-checkbox__button {
+  position: var(--amplify-components-checkbox-button-position);
+  align-items: var(--amplify-components-checkbox-button-align-items);
+  justify-content: var(--amplify-components-checkbox-button-justify-content);
+  color: var(--amplify-components-checkbox-button-color);
+
+  &::before {
+    content: var(--amplify-components-checkbox-button-before-content);
```
reesscot, line 18: This styling seems very specific to this checkbox styling implementation, do you think all of these need to be CSS variables? I don't think customers need to be able to override the `content` property.

packages/ui/src/theme/tokens/components/checkbox.js
```diff
@@ -0,0 +1,63 @@
...
+        backgroundColor: { value: '{colors.blue.40.value}' },
+      },
+      disabled: {
+        backgroundColor: { value: '{colors.neutral.20.value}' },
+      },
+    },
+  },
+  label: {
+    disabled: {
+      color: { value: '{colors.neutral.40.value}' },
```
reesscot, line 60: Can you double check the color contrast of this against a white background? How does it compare to the "disabled" color of the other fields?

reesscot (changes requested):
This is looking great, just some comments below.

docs/src/pages/ui/primitives/checkboxfield/react.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+### Controlled component
+
+```jsx
+import { useState } from 'react';
+
+import { CheckboxField } from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+
+const [checked, setChecked] = useState(false);
```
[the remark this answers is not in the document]
zchenwei: Yea, we should

docs/src/pages/ui/primitives/checkboxfield/react.mdx
zchenwei: I think I put `isDisabled` into `### Visual options`. For `isReadOnly`, I didn't mention it that because html `read-only` prop essentially means for `value`. A `read-only` checkbox can still be checked or unchecked, which may make users confused. I will add `hasError` examples, and looks like I will have to add error styles, likewise for `Radio`

docs/src/pages/ui/primitives/checkboxfield/react.mdx
zchenwei: Sounds great

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,92 @@
...
+  isRequired,
+  name,
+  onChange: onChangeProp,
+  size,
+  testId,
+  value,
+  ...rest
+}) => {
+  // controlled way should always override uncontrolled way
+  const initialChecked = checked !== undefined ? checked : defaultChecked;
```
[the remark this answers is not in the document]
zchenwei: It is slightly different from the `Radio` one. The `Radio` appears in a group and is subjected to the group value prop.

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,92 @@
...
+        />
+      </VisuallyHidden>
+      <Flex
+        aria-hidden="true"
+        as="span"
+        className={ComponentClassNames.CheckboxButton}
+        data-checked={dataChecked}
+        data-disabled={isDisabled}
+        data-focus={dataFocus}
+        testId={`${testId}-button`}
```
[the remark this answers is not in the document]
zchenwei: Good point, let me do that

packages/react/src/primitives/Checkbox/__tests__/Checkbox.test.tsx
```diff
@@ -0,0 +1,126 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import { Checkbox } from '../Checkbox';
+import { CheckboxProps } from '../../types/checkbox';
+import { ComponentClassNames } from '../../shared';
+
+describe('Checkbox test suite', () => {
+  const basicProps = { name: 'test', value: 'test', testId: 'test' };
```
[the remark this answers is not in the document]
zchenwei: Sounds great

packages/react/src/primitives/Checkbox/__tests__/Checkbox.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const getCheckbox = (props: CheckboxProps) => (
+    <Checkbox {...props}>test</Checkbox>
```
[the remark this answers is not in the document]
zchenwei: That's a valid point

packages/react/src/primitives/Checkbox/useCheckbox.ts
zchenwei: I am assuming you have added this function after I update with main

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,79 @@
...
+  children,
+  className,
+  defaultChecked,
+  descriptiveText,
+  direction = 'column',
+  errorMessage,
+  gap,
+  hasError = false,
+  id,
+  isEmphasized,
```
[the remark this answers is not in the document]
zchenwei: I looking into the checkbox guideline from react-spectrum, which has an emphasis option <- https://spectrum.adobe.com/page/checkbox/

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
zchenwei: Well, I think This is how we structure other fields? Users can pass the flex props directly instead of changing CSS variables

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
zchenwei: Yea, that make sense

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
zchenwei: Move to checkbox

packages/react/src/primitives/types/checkbox.ts
zchenwei: Sounds good, though user could still get around that by using `{}`, `null` or `undefined`. I was wondering if there is a way to tighten up the children type checking but it seems that it will always `& React.ReactNode` :(

packages/ui/src/theme/css/component/checkbox.scss
zchenwei: Oh yea, I think users wouldn't want to change the styling on `before`

packages/ui/src/theme/tokens/components/checkbox.js
```diff
@@ -0,0 +1,63 @@
...
+        borderColor: { value: '{colors.neutral.20.value}' },
+      },
+    },
+    focus: {
+      outlineColor: { value: '{colors.transparent.value}' },
+      outlineStyle: { value: 'solid' },
+      outlineWidth: { value: '{outlineWidths.medium.value}' },
+      outlineOffset: { value: '{outlineOffsets.medium.value}' },
+      borderColor: { value: '{colors.transparent.value}' },
+      boxShadow: { value: '0 0 0 3px {colors.focus.value}' },
```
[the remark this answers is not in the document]
zchenwei: Good point, it looks `2px` is a good choice

packages/ui/src/theme/tokens/components/checkbox.js
```diff
@@ -0,0 +1,63 @@
...
+  },
+  icon: {
+    borderRadius: { value: '20%' },
+    opacity: { value: '{opacities.0.value}' },
+    transform: { value: 'scale(0)' },
+    transitionProperty: { value: 'all' },
+    transitionDuration: { value: '{time.transition.medium.value}' },
+    transitionTimingFunction: { value: 'ease-in-out' },
+    checked: {
+      backgroundColor: { value: '{colors.neutral.60.value}' },
```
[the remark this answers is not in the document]
zchenwei: Sounds great! In that case, I think we can remove the emphasis style because it has no way to emphasize then.

packages/ui/src/theme/tokens/components/checkbox.js
zchenwei: Yup, double check and make it consistent with other fields

packages/react/src/primitives/types/checkbox.ts
reesscot: Hmm, is this because we are using React.FC type?

zchenwei: Yea, I think `React.FC<PropsType>` will enforce that

docs/src/pages/ui/primitives/checkboxfield/react.mdx
```diff
@@ -0,0 +1,246 @@
...
+/* styles.css */
+:root {
+  --amplify-components-checkbox-button-color: var(--amplify-colors-yellow-80);
+}
+/* OR */
+.amplify-checkbox__button {
+  color: var(--amplify-colors-yellow-80);
+}
+```
+
```
reesscot, line 161: Include actual example of global styling being applied here?

docs/src/pages/ui/primitives/checkboxfield/react.mdx
```diff
@@ -0,0 +1,246 @@
...
+## CSS Styling
+
+### Global styling
+
+To override styling on all Checkbox, you can set the Amplify CSS variables or use the built-in `.amplify-checkbox` class.
+
+```css
+/* styles.css */
+:root {
+  --amplify-components-checkbox-button-color: var(--amplify-colors-yellow-80);
```
reesscot, line 154: nit: yellow is probably going to have bad contrast, maybe blue instead?

docs/src/pages/ui/primitives/radiogroupfield/react.mdx
```diff
@@ -298,6 +298,41 @@ import '@aws-amplify/ui-react/styles.css';
   </RadioGroupField>
 </Example>
 
+### Validation error
```
reesscot, line 301: Thanks for adding this!

docs/src/pages/ui/primitives/selectfield/react.mdx
```diff
@@ -305,6 +305,41 @@ import '@aws-amplify/ui-react/styles.css';
   </SelectField>
 </Example>
 
+### Icon
+
+You can change the expand icon used in the select. Simply pass the `icon` prop. Also, you can change the icon color by passing the `iconColor` prop.
```
reesscot, line 310: Nice, thanks for adding this!

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,107 @@
...
+          isDisabled={isDisabled}
+          isReadOnly={isReadOnly}
+          isRequired={isRequired}
+          name={name}
+          onBlur={onBlur}
+          onChange={onChange}
+          onFocus={onFocus}
+          type="checkbox"
+          value={value}
+          {...rest}
```
reesscot, line 74: I just realized that the style props will be useless here given this is VisuallyHidden. Is there a different element where they would make sense, perhaps on the root label? It's challenging to decide where the style props make sense on these custom styled primitives.

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,107 @@
...
+      </VisuallyHidden>
+      <Flex
+        aria-hidden="true"
+        as="span"
+        className={ComponentClassNames.CheckboxButton}
+        data-checked={dataChecked}
+        data-disabled={isDisabled}
+        data-focus={dataFocus}
+        data-error={hasError}
+        testId={useTestId(testId, ComponentClassNames.CheckboxButton)}
```
reesscot, line 85: I think you have to use hooks in the main function body above the JSX. Can you use a hook inside JSX? I've never seen it before. This is probably a data point that we need to add https://www.npmjs.com/package/eslint-plugin-react-hooks

How about moving these above?
```js
const buttonTestID = useTestId(testId, ComponentClassNames.CheckboxButton);
```
```jsx
testId={buttonTestID}
```

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,107 @@
...
+          data-testid={useTestId(testId, ComponentClassNames.CheckboxIcon)}
+          size={size}
+        />
+      </Flex>
+      {children && (
+        <Text
+          as="span"
+          className={ComponentClassNames.CheckboxLabel}
+          data-disabled={isDisabled}
+          data-testid={useTestId(testId, ComponentClassNames.CheckboxLabel)}
```
reesscot, line 100: This hook call will throw an Invariant Violation given it's inside a conditional:
https://reactjs.org/docs/hooks-rules.html
`Don’t call Hooks inside loops, conditions, or nested functions.`

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -0,0 +1,71 @@
...
+        gap={gap}
+        hasError={hasError}
+        id={id}
+        isDisabled={isDisabled}
+        isReadOnly={isReadOnly}
+        isRequired={isRequired}
+        justifyContent={justifyContent}
+        name={name}
+        onChange={onChange}
+        testId={useTestId(testId, ComponentClassNames.Checkbox)}
```
reesscot, line 60: Same comment as above.

packages/react/src/primitives/types/checkbox.ts
reesscot: It's OK if it's a ReactNode, since that will allow them to style the label how they want. I think the main thing thought is to encourage them to provide it, and not use `<CheckboxField/>`.

packages/ui/src/theme/css/component/checkbox.scss
```diff
@@ -0,0 +1,104 @@
...
+
+.amplify-checkbox__icon {
+  opacity: var(--amplify-components-checkbox-icon-opacity);
+  transform: var(--amplify-components-checkbox-icon-transform);
+  border-radius: var(--amplify-components-checkbox-icon-border-radius);
+  transition-property: var(
+    --amplify-components-checkbox-icon-transition-property
+  );
+  transition-duration: var(
+    --amplify-components-checkbox-icon-transition-duration
```
reesscot, line 81: Please make the transition faster, it feels slow right now.

packages/ui/src/theme/tokens/components/checkbox.js
```diff
@@ -0,0 +1,62 @@
...
+        boxShadow: { value: '0 0 0 2px {colors.error.value}' },
+      },
+    },
+  },
+  icon: {
+    borderRadius: { value: '20%' },
+    opacity: { value: '{opacities.0.value}' },
+    transform: { value: 'scale(0)' },
+    transitionProperty: { value: 'all' },
+    transitionDuration: { value: '{time.transition.medium.value}' },
```
reesscot, line 46: I think short is probably better. Under 200ms is supposed to feel instantaneous according to UX research, medium is feeling too long.
```suggestion
    transitionDuration: { value: '{time.transition.short.value}' },
```

reesscot (changes requested):
Looking great, just had some minor comments on docs, and blocking comments on hook use.

packages/react/src/primitives/Checkbox/Checkbox.tsx
zchenwei: Yea, it's hard to make a decision. We hide the really `input` here. Will move it up to container maybe a good choice

packages/react/src/primitives/Checkbox/Checkbox.tsx
zchenwei: Oh yea, good catch. I forgot to update this. The function is not a hook initially.

docs/src/pages/ui/primitives/checkboxfield/react.mdx
```diff
@@ -68,29 +68,47 @@ Use the `size` prop to change the SelectField size. Available options are `small
...
+  <CheckboxField name="subscribe" value="yes">
     {'Subscribe'}
   </CheckboxField>
-  <CheckboxField name="subscribe" value="true" size="large">
+  <CheckboxField name="subscribe" value="yes" size="large">
     {'Subscribe'}
   </CheckboxField>
 </Example>
 
+### Value
```
reesscot, line 94: Thanks for adding!

packages/react/src/primitives/types/checkbox.ts
```diff
@@ -0,0 +1,18 @@
+import { FlexProps } from './flex';
+import { InputProps } from './input';
+
+export interface CheckboxProps extends FlexProps, InputProps {
+  /**
+   * The label text
+   */
+  children: string;
```
reesscot, line 8: I think this should match the other field [label](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/types/field.ts#L22) type:
```suggestion
  children: FieldProps['label'];
```

packages/react/src/primitives/types/checkbox.ts
```diff
@@ -0,0 +1,18 @@
...
+  /**
+   * The name of the input field in a checkbox (Useful for form submission).
+   */
+  name: string;
+
+  /**
+   * The value of the checkbox.
```
reesscot, line 15:
```suggestion
   * The submitted as value when checked (Useful for form submission.
   * Shows on form submission as key pair `name:value`
```

reesscot (approved):
Ship it!

PR #379, opened 2021-09-15
feat(angular): add i18n support

wlee221:
*Issue #, if available:*

*Description of changes:* 
- Add i18n support for Angular
- testable example for i18n
- Standardize auth strings we use in the authenticator
- `translate` helper function

I only applied I18n shared changes to Angular, because I wanted to prevent any breaking changes to pre-existing Vue and Angular I18n implementation as we near preview release. Vue and React I18n will still work as before. I added `@skip-[framework]` to track testable examples. 

I added a guided tour to demonstrate I18n process.

<details>

 <summary>Angular I18n checkboxes</summary>

- [x] confirm-sign-in
- [x] confirm-sign-up
- [x] federated-sign-in
- [x] force-new-password
- [x] reset-password
- [x] setup-totp
- [x] sign-in
- [x] sign-up
- [x] verify-user
- [x] confirm-reset-password
- [x] confirm-verify-user

</details>

### Appendix

Autocompletition!

https://user-images.githubusercontent.com/43682783/133669571-a3cf2193-1e74-4089-a12d-ac6acc238090.mp4

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #408, opened 2021-09-21
Default I18n Support

ericclemmons:
*Issue #, if available:* https://app.asana.com/0/0/1200973733971731/f

*Description of changes:*

- [x] Call `I18n.putVocabularies(translations)` by default
- [x] Document I18n
- [x] Add `.feature` tests
- [ ] ~Listen for `I18n.setLanguage` changes~
  
  Can't do this yet without a PR to send `Hub` events in Amplify JS

- [ ] ~Default using `window.navigation.language`~
 
  Not doing this explicitly because this would be a change in behavior where _by simply adding the new `Authenticator`_ could potentially change the entire page's language. (If the customer is using `I18n` anywhere else)

> Screen Shot 2021-09-21 at 4 52 48 PM
Screen Shot 2021-09-21 at 4 52 57 PM

### DIFF
--- a/docs/src/pages/ui/components/authenticator/index.page.mdx
+++ b/docs/src/pages/ui/components/authenticator/index.page.mdx
@@ -107,7 +107,7 @@ can sign in with Google, Amazon or Facebook.
 
 ### Internationalization (I18n)
 
-The `Authenticator` ships with translations for:
+The `Authenticator` ships with [translations](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/i18n/translations.ts) for:
 
 - `en` – English (default)
 - `zh` – Chinese
@@ -117,7 +117,7 @@ The `Authenticator` ships with translations for:
 - `ja` – Japenese
 - `es` – Spanish
 
-These translations can be customized using the [Amplify JS' `I18n`](https://docs.amplify.aws/lib/utilities/i18n/q/platform/js/) module:
+These [translations](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/i18n/translations.ts) can be customized using the [Amplify JS' `I18n`](https://docs.amplify.aws/lib/utilities/i18n/q/platform/js/) module:
 
 ```js
 I18n.setLanguage('fr');
--- a/docs/src/pages/ui/components/authenticator/index.page.mdx
+++ b/docs/src/pages/ui/components/authenticator/index.page.mdx
@@ -105,6 +105,37 @@ can sign in with Google, Amazon or Facebook.
 
 <Feature name="sign-in-federated" />
 
+### Internationalization (I18n)
+
+The `Authenticator` ships with translations for:
+
+- `en` – English (default)
+- `zh` – Chinese
+- `fr` – French
+- `de` – German
+- `it` – Italian
+- `ja` – Japenese
+- `es` – Spanish
+
+These translations can be customized using the [Amplify JS' `I18n`](https://docs.amplify.aws/lib/utilities/i18n/q/platform/js/) module:
+
+```js
+I18n.setLanguage('fr');
+
+I18n.putVocabularies({
+  fr: {
+    'Sign In': 'Se connecter',
+    'Sign Up': "S'inscrire",
+  },
+  es: {
+    'Sign In': 'Registrarse',
+    'Sign Up': 'Regístrate',
+  },
+});
+```
+
+<Feature name="i18n" />
+
 ## Headless Usage
 
 The `Authenticator`'s UI is powered by a state machine that you can hook into to customize behavior or create your own API.
--- a/examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
+++ b/examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
@@ -4,7 +4,6 @@ import { Amplify, I18n } from 'aws-amplify';
 
 Amplify.configure(awsExports);
 
-I18n.putVocabularies(translations);
 I18n.setLanguage('fr');
 
 function App({ state, send }) {
--- a/examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
+++ b/examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
@@ -1,4 +1,4 @@
-import { translations, withAuthenticator } from '@aws-amplify/ui-react';
+import { withAuthenticator } from '@aws-amplify/ui-react';
 import awsExports from '@environments/auth-with-username-no-attributes/src/aws-exports';
 import { Amplify, I18n } from 'aws-amplify';
 
--- a/examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
+++ b/examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
@@ -1,0 +1,19 @@
+import { translations, withAuthenticator } from '@aws-amplify/ui-react';
+import awsExports from '@environments/auth-with-username-no-attributes/src/aws-exports';
+import { Amplify, I18n } from 'aws-amplify';
+
+Amplify.configure(awsExports);
+
+I18n.putVocabularies(translations);
+I18n.setLanguage('fr');
+
+function App({ state, send }) {
+  return (
+    <>
+      <h1>Hello {state.context.user.username}</h1>
+      <button onClick={() => send('SIGN_OUT')}>Sign out</button>
+    </>
+  );
+}
+
+export default withAuthenticator(App);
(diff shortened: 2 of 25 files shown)


packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -5,8 +5,15 @@
...
 import { escapeRegExp } from 'lodash';
 
+let language = 'en-US';
+
 Given("I'm running the example {string}", (example: string) => {
-  cy.visit(example);
+  cy.visit(example, {
+    // See: https://glebbahmutov.com/blog/cypress-tips-and-tricks/#control-navigatorlanguage
+    onBeforeLoad(win) {
+      Object.defineProperty(win.navigator, 'language', { value: language });
```
ericclemmons, line 14: This will come in handy if & when we need to change language in a `.feature` file.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -31,6 +38,12 @@ When('I type a new {string}', (loginMechanism: string) => {
   );
 });
 
+When('I click the {string} tab', (label: string) => {
```
ericclemmons, line 41: @slaymance I made this more abstract to support I18n instead of `I click the Create Account tab`.

packages/react/src/components/Authenticator/index.tsx
```diff
@@ -23,6 +25,10 @@ export function Authenticator({
 
   const [state, send] = useActor(service);
 
+  React.useEffect(() => {
+    I18n.putVocabularies(translations);
+  }, []);
```
ericclemmons, line 30: I was worried that there would be an English-only render and a French render when the state changes.

Interestingly, this effect is called prior to rendering (like `componentWillMount`) so the Authenticator renders correctly the 1st time!

packages/react/src/index.tsx
```diff
@@ -8,4 +8,4 @@ export * from './theming';
 export * as components from './components';
 export * as primitives from './primitives';
 
-export { theme, translations } from '@aws-amplify/ui';
```
ericclemmons, line 11: I don't think we _need_ to have this exported by default, do we?

packages/e2e/features/ui/components/authenticator/i18n.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Internationalization (I18n)
+
+  The `Authenticator` contains translations using `I18n`.
+
+  Background:
+    Given I'm running the example "ui/components/authenticator/i18n"
+
+  @angular @todo-angular @react @vue @todo-vue
```
ErikCH, line 8: I can add the Vue implementation later and make sure it works

packages/react/src/components/Authenticator/index.tsx
ErikCH: 👍

ErikCH (approved):

packages/react/src/components/Authenticator/index.tsx
reesscot: TIL, I'm not sure I would have expected that. Thanks for calling this out!

ericclemmons: I was surprised as well! But the important take-away is that **we want to avoid side-effects from `import`ing**.

Meaning, `I18n.putVocabularies` **should not** be at the same level as `export...` or `import...`, because it creates side-effects that tree-shaking (particularly Angular) doesn't support well.

ericclemmons:
@wlee221 Ahhh, I see! I need to update your previous `auth-with-translations` e2e test to match the `i18n` one I created.  I totally forgot about this 🤦

examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
```diff
@@ -0,0 +1,18 @@
+import { withAuthenticator } from '@aws-amplify/ui-react';
+import awsExports from '@environments/auth-with-username-no-attributes/src/aws-exports';
+import { Amplify, I18n } from 'aws-amplify';
+
+Amplify.configure(awsExports);
+
+I18n.setLanguage('fr');
```
wlee221, line 7: I already created an angular example that tests (but not documentation) for I18n on [`ui/components/authenticator/auth-with-translation`](https://github.com/aws-amplify/amplify-ui/blob/16013d22bca8247b27a1d77b9feac92df33fa660/examples/angular/src/pages/ui/components/authenticator/auth-with-translations/auth-with-translations.component.ts). Do you think we can combine these into one route?

wlee221: I also wrote a `.feature` file which I marked `@todo-react` and `@todo-vue`:

https://github.com/aws-amplify/amplify-ui/blob/16013d22bca8247b27a1d77b9feac92df33fa660/packages/e2e/features/ui/components/authenticator/auth-with-translations.feature#L7-L16

wlee221: One cool thing I did on my test is that I directly pulled in the expected translations from `@aws-amplify/ui` to do the translation testing like so: 

https://github.com/aws-amplify/amplify-ui/blob/16013d22bca8247b27a1d77b9feac92df33fa660/packages/e2e/cypress/integration/ui/components/authenticator/auth-with-translations/auth-with-translations.steps.ts#L4-L12

That way the test files is in sync with the translation file that we'll be developing over time. But this is more wordy and just testing for header text translated might be sufficient.

wlee221 (changes requested):
The docs and test changes looks good! But I already have created a `.feature` file and `.step.js` file for `i18n` in #379. Sorry for not communicating this clearly.

Let's try to combine the one I have and yours in this PR into one `.feature` file along with its step file. Feel free to take any strategy that mine uses and combine it to yours if you'd like (or none, I won't be hurt 😄 )!

ericclemmons:
@wlee221 I agree on combining these, but was running into both development issues (`yarn angular dev` runs without errors) where I got errors from Cypress like:
>  Cannot find module '@aws-amplify/ui-angular' or its corresponding type declarations.
>
>  import { UiAngularModule } from '@aws-amplify/ui-angular';

and 

> Angular JIT compilation failed: '@angular/compiler' not loaded!

Let's talk today about getting this resolved 🙏

examples/next/pages/ui/components/authenticator/i18n/index.page.tsx
ericclemmons: I agree, so let's combine the best of both worlds with something like:

> `"Sign In" label is translated into "ja"`

docs/src/pages/ui/components/authenticator/index.page.mdx
```diff
@@ -105,6 +105,37 @@ can sign in with Google, Amazon or Facebook.
...
+I18n.putVocabularies({
+  fr: {
+    'Sign In': 'Se connecter',
+    'Sign Up': "S'inscrire",
+  },
+  es: {
+    'Sign In': 'Registrarse',
+    'Sign Up': 'Regístrate',
+  },
+});
```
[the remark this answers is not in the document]
ericclemmons: I agree, that would be consistent with these docs (and better for customers):
> https://docs.amplify.aws/ui/customization/translations/q/framework/angular/

Admittedly, this snippet came directly from https://docs.amplify.aws/lib/utilities/i18n/q/platform/js/#putvocabularies 🙈 

We'll likely want to have consistency between exports (e.g. `Translations` instead of `DefaultTexts`), but with this being authenticator-only, perhaps we should export `AuthenticatorTranslations` or similar 🤔

ericclemmons:
Also, because the Angular implementation doesn't use tabs the `auth-with-translations.feature` has to be updated as well from using a `button` role to a `tab` role.

wlee221 (approved):
Thanks for the angular adjustment! Looks good to me.

docs/src/pages/ui/components/authenticator/index.page.mdx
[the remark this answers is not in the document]
wlee221: Yeah, I wasn't sure of `DefaultTexts` either. The only reason I didn't use `Translations` was because the variable name `translations` was already taken: https://github.com/aws-amplify/amplify-ui/blob/7ccea320be5b62a0050cef9464a64536a1bf6b25/packages/ui/src/i18n/translations.ts#L101-L103.

Now that we're providing translation by default, maybe we rename that `translations` to something else so `Translations` isn't confuisng.

In terms of authenticator-only, we can name it `AuthenticatorTranslations`, then combine with other error string Amplify provides like [`AuthErrorStrings`](https://github.com/aws-amplify/amplify-js/blob/main/packages/auth/src/common/AuthErrorStrings.ts#L1). 


```ts
import { AuthErrorStrings } from 'aws-amplify';
export const Translations = {...AuthenticatorTranslations, ...AuthErrorStrings}
```

PR #409, opened 2021-09-21
Grid primitive

reesscot:
*Description of changes:*

This PR implements the Grid primitive. The `rowSpan` and `columnSpan` props don't map directly to CSS Properties, and thus need to be transformed to `grid-row: span ${spanValue}/span ${spanValue}` and `grid-column: span ${spanValue}/span ${spanValue}` CSS properties. The rowSpan and columnSpan props work can be responseive when the customer specifies styles like this:

```jsx
    <Grid
      templateColumns="300px 300px"
      autoRows="100px"
      justifyContent="center"
      gap="20px">
      <Card rowSpan={3}>1</Card> {/** <-- look here **/}
      <Card columnSpan={2}>2</Card>  {/** <-- look here **/}
    </Grid>
```

Or  using the array or object syntax:

```jsx
    <Grid 
      templateColumns="300px 300px"
      autoRows="100px"
      justifyContent="center"
      gap="20px">
      <Card rowSpan={[1,2]}>1</Card> {/** <-- look here **/}
      <Card columnSpan={{base: 1, large: 2}}>2</Card> {/** <-- look here **/}
    </Grid>
```


packages/react/__tests__/exports.ts
```diff
@@ -1866,23 +1868,16 @@ describe('@aws-amplify/ui-react', () => {
           "VisuallyHidden",
           "Wrapper",
           "components",
-          "convertStylePropsToStyleObj",
```
reesscot, line 1869: Removed these util functions, because I don't see any reason for them to be exposed on the public api.

packages/react/src/primitives/shared/__tests__/utils.test.tsx
```diff
@@ -30,119 +20,6 @@ const props: ViewProps = {
   className: 'my-section',
 };
 
-let breakpoints = theme.breakpoints.values;
```
reesscot, line 33: Moved these out to styleUtils.ts file

packages/react/src/primitives/types/style.ts
```diff
@@ -16,7 +17,7 @@ export type ResponsiveStyle<PropertyType> =
   | PropertyType[]
   | ResponsiveObject<PropertyType>;
 
-export interface BaseStyleProps {
+export interface BaseStyleProps extends GridItemStyleProps {
```
reesscot, line 20: This allows all primitives to be used as Grid items

packages/react/src/primitives/utils/testUtils.ts
```diff
@@ -0,0 +1,16 @@
+export const errorMessageWrapper = (fn: () => void, message: string) => {
```
reesscot, line 1: This function allows us to render a customer error message for an expect that throws.

packages/react/__tests__/exports.ts
hvergara: thanks!

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
+import * as React from 'react';
```
hvergara, line 1: any reason to use `*` instead of default export?

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+import autoprefixer from 'autoprefixer';
+import postcssJs from 'postcss-js';
```
hvergara, line 3: nit: `postcssJs` => `postcss`

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+  GridItemStyleProps,
+  ResponsiveObject,
+  ViewProps,
+} from '../types/index';
+
+import { getValueAtCurrentBreakpoint } from './responsive/utils';
+import { useBreakpoint } from './responsive/useBreakpoint';
+import { Breakpoint, Breakpoints } from '../types/responsive';
+
+import { useTheming } from '../../theming/';
```
hvergara, line 19:
```suggestion
import { useTheming } from '../../theming';
```

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export const prefixer = postcssJs.sync([autoprefixer]);
+
+import {
+  ComponentPropsToStylePropsMap,
+  ComponentPropsToStylePropsMapKeys,
+  GridItemStyleProps,
+  ResponsiveObject,
+  ViewProps,
+} from '../types/index';
```
hvergara, line 13:
```suggestion
} from '../types';
```

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+import { filterOutNullOrEmptyStringValues, isNullOrEmptyString } from './utils';
+
+/**
+ * Transforms style props to another target prop
+ * where the original is a simpler API than the target.
+ * This function will remove the original prop and
+ * replace target prop values with calculated
+ * E.g. rowSpan => row, columnSpan => column
+ * @param props T
+ * @returns
```
hvergara, line 29: are these from another doc?

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+  }, [rowSpan, columnSpan, convertGridSpan]);
+
+  return {
+    row: !isNullOrEmptyString(row) ? row : rowFromSpanValue,
+    column: !isNullOrEmptyString(column) ? column : columnFromSpanValue,
+    ...rest,
+  };
+};
+
+export const usePropStyles = (props: ViewProps, style: any) => {
```
hvergara, line 50:
```suggestion
export const usePropStyles = (props: ViewProps, style: React.CSSProperties) => {
```

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (Array.isArray(spanValue)) {
+      return spanValue.map((value) => getGridSpan(value));
+    } else {
+      const newObj: ResponsiveObject<string> = {};
+      Object.entries(spanValue).forEach(([key, value]) => {
+        newObj[key] = getGridSpan(value);
+      });
+      return newObj;
+    }
+  }
```
hvergara, line 117:
```suggestion
  if (Array.isArray(spanValue)) {
      return spanValue.map(getGridSpan);
  }
  
  if (typeof spanValue === 'object' && spanValue != null) {
    const newObj: ResponsiveObject<string> = {};
    Object.entries(spanValue).forEach(([key, value]) => {
      newObj[key] = getGridSpan(value);
    });
    return newObj;
  }
```

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+export const convertStylePropsToStyleObj: ConvertStylePropsToStyleObj = ({
+  props = {},
+  style = {},
+  breakpoint,
+  breakpoints,
+}) => {
+  ComponentPropsToStylePropsMapKeys.filter((stylePropKey) =>
+    filterOutNullOrEmptyStringValues(props[stylePropKey])
+  ).forEach((stylePropKey) => {
+    let value = getValueAtCurrentBreakpoint(
```
hvergara, line 141: if `value` never change, can be a `const`?

packages/react/src/primitives/shared/utils.ts
```diff
@@ -1,116 +1,21 @@
...
-  } = useTheming();
+export const isFn = (fn: unknown): fn is Function => typeof fn === 'function';
 
-  const breakpoint = useBreakpoint({
-    breakpoints,
-    breakpointUnit,
-    defaultBreakpoint: defaultBreakpoint as Breakpoint,
-  });
+export const isNullOrEmptyString = (value: unknown) =>
+  value == null || (typeof value === 'string' && !strHasLength(value));
```
hvergara, line 14: `strHasLength` already checks for value type (`string`)

```suggestion
  value == null || !strHasLength(value);
```

packages/react/src/primitives/types/flex.ts
```diff
@@ -1,34 +1,14 @@
...
-   */
-  alignItems?: ResponsiveStyle<Property.AlignItems>;
-
-  /**
-   * Sets the distribution of space between and around content items
-   */
-  alignContent?: ResponsiveStyle<Property.AlignContent>;
-
+import { CSSLayoutStyleProps, ResponsiveStyle } from './style';
+export interface FlexStyleProps extends CSSLayoutStyleProps {
```
hvergara, line 5:
```suggestion

export interface FlexStyleProps extends CSSLayoutStyleProps {
```

packages/react/src/primitives/types/style.ts
hvergara: great!

docs/src/components/DemoBox.tsx
```diff
@@ -1,17 +1,53 @@
...
   primitiveName: string;
+  startCollapsed?: boolean;
 }
 
 export const DemoBox: React.FC<DemoBoxProps> = ({
   primitiveName,
   children,
+  startCollapsed = true,
 }) => {
+  const [collapsed, toggleCollapsed] = React.useState(startCollapsed);
```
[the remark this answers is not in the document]
reesscot: good point, will update

docs/src/pages/ui/primitives/grid/demo.tsx
```diff
@@ -0,0 +1,158 @@
...
+          alignItems={gridContainerProps.alignItems}
+          autoColumns={gridContainerProps.autoColumns}
+          autoRows={gridContainerProps.autoRows}
+          columnGap={gridContainerProps.columnGap}
+          gap={gridContainerProps.gap}
+          rowGap={gridContainerProps.rowGap}
+          templateAreas={gridContainerProps.templateAreas}
+          templateColumns={gridContainerProps.templateColumns}
+          templateRows={gridContainerProps.templateRows}
+          justifyContent={gridContainerProps.justifyContent}
```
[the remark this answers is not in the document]
reesscot: I had this originally, but then React complains when we try to pass the setter functions to the DOM element. e.g., `setJustifyContent` is not a value DOM attribute

packages/react/src/primitives/shared/styleUtils.ts
reesscot: Yes, this is the new recommended way to import React. See:
https://epicreact.dev/importing-react-through-the-ages/

TLDR (my understanding): React doesn't actually expose a default export. We've been able to do this previously by using the [esModuleInterop](https://www.typescriptlang.org/tsconfig#esModuleInterop) setting which enables [allowSyntheticDefaultImports](https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports), which converts `import React from 'react'` to `import * as React from 'react'`. Doing a namespace import is ESM compatible and future proof.



https://twitter.com/housecor/status/1386839524362407943?lang=en

packages/react/src/primitives/shared/styleUtils.ts
reesscot: updated

packages/react/src/primitives/shared/styleUtils.ts
reesscot: This was a mistake. I've removed it since that information is in the Types

packages/react/src/primitives/shared/styleUtils.ts
reesscot: Good suggestion, will make the change

packages/react/src/primitives/shared/styleUtils.ts
reesscot: Good catch, will update

packages/react/src/primitives/shared/utils.ts
reesscot: Good point, updated.

zchenwei (changes requested):
Thanks for all the work! This is looking great and just left a few comments.

packages/react/src/primitives/types/grid.ts
```diff
@@ -0,0 +1,32 @@
+import { Property } from 'csstype';
+import { AriaProps, BaseComponentProps } from './base';
+
```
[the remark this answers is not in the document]
reesscot: I assume you mean the spacing here. Fixed.

packages/react/src/primitives/shared/utils.ts
```diff
@@ -1,116 +1,21 @@
...
-      prefixer(
-        convertStylePropsToStyleObj({ props, style, breakpoint, breakpoints })
-      ),
-    [props, style, breakpoints, breakpoint]
-  );
-};
-
-const filterOutNullOrEmptyStringValues = (value) =>
+export const filterOutNullOrEmptyStringValues = (value) =>
   value != null && (typeof value !== 'string' || strHasLength(value));
```
hvergara, line 17: I personally think we can rid of this function (as we can use `isNullOrEmptyString` as replacement). But, if you don't agree, we can simplify it a lot, just returning the opposite of `isNullOrEmptyString`:
 
```suggestion
export const filterOutNullOrEmptyStringValues = (value) => !isNullOrEmptyString(value);
```

packages/react/src/primitives/shared/utils.ts
```diff
@@ -1,116 +1,21 @@
...
-  const {
-    theme: {
-      breakpoints: {
-        values: breakpoints,
-        unit: breakpointUnit,
-        defaultBreakpoint,
-      },
-    },
-  } = useTheming();
+export const isFn = (fn: unknown): fn is Function => typeof fn === 'function';
```
hvergara, line 11: minor: can we call it `isFunction`? Also, I couldn't find any reference to this typeguard in our codebase. Why do we need this?

packages/react/src/primitives/shared/styleUtils.ts
hvergara: discussed offline: we'll change default exports in a separate PR

docs/src/pages/ui/primitives/grid/demo.tsx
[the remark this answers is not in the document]
hvergara: got it, thanks!

hvergara (approved):
Minor comments about some shared util functions (not a blocker though). Feel free to change it or not, I'm approving anyway :)

packages/react/src/primitives/shared/utils.ts
reesscot: Yes, will update the name. We are using it in some new PR's that haven't been merged yet.

packages/react/src/primitives/types/grid.ts
```diff
@@ -0,0 +1,32 @@
...
+  columnSpan?: ResponsiveStyle<number | 'auto'>;
+  columnStart?: ResponsiveStyle<Property.GridColumnStart>;
+  row?: ResponsiveStyle<Property.GridRow>;
+  rowEnd?: ResponsiveStyle<Property.GridRowEnd>;
+  rowSpan?: ResponsiveStyle<number | 'auto'>;
+  rowStart?: ResponsiveStyle<Property.GridRowStart>;
+}
+
+export interface GridProps
+  extends BaseComponentProps,
```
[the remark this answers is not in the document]
reesscot: Fixed, and updated to extend ViewProps

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+    });
+    return newObj;
+  }
+  return null;
+};
+
+export const getGridSpan = (spanValue: number | 'auto'): string => {
+  return spanValue === 'auto'
+    ? 'auto'
+    : `span ${spanValue} / span ${spanValue}`;
```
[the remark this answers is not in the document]
reesscot: Fixed

packages/react/src/primitives/shared/utils.ts
reesscot: Good idea, removed.

zchenwei (approved):
LGTM!

PR #410, opened 2021-09-22
feat(angular): Sign in parity style

wlee221:
*Issue #, if available:*

*Description of changes:* Deferring tab until I style each screen. This will be a bit bigger than my other style PRs to come because here I do all the initial styling setup.

Screen Shot 2021-09-22 at 9 24 20 AM

Screen Shot 2021-09-22 at 9 24 06 AM


packages/angular/projects/ui-angular/package.json
```diff
@@ -4,19 +4,21 @@
   "scripts": {
     "build": "yarn --cwd ../../ build",
     "dev": "yarn --cwd ../../ dev",
-    "test": "yarn --cwd ../../ test"
+    "test": "yarn --cwd ../../ test",
+    "generate:css": "npx tailwindcss -i ./styles/component.tailwind.css -o ./styles/component.css"
```
wlee221, line 8: Angular libraries built with `@angular/cli` does not support tailwind: https://github.com/ng-packagr/ng-packagr/issues/1943. There are some workarounds in the thread, but the simplest solution I found was to geerate the css directly as we only use tailwind for authenticator styling. Added a command here to keep track of how I generated it. 

It would be nice if @aws-amplify/ui can output a `authenticator.css` that has tailwind preprocessed so that it's readily usable by all frameworks.

packages/angular/projects/ui-angular/theme.css
```diff
@@ -1,99 +1,2 @@
...
-[data-amplify-error]:last-child {
-  margin-bottom: 1rem;
-}
-
-[data-amplify-federated] {
-  display: flex;
-  flex-direction: column;
-  margin-bottom: 1rem;
-}
+@import '@aws-amplify/ui/dist/styles.css';
```
wlee221, line 1: Angular app built with `@angular/cli` has trouble finding `'@aws-amplify/ui/styles.css'`, though it is declared through `exports` and `browser` field. 

Adding `dist/` to unblock upcoming styling PRs, but I'll revisit this later (or open an issue on @angular/cli). Possibly related: https://github.com/angular/angular-cli/issues/3500

packages/angular/projects/ui-angular/src/lib/components/amplify-sign-in/amplify-sign-in.component.html
```diff
@@ -2,41 +2,64 @@
...
-            {{ createAccountText }}
-          </button>
-        </div>
-        <ng-container
-          [ngTemplateOutlet]="customComponents.signInButton || signInButton"
-          [ngTemplateOutletContext]="context"
-        >
-        </ng-container>
-      </div>
+      <amplify-error *ngIf="remoteError">
```
ErikCH, line 58: This is a nitpick but you deleted the styles for `amplify-error`. So it no longer appears as red text, it's now black for errors

packages/angular/projects/ui-angular/src/lib/components/amplify-sign-in/amplify-sign-in.component.html
```diff
@@ -2,41 +2,64 @@
...
+      [disabled]="isPending"
+    >
+      <h3 class="amplify-heading">{{ this.headerText }}</h3>
       <amplify-user-name-alias></amplify-user-name-alias>
       <amplify-form-input
         data-amplify-password
         name="password"
         type="password"
         autocomplete="current-password"
       ></amplify-form-input>
```
ErikCH, line 30: Are you planning on adding the show password button when you add the tabs?

packages/angular/projects/ui-angular/src/lib/components/amplify-sign-in/amplify-sign-in.component.html
wlee221: Ah, good catch! Fixed in https://github.com/aws-amplify/amplify-ui/pull/410/commits/00c6d9f477bee8706cc13fece3f313b5c7c233e0.

image

packages/angular/projects/ui-angular/src/lib/components/amplify-sign-in/amplify-sign-in.component.html
wlee221: Right, I need to add that in. I'll do that in my sign up PR and create a password-control primitive like you did for the Vue implementation.

ErikCH (approved):

PR #414, opened 2021-09-22
XState Facade

ericclemmons:
*Issue #, if available:* https://app.asana.com/0/0/1200982218231970/f

*Description of changes:*

- [x] `user` from `state.context.user`
- [x] `signOut` from `send('SIGN_OUT')`
- [ ] `state` that should be `signIn.pending` (or similar)


packages/ui/src/helpers/auth.ts
```diff
@@ -144,3 +144,15 @@ export const getSendEventAliases = (send: Sender<AuthEvent>) => {
     submit: sendToMachine('SUBMIT'),
   } as const;
 };
+
+export const getServiceFacade = ([state, send]) => {
+  const { user } = state.context as AuthContext;
```
wlee221, line 149: `state.context.user` isn't always up to date as `actorState.context.user` because the parent state context isn't updated until the actor finishes (this can be improved). 

This is okay with the use cases in this PR, but can be a problem later if customer modifies other screens like setup-totp.

```suggestion
  const { user } = getActorContext(state) as AuthActorContext;
```

wlee221: A long term solution would be have a single source of truth that contains all up-to-date context in both parent and child actors -- basically 
```tsx
const getContext(state) = {
  const parentContext = state;
  const childContext = getActorContext(state);
  return { ...parentContext, ...childContext }; // child context should take precedence.
}
```
Added this to e2 tasks.

ericclemmons: That sounds desirable, actually?  What I'm hearing is that even though the user is logged in, the Authenticator is still going through its Setup TOTP flow, so the `user` isn't yet available because the machine isn't _done_.

That sounds like customers could do something like this then:

```js
function App() {
  const { user } = useAuthenticator();

  if (!user) {
    return <Authenticator>;
  }

  return <RestOfMyApp />
}
```

ErikCH (approved):
Good first start with our facades

packages/ui/src/helpers/auth.ts
wlee221: Yeah, I see your point, especially for `{user, signOut}` use case. Let's wait until we actually see a need to do so!

PR #415, opened 2021-09-23
feat(angular): Sign up + Confirm sign up parity style

wlee221:
*Issue #, if available:*

*Description of changes:*

- [x] Add amplify-button primitive: `<button amplify-button></button>`
- [x] Add amplify-text-field primitive: `<amplify-text-field></amplify-text-field>`
- [x] Add amplify-password-field-primitive: `<amplify-password-field></amplify-password-field>`
- [x] Rename amplify-form-input to amplify-form-field
- [x] Add styles to sign-up
- [x] Add styles to confirm-sign-up

image

image


packages/angular/projects/ui-angular/src/lib/components/amplify-confirm-sign-up/amplify-confirm-sign-up.component.html
```diff
@@ -2,41 +2,36 @@
   Define default contents here
 -->
 <ng-template #confirmSignUpButton>
-  <button data-amplify-button type="submit">{{ confirmText }}</button>
+  <button amplify-button variation="primary" fullWidth="true" type="submit">
+    {{ confirmText }}
+  </button>
```
wlee221, line 7: I'm using directives to declare `amplify-button` primitive. That way it does not put a `amplify-button` element wrapper around the button, ie. we get `<button></button>` instead of `<amplify-button><button></button></amplify-button>`. 

This allows us to directly place attributes on the `button`, and ensures that css grids and flexbox style are applied correctly to the button and not to its wrapper.

This pattern is used for angular [bootstrap](https://ng-bootstrap.github.io/#/components/buttons/examples) and [material](https://material.angular.io/components/button/examples). I'll be using this pattern for primitives that consists of one html element (button, label, ...).

wlee221: This is only used internally for now and its spec can change later.

packages/angular/projects/ui-angular/src/lib/primitives/amplify-form-field/amplify-form-field.component.ts
```diff
@@ -18,10 +18,10 @@ import { StateMachineService } from '../../services/state-machine.service';
  * that does not make any auth-related inference.
  */
 @Component({
-  selector: 'amplify-form-input',
-  templateUrl: './amplify-input.component.html',
+  selector: 'amplify-form-field',
+  templateUrl: './amplify-form-field.component.html',
 })
-export class AmplifyInputComponent implements OnInit {
```
wlee221, line 24: This is not really a primitive because it's opinionated to auth usage. I'll move this out of the primitive folder in a later PR.

packages/angular/projects/ui-angular/src/lib/components/amplify-confirm-sign-up/amplify-confirm-sign-up.component.html
ErikCH: Nice clever! Yeah I saw the selector ` selector: 'button[amplify-button]',` for you Amplify button component and that makes sense now. I think you can have that just `[amplify-button]` and it would work too not just buttons. But either way good job

ErikCH (approved):
👍

packages/angular/projects/ui-angular/src/lib/components/amplify-confirm-sign-up/amplify-confirm-sign-up.component.html
wlee221: Yeah! That would be nice too, so that it can be applied to links and other elements as well. I'll take note 📝

PR #417, opened 2021-09-23
Catalog

hvergara:
*Description of changes:*
This PR introduces a Primitives catalog generation script, which output can be used by consumers to know which properties (and types) are available for every component.

Example:
```json
{
  "Alert": {
    "properties": {
      "variation": {
        "type": "string",
        "enum": [
          "info",
          "error",
          "warning",
          "success"
        ]
      },
      "isDismissible": {
        "type": "boolean"
      },
      "hasIcon": {
        "type": "boolean"
      },
      "iconSize": {
        "type": "string"
      }

// more...

```


packages/react/scripts/generatePrimitivesCatalog.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Remove double quotes from String literal types
+ */
+const cleanStringLiteral = (literal: string) => literal.replace(/"/g, '');
+
+/**
+ * Filter browser vendor prefixes (-webkit, -moz, -ms, ...)
+ */
+const isValidStringValue = (value: string) => !value.startsWith('-');
```
reesscot, line 29: Curious where browser prefixes are being found in the Types

reesscot: Oh, I think these are included in the React CSS property types, aren't they?

reesscot (approved):
LGTM

packages/react/scripts/generatePrimitivesCatalog.ts
hvergara: yep!

PR #423, opened 2021-09-23
<Authenticator loginMechanisms>

ericclemmons:
*Issue #, if available:*  https://app.asana.com/0/0/1200982218231970/f

*Description of changes:*

Until zero-config is available from `amplify pull` for all backends (new and existing), the `loginMechanisms` prop is an escape-hatch for customers to use to explicitly change the Authenticator UI.

- [x] `loginMechanisms={[...]}`, similar to the previous `usernameAlias="..."`
- [x] Removes `Amplify.configure({ ...awsExports, auth: { login_mechanisms: [...] } })` 


packages/e2e/cypress/fixtures/force-change-password.json
```diff
@@ -0,0 +1,8 @@
+{
+  "ChallengeName": "NEW_PASSWORD_REQUIRED",
```
ericclemmons, line 2: @slaymance I went to this user pool, created a test user for this state, and gave them a temporary password.

Is this the same as `FORCE_CHANGE_PASSWORD`?

packages/e2e/features/ui/components/authenticator/sign-in-with-username.feature
```diff
@@ -33,7 +33,9 @@ Feature: Sign In with Username
 
   # FORCE_CHANGE_PASSWORD tests are skipped as the temporary passwords used for these
   # test accounts will expire in Cognito.
+  @angular @react @vue @todo-angular @todo-vue
   Scenario: Sign in with force change password credentials
+    Given intercept '{ "headers": { "X-Amz-Target": "AWSCognitoIdentityProviderService.RespondToAuthChallenge" } }' with fixture "force-change-password"
```
ericclemmons, line 38: This is pretty easy to do! Use DevTools' Network tab, find a pattern to match on, and copy/paste 👍

packages/react/src/components/Authenticator/index.tsx
```diff
@@ -20,12 +20,27 @@ import { SignIn } from './SignIn';
 import { SignUp } from './SignUp';
 import { ConfirmVerifyUser, VerifyUser } from './VerifyUser';
 
+type AuthenticatorProps = {
+  children: (facade: ReturnType<typeof getServiceFacade>) => JSX.Element;
+  className?: string;
+  initialState?: 'signIn' | 'signUp' | 'resetPassword';
```
ericclemmons, line 26: Not using this yet, but will be testing out these values _soon_.

packages/ui/src/machines/authMachine.ts
```diff
@@ -17,7 +17,9 @@ export const authMachine = createMachine<AuthContext, AuthEvent>(
     initial: 'idle',
     context: {
       user: undefined,
-      config: undefined,
+      config: {
+        login_mechanisms: ['username'],
```
ericclemmons, line 21: Defaulted in the state machine so that all view code can pass along their `loginMechanisms`, even if undefined.

**Note – I'm only keeping `config.login_mechanisms` because of how many references there are to it!  Ideally, all `context.*` keys are _flat_, so we don't have to worry about deep-merging arrays**.

packages/ui/src/machines/authMachine.ts
```diff
@@ -36,7 +38,7 @@ export const authMachine = createMachine<AuthContext, AuthEvent>(
           {
             src: 'getAmplifyConfig',
             onDone: {
-              actions: 'setAuthConfig',
+              actions: 'applyAmplifyConfig',
```
ericclemmons, line 41: Renaming from `set` because this isn't a dumb setter.

packages/ui/src/machines/authMachine.ts
```diff
@@ -110,9 +112,33 @@ export const authMachine = createMachine<AuthContext, AuthEvent>(
...
+                default:
+                  console.warn(`Unknown login mechanism: ${login}`);
+              }
+            }
+          );
+
+          return {
+            loginMechanisms,
+            // Override config defaults with explicit values provided to state machine
+            ...context.config,
```
ericclemmons, line 140: ^ Note this part. This may not work as expected until we test zero-config stuff.  For example, is `USERNAME` always present? Should defaulting `['username']` in the state machine happen _here_ rather than on init?

Should `context.config.login_mechanisms = getDefaultLoginMechanisms(Amplify.configure())` at run-time?

I don't know yet until the feature fully lands and we basically _remove_ all `loginMechanism` usage from our examples and re-test.

packages/e2e/features/ui/components/authenticator/sign-in-with-username.feature
wlee221: ✨ ✨

wlee221: You know, we can probably mock even more things like `Auth.setupTOTP` responses, we can render QR code, and actually test if the image source contains the right source. That's just too exciting.

packages/ui/src/machines/authMachine.ts
wlee221: imo, i think what you have is fine as-is for now, until we get the feature at hand. And once it lands, my understanding is that I expect all aws-exports to have the zero-config values if customer ran `amplify pull`, so I don't see much need for `getDefaultLoginMechanisms` on runtime.

wlee221 (approved):
LGTM, added some comments 👍

packages/ui/src/machines/authMachine.ts
wlee221: Btw, with the above implementation `loginMechanisms` will always be overwritten by `...context.config` because it's not deep merging. Is this is intended so that zero config doesn't take effect for now (and we'll flatten config object as suggested in above comment)?

ericclemmons:
Test failures happened due to Cognito, _again_:

> Screen Shot 2021-09-23 at 5 34 57 PM

Tomorrow I'll get a fresh sample of a sign up so I can mock it.

packages/ui/src/machines/authMachine.ts
```diff
@@ -110,9 +112,33 @@ export const authMachine = createMachine<AuthContext, AuthEvent>(
...
-        config(_, event) {
-          return event.data.auth;
+      applyAmplifyConfig: assign({
+        config(context, event) {
+          const loginMechanisms = event.data.aws_cognito_login_mechanisms?.map(
+            (login) => {
+              switch (login) {
+                case 'PREFERRED_USERNAME':
+                case 'USERNAME':
+                  return 'username';
```
ericclemmons, line 122: @lazpavel So this should just be:

```suggestion
                case 'PREFERRED_USERNAME':
                  return 'username';
```

?

lazpavel: yes

packages/ui/src/machines/authMachine.ts
ericclemmons: @wlee221 You're 100% correct.  The only thing we'll need to do is map the UPPERCASE values to our lowercase input equivalents.

But this logic isn't actually correct (yet). Fixing this in 08bfd89bde23986f648518c1918b7905024302d3 so that zero-config will "just work" 🙏

ericclemmons:
@wlee221 I could use your help resolving Angular's tests: https://github.com/aws-amplify/amplify-ui/runs/3701630782?check_suite_focus=true

wlee221:
Angular test looks good now! The only meaningful code change I made is https://github.com/aws-amplify/amplify-ui/pull/425/commits/bc705f1af1e3084c069122a1dd5aa4af0cfc27c7.

ErikCH (approved):
Added Vue fixes, hopefully should build and pass e2e tests now.

ericclemmons:
@ErikCH Everything passed and the Vue examples look _clean_ ✨  THANK YOU!!

PR #425, opened 2021-09-24
Angular loginMechanisms @Input prop

wlee221:
*Issue #, if available:*

*Description of changes:*

Adds `loginMechanism` `@Input` prop to Angular authenticator and updates the examples.

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/services/state-machine.service.ts
+++ b/packages/angular/projects/ui-angular/src/lib/services/state-machine.service.ts
@@ -6,6 +6,7 @@ import {
   authMachine,
   AuthMachineState,
   getSendEventAliases,
+  LoginMechanism,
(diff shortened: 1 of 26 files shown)


packages/angular/projects/ui-angular/src/lib/services/state-machine.service.ts
```diff
@@ -22,15 +23,27 @@ export class StateMachineService {
...
+    })
       .onTransition((state) => {
         this._user = state.context.user;
         this._authState = state;
       })
       .start();
-    this._services = getSendEventAliases(interpreter.send);
-    this._authService = interpreter;
+
+    console.log(loginMechanisms, machine, authService);
```
wlee221, line 44: Not again :( 
```suggestion
```

PR #430, opened 2021-09-27
fixing flash of content on auto sign in

eddiekeller:
*Issue #, if available:*

https://app.asana.com/0/1201058552570744/1200975648174197/f

*Description of changes:*

Fixing the flash of content when automatically signing in after a successful `confirmSignUp`

### DIFF
--- a/packages/ui/src/machines/actors/auth/signUp.ts
+++ b/packages/ui/src/machines/actors/auth/signUp.ts
@@ -143,7 +143,7 @@ export const signUpActor = createMachine<SignUpContext, AuthEvent>(
           const { username, password } = context.authAttributes;
 
           return {
-            user: event.data.user,
+            user: get(event, 'data.user') || context.user,
             authAttributes: { username, password },
           };
         },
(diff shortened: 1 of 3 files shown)


packages/ui/src/machines/actors/auth/signUp.ts
```diff
@@ -175,7 +174,12 @@ export const signUpActor = createMachine<SignUpContext, AuthEvent>(
         const username =
           get(user, 'username') || get(authAttributes, 'username');
 
-        return Auth.confirmSignUp(username, code);
+        const { password } = authAttributes;
+
+        await Auth.confirmSignUp(username, code);
+        const result = await Auth.signIn(username, password);
+
+        return result;
```
ericclemmons, line 182: What do you think about this proposal?

- `Auth.confirmSignUp`
  - If it fails, the state machine should know.
  - It's expected to succeed and proceed...
  - `Auth.signIn`
    - If it fails, `Auth.confirmSignUp` still succeeded, so redirect and let the default flow take over.
    - If it succeeds, we assign `user`, redirect, but this time the `user` in in-memory and the App should show.

```suggestion
        await Auth.confirmSignUp(username, code);

		try {
          const { password } = authAttributes;
		  return await Auth.signIn(username, password);
		} catch (e) {
		  console.warn(e);
		}
```

ericclemmons (approved):
This looks good!  I have on small suggestion on which errors we want to bubble up to the machine vs. silently ignore.

Otherwise, no blockers, only a suggestion.

packages/ui/src/machines/actors/auth/signUp.ts
eddiekeller: This is a good idea, but I will also need to modify the `data` returned by the `resolved` action in the `signUp` machine too. So I won't commit this suggestion, but I will take it and add in the additional change, then send up a revision.

ericclemmons: I thought the two were equivalent in returning the same `result`, but wrapping a `try/catch` around `signIn` since that's an optional, optimistic call?

Either way, I trust your judgement on this over mine 🙏

PR #433, opened 2021-09-28
Revert "build(deps): bump mixme from 0.5.1 to 0.5.2"

zchenwei:
Reverts aws-amplify/amplify-ui#403


wlee221:
Why is this being reverted?

zchenwei:
> Why is this being reverted?

Oops, I think I might open it by mistake. Close it

PR #450, opened 2021-09-29
Fix SSR rendering of Fields and Tabs w/ IdProvider

reesscot:
*Description of changes:*
This PR is a proposal for fixing a Tabs and Fields SSR Id mismatch issue. This is an issue where the Id generated on the server side doesn't match the one generated on the client side, causing SSR hydration issues. 

Error message for tabs:
image

This can also occur with the Field components.

*Proposed fix for Tabs*:
This PR adds the IdProvider to the AmplifyProvider so that the Tabs primitive generates the same Id on server side and client side. See: https://www.radix-ui.com/docs/primitives/overview/server-side-rendering

*Proposed fix for Fields (TextField, SelectField, etc)*: 
This PR updates the id generation used by Fields to use the `AmplifyProvider` with `IdProvider` instead of nanoid. Docs are also updated to indicate the requirement for  using `AmplifyProvider`.

Screen Shot 2021-09-29 at 5 00 07 PM

One potential drawback to this approach is that it generates an id with `radix-id-` prefix, which could be confusing to customers. @dbanksdesign 

TODO:

- [x] Add docs requiring `AmplifyProvider` for Tabs and Field components
- [x] Fix unit tests 


packages/react/src/primitives/shared/utils.ts
```diff
@@ -27,18 +22,11 @@ export const getConsecutiveIntArray = (start: number, end: number) => {
...
  */
-export const useAmplifyFieldID = (id?: string) => {
-  return React.useMemo(() => {
-    if (id) {
-      return id;
-    }
-    return `amplify-field-${nanoid()}`;
-  }, [id]);
-};
+export const useAmplifyFieldID = (id?: string) => useId(id);
```
zchenwei, line 29: nit: I was wondering if this function only specific to field. Should we make it more generic like `useAmplifyID` so that it could be used to generated random id for other primitives as well?

packages/react/package.json
```diff
@@ -33,6 +33,7 @@
   "dependencies": {
     "@aws-amplify/ui": "3.0.1-next.3",
     "@aws-amplify/ui-react-v1": "npm:@aws-amplify/ui-react@1.2.9",
+    "@radix-ui/react-id": "^0.1.0",
     "@radix-ui/react-tabs": "0.0.16",
     "@xstate/react": "^1.4.0",
     "autoprefixer": "^10.3.1",
```
zchenwei, line 39: `yarn remove nanoid`?

zchenwei (commented):
This is looking good. Just a few questions

zchenwei (approved):

packages/react/src/primitives/shared/utils.ts
reesscot: That's a good idea, I'll update.

packages/react/package.json
reesscot: Yes, good catch.

ericclemmons (commented):
Can you validate a couple things before we ✅  this?

1. Run https://github.com/aws-amplify/amplify-ui/tree/main/examples/next (like http://localhost:3000/ui/components/authenticator/sign-in-with-username) and confirm the error goes away. (There's likely some changes that need to be made)
2. Update any of the docs (e.g. https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/pages/ui/getting-started/installation/dependencies.react.mdx) since it sounds like our primitives **require** `<AmplifyProvider>` being added at the top-level.

reesscot:
@ericclemmons Error message is gone
Screen Shot 2021-09-30 at 10 00 08 AM

Will update docs now.

docs/src/components/Example.tsx
```diff
@@ -1,13 +1,20 @@
...
+  children: React.ReactNode;
+  className?: string;
+}) {
   return (
     <View
       backgroundColor="rgba(249, 250, 251, 1)"
       borderRadius="0.25rem"
       boxShadow="inset 0 2px 4px 0 rgba(0, 0, 0, 0.06)"
       className={className}
-      marginBottom="1.5rem"
```
reesscot, line 10: Note: we don't support `marginButton` style prop at this time on the primitives.

docs/src/pages/ui/primitives/passwordfield/react.mdx
```diff
@@ -25,13 +25,15 @@ The PasswordField primitive allows users to input passwords. It also features fu
 
 ## Usage
 
+<Fragment>{() => import('./../shared/requiredAmplifyProvider.mdx')}</Fragment>
+
```
ericclemmons, line 29: What were your thoughts on my recommendation to document this once as part of the installation for React vs. on every component?
> https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/pages/ui/getting-started/installation/dependencies.react.mdx

Radix does the same – exclude the provider from primitive pages:
- https://www.radix-ui.com/docs/primitives/components/tabs
- https://www.radix-ui.com/docs/primitives/overview/server-side-rendering

reesscot: Yeah, after I pushed this I was thinking about that potentially being a better option. Will revisit this later today.

reesscot: @ericclemmons Updated.

docs/src/pages/ui/primitives/tabs/demo.tsx
```diff
@@ -48,7 +48,7 @@ export const TabsDemo = () => {
 export const ControlledTabExample = () => {
   const [index, setIndex] = React.useState(0);
   return (
-    <Tabs currentIndex={index} onChange={(i) => setIndex(i)}>
+    <Tabs currentIndex={index} onChange={(i) => setIndex(i as number)}>
```
ericclemmons, line 51: I believe `React.useState<number>(0)` will help the inference so that `setIndex` doesn't require casting 🤔

ericclemmons (approved):
Thanks for your diligence on this @reesscot & getting it resolved 💪

docs/src/pages/ui/primitives/tabs/demo.tsx
reesscot: I thought so too but TS disagreed. 🤷‍♂️

zchenwei: Typing the parameter may be better than casting?

`<Tabs currentIndex={index} onChange={(i: number) => setIndex(i)}>`

reesscot: @zchenwei Good call, updated.

docs/src/components/Example.tsx
```diff
@@ -1,13 +1,20 @@
 import { View } from '@aws-amplify/ui-react';
+import React from 'react';
```
hvergara, line 2: I understand that we're moving away from default imports, right? can we start with this one?
```suggestion
import * as React from 'react';
```

packages/react/src/primitives/shared/utils.ts
```diff
@@ -27,18 +22,14 @@ export const getConsecutiveIntArray = (start: number, end: number) => {
...
-    if (id) {
-      return id;
-    }
-    return `amplify-field-${nanoid()}`;
-  }, [id]);
-};
+export const useAmplifyID = (id?: string) => useId(id);
+
+// Prefixed autogenerated id created by useAmplifyID
+export const AUTO_GENERATED_ID_PREFIX = 'radix-id-';
```
hvergara, line 32: If I understand correctly, this variable is used to find elements IDs that matches the prefix generated by `useId`, right? 

https://github.com/radix-ui/primitives/blob/d3f1d5839888110939e61a0ccea6c6944205bb90/packages/react/id/src/id.tsx#L43

packages/react/src/primitives/shared/utils.ts
```diff
@@ -27,18 +22,14 @@ export const getConsecutiveIntArray = (start: number, end: number) => {
...
  */
-export const useAmplifyFieldID = (id?: string) => {
-  return React.useMemo(() => {
-    if (id) {
-      return id;
-    }
-    return `amplify-field-${nanoid()}`;
-  }, [id]);
-};
+export const useAmplifyID = (id?: string) => useId(id);
```
hvergara, line 29: suggestion: what `Amplify ID` means? would you be open to rename it to `useStableId` or `useProviderId` or something more meaningful?

packages/react/src/primitives/shared/utils.ts
reesscot: Yes, that's correct. I colocated it with `useAmplifyID` so that it would be clearer. Should I add an additional comment?

packages/react/src/primitives/shared/utils.ts
reesscot: I like `useStableId` a lot, will update.

hvergara (approved):
thanks for considering my comments. approved!

dbanksdesign (approved):
LGTM! I think a future improvement could be just implementing the IDProvider ourselves so we don't have the 'radix-id' in our IDs, but we can do that later it is not a huge deal.

ericclemmons:
Yeah, I consider auto-generated IDs to be a bit of a smell. Ideally, these tab panels & names would have explicit `key`s to them that would be consistent between the server & client.

I don't know why Radix choose to go with auto-generation here specifically.

PR #453, opened 2021-09-30
Update Sign Up examples to use initialState="signUp"

ericclemmons:
*Description of changes:*

- [x] Instead of `authMachine`, use `createAuthenticatorMachine({ initialState?, loginMechanisms? })`
- [x] Fix `SignInSignUpTabs.tsx` to be controlled
- [x] Update Angular
- [x] Update Vue 
- [x] Update `sign-up-*` tests to use `initialState` instead of clicking the tab


ericclemmons:
@wlee221 @ErikCH I could use your help figuring out the types for these changes. It _looked_ right, but clearly not :D

wlee221:
Ah the input binding should be 

`[initialState]="'signUp'"`, or just `initialState="signUp"`.

Putting brackets around the `input` signals that the value would be a javascript value, not a string html literal. And the surrounding double quote `" value "` wraps that javascript value (think of it as React `{ myVariable }`).

packages/vue/src/components/authenticator.vue
```diff
@@ -300,8 +298,11 @@ onBeforeMount(() => {
 
 const attrs = useAttrs();
 
-const { loginMechanisms } = withDefaults(
-  defineProps<{ loginMechanisms?: LoginMechanism[] }>(),
+const { initialState, loginMechanisms } = withDefaults(
+  defineProps<{
+    initialState: AuthenticatorMachineOptions['initialState'];
+    loginMechanisms: AuthenticatorMachineOptions['loginMechanisms'];
```
ErikCH, line 304: These should be optional

```suggestion
    initialState?: AuthenticatorMachineOptions['initialState'];
    loginMechanisms?: AuthenticatorMachineOptions['loginMechanisms'];
```

ErikCH (approved):
👍

PR #476, opened 2021-10-05
Starting converting theme to typescript

dbanksdesign:
*Issue #, if available:*

*Description of changes:* This is the first part of a multi-part update of our theming infrastructure. To make the changes more manageable, this is just converting the default theme from CommonJS modules to Typescript and using explicit types rather than relying on a generated file or `typeof`. I tried to minimize other changes. After this lands the next steps are:
1. Update theming code in the react package to properly handle references
2. Share the CSS generation logic between ui and react packages
3. Get theme package ready to be consumed by React Native


packages/react/src/theming/extendTheming.ts
```diff
@@ -1,15 +1,18 @@
 import merge from 'deepmerge';
+import { AmplifyTheme, AmplifyThemeOverrides } from '@aws-amplify/ui';
 
-import { defaultTheme, defaultTokens } from './defaultTheming';
-import { DeepPartial, Theme, Tokens } from './types';
```
dbanksdesign, line 4: Our own DeepPartial helper wasn't working properly for me. I added one from `type-fest` in the ui package and had that export a DeepPartial of the Theme.

packages/ui/package.json
```diff
@@ -21,12 +21,13 @@
     "build": "yarn build:sd && yarn build:css && yarn build:js",
     "build:css": "./scripts/generateCSS.js",
     "build:js": "tsup --minify",
-    "build:sd": "style-dictionary build --config ./sd.config.js",
+    "build:sd": "ts-node --transpile-only sd.config.ts",
```
dbanksdesign, line 24: Running Style Dictionary in Typescript

cool

packages/ui/sd.config.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Style Dictionary config
+ */
+
+import StyleDictionary from 'style-dictionary';
+import { theme, CSS_VARIABLE_PREFIX } from './src/theme';
+
+const CSS_VARIABLE_SCOPE = ':root, .amplify-theme';
+
+StyleDictionary.extend({
+  tokens: theme,
```
dbanksdesign, line 11: Using the typescript files directly imported rather than with file paths.

packages/ui/src/theme/tokens/borderWidths.ts
```diff
@@ -0,0 +1,13 @@
+import { DesignToken } from './types/designToken';
+
+export interface BorderWidths {
```
dbanksdesign, line 3: Originally I had all the types in the types folder, but felt weird to have them so far from being used. Is this ok or should they go into a types folder?

packages/ui/src/theme/tokens/colors.ts
```diff
@@ -1,4 +1,57 @@
...
+  white: DesignToken;
+  black: DesignToken;
+
+  // Semantic colors
+  font: FontColors;
+  background: BackgroundColors;
+  border: BorderColors;
+
+  // escape hatch to allow customers to add arbitrary colors
+  [key: string]: any;
```
dbanksdesign, line 51: Should we remove this?

packages/ui/tsconfig.json
```diff
@@ -1,4 +1,9 @@
 {
+  "ts-node": {
+    "compilerOptions": {
+      "module": "CommonJS"
```
dbanksdesign, line 4: This is because Style Dictionary is a CommonJS module

dbanksdesign (commented):
welcome aboard

TypeScript Theme!

examples/next/pages/examples/listings/ExperienceCard.tsx
```diff
@@ -4,7 +4,7 @@ export const ExperienceCard = ({ img, title }) => (
   <Card
     width="20rem"
     padding="0"
-    borderRadius={theme.radii.large}
+    borderRadius={theme.radii.large.value}
```
dbanksdesign, line 7: Note: temporarily adding `.value` to get the example to work. In the next PR I will address this and the reference issue so users won't have to add `.value`.

examples/next/pages/examples/listings/ExperienceCard.tsx
```diff
@@ -4,7 +4,7 @@ export const ExperienceCard = ({ img, title }) => (
   <Card
     width="20rem"
     padding="0"
-    borderRadius={theme.radii.large}
+    borderRadius={`${theme.radii.large.value}`}
```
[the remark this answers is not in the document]
dbanksdesign: This was actually a Typescript thing because `theme.radii.large` is of type `DesignToken` which has a `value` that can be a string or number. So I think this could be changed to `theme.radii.large.value as string`? Also one of my todos was to add more design token types like `RadiusDesignToken` which could be more explicit that its value is a string.

packages/ui/package.json
```diff
@@ -21,12 +21,13 @@
...
     "dev": "tsup --watch src",
     "test": "echo \"Skipped: no test specified\""
   },
   "dependencies": {
     "lodash": "^4.17.21",
+    "type-fest": "^2.3.4",
```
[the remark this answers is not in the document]
dbanksdesign: Yup that is my bad

packages/react/src/primitives/shared/responsive/__tests__/getMediaQueries.test.ts
```diff
@@ -2,12 +2,12 @@ import { Breakpoints } from '../../../types/responsive';
 import { getMediaQueries } from '../getMediaQueries';
 
 const unSortedBreakpoints: Breakpoints = {
-  xl: 80,
-  small: 30,
-  base: 0,
-  large: 62,
-  medium: 48,
-  xxl: 96,
+  xl: { value: 80 },
```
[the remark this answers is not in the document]
dbanksdesign: Correct. Previously we were using Style Dictionary to generate a typescript object that had no `.value` but instead just had the value itself, take a look at `packages/ui/src/theme/theme.ts`after you run a build on the ui package. Here is a truncated part of that file:

```typescript
export default {
  "components": {
    "alert": {
      "backgroundColor": "hsl(210, 5%, 90%)",
      "padding": "0.75rem 1rem",
```

As opposed to directly consuming the source typescript and not using Style Dictionary to generate it:

```typescript
export const alert = {
  //...
}
```

hvergara (approved):
Looks good to me! I guess there will be some follow-up changes but this is a great start

packages/react/src/theming/defaultTheming.ts
```diff
@@ -1,6 +1,6 @@
-import { theme, tokens } from '@aws-amplify/ui';
+import { theme as defaultTheme } from '@aws-amplify/ui';
 import { getCSSVariablesFromTokens } from './utils';
 
-const defaultCSSVariables = getCSSVariablesFromTokens(tokens);
```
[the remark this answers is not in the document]
dbanksdesign: Changing to `.value` doesn't solve that just yet... That would be part 2. But now thinking about it, without part 2 this breaks the theme Provider even more so I might need to add those changes... But yes, right now actually our default theme typescript object contains no references because they got resolved by style dictionary at build-time. So in the example above if you overrode `colors.background.tertiary` it would not affect the alert component because at that time that token is already resolved to a specific color. You could still use `({ theme }) => theme.colors.background.tertiary` and that would work as expected, but it just wouldn't affect the style of primitives.

packages/react/__tests__/exports.ts
```diff
@@ -10,6 +10,7 @@ describe('@aws-amplify/ui-react', () => {
         Array [
           "Alert",
           "AmplifyProvider",
+          "AmplifyThemeOverrides",
```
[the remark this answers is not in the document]
dbanksdesign: I think you are right,  I don't think we need to export this type if it is part of `AmplifyProviderProps`

PR #477, opened 2021-10-05
ButtonGroup Primitive

zchenwei:
*Description of changes:*

This PR will be adding `ButtonGroup` primitive. A `ButtonGroup` can be use to group buttons with the same `size` or `variation`. A `ButtonGroup` is also a flex container, so any flex props can apply to it for layout purpose.

Screen Shot 2021-10-05 at 3 26 59 PM

Screen Shot 2021-10-05 at 3 27 21 PM

### DIFF
--- a/packages/react/src/primitives/types/buttonGroup.ts
+++ b/packages/react/src/primitives/types/buttonGroup.ts
@@ -1,0 +1,9 @@
+import { AriaProps } from './base';
+import { BaseStyleProps } from './style';
+import { ButtonProps } from './button';
+import { FlexProps } from './flex';
+export interface ButtonGroupProps
+  extends AriaProps,
+    BaseStyleProps,
+    FlexProps,
+    Pick<ButtonProps, 'size' | 'variation'> {}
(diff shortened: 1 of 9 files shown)


packages/react/src/primitives/types/buttonGroup.ts
```diff
@@ -0,0 +1,14 @@
...
+import { BaseStyleProps } from './style';
+import { ButtonProps } from './button';
+import { FlexProps } from './flex';
+
+export interface ButtonGroupProps
+  extends AriaProps,
+    BaseStyleProps,
+    FlexProps,
+    Pick<ButtonProps, 'size' | 'variation'> {
+  children: React.ReactNode;
```
hvergara, line 13: is this 100% necessary? Because `React.FC<X>` should add this property.

zchenwei: Me and Scott discussed this before. This aims to make the `children` required, which is what we expected, to prevent our users using it like `<ButtonGroup />`. React.FC<X> doesn't force the `children` to be present and thus users won't get any syntax error if they try to do so.

Screen Shot 2021-10-06 at 12 05 24 PM

versus

Screen Shot 2021-10-06 at 12 07 09 PM

zchenwei: So I think it makes sense to make it required for group related primitive. Same applies to `ToggleButton`

hvergara: awesome! thanks. I don't have more comments, so I'm approving it.

hvergara (approved):

PR #483, opened 2021-10-07
Run patch-package *before* build

ericclemmons:
This reverts commit b1e43472f00cc1f3b9c89f945174d88909e613fa.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "private": true,
   "scripts": {
-    "install": "patch-package",
-    "postinstall": "yarn build",
+    "install": "yarn build",
+    "postinstall": "patch-package",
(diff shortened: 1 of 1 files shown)


ericclemmons:
What's odd is that the build is being skipped:
https://github.com/aws-amplify/amplify-ui/runs/3830354495#step:5:155

But _why_? Yes, the `unit` tests also do a build, but that's not happening in the `e2e` tests explicitly.  So I'm not sure what the cached entry is for Ultra that's say "No changes, no need to build this lib!"

On a related note, though, **all of our tests should be using a published version in Verdaccio (or a PR preview!)** rather than relying on a build each time.

That way all examples could do `yarn install --prod`, `yarn build && yarn start` while using the _real_ packages.

Issue #508, opened 2021-10-13
Vue + React: Sign out does not work


ErikCH:
After signing in, the sign out button does not work in the examples.  You can't press the sign out button

ErikCH:
Fixed with last PR

PR #512, opened 2021-10-13
Fixed sign out for Vue and Angular

ErikCH:
*Issue #, if available:* Fixes issue with signing out #508

*Description of changes:*
Sign out button is obscured by Authenticator modal. This will make it so the modal will not show when user is authenticated. 


ericclemmons:
Can you add tests to our features that clicks the Sign Out button?

We have it in one of them already, but need to prevent against regressions.

Issue #509, opened 2021-10-13
Authenticator Modal Takes Up Whole Page And Covers Other Elements


ErikCH:
After adding in the authenticator to a Vue, React or Angular project, the `div` with the `data-amplify-modal` covers the whole page, obscuring any other HTML elements on the page (Other then the authenticator)

ericclemmons:
@ErikCH Do you know off-hand what the default experience is for https://docs.amplify.aws/ui/auth/authenticator/q/framework/vue/?

I remember when we launched last year, one of the first issues was that the form wasn't centered on the screen & we had to wrap it in a container.

ErikCH:
The legacy Vue UI authenticator allows you to enter data above or below it. It is centered horizontally and vertically by using flex with a min-height. 

I like the `variation` approach with adding in `modal`. I can add that to vue perhaps?

ericclemmons:
This will help out with the docs, too.

- [ ] Docs will just use an `Authenticator`.
- [ ] Remove `.example [data-authenticator]` overrides for docs.
- [ ] Centering within the given space for the Authenticator is desirable (e.g. `place-items: center` or `place-self: center` I think), but having `min-height` isn't IMO.
- [ ] Docs will need `Configuration > variant` added.
- [ ] `<Authenticator variant="modal">` will offer the modal experience. 
- [ ] Styles should target `[data-amplify-authenticator] [data-variant=modal]`
    
    I don't know if we need to change the markup vs. leaving the default unstyled?

PR #516, opened 2021-10-14
Vue + Angular: Add new Variation Modal Prop

ErikCH:
*Issue #, if available:* Per conversation in #509 added a new `variation` prop for Angular and Vue that will only add the modal if this prop is added with the string `modal`.

*Description of changes:*
Vue and Angular users can choose if they would like to setup their authenticator as a modal. 

### DIFF
--- a/docs/src/styles/styles.css
+++ b/docs/src/styles/styles.css
@@ -61,7 +61,7 @@ ul {
   padding: 1rem;
 }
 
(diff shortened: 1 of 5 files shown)


ErikCH:
> Just got one naming nit, but it looks good! Thanks for working cross-platform 💪

Great! I'll make a quick change to the naming, and push

PR #532, opened 2021-10-19
Convert useAuthenticator hook & services to public APIs

ericclemmons:
*Description of changes:*

This PR paves the way for custom components & public APIs using React's `useAuthenticator` hook: https://github.com/aws-amplify/amplify-ui/pull/552

- [x] `<AuthenticatorProvider {...initialState, components, loginMechanisms, services } />` 
- [x] `useAuthenticator` **no longer returns `[state, send]`** and instead returns only public APIs.
- [x] Replace `state.matches(...)` with `route` (`signIn` | `verifyUser` | ...). This doesn't include pending states, which will be indicated via an `isPending` facade.
- [x] Still providing `_send` and `_state` but `/** @deprecated */` so we can track & remove these private values.
- [x] Add `isPending` facade.  **We want to avoid any usage of `state.matches(...)` in components!**
- [x] Rename `signIn` and `signUp` facades to `transitionToSignIn` or `toSignIn` or `gotoSignIn` or something for @wlee221 
- [x] Get React tests passing with the current refactor
- [x] Get all tests working

### Future Changes

- [ ] https://github.com/aws-amplify/amplify-ui/pull/552
- [ ] https://github.com/aws-amplify/amplify-ui/issues/549
- [ ] Replace all usage of `state.matches` in components
- [ ] I'm not a fan of `submitForm` and `updateForm` vs. explicitly saying `updateSignUpForm` or `submitSignUpForm`.
- [ ] Update the docs example to _actually work_.
- [ ] Replace all usage of `send(...)` in components
- [ ] Because `loginMechanisms` is pretty much reverted (https://docs.amplify.aws/cli/migration/cli-auth-signup-changes/), `usernameAliases = [email | phone_number]` is consistent with:

	> https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPool.html#API_CreateUserPool_RequestSyntax

- [ ] `socialProviders = ['amazon', 'facebook', 'google']` would differentiate the two.

### DIFF
--- a/packages/react/__tests__/exports.ts
+++ b/packages/react/__tests__/exports.ts
@@ -9,7 +9,6 @@ describe('@aws-amplify/ui-react', () => {
       expect(Object.keys(exported)).toMatchInlineSnapshot(`
         Array [
           "Alert",
-          "AmplifyContext",
           "AmplifyProvider",
           "Authenticator",
           "Badge",
--- a/packages/react/src/components/AmplifyProvider/index.tsx
+++ b/packages/react/src/components/AmplifyProvider/index.tsx
@@ -39,5 +39,3 @@ export function AmplifyProvider({
     </AmplifyContext.Provider>
   );
 }
-
-export { AmplifyContext };


ericclemmons:
I'm going to get tests passing on this refactor and _then_ do a PR documenting & supporting this use-case.

That way, that PR can be done to support multiple frameworks.

PR #580, opened 2021-10-27
Use safe imports of 'aws-amplify' library

johnpc:
This is a copy of the https://github.com/aws-amplify/amplify-ui/pull/577, which made against my personal fork and didn't have e2e permissions.

-------

When you use an import like `import {I18n} from 'aws-amplify'`, it includes all the code from 'aws-amplify' which can result in errors like:

```
Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Storage is disabled inside 'data:' URLs.
    at <anonymous>:292337:28
    at LibraryFactory (<anonymous>:294621:4)
    at ExportLibrary (<anonymous>:292323:20)
    at Object.<anonymous> (<anonymous>:292325:3)
    at Object../node_modules/paho-mqtt/paho-mqtt.js (<anonymous>:294625:30)
    at __webpack_require__ (<anonymous>:20:30)
    at Module../node_modules/@aws-amplify/pubsub/lib-esm/Providers/MqttOverWSProvider.js (<anonymous>:28992:67)
    at __webpack_require__ (<anonymous>:20:30)
```

when called from a context such as a `data:` protocol that does not support `localStorage` apis.

The heart of the issue is that AppSync code in AmplifyJS depends on library called Paho (https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/Providers/MqttOverWSProvider.ts#L13) which, when imported, uses the `localStorage` api.

I want to import `withAuthenticator` using `import {withAuthenticator} from '@aws-amplify/ui-react';`, but this currently fails because the Authenticator component pulls in I18n unsafely:

https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/components/Authenticator/Authenticator.tsx#L2

```
import { I18n } from 'aws-amplify';

```

My use case works when I selectively _don't_ import AppSync modules:

```
import { I18n } from '@aws-amplify/core';
```

This commit updates imports to avoid this issue.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ Whenever `yarn angular build` is run, such as part of postinstall script, you wa
    ```js
    // examples/next/pages/components/authenticator/sign-up-with-username.tsx
    import { Authenticator } from '@aws-amplify/ui-react';
-   import { Amplify } from 'aws-amplify';
+   import { Amplify } from '@aws-amplify/core';
    import awsExports from '../../../../../environments/auth-with-username-no-attributes/src/aws-exports';
 
    Amplify.configure(awsExports);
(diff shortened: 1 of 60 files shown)


reesscot:
closing in favor of https://github.com/aws-amplify/amplify-ui/pull/1114

PR #589, opened 2021-10-28
DataStore predicate creation utility

hvergara:
*Description of changes:*
This PR introduce an utility to create DataStore-compatible predicates from an object representation.

Turns:
```js
{
  and: [
    {
       field: "name",
       operator: "beginsWith",
       operand: "John"
     },
     {
       field: "age",
       operator: "gt",
       operand: "25"
     }
  ]
}
```

Into:
```ts
p => p.and(
   p => p.name("contains", "John").age("gt", "25")
)
```


Milan-Shah (approved):

ericclemmons (commented):
The utility looks good!  Should this live in https://github.com/aws-amplify/amplify-js/tree/main/packages/datastore as an export of `@aws-amplify/datastore`?

I'm not seeing where our UI components leverage this?

ericclemmons (approved):
Approving because my questions are non-blocking 🙏

alharris-at (commented):
Logic looks sane to me. Only comment is around exposing types up.

packages/react/src/primitives/shared/datastore.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Given an array of predicates, compose them in sequential order
+ */
+const mergePredicates = <Model extends PersistentModel>(
+  predicates: ProducerModelPredicate<Model>[]
+): ProducerModelPredicate<Model> =>
+  predicates.reduce(
+    (previous, current) => (predicate) => current(previous(predicate)),
+    (predicate) => predicate
+  );
```
slaymance, line 25: This function is a bit confusing to me. Does it return a function that accepts a value and returns a boolean indicating whether that value passes all composed predicates?

hvergara: from a list of predicates, will return a composed version:

`f(x), g(x), h(x) => f(g(h(x)))`

packages/react/src/primitives/shared/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './constants';
+export { createDataStorePredicate } from './datastore';
```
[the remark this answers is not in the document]
hvergara: sure! good idea, will add in next commit

packages/react/src/primitives/shared/datastore.ts
slaymance: I see, I guess I just haven't seen predicates composed like this. Ultimately, a predicate will convert whatever type of value it's given into a `boolean` so, in your example above, `f` and `g` would be accepting a `boolean` as their argument instead of the original type of the parameter `x`.

PR #593, opened 2021-10-28
SliderField Primitive

zchenwei:
*Description of changes:*

This PR will be adding a `SliderField` primitive.


packages/react/src/primitives/SliderField/SliderField.tsx
```diff
@@ -0,0 +1,135 @@
...
+          >
+            <Range
+              className={ComponentClassNames.SliderFieldRange}
+              data-testid={SLIDER_RANGE}
+              style={{ backgroundColor: filledTrackColor }}
+            />
+          </Track>
+          <Thumb
+            aria-describedby={fieldId}
+            aria-valuetext={ariaValuetext}
```
zchenwei, line 126: Sometimes, a slider is used to choose a value that is not, semantically, a number. <- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role

docs/src/pages/ui/primitives/sliderField/react.mdx
```diff
@@ -0,0 +1,366 @@
...
+```
+
+<Example>
+  <SliderField
+    label="Slider"
+    defaultValue={50}
+    min={0}
+    max={100}
+    step={1}
+    errorMessage="The is an error message."
```
slaymance, line 245:
```suggestion
    errorMessage="This is an error message."
```

slaymance (approved):
Great work here! Left a few non-blocking comments for your consideration.

packages/react/src/primitives/SliderField/SliderField.tsx
reesscot: TIL, thanks for sharing!

packages/react/src/primitives/types/sliderField.ts
```diff
@@ -0,0 +1,21 @@
...
+  step?: number;
+  orientation?: SliderOrientation;
+  isValueHidden?: boolean;
+  trackWidth?: string;
+  emptyTrackColor?: Property.Color;
+  filledTrackColor?: Property.Color;
+  thumbColor?: Property.Color;
+  value?: number;
+  defaultValue?: number;
+  onChange?: (value: number) => void;
```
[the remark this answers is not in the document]
reesscot: +1, it looks like radix type supports an array:
```
onValueChange?(value: number[]): void
```

packages/react/src/primitives/types/fieldGroup.ts
```diff
@@ -1,7 +1,10 @@
 import { FlexProps } from './flex';
 
+type FieldGroupOrientation = 'horizontal' | 'vertical';
```
reesscot, line 3: Types used in other exported types must be exported

```suggestion
export type FieldGroupOrientation = 'horizontal' | 'vertical';
```

packages/ui/src/theme/css/component/fieldGroup.scss
```diff
@@ -94,6 +104,10 @@
 .amplify-field-group__field-wrapper {
   position: relative;
   width: 100%;
+
+  &[data-orientation='vertical'] {
+    width: fit-content;
```
reesscot, line 109: Can you confirm this is working across browsers? I don't think we are auto-prefixing our scss, but I may be wrong. Looks like FF requires prefix like `-moz-fit-content` https://caniuse.com/mdn-css_properties_width_fit-content

packages/ui/src/theme/css/component/sliderField.scss
```diff
@@ -0,0 +1,83 @@
...
+  &[data-orientation='vertical'] {
+    flex-direction: column;
+    width: var(--amplify-components-sliderfield-root-horizontal-width);
+  }
+}
+
+.amplify-sliderfield__track {
+  position: relative;
+  flex-grow: 1;
+  border-radius: 9999px;
```
reesscot, line 34: Should this value come from the token files?

packages/react/src/primitives/SliderField/SliderField.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { splitPrimitiveProps } from '../shared/styleUtils';
+import { isFunction, useStableId } from '../shared/utils';
+
+export const SLIDER_LABEL = 'slider-label';
+export const SLIDER_ROOT = 'slider-root';
+export const SLIDER_TRACK = 'slider-track';
+export const SLIDER_RANGE = 'slider-range';
+
+export const SliderField: Primitive<SliderFieldProps, typeof Root> = ({
+  ariaValuetext,
```
reesscot, line 22: nit: alphabetize

reesscot (changes requested):
Looks great, just a few comments/requests

docs/src/pages/ui/primitives/sliderField/react.mdx
zchenwei: Good catch

docs/src/pages/ui/primitives/sliderField/react.mdx
```diff
@@ -0,0 +1,366 @@
...
+  return (
+    <SliderField
+      ariaValuetext={dayNames[value]}
+      label="Day slider"
+      min={0}
+      max={6}
+      step={1}
+      value={value}
+      onChange={setValue}
+      labelHidden
```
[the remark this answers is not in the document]
zchenwei: I think the `Accessibility / Label behavior` section aims for that and I have an example for showing the label there

packages/ui/src/theme/css/component/fieldGroup.scss
zchenwei: We do auto-prefix our scss. <- https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/scripts/generateCSS.js#L9

packages/react/src/primitives/types/sliderField.ts
```diff
@@ -0,0 +1,21 @@
...
+
+type SliderOrientation = 'horizontal' | 'vertical';
+
+export interface SliderFieldProps extends TextInputFieldProps, ViewProps {
+  min?: number;
+  max?: number;
+  step?: number;
+  orientation?: SliderOrientation;
+  isValueHidden?: boolean;
+  trackWidth?: string;
```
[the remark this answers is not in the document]
zchenwei: I would update it to `trackSize`

packages/ui/src/theme/css/component/sliderField.scss
zchenwei: Well, I was thinking this was a special value. But it does make sense to make it as a token, since users may don't want the track to be rounded

packages/react/src/primitives/types/sliderField.ts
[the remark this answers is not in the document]
reesscot: We don't have a use case for multiple thumbs at this time, so let's it simple for now. In the future we could still avoid a breaking change, and make it backwards compatible by returning `number | number[]`.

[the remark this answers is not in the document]
zchenwei: Yea, that's what I am thinking. `number | number[]` makes more sense

reesscot (approved):
LGTM

Issue #594, opened 2021-10-29
Removing the "Create Account" tab


StephenK9:
When using I18n, where do we put the following code?

`    I18n.putVocabulariesForLanguage('en', {
        'Sign In': 'Login', // Tab header
        'Sign in': 'Log in', // Button label
        'Sign in to your account': 'Welcome Back!',
        Username: 'Enter your username', // Username label
        Password: 'Enter your password', // Password label
        'Forgot your password? ': 'Reset Password',
      });`
I've imported I18n and added the code above to my App.js file, but nothing changes when I do.

ErikCH:
Hi @StephenK9 !

Thanks for the feedback. I believe this is a bug, and I'm pushing up a fix write now for it! I'll let you know as soon as it's fixed. 

Did you have a question on removing the `Create Account` tab too?

StephenK9:
> Did you have a question on removing the `Create Account` tab too?

Yes, thanks! How do we explicitly remove that tab? I do not see any place in the docs that explicitly states how to remove that tab. The only image I see of the login screen without the 'create account' tab is the the one that falls under the i18N sections: https://ui.docs.amplify.aws/ui/components/authenticator?platform=react#labels--text

ErikCH:
Hi @StephenK9 !  We have pushed up the fix for custom translations, and update the documentation. Please try that.

As for removing tabs, the easiest way to do that is to add this to your style sheet.

```
.amplify-tabs {
  display: none;
}
```

Then make sure when you initialize the authenticator that you send over `initialState: 'signIn`.

```
export default withAuthenticator(App, {
  initialState: 'signIn',
});
```

Let me know if that works for you. We'll be adding that to the documentation soon.

Milan-Shah:
@StephenK9 Please let us know if you have any issue applying @ErikCH 's suggestion above! Marking the issue as resolved, but do let us know if you have any more issues with Authenticator.

PR #597, opened 2021-10-29
Custom translations are broken

ErikCH:
*Issue #, if available:*
Issue: #594 

*Description of changes:*
This will fix an issue where the `I18n.putVocabularies(translations)` is being ran after `I18n.putVocabulariesForLanguage()` overwriting it.

My solution is to run it during import instead of when the `authenticator` is created

### DIFF
--- a/packages/angular/projects/ui-angular/src/index.ts
+++ b/packages/angular/projects/ui-angular/src/index.ts
@@ -28,3 +28,11 @@ export * from './lib/directives/amplify-slot.directive';
  */
 export * from './lib/services/authenticator-context.service';
 export * from './lib/services/state-machine.service';
+
+/**
+ * Translations
+ */
+
+import { I18n } from 'aws-amplify';
+import { translations } from '@aws-amplify/ui';
+I18n.putVocabularies(translations);
--- a/packages/angular/projects/ui-angular/src/index.ts
+++ b/packages/angular/projects/ui-angular/src/index.ts
@@ -28,11 +28,3 @@ export * from './lib/directives/amplify-slot.directive';
  */
 export * from './lib/services/authenticator-context.service';
 export * from './lib/services/state-machine.service';
-
-/**
- * Translations
- */
-
-import { I18n } from 'aws-amplify';
-import { translations } from '@aws-amplify/ui';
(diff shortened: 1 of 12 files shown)


packages/angular/projects/ui-angular/src/index.ts
```diff
@@ -28,3 +28,11 @@ export * from './lib/directives/amplify-slot.directive';
...
 export * from './lib/services/authenticator-context.service';
 export * from './lib/services/state-machine.service';
+
+/**
+ * Translations
+ */
+
+import { I18n } from 'aws-amplify';
+import { translations } from '@aws-amplify/ui';
+I18n.putVocabularies(translations);
```
ErikCH, line 38: This gets run at import instead of when the Authenticator loads.

ericclemmons: We need to be careful about this because it means there are side effects and no longer tree-shakable. 

Angular had specific issues with this.

ErikCH: The other way is we tell the customer to run the `I8n.putVocabulatries(translations)` themselves What do you think @ericclemmons

ericclemmons: @ErikCH You know, I think you found the reason where our good intentions (automatically translated) create problems.

`I18n` is a naive utility that doesn't really have a way for us to say, "Here are some defaults, if you don't already have them".  Which is what we want.

**My recommendation for resolution is the following**:

- [ ] Remove `I18n.putVocabularies` from our implementations to fix this bug.
- [ ] Update https://ui.docs.amplify.aws/ui/components/authenticator?platform=react#internationalization-i18n with instructions on how to add opt-in to translations for customer apps.
- [ ] Update the `/ui/components/authenticator/i18n` examples to match the docs

This allows them to choose when the translations are added, avoiding race-conditions.

ErikCH: 👍

ErikCH: I pushed up the changes

wlee221 (approved):
This lgtm, thanks @ErikCH!

PR #598, opened 2021-11-01
feat(angular): implement amplify-slot component in Angular

wlee221:
*Issue #, if available:*

*Description of changes:*


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/amplify-authenticator/amplify-authenticator.component.html
```diff
@@ -54,99 +14,85 @@
...
       >
         <!-- signIn component -->
-        <ng-container
-          [ngTemplateOutlet]="customComponents.signIn || signIn"
-          [ngTemplateOutletContext]="context"
-          *ngIf="actorState?.matches('signIn')"
-        ></ng-container>
+        <amplify-slot name="sign-in" *ngIf="actorState?.matches('signIn')">
+          <amplify-sign-in></amplify-sign-in>
+        </amplify-slot>
```
wlee221, line 19: No more of these repeated `ngTemplateOutlet` logics!

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/amplify-authenticator/amplify-authenticator.component.html
```diff
@@ -1,43 +1,3 @@
-<!-- 
-  Define default contents here.
--->
-<ng-template #signIn>
-  <amplify-sign-in></amplify-sign-in>
-</ng-template>
```
wlee221, line 6: Default contents no longer need to be defined separately, now that they are explicitly defined where it's being placed.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/amplify-confirm-sign-in/amplify-confirm-sign-in.component.html
```diff
@@ -1,13 +1,4 @@
-<!-- 
-  Define default contents here
--->
-<ng-template #confirmSignInButton>
-  <button amplify-button variation="primary" fullWidth="true" type="submit">
-    {{ confirmText }}
-  </button>
-</ng-template>
```
wlee221, line 8: As we're closing GA, I wanted to be more careful about which slots we expose. As discussed, I'll incrementally open slots as we resolve customer use-cases. Currently, these slots are

1. `sign-up-form-fields` (in upcoming PR!)
2. top level`sign-in`, `sign-up`, ...

but we will be adding more slots like headers and footers as we go on.

packages/angular/projects/ui-angular/src/lib/utilities/amplify-slot/amplify-slot.component.ts
```diff
@@ -0,0 +1,34 @@
+import {
```
wlee221, line 1: This file (and its template above) is the actual meaningful code change here! This uses `CustomComponentsService` to get the custom components customers provided at the `amplify-authenticator` level.

packages/angular/projects/ui-angular/src/lib/services/custom-components.service.ts
```diff
@@ -16,12 +17,4 @@ export class AuthPropService {
...
     this._customComponents = { ...this._customComponents, ...customComponents };
   }
-
-  public get props(): PropContext {
-    return this._props;
-  }
-
-  public set props(props: PropContext) {
-    this._props = { ...this._props, ...props };
-  }
```
wlee221, line 26: This was previously used to share props between subcomponents, but isn't needed currently as all the props are being passed to `amplify-authenticator` anyway. Will add back in as needed.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/amplify-authenticator/amplify-authenticator.component.html
ErikCH: Nice!

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/amplify-confirm-sign-in/amplify-confirm-sign-in.component.html
ErikCH: Yeah, I was thinking about this in Vue, if I should remove all the slots I have, and only include a few. I think both approaches work. This is fine

packages/angular/projects/ui-angular/src/lib/services/custom-components.service.ts
ErikCH: Nice refactor

ErikCH (approved):
👍

wlee221:
Thanks @ErikCH for the review!

PR #601, opened 2021-11-02
useAuth React hook

hvergara:
*Description of changes:*
This PR introduces a new `useAuth` hook, containing metadata from the current Cognito authenticated user.

- [x] Implementation
- [x] Unit tests
- [x] Documentation


elorzafe:
@hvergara PR looks very clean!

I just have a few questions.
1. What can we expect if there is a state change, for example when the user sign-in or the user sign-out. Do we want to change notify that change on the hook?
2. Similar to the previous point, if the customer change the attributes on runtime, using `Auth.updateUserAttribute` or when pulling the latest information there was a change, what should be the expected behavior?
3. If the customer used `federatedSignIn` and passing the token, what should be the error message?

ericclemmons (commented):
Can you share some background on the requirements for this hook that I may be unaware of?

My experience has been that customers rely on several aspects of `Auth` beyond just the `user`, such as `session` & events.

Can we reconsider `useAuth` similar to https://github.com/aws-amplify/amplify-js-samples-staging/blob/master/samples/react/multi-category/photo-albums-e2e/src/useAuth.js?

A couple things that stand out:

- `Hub` events so `useAuth` is _live_ and components that rely on `user` will continue to get the latest, correct value.
- Same for `session`, which is used when sending credentials manually to other services.
- `Auth.currentAuthenticatedUser()` returns a `CognitoUser` object which has properties like:
  - `username`
  - `attributes`

  So is there a need to generate a new map?

These changes would make this hook more of an abstraction for what customers already have to manage in their apps, but without the boilerplate.

hvergara:
Thanks for your comments, Francisco!

>  What can we expect if there is a state change, for example when the user sign-in or the user sign-out. Do we want to change notify that change on the hook?

Added some logic to handle `signIn` & `signOut` Auth events

> Similar to the previous point, if the customer change the attributes on runtime, using `Auth.updateUserAttribute` or when pulling the latest information there was a change, what should be the expected behavior?

I looked into the official Auth implementation and there's no Auth Hub events triggered when attributes are updated. Do you have any idea on how to listen on attribute changes?

> 3. If the customer used `federatedSignIn` and passing the token, what should be the error message?

Let's sync to help me understand this use case

packages/react/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,56 @@
...
+        return handleUser(undefined);
+      default:
+        break;
+    }
+  };
+
+  const fetch = () => {
+    Auth.currentAuthenticatedUser()
+      .then(handleUser)
+      .catch(setError)
```
ericclemmons, line 38: I'm thinking about what happens in federated flows 🤔 

If this fails the first time, then `setError(error)` & `setLoading(false)` is called:
```js
{
  error: someError,
  isLoading: false,
  user: undefined,
}
```


But then if the auth event eventuaally succeeds and `Hub.listen('auth', handleAuth)` is called, then `handleUser(user)` & `setLoading(false)` is called:

```js
{
  error: someError, // Oh no!
  isLoading: false,
  user: someUser,
}
```

This seems to have the following behavior:

- `loading` is `true` on initial load, but not on multiple `handleAuth` invocations. (I assume this is desirable for an app to wait until we resolve/reject)
- `error`s aren't cleared when `setUser` is called.

You may consider using a single state object or reducer for `{ error, isLoading, user }` to solve for interdependent values:

```js
const [result, setResult] = useState<AuthResult>({
  error: undefined,
  isLoading: true,
  user: undefined,
});

...

Auth.currentAuthenticatedUser()
  .then(user => setResult({ isLoading: false, user })
  .catch(error => setResult({ error, isLoading: false })
```

hvergara: really like this approach, will change in next commit :)

hvergara: Updated. Thanks for your suggestions <3

ericclemmons (approved):
Looking good!  With `Hub`, I don't know if `fetch` is necessary, so if it absolutely isn't it may be in our best interest to have the smallest API initially and better understand the use-case that's either (1) already handled via `Hub` or (2) should happen automatically within the hook.

Issue #569, opened 2021-10-23
[Feedback] First impression on the new authenticator / UI.


artidata:
**Page**: [`/ui`](https://ui.docs.amplify.aws/ui)

**Feedback**: It has a pretty good design. However, the documentation has a lot to work on, No instruction on how to add more fields on the sign up form. I wish to be able to override the components from other UI frameworks. I will definitely migrate just for the good design.

ericclemmons:
Working on the fix @artidata , I'll update here once it's out 🙏

ericclemmons:
Alrighty, #606 fixes the bug, adds several tests, and we'll get this resolved for React, Angular, & Vue 🙏

artidata:
Great, the fixes work. Migrating sooner than expected. 👍

PR #606, opened 2021-11-02
Confirm Sign Up tests

ericclemmons:
*Issue #, if available:* Fixes https://github.com/aws-amplify/amplify-ui/issues/569

> Screen Shot 2021-11-01 at 2 52 25 PM

*Description of changes:*

- [x] Update `confirm-sign-up.feature` tests to run against `@angular @react @vue`
- [x] Introduce new `I mock {string} with fixture {string}` behavior
- [x] Expose `window.Amplify.*` for the new behavior 
- [x] Fix React tests (@ericclemmons)
- [x] Fix Vue tests (@ErikCH)
- [x] Fix Angualr tests (@wlee221)


ericclemmons (commented):
@ErikCH I've made a small change for Vue, but I'm noticing that Cypress is having issues typing into the `Confirmation Code` field.  Can you help debug these?
Screen Shot 2021-11-02 at 1 37 54 PM
Screen Shot 2021-11-02 at 1 38 06 PM
Screen Shot 2021-11-02 at 1 38 29 PM

ericclemmons (commented):
@wlee221 I'm seeing similar issues as Vue when I run the e2e tests – typing into the `Confirmation Code` field doesn't seem to work:
Screen Shot 2021-11-02 at 1 50 28 PM
Screen Shot 2021-11-02 at 2 02 23 PM

wlee221:
@ericclemmons Ah, I think I know what the problem is!

1. you should preferably use

```diff
- cy.findByLabelText('Confirmation Code').type('invalidcode');
+ cy.findInputField('Confirmation Code').type('invalidcode');
```
which searches for accessible textbox instead of the label.

2. In Vue and Angular, code field has `type="number"`, so `dom-testing-library` thinks it's a spinbutton instead of a textbox.

I think we should make `type` consistent first, then we can figure out the tests.

ericclemmons:
@wlee221 I attempted that change locally, but it didn't seem to resolve it.  I agree, the `type` should be consistent and as a `text`. (Though it's a number, the spin controls don't help because they're not incrementing like a quantity)

wlee221 (approved):

PR #612, opened 2021-11-02
Menu primitive

reesscot:
*Description of changes:*
This PR adds a `Menu` primitive. Menu provides an accessible, interactive menu for selecting actions within an application. The dropdown menu is collision-aware and will automatically change location based on available space.

`align` and `size`:

https://user-images.githubusercontent.com/6165315/139949939-9ca4beed-1e3b-4cb9-b733-e396c9bebc47.mp4

Keyboard accessibility:

### TODO

- [x] Add unit tests


docs/src/components/MenuPropControls.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <option value="end">end</option>
+      </SelectField>
+
+      <SelectField
+        label="size"
+        name="size"
+        placeholder="default"
+        onChange={(event) => setSize(event.target.value as MenuProps['size'])}
+      >
+        <option value="small">small</option>
```
dbanksdesign, line 38: Should we add a default option here so the user can get back to the default size if after they change it to small or large?

docs/src/pages/ui/primitives/menu/examples/controlledExample.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react';
```
dbanksdesign, line 1: I love splitting these examples up into their own files!

docs/src/pages/ui/primitives/menu/react.mdx
```diff
@@ -0,0 +1,224 @@
...
+
+## Demo
+
+<MenuDemo />
+
+## Usage
+
+Import the `Menu` and `MenuItem` components and styles. Use `MenuItem` component to configure menu options.
+
+```tsx file=./examples/basicExample.tsx
```
dbanksdesign, line 24: Ohhh so nice and clean and we can ensure the example code functions properly!

docs/src/pages/ui/primitives/menu/react.mdx
```diff
@@ -0,0 +1,224 @@
...
+
+```
+
+<Example>
+  <ControlledExample />
+</Example>
+
+## CSS Styling
+
+### Target classes
```
dbanksdesign, line 83: I love this! Is there any way we can automate this from the ComponentClassNames or something?

docs/src/pages/ui/primitives/menu/react.mdx
```diff
@@ -0,0 +1,224 @@
...
+<Menu>
+  <MenuItem>Download</MenuItem>
+  <MenuItem>Create a Copy</MenuItem>
+  <MenuItem>Mark as Draft</MenuItem>
+</Menu>;
+```
+
+<Example>
+  <Menu
+    className="global-styling-menu"
```
dbanksdesign, line 123: You could set the className on the `<Example>` and then it would function more like the example code

packages/ui/src/theme/__tests__/defaultTheme.test.ts
```diff
@@ -35,7 +35,7 @@ describe('@aws-amplify/ui', () => {
         --amplify-components-button-border-radius: var(--amplify-components-fieldcontrol-border-radius);
         --amplify-components-button-transition-duration: var(--amplify-components-fieldcontrol-transition-duration);
         --amplify-components-button-font-size: var(--amplify-components-fieldcontrol-font-size);
-        --amplify-components-button-line-height: var(--amplify-components-fieldcontrol-font-size);
+        --amplify-components-button-line-height: var(--amplify-components-fieldcontrol-line-height);
```
dbanksdesign, line 38: 🙌

packages/ui/src/theme/__tests__/defaultTheme.test.ts
```diff
@@ -56,7 +56,7 @@ describe('@aws-amplify/ui', () => {
         --amplify-components-button-loading-color: var(--amplify-colors-font-tertiary);
         --amplify-components-button-loading-background-color: transparent;
         --amplify-components-button-loading-border-color: var(--amplify-colors-border-tertiary);
-        --amplify-components-button-disabled-color: var(--amplify-colors-font-tertiary);
+        --amplify-components-button-disabled-color: var(--amplify-colors-font-disabled);
```
dbanksdesign, line 59: 🙌

packages/ui/src/theme/tokens/components/button.js
```diff
@@ -1,3 +1,5 @@
+const { boxShadow } = require('./menu');
```
dbanksdesign, line 1: I don't see this being used in the file, unless I'm missing something?

dbanksdesign (commented):
Just a few small comments. Overall looks great though, lots of small improvements too!

docs/src/components/MenuPropControls.tsx
reesscot: It's added by default already via the `placeholder` prop. I actually added it manually and found it to be duplicated. I'm still unsure of whether we should have the `placeholder` API on the SelectField primitive.

docs/src/pages/ui/primitives/menu/examples/controlledExample.tsx
reesscot: Yeah, it makes for a much more maintainable examples I think!

docs/src/pages/ui/primitives/menu/react.mdx
reesscot: I was wondering that as well. Maybe we could grab all the ComponentClassNames that start with the `Menu` prefix...

docs/src/pages/ui/primitives/menu/react.mdx
reesscot: I could move `triggerClassName` but not `className`. The reason is that `className` gets added to the Menu Content, which is rendered in a React Portal outside the Example div, so I wouldn't be able to target the content from a class on the Example div.

packages/ui/src/theme/tokens/components/button.js
reesscot: Weird, looks like an IDE autoimport mistake on my part. Will remove.

packages/react/src/primitives/types/menu.ts
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Alignment of menu against trigger
+   * @default "start"
+   */
+  align?: 'start' | 'center' | 'end';
+
+  /**
+   * Handle open and close event of menu
+   */
+  onOpenChange?: (open: boolean) => void;
```
hvergara, line 17:
```suggestion
  onOpenChange?: (isOpen: boolean) => void;
```

packages/react/src/primitives/types/menu.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ButtonProps } from './button';
+import { FlexProps } from './flex';
+import { Sizes } from './base';
+
+export interface MenuProps extends FlexProps {
+  /**
+   * Alignment of menu against trigger
+   * @default "start"
+   */
+  align?: 'start' | 'center' | 'end';
```
hvergara, line 12: is there a better name for this property? can be easily confused with other `*align` style properties

reesscot: Agreed, do you have any suggestions? Perhaps `menuAlign`?

packages/react/src/primitives/types/menu.ts
reesscot: Fixed

docs/src/pages/ui/primitives/menu/examples/sizeExample.tsx
```diff
@@ -0,0 +1,20 @@
...
+      <Menu>
+        <MenuItem>Download</MenuItem>
+        <MenuItem>Create a Copy</MenuItem>
+      </Menu>
+      <Menu size="large">
+        <MenuItem>Download</MenuItem>
+        <MenuItem>Create a Copy</MenuItem>
+      </Menu>
+    </Flex>
+  );
```
slaymance, line 19: I noticed that, while the icons for these different menus are different sizes, the actual size of each menu is the same. It seems like Flex stylings will overwrite stylings from the size attribute.

reesscot: Hmm, I was having this problem before, let me make sure it's still fixed.

reesscot: Dropdowns are sizing correctly, and the buttons are scaled based on the icon sizes. I'll create a ticket to address icon size.

packages/react/src/primitives/types/menu.ts
reesscot: Updated to `menuAlign`

docs/src/pages/ui/primitives/menu/examples/sizeExample.tsx
slaymance: It looks like the heights are actually different and it's just the widths that are the same since they inherit from Flex.

packages/react/src/primitives/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { ComponentClassNames } from '../shared/constants';
+import { MenuButton } from './MenuButton';
+import { MenuItemProps } from '../types';
+
+export const MENU_ITEM_TEST_ID = 'amplify-menu-item-test-id';
+
+export const MenuItem = React.forwardRef<
+  HTMLDivElement,
+  MenuItemProps & { children?: React.ReactNode }
```
zchenwei, line 13: Why not just add `children` to `MenuItemProps`?

packages/react/src/primitives/Menu/MenuButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+  ) => {
+    const propStyles = usePropStyles(rest, style);
+    const nonStyleProps = useNonStyleProps(rest);
+
+    return (
+      <button
+        ref={ref}
+        className={classNames(
+          ComponentClassNames.Button,
+          ComponentClassNames.FieldGroupControl,
```
zchenwei, line 40: Why apply a `ComponentClassNames.FieldGroupControl,` here?

packages/react/src/primitives/Menu/MenuButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+        )}
+        data-fullwidth={isFullWidth}
+        data-loading={isLoading}
+        data-size={size}
+        data-variation={variation}
+        disabled={isDisabled || isLoading}
+        type={type}
+        data-testid={testId}
+        aria-label={ariaLabel}
+        id={id}
```
zchenwei, line 51: Why not just have `id` and `role` pass along with `...rest`? It seems we don't need to do something special to them

packages/react/src/primitives/Menu/Menu.tsx
```diff
@@ -0,0 +1,55 @@
...
+            triggerClassName
+          )}
+        >
+          <IconMenu size={size} />
+        </MenuButton>
+      )}
+    </DropdownMenuTrigger>
+    <DropdownMenuContent align={menuAlign}>
+      <ButtonGroup
+        className={classNames(ComponentClassNames.Menu, className)}
```
zchenwei, line 46: I was thinking if ` className={classNames(ComponentClassNames.Menu, className)}` should apply to the outer `<DropdownMenuContent align={menuAlign}>`. It looks wired that the outer element has `role='menu'` applied to, but the class `amplify-menu` goes into to the button group, which is a flex container

packages/react/src/primitives/shared/constants.ts
```diff
@@ -41,6 +41,9 @@ export enum ComponentClassNames {
   Label = 'amplify-label',
   Link = 'amplify-link',
   Loader = 'amplify-loader',
+  Menu = 'amplify-menu',
+  MenuItem = 'amplify-menu__item',
+  MenuTrigger = 'amplify-menu-trigger',
```
zchenwei, line 46: What do you think of adding a comment here? At least for me, this is where I feel confused and had the question why `'amplify-menu-trigger'`  instead of `'amplify-menu__trigger'`. For external users, I think we have an awesome `Target Classes` section for them to reference

packages/react/src/primitives/Menu/Menu.tsx
```diff
@@ -0,0 +1,55 @@
...
+}) => (
+  <DropdownMenu onOpenChange={onOpenChange} open={isOpen}>
+    <DropdownMenuTrigger asChild={true}>
+      {trigger ?? (
+        <MenuButton
+          role="button"
+          size={size}
+          testId={MENU_TRIGGER_TEST_ID}
+          className={classNames(
+            ComponentClassNames.MenuTrigger,
```
zchenwei, line 36: I am ok with the way we assign the classnames for trigger and menu here. We had better not break the BEM convention.

zchenwei (commented):
This is looking great! Left a few comments for you us to discuss

packages/react/src/primitives/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { DropdownMenuItem } from '@radix-ui/react-dropdown-menu';
+
+import { ComponentClassNames } from '../shared/constants';
+import { MenuButton } from './MenuButton';
+import { MenuItemProps } from '../types';
+
+export const MENU_ITEM_TEST_ID = 'amplify-menu-item-test-id';
+
+export const MenuItem = React.forwardRef<
+  HTMLDivElement,
```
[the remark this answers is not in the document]
reesscot: It must be `HTMLDivElement` because the element where the ref is applied is an `HTMLDivElement`.

```
const DropdownMenuItem: React.ForwardRefExoticComponent<DropdownMenuItemProps & React.RefAttributes<HTMLDivElement>>
```

packages/react/src/primitives/Menu/MenuItem.tsx
reesscot: Good point, updated.

packages/react/src/primitives/Menu/MenuItem.tsx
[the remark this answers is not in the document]
zchenwei: Right, but there is no `div` get rendered in this use case. This will merge the original component props with the props of the supplied element/component and change the underlying DOM node. I think the ref will eventually go to the button, and pass in a `div` ref doesn't make sense I think

packages/react/src/primitives/Menu/MenuButton.tsx
reesscot: It's not needed, removed.

packages/react/src/primitives/Menu/MenuButton.tsx
reesscot: good call, removed.

packages/react/src/primitives/Menu/Menu.tsx
reesscot: I tried to do that, but it actually breaks the Radix menu because it overrides the classes they are using. So it's better to keep it on a element we control.

packages/react/src/primitives/shared/constants.ts
reesscot: After online discussion, I've updated the classnames to `amplify-menu-content` and `amplify-menu-trigger` to make it more readable.

zchenwei (approved):
LGTM！

packages/react/src/primitives/types/menu.ts
hvergara: thanks <3

hvergara (approved):
Looks good to me, approved :)

PR #618, opened 2021-11-03
Make style prop types more flexible

hvergara:
*Description of changes:*
This PR makes style props more flexible: they always will accept an string, but without losing type inference & auto-completion ;)

### DIFF
--- a/packages/react/src/primitives/types/style.ts
+++ b/packages/react/src/primitives/types/style.ts
@@ -6,11 +6,15 @@ import { ImageStyleProps } from './image';
 import { TextAreaStyleProps } from './textArea';
 
 /**
- * Extract string-like types. Defaulted to string if prop doesn't allow it
+ * Extract style compatible types (string literal | number | any string)
  */
-export type StyleProp<PropertyType> = PropertyType extends string
-  ? PropertyType
-  : string;
+export type StyleProp<PropertyType> =
+  | (PropertyType extends number
+      ? number
+      : PropertyType extends string
+      ? PropertyType
+      : never)
+  | (string & {});
 
 export interface ResponsiveObject<PropertyType> {
   base?: PropertyType;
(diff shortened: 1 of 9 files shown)


packages/react/src/primitives/types/style.ts
```diff
@@ -5,6 +5,13 @@ import { GridItemStyleProps, GridContainerStyleProps } from './grid';
 import { ImageStyleProps } from './image';
 import { TextAreaStyleProps } from './textArea';
 
+/**
+ * Extract string-like types. Defaulted to string if prop doesn't allow it
+ */
+export type StyleProp<PropertyType> = PropertyType extends string
```
hvergara, line 11: This is the important part: this generic type will extract the string-like types only, allowing to keep IDE auto-completion, and defaulting to string if no string-like type is found.

Examples:
- `StyleProp<"a" | "b" | "c"> => "a" | "b" | "c"`
- `StyleProp<"a" | "b" | "c" | number> => "a" | "b" | "c" | string`
- `StyleProp<"a" | "b" | "c" | string> => "a" | "b" | "c" | string`
- `StyleProp<any> => string`
- `StyleProp<unknown> => string`

reesscot: Can't we support `number` in addition to string?

hvergara: Fixed in last commit. Now number-compatible types are: `number | string | ("other" | "string" | "literals")`

dbanksdesign (approved):
LGTM! :shipit:

reesscot (approved):
LGTM!

reesscot (changes requested):
Looks great, but needs a changeset before merging.

hvergara:
> Looks great, but needs a changeset before merging.

Added :)

reesscot (approved):

PR #635, opened 2021-11-04
Expander Primitive

zchenwei:
*Description of changes:*

This PR will be adding an `Expander` primitive.


packages/ui/src/theme/css/component/expander.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.amplify-expander__item {
+  display: block;
+  overflow: hidden;
+  margin-top: 1px;
+  box-shadow: var(--amplify-components-expander-item-box-shadow);
+
+  &:first-of-type {
+    margin-top: 0;
+    border-start-start-radius: var(
```
zchenwei, line 17: I may need to use media query here later because Safari doesn't support logical border radius well

packages/react/src/primitives/Expander/ExpanderItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+            size="large"
+          />
+        </Trigger>
+      </Header>
+      <Content
+        aria-labelledby={triggerId}
+        className={ComponentClassNames.ExpanderContent}
+        id={contentId}
+      >
+        <View
```
zchenwei, line 56: Need an extra container here to add `paddingTop` and `paddingBottom`. This is how radix demonstrate on their examples

docs/src/pages/ui/primitives/expander/react.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+<Example>
+  <UncontrolledExpander type="single" isCollapsible />
+</Example>
+
+## CSS Styling
+
+### Global styling
+
+To override styling on all Expanders, you can set the Amplify CSS variables with the built-in `.amplify-expander`, `.amplify-expander__item`, `.amplify-expander__header`, `.amplify-expander__trigger`, `.amplify-expander__icon`, `.amplify-expander__content` and `.amplify-expander__content__text` class.
```
zchenwei, line 231: Looks like a `Target Class` section here would be better

packages/react/src/primitives/Expander/ExpanderItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const triggerId = useStableId();
+  const contentId = useStableId();
+  const { rest } = splitPrimitiveProps(_rest);
+  return (
+    <Item
+      className={classNames(ComponentClassNames.ExpanderItem, className)}
+      data-testid={EXPANDER_ITEM}
+      {...rest}
+    >
+      <Header
```
zchenwei, line 33: Maybe we can set `asChild={true}` later and use our own `Heading` as wrapper, which will allow users to specify the heading level. But I ran into some errors when I tried to do so because of `ref`. Hopefully this could get resolved once we start forwarding `ref`

packages/ui/src/theme/css/component/expander.scss
```diff
@@ -0,0 +1,131 @@
...
+  padding-inline-start: var(
+    --amplify-components-expander-content-padding-inline-start
+  );
+  padding-inline-end: var(
+    --amplify-components-expander-content-padding-inline-end
+  );
+
+  &[data-state='open'] {
+    animation-name: amplify-expander-slide-down;
+    animation-timing-function: cubic-bezier(0.87, 0, 0.13, 1);
```
zchenwei, line 78: This timing function looks good

packages/react/src/primitives/Expander/ExpanderItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+        className={ComponentClassNames.ExpanderHeader}
+        data-testid={EXPANDER_HEADER}
```
hvergara, line 35: why do we need these test ids? only for testing purposes? I'd prefer to search by classname on tests rather than injecting an static value just to have a hook when running our tests suites. What do you think?

packages/react/src/primitives/Expander/ExpanderItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Primitive } from '../types/view';
+import { ExpanderItemProps } from '../types/expander';
+import { ComponentClassNames } from '../shared/constants';
+import { splitPrimitiveProps } from '../shared/styleUtils';
+import { useStableId } from '../shared/utils';
+
+export const EXPANDER_ITEM = 'expander-item';
+export const EXPANDER_HEADER = 'expander-header';
+export const EXPANDER_ICON = 'expander-icon';
+// export const EXPANDER_CONTENT = 'expander-content';
```
hvergara, line 15: should we remove this one?

dbanksdesign, line 15: nit: remove commented out code

packages/react/src/primitives/Expander/Expander.tsx
```diff
@@ -0,0 +1,53 @@
...
+  // It is not in use but remove it from rest to avoid type errors
+  onValueChange,
+  testId,
+  type = 'single',
+  value,
+  ..._rest
+}) => {
+  const { rest } = splitPrimitiveProps(_rest);
+
+  const expander =
```
hvergara, line 24: is the `isCollapsible` property the only difference between these two? I wonder if there's a simpler way to write this conditional render 🤔

packages/ui/src/theme/css/component/expander.scss
```diff
@@ -0,0 +1,131 @@
...
+    );
+    border-end-end-radius: var(
+      --amplify-components-expander-item-border-end-end-radius
+    );
+  }
+
+  &:focus-within {
+    position: relative;
+    z-index: 1;
+    box-shadow: var(--amplify-components-expander-item-focus-box-shadow);
```
zchenwei, line 38: This will make the focus styling over other border styling. It is a really nice trick that I recently learned

packages/react/src/primitives/Expander/ExpanderItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+          aria-controls={contentId}
+          className={ComponentClassNames.ExpanderTrigger}
+          id={triggerId}
+        >
+          {title}
+          <IconExpandMore
+            aria-hidden
+            className={ComponentClassNames.ExpanderIcon}
+            data-testid={EXPANDER_ICON}
+            size="large"
```
dbanksdesign, line 47: With the icon size change scott pushed we might not need this, but we also might need to make sure the size looks correct.

packages/react/src/primitives/types/expander.ts
```diff
@@ -0,0 +1,42 @@
+import { ViewProps } from '../types/view';
+
+type ExpanderType = 'single' | 'multiple';
+
+export interface ExpanderProps extends ViewProps {
+  /**
+   * The value of the item(s) to expand.  Use on uncontrolled component.
```
dbanksdesign, line 7: Great doc comments!

packages/ui/src/theme/tokens/components/expander.js
```diff
@@ -0,0 +1,76 @@
...
+      offsetX: '0',
+      offsetY: '1px',
+      blurRadius: '10px',
+      color: '{colors.overlay.30.value}',
+    },
+  },
+  width: { value: '100%' },
+  item: {
+    marginTop: { value: '1px' },
+    boxShadow: {
```
dbanksdesign, line 16: We could use references to a box shadows.

packages/ui/src/theme/tokens/components/expander.js
```diff
@@ -0,0 +1,76 @@
...
+      value: {
+        offsetX: '0',
+        offsetY: '1px',
+        blurRadius: '0',
+        color: '{colors.overlay.20.value}',
+      },
+    },
+    borderStartStartRadius: { value: '0.5rem' },
```
dbanksdesign, line 24: Could these be references to radii?

packages/ui/src/theme/tokens/components/expander.js
```diff
@@ -0,0 +1,76 @@
...
+          offsetY: '0',
+          blurRadius: '0',
+          spreadRadius: '2px',
+          color: '{colors.border.focus.value}',
+        },
+      },
+    },
+  },
+  trigger: {
+    height: { value: '3rem' },
```
dbanksdesign, line 41: What happens if the content in the trigger is too long and breaks to multiple lines?

packages/ui/src/theme/tokens/components/expander.js
```diff
@@ -0,0 +1,76 @@
...
+    paddingInlineStart: { value: '1.5rem' },
+    paddingInlineEnd: { value: '1.5rem' },
```
dbanksdesign, line 43: Could we use space references here and for other paddings?

packages/ui/src/theme/tokens/time.ts
```diff
@@ -9,6 +14,11 @@ export interface Time {
 }
 
 export const time: Time = {
+  animation: {
```
dbanksdesign, line 17: What is the difference between 'animation' and 'transition'?

dbanksdesign (commented):
Just a few comments, mostly on the design tokens. But it is looking pretty good! Love the animations

packages/react/src/primitives/Expander/ExpanderItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import { IconExpandMore } from '../Icon';
+import { View } from '../View';
+import { Primitive } from '../types/view';
+import { ExpanderItemProps } from '../types/expander';
+import { ComponentClassNames } from '../shared/constants';
+import { splitPrimitiveProps } from '../shared/styleUtils';
+import { useStableId } from '../shared/utils';
+
+export const EXPANDER_ITEM = 'expander-item';
```
reesscot, line 12: non-blocking nit: could have TEST_ID suffix?

packages/react/src/primitives/Expander/ExpanderItem.tsx
reesscot: Yeah, maybe wait on supporting custom Header until we forward refs. Want to create a ticket?

packages/react/src/primitives/Expander/ExpanderItem.tsx
reesscot: We are using test ids in many places. I think it's a more stable API than searching by classnames. It draws a clear line between things used for styling (classnames) and things used for testing (data-testid). This allows both to be changed without side affects. It's also a best practice encouraged by React Testing Library.

reesscot (commented):
I'm still reviewing, but publishing some comments

packages/ui/src/theme/css/component/expander.scss
```diff
@@ -0,0 +1,131 @@
...
+  background-color: var(--amplify-components-expander-background-color);
+  border-radius: var(--amplify-components-expander-border-radius);
+  box-shadow: var(--amplify-components-expander-box-shadow);
+  width: var(--amplify-components-expander-width);
+}
+
+.amplify-expander__item {
+  display: block;
+  overflow: hidden;
+  margin-top: 1px;
```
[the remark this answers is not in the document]
reesscot: If we need this, would prefer padding. https://mxstbr.com/thoughts/margin/

packages/react/src/primitives/Expander/ExpanderItem.tsx
zchenwei: Creating

packages/react/src/primitives/Expander/ExpanderItem.tsx
hvergara: I get the benefits of using `test-ids`, but developer's HTML will contain unknown`test-id` attributes that might cause some confusion (at first). This is not a blocker, btw.

packages/ui/src/theme/css/component/expander.scss
reesscot: Can we do this now, please, so we don't forget?

packages/react/src/primitives/types/expander.ts
```diff
@@ -0,0 +1,42 @@
...
+   */
+  defaultValue?: string | string[];
+
+  /**
+   * The controlled value of the item(s) to expand. Must be used in conjunction with onChange.
+   */
+  value?: string | string[];
+
+  /**
+   * Determines whether the opened item can be collapsed if this is a single expander.
```
reesscot, line 17: Only applies to `type`single
```suggestion
   * Determines whether an opened item can be collapsed
   * Note: only applies when `type` is `single`
```

packages/react/src/primitives/types/expander.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface ExpanderItemProps extends ViewProps {
+  /**
+   * The content of the heading.
+   */
+  title?: string;
+
+  /**
+   * A unique value for the item.
+   */
+  value: string;
```
reesscot, line 41: Is this more of a key than a value? Is it a value because it's used for accessibility?

packages/ui/src/theme/css/component/expander.scss
reesscot:
```
  // required for Safari 14 and below
  // see: https://caniuse.com/mdn-css_properties_border-end-end-radius
  @supports not (border-end-end-radius: 0) {
  }
```

packages/ui/src/theme/css/component/expander.scss
reesscot: Nice! Can you please add a brief comment on how this works?

packages/ui/src/theme/css/component/expander.scss
zchenwei: yup, that's my plan

packages/ui/src/theme/tokens/time.ts
reesscot: +1, seems like we can use the same times for both

packages/ui/src/theme/tokens/components/fieldControl.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   borderStyle: { value: 'solid' },
   borderColor: { value: '{colors.border.primary.value}' },
   borderWidth: { value: '{borderWidths.small.value}' },
-  borderRadius: { value: '{radii.medium.value}' },
+  borderRadius: { value: '{radii.small.value}' },
```
reesscot, line 5: Does this change affect other components? Any reason why we want to make the radii smaller?

packages/ui/src/theme/css/component/expander.scss
```diff
@@ -0,0 +1,131 @@
...
+    transform: rotate(180deg);
+  }
+}
+
+@keyframes amplify-expander-slide-down {
+  from {
+    height: 0;
+  }
+  to {
+    height: fit-content;
```
reesscot, line 120: Can you confirm that this works in FF and Safari? FF doesn't appear to support `height: fit-content`. See: https://caniuse.com/?search=fit-content

reesscot (changes requested):
This is looking great! Just had a few comments/requests

docs/src/pages/ui/primitives/expander/react.mdx
```diff
@@ -0,0 +1,331 @@
...
+</Expander>;
+```
+
+<Example>
+  <UncontrolledExpander type="multiple" defaultValue={['item-1', 'item-2']} />
+</Example>
+
+### Allow collapsing
+
+You may want to enable the Expanders with type `single` to collapse all items. In this case, set the `isCollapsible` prop to true.
```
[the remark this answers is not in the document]
zchenwei: Type single only allows one section to be open. But once the item gets open, it cannot be closed unless this prop is set to true

packages/react/src/primitives/Expander/Expander.tsx
zchenwei: No, it will start crying

packages/ui/src/theme/tokens/components/fieldControl.js
zchenwei: I updated the tokens. The `medium` value goes into `small` now

packages/react/src/primitives/Expander/Expander.tsx
```diff
@@ -0,0 +1,53 @@
...
+    type === 'multiple' ? (
+      <Root
+        className={classNames(ComponentClassNames.Expander, className)}
+        data-testid={testId}
+        defaultValue={defaultValue as string[]}
+        onValueChange={onChange}
+        type={type}
+        value={value as string[]}
```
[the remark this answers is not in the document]
zchenwei: No, we cannot. The value type is `string | string[]`, while Radix is using a union on interface level and only accept either `string` or `string[]` for different `type`.

packages/react/src/primitives/Expander/Expander.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { splitPrimitiveProps } from '../shared/styleUtils';
+
+export const Expander: Primitive<ExpanderProps, typeof Root> = ({
+  children,
+  className,
+  defaultValue,
+  isCollapsible,
+  onChange,
+  // It is not in use but remove it from rest to avoid type errors
+  onValueChange,
```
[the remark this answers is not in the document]
zchenwei: It is a prop from `Root` and will cause type errors after using `Primitive` to merge props. The type of this function is `((value: string) => void) | ((value: string[]) => void)`, which is a result of union on interface level by Radix. But what we want is `(value: string | value: string[]) => void`, and here comes the `onChange`

packages/ui/src/theme/tokens/time.ts
dbanksdesign: I like it, so we would just have short, medium, long directly under time? Let's go with that.

dbanksdesign (approved):
LGTM! :shipit:

packages/ui/src/theme/tokens/time.ts
zchenwei: Yea, exactly 💯

reesscot (approved):
Shipit!

PR #636, opened 2021-11-05
useStorageURL React hook

hvergara:
*Description of changes:*
This PR introduces a new `useStorageURL` React hook to easily get a signed URL for an Amplify Storage object

- [x] Implementation
- [x] Unit tests
- [x] Documentation

### DIFF
--- a/packages/react/src/hooks/useStorageURL.ts
+++ b/packages/react/src/hooks/useStorageURL.ts
@@ -1,0 +1,32 @@
+import { S3ProviderGetConfig, Storage } from '@aws-amplify/storage';
+import { useEffect, useState } from 'react';
+
+export interface UseStorageURLResult {
+  url?: string;
+  error?: Error;
+  isLoading: boolean;
+}
+
+export const useStorageURL = (key: string, options?: S3ProviderGetConfig) => {
+  const [result, setResult] = useState<UseStorageURLResult>({
+    isLoading: true,
+  });
+
+  const fetch = () => {
+    setResult({ isLoading: true });
+
+    const promise = Storage.get(key, options);
+
+    // Attempt to fetch storage object url
+    promise
+      .then((url) => setResult({ url, isLoading: false }))
+      .catch((error) => setResult({ error, isLoading: false }));
+
+    // Cancel current promise on unmount
+    return () => Storage.cancel(promise);
+  };
+
+  useEffect(fetch, []);
+
+  return { ...result, fetch };
+};
--- a/packages/react/src/hooks/useStorageURL.ts
+++ b/packages/react/src/hooks/useStorageURL.ts
@@ -12,6 +12,10 @@ export const useStorageURL = (key: string, options?: S3ProviderGetConfig) => {
     isLoading: true,
   });
 
+  // Used to prevent an infinite loop on useEffect, because `options`
+  // will have a different reference on every render
+  const serializedOptions = JSON.stringify(options);
+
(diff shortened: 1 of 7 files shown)


packages/react/src/hooks/useStorage.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const [error, setError] = useState<Error>();
+
+  const fetch = () => {
+    Storage.get(key, options).then(setURL).catch(setError);
+  };
+
+  useEffect(fetch, []);
+
+  return {
+    url,
```
ericclemmons, line 15: In your screenshot, `<Image src={url}>` was immediately used. But if this is an async call, what happens to images? Is there a "flash", or do we expect `` components to have an empty space from ``?

packages/react/src/hooks/useStorage.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const useStorageURL = (key: string, options: S3ProviderGetConfig) => {
+  const [url, setURL] = useState<string>();
+  const [error, setError] = useState<Error>();
+
+  const fetch = () => {
+    Storage.get(key, options).then(setURL).catch(setError);
+  };
+
+  useEffect(fetch, []);
```
ericclemmons, line 12: When components are unmounted, these requests can be cancelled:
> https://docs.amplify.aws/lib/storage/cancel-requests/q/platform/js/

ericclemmons (commented):
Nice work!  I only had one question about async behavior & another optimization opportunity.

packages/react/src/hooks/useStorage.tsx
hvergara: At first render, `url` will be undefined. Users can read the value of `url` to determine what to render in that case.

```
const { url } = useStorageURL(key, options);

if (!url) {
  return <Placeholder />;
}

return <Image src={url} />
```

packages/react/src/hooks/useStorage.tsx
hvergara: resolved! thanks for the suggestion <3

packages/react/src/hooks/useStorage.tsx
ericclemmons: That works for me! In documentation, the async nature of it should be clear so long as we do `const { isLoading, url } = useStorageUrl(key)` 👍

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -0,0 +1,32 @@
...
+    // Attempt to fetch storage object url
+    promise
+      .then((url) => setResult({ url, isLoading: false }))
+      .catch((error) => setResult({ error, isLoading: false }));
+
+    // Cancel current promise on unmount
+    return () => Storage.cancel(promise);
+  };
+
+  useEffect(fetch, []);
```
ayush987goyal, line 29: I think you should either:
1. Wrap the `fetch` function in a `useCallback` or
2. Inline the function in the `useEffect` hook itself

Currently the dependencies are not appropriately addressed. The `key` and `options` props are supposed to go into the dependency array of the `useEffect` hook since you would want to get a new URL when the `key` or `options` change.

hvergara: good call! added `key` and a serialized version of `options` as part of the dependency list

ayush987goyal (approved):

PR #639, opened 2021-11-05
chore: Mark "Preferred Username" custom input as required

### LINKED ISSUE #590: (@next) Authenticator sends field data to Cognito even if field was emptied
### To reproduce

1. run `amplify-pull` on `environments/auth-with-email-and-custom-attributes`
2. `yarn react dev && yarn next-example dev`
3. Navigate to `http://localhost:3000/ui/components/authenticator/custom-sign-up-fields`
4. Type any string to "Preferred Username" and then delete it to make it empty
5. Fill out the rest of the form and submit

You'll see that sign up request goes through, even though it should error because `Preferred Username` is empty. Inspecting the network tab shows that authenticator sent a response with empty string, which is valid to Cognito's pov.

image

### Solution
[`handleInput`](https://github.com/aws-amplify/amplify-ui/blob/3143deff199a9aab367f253020205d3e2f25fc5b/packages/ui/src/machines/authenticator/actions.ts#L97-L106) should delete the current input field key/value off from context if the new value is empty, instead of setting it to an empty string.

wlee221:
*Issue #, if available:* Fixes #590

*Description of changes:* Marks "Preferred Username" custom input as required. Also adds a cypress util for testing if an input is invalid based on native html validations using `inputHTMLElement.checkValidity()`.

### DIFF
--- a/examples/vue/src/pages/ui/components/authenticator/custom-sign-up-fields/index.vue
+++ b/examples/vue/src/pages/ui/components/authenticator/custom-sign-up-fields/index.vue
@@ -41,7 +41,6 @@ const services = {
         name="preferred_username"
         placeholder="Preferred Username"
         type="text"
-        :required="false"
       ></amplify-text-field>
       <authenticator-sign-up-form-fields />
       <amplify-check-box :errorMessage="validationErrors.acknowledgement" />
--- a/packages/e2e/cypress/integration/common/shared.ts
+++ b/packages/e2e/cypress/integration/common/shared.ts
@@ -144,3 +144,9 @@ Then('the {string} button is disabled', (name: string) => {
     name: new RegExp(`^${escapeRegExp(name)}$`, 'i'),
   }).should('be.disabled');
 });
+
+Then('the {string} field is invalid', (name: string) => {
+  cy.findInputField(name)
+    .then(($el) => $el.get(0).checkValidity())
+    .should('be.false');
+});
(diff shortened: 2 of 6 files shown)


packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -144,3 +144,9 @@ Then('the {string} button is disabled', (name: string) => {
     name: new RegExp(`^${escapeRegExp(name)}$`, 'i'),
   }).should('be.disabled');
 });
+
+Then('the {string} field is invalid', (name: string) => {
+  cy.findInputField(name)
+    .then(($el) => $el.get(0).checkValidity())
```
wlee221, line 150: Need to do this because `cy.findInputField`, which uses `dom-testing-library-cypress` under the hood returns `JQuery<HTMLInputElement>`.

examples/vue/src/pages/ui/components/authenticator/custom-sign-up-fields/index.vue
```diff
@@ -41,7 +41,6 @@ const services = {
         name="preferred_username"
         placeholder="Preferred Username"
         type="text"
-        :required="false"
```
ErikCH, line 44: Thanks for fixing this in Vue too!

packages/e2e/cypress/integration/common/shared.ts
ErikCH: Good catch!

ErikCH (approved):
🚢

Issue #637, opened 2021-11-05
Custom Header/Footer slots


ericclemmons:
Closes #259, #498, #257

Customizing text via `I18n` isn't enough.  Slots are needed to **prepend, append, & replace** entire elements.

- [x] #647 
- [x] #695
- [x] #661
- [x] #690
- [x] Updated docs

---

These largely exist in the internal implementation, but need to be exposed & documented.

## Slot Names (React)

Here I mocked up the React documentation with where the slots would go:

> ![Screen Shot 2021-11-05 at 10 38 03 AM](https://user-images.githubusercontent.com/15182/140538173-3c93ec43-1a19-49b9-b6f8-5ba8163b3f29.png)
> ![Screen Shot 2021-11-05 at 10 43 51 AM](https://user-images.githubusercontent.com/15182/140538480-979af72f-5e72-44b4-b02b-d03eb4ed49b2.png)
> ![Screen Shot 2021-11-05 at 10 44 14 AM](https://user-images.githubusercontent.com/15182/140538483-7c40d7b2-7da5-4d33-9026-d9d30cc76638.png)

## Slot Implementation (React)

This example implementation does the following:

- Adds a Logo above the entire Authenticator in the `Header`
- Adds a legal text below the entire Authenticator in the `Footer`
- Appends the `SignIn.Header` with a subtitle
- Injects a custom `SignIn.Footer` below the form

```jsx
const customComponents = {
  Header() {
    return <img src="logo.png" />
  },
  Footer() {
    return <p>Copyright 2021</p>
  },
  SignIn: {
    Header() {
  	  return (
		<>
		  <Authenticator.SignIn.Header />
		  <h3>Internal Use Only!</h3>
		</>
	  )
    },
    Footer() {
      return <p>Call 1-800-help-desk for support</p>
    }
  }
}

function App() {
  return (
    <Authenticator components={customComponents}>
	  {({ user }) => <h1>Howdy {user.username}</h1>}
	</Authenticator>
  )
}
```

ericclemmons:
Closed via https://github.com/aws-amplify/amplify-ui/pull/695

PR #362, opened 2021-09-10
Error message styling

eddiekeller:
*Issue #, if available:*

https://app.asana.com/0/1200881947262804/1200960625209728/f

*Description of changes:*

Adding `Alert` primitives for `RemoteError` messages. These are distinct from `ValidationError` messages. Remote errors come from Cognito, validation errors are operations we perform (password matching, for instance).

Screen Shot 2021-09-10 at 11 26 20 AM
Screen Shot 2021-09-10 at 11 50 46 AM

EDIT: screenshot showing new width of authenticator to better match the mocks --

Screen Shot 2021-09-10 at 12 01 13 PM

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #647, opened 2021-11-09
Vue: Add new Slots

ErikCH:
*Issue #, if available:*
#637

*Description of changes:*
Added new slots and cleaned up old slots for all the components.

image

```
<template>
  <authenticator>
    <template v-slot:header>
      <h1 class="text-center">Authenticator Header</h1>
    </template>
    <template v-slot:footer>
      <h1>Authenticator Footer</h1>
    </template>
    <template v-slot:sign-in-header>
      <h2>Sign In Header</h2>
    </template>
    <template v-slot:sign-in-footer>
      <h2>Sign In Footer</h2>
    </template>
    <template v-slot="{ user, signOut }">
      <h1>Hello {{ user.username }}!</h1>
      <button @click="signOut">Sign Out</button>
    </template>
  </authenticator>
</template>
```

### DIFF
--- a/packages/ui/src/theme/css/component/authenticator.scss
+++ b/packages/ui/src/theme/css/component/authenticator.scss
@@ -41,8 +41,10 @@
   background-color: var(--amplify-colors-background-primary);
   box-shadow: var(--amplify-shadows-small);
   place-self: center;
-  width: calc(var(--amplify-breakpoints-values-small) * 1rem);
-  max-width: 460px;
+  width: 460px;
+  @media (max-width: 30em) {
+    width: 100%;
+  }
 }
 
 [data-amplify-authenticator] [data-amplify-form] {
--- a/packages/ui/src/theme/css/component/authenticator.scss
+++ b/packages/ui/src/theme/css/component/authenticator.scss
@@ -42,6 +42,7 @@
   box-shadow: var(--amplify-shadows-small);
   place-self: center;
   width: calc(var(--amplify-breakpoints-values-small) * 1rem);
+  max-width: 460px;
 }
 
 [data-amplify-authenticator] [data-amplify-form] {
--- a/packages/ui/src/theme/css/component/authenticator.scss
+++ b/packages/ui/src/theme/css/component/authenticator.scss
@@ -41,8 +41,8 @@
   background-color: var(--amplify-colors-background-primary);
   box-shadow: var(--amplify-shadows-small);
   place-self: center;
-  width: 460px;
-  @media (max-width: 30em) {
+  width: 30rem;
+  @media (max-width: 30rem) {
     width: 100%;
   }
 }
--- a/packages/vue/src/components/authenticator.vue
+++ b/packages/vue/src/components/authenticator.vue
@@ -4,6 +4,7 @@
     data-amplify-authenticator
     v-if="!state?.matches('authenticated')"
   >
+    <slot name="header"> </slot>
     <div data-authenticator-variation="modal" v-if="variationModal" />
     <div data-amplify-container>
       <base-two-tabs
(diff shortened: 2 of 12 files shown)


packages/ui/src/theme/css/component/authenticator.scss
```diff
@@ -42,6 +42,7 @@
   box-shadow: var(--amplify-shadows-small);
   place-self: center;
   width: calc(var(--amplify-breakpoints-values-small) * 1rem);
+  max-width: 460px;
```
ErikCH, line 45: Let's say for now we set this to 460px?

ErikCH: Which is the size of the legacy authenticator.

packages/ui/src/theme/css/component/authenticator.scss
```diff
@@ -41,7 +41,10 @@
   background-color: var(--amplify-colors-background-primary);
   box-shadow: var(--amplify-shadows-small);
   place-self: center;
-  width: calc(var(--amplify-breakpoints-values-small) * 1rem);
+  width: 460px;
+  @media (max-width: 30em) {
+    width: 100%;
+  }
```
ErikCH, line 47: Let's say we set this to 460px, and then reduce less then < 30em ? Also, `--amplify-breakpoints-values-small` doesn't exist, not sure if it's not being generated correctly 🤷

packages/vue/src/components/confirm-sign-in.vue
```diff
@@ -11,9 +11,11 @@
           style="flex-direction: column"
           :disabled="actorState.matches('confirmSignIn.pending')"
         >
-          <base-heading :level="3" class="amplify-heading">
-            {{ confirmSignInHeading }}
-          </base-heading>
```
ErikCH, line 16: I added in the change so we are no longer showing the header for sign in / sign up. I left them for the other pages though, since we don't have tabs on them like confirm sign up, etc.

packages/ui/src/theme/css/component/authenticator.scss
wlee221: I believe we should use the same units here (`rem` should be more accurate to the device text size).

```suggestion
  width: 30rem;
  @media (max-width: 30rem) {
    width: 100%;
  }
```

This was previously set to 36rem (#362), but I'll defer on your judgement on which value looks better.

ErikCH: OK! I changed it to 30rem! I think it looks good at that.

packages/vue/src/components/confirm-sign-in.vue
ErikCH: Scratch that, I'm adding back the header for now, since it breaks the tests. I'll have that in a different PR.

packages/vue/src/components/authenticator.vue
```diff
@@ -4,6 +4,7 @@
     data-amplify-authenticator
     v-if="!state?.matches('authenticated')"
   >
+    <slot name="header"> </slot>
```
ericclemmons, line 7: Will this ` ` space cause it to render anything by default?  I know that sometimes `padding` can get a "ghost element" when there's excess whitespace.

ericclemmons (approved):
This looks great!

Only a small question, but I went ahead & updated https://github.com/aws-amplify/amplify-ui/issues/637 to include this PR since we'll have ~4 PRs to ultimately close it.

packages/vue/src/components/authenticator.vue
ErikCH: OK, I just double checked, and it did NOT render anything there with the whitespace. So we should be fine.

PR #652, opened 2021-11-09
build: prepend commit hash before date to @next versions

wlee221:
*Issue #, if available:*

*Description of changes:* Prepends commit hash so that the versions look like 

```
next-[commit_sha]-timestamp
```

which will increase the specificity of each versions in case customers are interested in pinning to one of them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "build:icons": "yarn react build:icons",
     "test": "yarn react test && yarn e2e test",
     "prepare": "husky install",
-    "version:next": "yarn changeset version --snapshot next && yarn build",
+    "version:next": "yarn changeset version --snapshot next-$(git rev-parse --short=7 HEAD) && yarn build",
     "publish:next": "yarn changeset publish --tag next",
     "version:latest": "echo \"not implemented until stable release\"",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "build:icons": "yarn react build:icons",
     "test": "yarn react test && yarn e2e test",
     "prepare": "husky install",
-    "version:next": "yarn changeset version --snapshot next && yarn build",
+    "version:next": "yarn changeset version --snapshot next-$(git rev-parse --short=7 HEAD) && yarn build",
```
wlee221, line 20: Put `--short=7` as length 7 short hash seemed standard in GitHub.

wlee221:
Here's what my POC versions look like with this change: https://www.npmjs.com/package/@wlee221-changeset-poc/core.

package.json
ericclemmons: For comparison,

https://github.com/aws-amplify/amplify-ui/blob/6a0c6e32dbb253c49291a1d7bc6cbd222a8c5001/docs/next.config.js#L4

yields `623-signup-attributes`, while your change yields `746c0a4`

👍

ericclemmons (approved):

PR #660, opened 2021-11-09
Docs - update secondary nav & grid logic

reesscot:
*Description of changes:*
This PR updates the secondary navigation and components grid logic. It alphabetizes all the links and adds some missing components.

Screen Shot 2021-11-09 at 3 03 21 PM

Screen Shot 2021-11-09 at 3 04 04 PM

### DIFF
--- a/docs/src/data/links.ts
+++ b/docs/src/data/links.ts
@@ -1,0 +1,134 @@
+export interface ComponentNavItem {
+  href: string;
+  label: string;
+  body: string;
+}
+
+const sortByLabel = (a: ComponentNavItem, b: ComponentNavItem) =>
+  a.label < b.label ? -1 : a.label > b.label ? 1 : 0;
+
+export const baseComponents: ComponentNavItem[] = [
+  {
+    href: '/components/view',
+    label: 'View',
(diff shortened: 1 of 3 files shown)


docs/src/data/links.ts
```diff
@@ -0,0 +1,134 @@
...
+    href: '/components/s3album',
+    label: 'S3 Album',
+    body: 'Renders a list of S3Image objects.',
+  },
+].sort(sortByLabel);
+
+export const dataDisplayComponents = [
+  { href: '/components/badge', label: 'Badge', body: `` },
+  { href: '/components/rating', label: 'Rating', body: `` },
+].sort(sortByLabel);
```
dbanksdesign, line 64: Is there any performance worry here? It is probably nothing, but should we just sort these by hand in the source?

dbanksdesign (approved):
LGTM! Just one question about the sorting thing, not a big deal.

docs/src/data/links.ts
reesscot: Since the docs are SSG/SSR, I don't think there's much performance impact.

PR #650, opened 2021-11-09
Render required attributes by default in React

ericclemmons:
*Issue #, if available:* Closes #623

*Description of changes:*

- [x] `SignUpAttributes` type for https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html
- [x] `signUpAttributes: SignUpAttribute[]` to authenticator machine
- [x] E2E test for all supported fields
- [x] React implementation

	Notice how these have labels, mostly because placeholders & pre-filled values aren't clear enough about _what the field is for_.  (`birthday`, `given_name`, `name`, `nickname` are all examples of this)

	> Screen Shot 2021-11-09 at 10 43 15 AM

- [x] `console.debug` for any missing implementations that customers have to provide:

	```console
	Authenticator does not have a default implementation for address. Customize Authenticator.SignUp.FormFields to add your own.
    Authenticator does not have a default implementation for gender. Customize Authenticator.SignUp.FormFields to add your own.
    Authenticator does not have a default implementation for locale. Customize Authenticator.SignUp.FormFields to add your own.
    Authenticator does not have a default implementation for picture. Customize Authenticator.SignUp.FormFields to add your own.
    Authenticator does not have a default implementation for updated_at. Customize Authenticator.SignUp.FormFields to add your own.
    Authenticator does not have a default implementation for zoneinfo. Customize Authenticator.SignUp.FormFields to add your own.
	```

- [x] Update documentation for `signUpAttributes`
- [x] Remove `preferred_username` from https://ui.docs.amplify.aws/ui/components/authenticator?platform=react#sign-up-fields

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #623, opened 2021-11-03
Render Standard Attribute form fields by default


ericclemmons:
The goal here is for fields like `preferred_username` to render by default instead of requiring custom configuration:
> https://ui.docs.amplify.aws/ui/components/authenticator?platform=react#sign-up-fields

See the current list & references:
https://github.com/aws-amplify/amplify-ui/blob/1fc955c4e32a2868fa85078befa2f7e2fe95b5a1/packages/ui/src/machines/authenticator/signUp.ts#L250-L275

Closes #590 

- [x] #609 so that we can re-use `signupAttributes` (or whatever the key is)
- [x] **Must** include `preferred_username`, `email`, and `phone_number`
- [x] Render free-form `string` attributes as a text field
- [x] Leave non-standard fields like `gender` for custom implementations

PR #668, opened 2021-11-10
Render required attributes by default in Angular

ericclemmons:
*Issue #, if available:* #623, extending #650

*Description of changes:*

- [x] Remove `@todo-angular` placeholders
- [x] Add `labelHidden` prop to form fields so that we can display labels on sign up
- [x] Add `signUpAttributes` prop to `amplify-authenitcator`
- [x] Modify `amplify-sign-up-form-fields` component to render configured attributes

image

### DIFF
--- a/packages/ui/src/types/authMachine.ts
+++ b/packages/ui/src/types/authMachine.ts
@@ -34,12 +34,10 @@ export interface SignInContext extends BaseFormContext {
 }
 
 // https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html
-export const SignUpAttributes = [
-  'address',
+export const signUpFieldsWithDefault = [
   'birthdate',
   'email',
   'family_name',
-  'gender',
   'given_name',
   'locale',
   'middle_name',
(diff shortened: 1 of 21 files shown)


packages/ui/src/types/authMachine.ts
```diff
@@ -132,10 +143,17 @@ export type LoginMechanism = typeof LoginMechanismArray[number];
...
-// other non-alias inputs that Cognito would require
-export type AuthInputNames = LoginMechanism | 'confirmation_code' | 'password';
-
-export type AuthInputAttributes = Record<AuthInputNames, InputAttributes>;
+// Auth fields that we provide default fields with
+export type AuthFieldsWithDefaults =
+  | LoginMechanism
+  | SignUpFieldsWithDefaults
+  | 'confirmation_code'
+  | 'password';
```
wlee221, line 151: This... can use a lot of clean ups. Punting it to post-GA per time restraints, but ideally we should use fewer variables to depict this. Also could implement `fieldHasDefaults(field: string): field is AuthFieldWithDefaults` to make sign up form field generation easier.

ErikCH: 👍

ErikCH (approved):
Looks good

PR #670, opened 2021-11-10
Update react styles

ErikCH:
*Issue #, if available:*

The background color and border is missing from the React UI Authenticator

*Description of changes:*

Added back in styles with wrapper div

New drop shadow box with fixed styles

image

### DIFF
--- a/packages/react/src/components/Authenticator/Router/index.tsx
+++ b/packages/react/src/components/Authenticator/Router/index.tsx
@@ -35,39 +35,41 @@ export function Router({ className, children }: RouterProps) {
         <View data-authenticator-variation="modal" />
 
         <View data-amplify-container="">
-          {(() => {
-            switch (route) {
-              case 'idle':
-                return null;
-              case 'confirmSignUp':
-                return <ConfirmSignUp />;
-              case 'confirmSignIn':
-                return <ConfirmSignIn />;
-              case 'setupTOTP':
-                return <SetupTOTP />;
-              case 'signIn':
-              case 'signUp':
-                return <SignInSignUpTabs />;
-              case 'forceNewPassword':
-                return <ForceNewPassword />;
-              case 'resetPassword':
-                return <ResetPassword />;
-              case 'confirmResetPassword':
-                return <ConfirmResetPassword />;
-              case 'verifyUser':
-                return <VerifyUser />;
-              case 'confirmVerifyUser':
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Authenticator/Router/index.tsx
```diff
@@ -35,39 +35,41 @@ export function Router({ className, children }: RouterProps) {
...
-                return <ForceNewPassword />;
-              case 'resetPassword':
-                return <ResetPassword />;
-              case 'confirmResetPassword':
-                return <ConfirmResetPassword />;
-              case 'verifyUser':
-                return <VerifyUser />;
-              case 'confirmVerifyUser':
-                return <ConfirmVerifyUser />;
+          <View data-amplify-body="">
```
ErikCH, line 38: Added this in to fix styles, to match Angular and Vue implementation

packages/ui/src/theme/css/component/authenticator.scss
```diff
@@ -53,7 +53,7 @@
 
 [data-amplify-authenticator] [data-amplify-container] > [data-amplify-body] {
   background-color: var(--amplify-colors-background-primary);
-  box-shadow: var(--amplify-shadows-small);
+  box-shadow: var(--amplify-shadows-medium);
```
ErikCH, line 56: The box shadow was barely visible on the docs site. This will make it more visible.

packages/react/src/components/Authenticator/Router/index.tsx
ericclemmons: Is there a difference between `data-amplify-container` and `data-amplify-body`?  If not, we can rename the 1st one to match.

ericclemmons (approved):
Not blocking, but an opportunity to keep all frameworks in sync.

packages/react/src/components/Authenticator/Router/index.tsx
ErikCH: the `data-amplify-container` is used all over the place. I'd say we leave it as is for now, but eventually we'll need to go through and clean up all the `data-` attributes

wlee221: `data-amplify-container` contains `data-amplify-body`, whereas `data-amplify-header` and `data-amplify-footer` contain the appropriate top-level Authenticator slots. This can be revisited though.

PR #673, opened 2021-11-10
ci: Add a workflow for tracking `@todo-`s

wlee221:
*Issue #, if available:*

*Description of changes:* Currently, we have `@todo` items fail on `main` to track `@todo` tags. But failing `e2e` tests is not necessary because `@todo` tests are _expected_ to fail and are not breaking functionality. Instead, we really care about _tracking_ those @todos.

This adds a separate workflow for tracking @todo- features. e2e tests will no longer fail on them and continuous deployment will not be blocked by it.

See https://github.com/aws-amplify/amplify-ui/runs/4172243825?check_suite_focus=true for an example output:

image

---

### DIFF
--- a/.github/scripts/track-todos.sh
+++ b/.github/scripts/track-todos.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# total number of .feature files that has unresolved @todo-tags
+nTodoFiles=$(grep -R -e "@todo-*" ./packages/e2e/features/ -l | wc -l | sed -e 's/^[ \t]*//');
+
+red="\033[0;31m"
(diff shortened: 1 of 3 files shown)


wlee221:
Open questions:

- Are there circumstances in which we _want_ e2e tests to run TODOs?
- Later, we can create an issue that shows coverage of our TODOs.

ericclemmons (approved):
This is really clever (and fast!)

Originally I figured Cypress could run _only_ `@todo-*` tags to see what would fail and why, but this is better since we'd want a new PR that converts them anyway.

PR #681, opened 2021-11-11
Updating docs colors to meet contrast standards

dbanksdesign:
*Issue #, if available:*

*Description of changes:* 
* Fixing selection background color in dark mode
* Fixing light/dark transition bug in ToC
* Updating syntax highlighting colors to add more contrast for a11y

https://user-images.githubusercontent.com/321279/141269896-3c33b03f-7e88-45af-b55f-417247abbe8e.mp4

### DIFF
--- a/docs/src/theme.ts
+++ b/docs/src/theme.ts
@@ -96,6 +96,18 @@ export const theme = createTheme({
             secondary: { value: '{colors.neutral.20.value}' },
             tertiary: { value: '{colors.neutral.20.value}' },
           },
+
+          overlay: {
+            10: { value: 'hsla(0, 0%, 100%, 0.1)' },
+            20: { value: 'hsla(0, 0%, 100%, 0.2)' },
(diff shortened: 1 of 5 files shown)


docs/src/theme.ts
```diff
@@ -96,6 +96,18 @@ export const theme = createTheme({
             secondary: { value: '{colors.neutral.20.value}' },
             tertiary: { value: '{colors.neutral.20.value}' },
           },
+
+          overlay: {
```
reesscot, line 100: Are these used somewhere I'm missing, or planning to use later?

reesscot (approved):
LGTM!

docs/src/theme.ts
dbanksdesign: We use these in a few places, in the toggle button and pagination, and I was using it for the text selection background (the issue Tim had). We should probably included these dark mode overrides in the default theme so users can get dark mode out of the box though.

PR #685, opened 2021-11-11
chore: adding flex and margin style props

zchenwei:
*Description of changes:*

1. Adding flex and margin style props
2. Making all props required in `ComponentPropToStyleProp` so that any missing style props in `ComponentPropsToStylePropsMap` will be caught.
3. Adding interface to type `useAmplify` hook explicitly

### DIFF
--- a/packages/react/src/primitives/types/style.ts
+++ b/packages/react/src/primitives/types/style.ts
@@ -38,44 +38,38 @@ export type ResponsiveStyle<PropertyType> =
   | ResponsiveObject<StyleProp<PropertyType>>;
 
 export interface BaseStyleProps extends FlexItemStyleProps, GridItemStyleProps {
-  alignSelf?: ResponsiveStyle<Property.AlignSelf>;
+  alignSelf?: ResponsiveStyle<StyleToken<Property.AlignSelf>>;
   backgroundColor?: ResponsiveStyle<StyleToken<Property.BackgroundColor>>;
-  border?: ResponsiveStyle<Property.Border>;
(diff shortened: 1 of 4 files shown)


packages/react/src/primitives/types/style.ts
```diff
@@ -53,6 +53,13 @@ export interface BaseStyleProps extends FlexItemStyleProps, GridItemStyleProps {
   height?: ResponsiveStyle<StyleToken<Property.Height>>;
   left?: ResponsiveStyle<StyleToken<Property.Left>>;
   letterSpacing?: ResponsiveStyle<Property.LetterSpacing>;
+  lineHeight?: ResponsiveStyle<Property.LineHeight>;
+  margin?: ResponsiveStyle<Property.Margin>;
+  maxHeight?: ResponsiveStyle<Property.MaxHeight>;
+  maxWidth?: ResponsiveStyle<Property.MaxWidth>;
+  minHeight?: ResponsiveStyle<Property.MinHeight>;
+  minWidth?: ResponsiveStyle<Property.MinWidth>;
+  opacity?: ResponsiveStyle<Property.Opacity>;
```
dbanksdesign, line 62: Could we change these lines to `ResponsiveStyle<StyleToken<Property.*>>;` ?

dbanksdesign (approved):
Not blocking but we could have the new style props accept design tokens. :shipit:

packages/react/src/primitives/types/style.ts
zchenwei: Oh yea, I would be happy to do that

reesscot (approved):
🚢!

PR #686, opened 2021-11-11
primitives: update exports and add displayname

reesscot:
*Description of changes:*
This PR updates the following:
* exports and adds a `displayName` for all Primitives. See: [React docs](https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools)
* converts all `export * from './primitive'` exports to named exports to prevent accidentally shipping internal APIs

This work is needed for the React.forwardRef support work, but I wanted to split it out into a separate PR to make it easier to review.


dbanksdesign (approved):
LGTM! :shipit:

PR #689, opened 2021-11-12
chore: reorganizing primitive exports

zchenwei:
*Issues:*

```
interface UseAmplifyOutput {
  components: React.ReactNode;
  theme: Theme;
}
```

Typing components roughly with `React.ReactNode` will confuse Typescript for a ReactNode is not necessarily callable.

Screen Shot 2021-11-11 at 3 29 57 PM

The workaround here is to reorganize the existing exports and separate out the component-only exports from primitives package for explicit typing purpose and this way also provides us with intellisense hints 

*Description of changes:*

Reorganize primitive exports and update the return type of `useAmplify` hook. 

Intellisense hints:

Screen Shot 2021-11-11 at 4 27 27 PM


packages/react/src/primitives/primitives.ts
```diff
@@ -0,0 +1,40 @@
+export { Alert } from './Alert';
```
reesscot, line 1: Is this change necessary, now that our exports are all named exports?

packages/react/src/hooks/useAmplify.ts
```diff
@@ -3,10 +3,10 @@ import * as React from 'react';
 import { Theme } from '@aws-amplify/ui';
 
 import { AmplifyContext } from '../components/AmplifyProvider/AmplifyContext';
-import * as primitives from '../primitives';
+import * as primitives from '../primitives/primitives';
```
reesscot, line 6: Feels awkward to export from `primitives/primitives`

reesscot (changes requested):
Can you rebase from main and see if the index -> primitives file change is necessary?

packages/react/src/primitives/primitives.ts
zchenwei: I don't see the relation with named exports. There will be hooks exported  which we don't want to include in primitives?

packages/react/src/hooks/useAmplify.ts
zchenwei: Rename to `primitives/components.ts`

reesscot: This file name feels better

packages/react/src/primitives/primitives.ts
reesscot: OK, I understand this a bit better. Still feels awkward, but it makes sense.

reesscot (approved):

PR #692, opened 2021-11-12
Fixing theme type export

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Fixing the issue with the Theme type. Open to suggestions on how to fix this in a better way. 

### DIFF
--- a/packages/react/src/components/AmplifyProvider/index.tsx
+++ b/packages/react/src/components/AmplifyProvider/index.tsx
@@ -2,9 +2,10 @@ import { IdProvider } from '@radix-ui/react-id';
 import { ReactNode } from 'react';
 
 import { AmplifyContext } from './AmplifyContext';
-import { createTheme, defaultTheme, Theme } from '@aws-amplify/ui';
+import { createTheme, defaultTheme, Theme as UiTheme } from '@aws-amplify/ui';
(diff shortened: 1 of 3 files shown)


hvergara (approved):

reesscot (approved):
LGTM

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -2,9 +2,10 @@ import { IdProvider } from '@radix-ui/react-id';
 import { ReactNode } from 'react';
 
 import { AmplifyContext } from './AmplifyContext';
-import { createTheme, defaultTheme, Theme } from '@aws-amplify/ui';
+import { createTheme, defaultTheme, Theme as UiTheme } from '@aws-amplify/ui';
 import * as primitives from '../../primitives';
 
+export type Theme = UiTheme;
```
ericclemmons, line 8: I could've sworn `export { Theme } from "@aws-amplify/ui"` would work, but that's a nit.

You can manually validate by checking the `dist/index.d.ts` output after `yarn build`, too.

ericclemmons (approved):
Looks like it should work!  Approving so we can unblock others...

PR #696, opened 2021-11-12
Conditional exports: ESM/CommonJS

hvergara:
*Description of changes:*
Fix exports section from our UI/React package.json definitions to handle both ESM/CommonJS cases on import

https://nodejs.org/api/packages.html#conditional-exports

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -4,8 +4,14 @@
   "main": "dist/index.js",
   "module": "dist/esm/index.js",
   "exports": {
-    ".": "./dist/esm/index.js",
-    "./legacy": "./dist/esm/legacy.js",
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/index.js"
+    },
+    "./legacy": {
+      "import": "./dist/esm/legacy.js",
+      "require": "./dist/index.js"
+    },
     "./styles.css": "./dist/index.css"
   },
   "browser": {
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -10,7 +10,7 @@
     },
     "./legacy": {
       "import": "./dist/esm/legacy.js",
-      "require": "./dist/index.js"
+      "require": "./dist/legacy.js"
     },
     "./styles.css": "./dist/index.css"
   },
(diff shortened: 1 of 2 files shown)


dpilch (approved):

packages/react/package.json
```diff
@@ -4,8 +4,14 @@
   "main": "dist/index.js",
   "module": "dist/esm/index.js",
   "exports": {
-    ".": "./dist/esm/index.js",
-    "./legacy": "./dist/esm/legacy.js",
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/index.js"
+    },
+    "./legacy": {
```
ericclemmons, line 11: Does this require any changes to
https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/legacy/package.json?

It sounds like there's inconsistent behavior between environments:

- ESM bundlers (e.g. `esbuild`) respect `exports`
- CJS bundlers (some webpack configs) ignore `exports` and read `main`.
- Node, however, originally read `main` but newer versions read `exports` _even if not within a strict module structure_.

What I'm wondering is _how can we test or validate this_?

ericclemmons (approved):
Whoops! Meant to approve.  If e2e approves, so do I!

@dpilch @hvergara Can someone create an issue articulating the failure scenario so we can add a test for it?

ESM/CJS is _so hard to get right_, I don't want to keep this as a blindspot.

dpilch:
Minimum reproduction before fix:
1. `mkdir tmp && cd tmp`
2. `npm init`
3. Add `"type": "module",` to `package.json`
4. Create `index.js` with the following:
```
import * as AmplifyUI from '@aws-amplify/ui-react';

console.log(AmplifyUI);
```
5. `node index.js` (`SyntaxError: Cannot use import statement outside a module`)
6. Create `index.cjs` with the following:
```
const AmplifyUI = require('@aws-amplify/ui-react');

console.log(AmplifyUI);
```
7. 5. `node index.js` (Same `SyntaxError: Cannot use import statement outside a module`)

reesscot (approved):
LGTM

PR #698, opened 2021-11-12
Add forwardRef support to Button & View

reesscot:
*Description of changes:*
This PR adds React.forwardRef support to the Button and View primitives to start. I'm not including all the primitive components in this PR for ease of review. When approved, I will cut follow up PRs for the other primitives.

*Why this PR*
React offers the `ref` API as a way to access the underlying DOM element for a component. This PR forwards a ref attached on one of the Amplify UI primitives down to the underlying element. This allows customers to have an escape hatch to manage focus, animations, check if an element is in viewport, etc.

Example:

```tsx
const CustomerComponent = () =>{
  const ref = React.useRef<HTMLButtonElement>(null);
  
  // In response to some other app action, force focus on button:
  // ref.current.focus();
  
  return(
    <Button ref={ref>Hello</Button>
  )
}

```

### DIFF
--- a/.changeset/dirty-keys-eat.md
+++ b/.changeset/dirty-keys-eat.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": minor
+---
+
+Add forwardRef support to Button & View
--- a/packages/react/src/primitives/Button/Button.tsx
+++ b/packages/react/src/primitives/Button/Button.tsx
@@ -1,37 +1,47 @@
 import classNames from 'classnames';
+import * as React from 'react';
+
 import { ComponentClassNames } from '../shared/constants';
-import { ButtonProps, Primitive } from '../types';
+import { ButtonProps, PrimitiveWithForwardRef } from '../types';
 import { View } from '../View';
 
-export const Button: Primitive<ButtonProps, 'button'> = ({
-  className,
-  children,
-  isFullWidth = false,
-  isDisabled,
-  isLoading,
-  loadingText = '',
-  size,
-  type = 'button',
-  variation,
-  ...rest
-}) => (
-  <View
-    as="button"
-    className={classNames(
-      ComponentClassNames.Button,
-      ComponentClassNames.FieldGroupControl,
-      className
-    )}
-    data-fullwidth={isFullWidth}
-    data-loading={isLoading}
-    data-size={size}
-    data-variation={variation}
-    isDisabled={isDisabled || isLoading}
-    type={type}
-    {...rest}
-  >
-    {isLoading && loadingText ? <span>{loadingText}</span> : children}
-  </View>
-);
+export const ButtonInner: PrimitiveWithForwardRef<ButtonProps, 'button'> = (
+  {
+    className,
+    children,
+    isFullWidth = false,
+    isDisabled,
+    isLoading,
+    loadingText = '',
+    size,
+    type = 'button',
+    variation,
+    ...rest
+  },
+  ref
+) => {
+  return (
+    <View
+      ref={ref}
+      as="button"
+      className={classNames(
+        ComponentClassNames.Button,
+        ComponentClassNames.FieldGroupControl,
+        className
+      )}
+      data-fullwidth={isFullWidth}
+      data-loading={isLoading}
+      data-size={size}
+      data-variation={variation}
+      isDisabled={isDisabled || isLoading}
+      type={type}
+      {...rest}
+    >
+      {isLoading && loadingText ? <span>{loadingText}</span> : children}
+    </View>
+  );
+};
+
+export const Button = React.forwardRef(ButtonInner);
 
 Button.displayName = 'Button';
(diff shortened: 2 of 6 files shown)


.changeset/dirty-keys-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aws-amplify/ui-react": minor
```
reesscot, line 2: Should this be major since it's technically a change in behavior?

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -1,37 +1,47 @@
...
+    isFullWidth = false,
+    isDisabled,
+    isLoading,
+    loadingText = '',
+    size,
+    type = 'button',
+    variation,
+    ...rest
+  },
+  ref
```
reesscot, line 21: Added `ref` as second parameter

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -1,37 +1,47 @@
...
-    data-size={size}
-    data-variation={variation}
-    isDisabled={isDisabled || isLoading}
-    type={type}
-    {...rest}
-  >
-    {isLoading && loadingText ? <span>{loadingText}</span> : children}
-  </View>
-);
+export const ButtonInner: PrimitiveWithForwardRef<ButtonProps, 'button'> = (
```
reesscot, line 8: We may want to rename `PrimitiveWithForwardRef` to just `Primitive` when we've migrated all the other components. At this point I didn't want to touch the other primitives.

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -1,37 +1,47 @@
...
+    size,
+    type = 'button',
+    variation,
+    ...rest
+  },
+  ref
+) => {
+  return (
+    <View
+      ref={ref}
```
reesscot, line 25: This is the only change to the returned jsx, setting the `ref` prop to be forwarded.

packages/react/src/primitives/types/view.ts
```diff
@@ -7,6 +7,20 @@ type MergeProps<A, B> = A & Omit<B, keyof A>;
 
 export type ElementType = React.FC<any> | keyof JSX.IntrinsicElements;
 
+/**
+ * Convert string element type to DOMElement Type
+ * e.g. 'button' => HTMLButtonElement
+ */
+export type HTMLElementType<Element extends ElementType> =
```
reesscot, line 14: Hat tip to @hvergara for this type magic

packages/react/src/primitives/Button/Button.tsx
hvergara: cool!

packages/react/src/primitives/types/view.ts
```diff
@@ -18,15 +32,32 @@ export type PrimitiveProps<
   Props extends ViewProps,
   Element extends ElementType
 > = MergeProps<
-  Omit<Props, 'as'> & { as?: Element | Props['as'] },
-  ElementProps<Element>
+  Omit<Props, 'as'> & {
+    as?: Element | Props['as'];
+  },
+  Omit<ElementProps<Element>, 'ref'> // exclude `ref?: LegacyRef` included in DetailedHTMLProps
```
reesscot, line 38: This Omit is required because `JSX.IntrinsicElements[Element]` resolves to React.DetailedHTMLProps, which resolves to `ClassAttributes`, which contains the `ref?: LegacyRef<T>`. AFAIK, it's invalid to include the `ref` prop in a functional component, and causes type errors on functional component primitives. For example, `Text` throws a type error when passing props to View because `React.LegacyRef` and `React.Ref` are not compatible.

hvergara: 🙌🏽

hvergara (approved):
Looks great! thank you so much

packages/react/src/primitives/types/view.ts
```diff
@@ -7,6 +7,20 @@ type MergeProps<A, B> = A & Omit<B, keyof A>;
 
 export type ElementType = React.FC<any> | keyof JSX.IntrinsicElements;
 
+/**
+ * Convert string element type to DOMElement Type
+ * e.g. 'button' => HTMLButtonElement
+ */
+export type HTMLElementType<Element extends ElementType> =
+  Element extends keyof JSX.IntrinsicElements
+    ? JSX.IntrinsicElements[Element] extends React.DetailedHTMLProps<
```
zchenwei, line 16: This could potentially be `Element extends keyof JSX.IntrinsicElements ? React.ElementRef<Element> : HTMLDivElement;` ? <- https://flow.org/en/docs/react/types/#toc-react-elementref

And can you explain why always return `HTMLDivElement` for `React.FC<any>`? 😕

zchenwei (commented):
This is looking good. Just left one comment

packages/react/src/primitives/types/view.ts
reesscot: It might be more accurate to return `HTMLElement` instead of `HTMLDivElement`. Thoughts @hvergara?

zchenwei: I am asking because we do have `SVGELement`, though the primitives don't need ref in large part

hvergara: yeah, good idea

PR #700, opened 2021-11-12
refactor: removing labelled by children in favor of label prop

zchenwei:
*Issue #, if available:*

The `CheckboxField` is accepting `children` as label which is inconsistent with other field primitives

*Description of changes:*

Removing labelled by `children` support in favor of passing a `label` prop

### DIFF
--- a/docs/src/pages/components/checkboxfield/react.mdx
+++ b/docs/src/pages/components/checkboxfield/react.mdx
@@ -1,8 +1,10 @@
 import { CheckboxField } from '@aws-amplify/ui-react';
 
-import { ControlledCheckbox } from './examples';
+import { ControlledCheckbox } from './examples/controlledExample';
+import { ValueExample } from './examples/valueExample';
 import { Demo } from './demo';
 import { Example } from '@/components/Example';
+import { Fragment } from '@/components/Fragment';
 
 CheckboxField is used to mark an individual item as selected, or to select multiple items from a list of individual items.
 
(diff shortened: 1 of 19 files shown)


docs/src/pages/components/checkboxfield/react.mdx
```diff
@@ -14,46 +16,22 @@ CheckboxField is used to mark an individual item as selected, or to select multi
...
-      name="subscribe-controlled"
-      value="yes"
-      checked={checked}
-      onChange={(e) => setChecked(e.target.checked)}
-    >
-      Subscribe
-    </CheckboxField>
-  );
-};
+```tsx file=./examples/controlledExample.tsx
```
dbanksdesign, line 33: 🙌

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
```diff
@@ -43,28 +28,11 @@ export const CheckboxField: Primitive<CheckboxFieldProps, 'input'> = ({
       testId={testId}
     >
       <Checkbox
-        alignContent={alignContent}
```
dbanksdesign, line 46: I'm assuming we don't need any of these because of the rest spread? Much cleaner!

packages/react/src/primitives/types/checkbox.ts
```diff
@@ -6,7 +6,14 @@ export interface CheckboxProps extends FlexProps, InputProps {
...
    * The label text
    */
-  children: FieldProps['label'];
+  label: FieldProps['label'];
+
+  /**
+   * Visually hide label (not recommended in most cases)
+   * @default false
+   */
+  labelHidden?: boolean;
```
dbanksdesign, line 15: @reesscot @hvergara should this be isLabelHidden? It looks like field, and search field use `labelHidden`, switchField uses `isLabelHidden`

dbanksdesign (approved):
LGTM! Just one question about `labelHidden` v. `isLabelHidden`

packages/react/src/primitives/CheckboxField/CheckboxField.tsx
zchenwei: Yup, correct. Just let the rest spread take care of those

packages/react/src/primitives/types/checkbox.ts
reesscot: Yeah, I think there's an issue in the backlog to update them all to `isLabelHidden`

PR #703, opened 2021-11-12
chore(angular): Move form-field and user-name-alias to authenticator/

wlee221:
*Issue #, if available:*

*Description of changes:* Moves `amplify-form-field` and `user-name-alias` from `primitives/` directory to `authenticator/` because these components are opinionated for authenticator use and thus are not primitives.


ErikCH (approved):
👍

PR #706, opened 2021-11-14
Badge docs cleanup

dbanksdesign:
*Issue #, if available:*

*Description of changes:*

https://user-images.githubusercontent.com/321279/141669514-6aafb44e-527a-498c-b050-bb9114f325eb.mp4


docs/src/pages/components/badge/examples/BadgeStyleProps.tsx
```diff
@@ -0,0 +1,13 @@
+import { Badge, useTheme } from '@aws-amplify/ui-react';
+
+export const BadgeStyleProps = () => {
+  const { tokens } = useTheme();
+  return (
+    <Badge
+      backgroundColor={tokens.colors.brand.secondary[20]}
```
hvergara, line 7: 😍

docs/src/pages/components/badge/demo.tsx
```diff
@@ -1,21 +1,34 @@
...
-import { useBadgeProps } from '@/components/useBadgeProps';
-import { Example } from '@/components/Example';
+import { BadgePropControls } from './BadgePropControls';
+import { useBadgeProps } from './useBadgeProps';
+import { Demo } from '@/components/Demo';
+
+const propsToCode = (badgeProps) => {
+  return (
+    `<Badge` +
+    (badgeProps.size ? `\n  size="${badgeProps.size}"` : '') +
```
hvergara, line 11: should we escape these values? Bad people could abuse props values to inject unexpected props to generated code.

docs/src/components/Demo.tsx
```diff
@@ -16,14 +37,16 @@ export const Demo = ({ children, propControls, themeControls, code }) => {
   return (
     <Card className="docs-component-demo">
       <Flex direction="row" alignItems="stretch">
-        <Flex direction="column" width="50%">
-          {children}
+        <Flex direction="column" flex="1">
+          <View>{children}</View>
```
hvergara, line 41: not a blocker: why do we need to enclose children within a View?

hvergara (approved):

docs/src/pages/components/alert/react.mdx
```diff
@@ -100,7 +111,9 @@ Use the `isDismissible` prop to control whether the user can dismiss the Alert.
 Use the `onDismiss` prop to pass a function that will run when the Alert is dismissed. Note that `isDismissible` must be set to `true`.
 
 <Example>
-  <OnDismissAlertExample />
+  <View>
```
reesscot, line 114: Why is this `View` required?

reesscot (approved):
LGTM

docs/src/pages/components/badge/demo.tsx
dbanksdesign: That is a good point.. do you think we need that for when the values come from a select field or just when they come from a text field?

hvergara: I feel we need to quote those values regardless of where the value come from. A simple quote/double quote escaping should be enough.

docs/src/components/Demo.tsx
dbanksdesign: Both this, and Scott's comment below are because of wrapping things in a Flex kind of breaks the inline/block model. If we set `alignItems` to `flex-start` then any element that is normally a block, like an Alert because it is a div, gets shrunk to its contents width. If it is set to 'normal' or 'stretch' then it will stretch inline-block elements like the badge. There might be some magical combination of properties to get both scenarios to work, but wrapping the contents in a div unsets the flex-iness of its children

PR #716, opened 2021-11-15
Remove AmplifyProvider custom components feature

reesscot:
*Description of changes:*
This PR removes the `AmplifyProvider` custom components feature. The original intention of this feature was to allow customers to completely override any primitive used below the `AmplifyProvider`. However, it doesn't work correctly as currently implemented as we import primitives used in other composed primitives directly from relative paths. Additionally, there is little evidence of a customer need for this feature that cannot be met primarily by overriding the provided CSS variables. This PR removes the feature for GA launch, but we may return to revisit this feature in the future depending on customer feedback.

Further background:
https://github.com/aws-amplify/amplify-ui/pull/582
(original git blame commits were overwritten)

### DIFF
--- a/.changeset/sharp-cherries-knock.md
+++ b/.changeset/sharp-cherries-knock.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": minor
+---
+
+Remove AmplifyProvider custom components feature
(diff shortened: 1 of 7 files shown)


hvergara (approved):

zchenwei (approved):
LGTM!

PR #682, opened 2021-11-11
Reduced bundle size by changing imports of lodash

ErikCH:
*Issue #, if available:*
#599

*Description of changes:*
I noticed the imports of lodash in the `@aws-amplify/ui` package was inconsistent. I updated it so all the lodash imports looked the same. This also helped reduce the bundle size in the Vue packages by 50kb.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #719, opened 2021-11-16
ForwardRef support - base primitives

reesscot:
*Description of changes:*
This PR adds forward ref support for base primitives: `Divider`, `Heading`, `Icon`, `Image`,  `Text` by wrapping them in `React.forwardRef`. I'm limiting the number of primitives changed per PR to make them easier to review.

```jsx
  
  // ref.current.getBoundingClientRect()
  
  return (
    <Image ref={ref} />
  );
};
```

### DIFF
--- a/packages/react/src/primitives/Text/__test__/Text.test.tsx
+++ b/packages/react/src/primitives/Text/__test__/Text.test.tsx
@@ -1,8 +1,10 @@
-import { Text } from '../Text';
+import * as React from 'react';
+import { kebabCase } from 'lodash';
 import { render, screen } from '@testing-library/react';
+
+import { Text } from '../Text';
 import { ComponentClassNames } from '../../shared';
 import { ComponentPropsToStylePropsMap } from '../../types';
-import { kebabCase } from 'lodash';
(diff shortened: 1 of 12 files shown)


packages/react/src/primitives/Text/__test__/Text.test.tsx
```diff
@@ -1,8 +1,10 @@
-import { Text } from '../Text';
+import * as React from 'react';
+import { kebabCase } from 'lodash';
```
dbanksdesign, line 2: Should we change this to `import kebabCase from 'lodash/kebabCase'`? Following this PR: #682 I know this isn't part of the change, but something we should track.

dbanksdesign (approved):
This looks great! Thanks for getting this in and adding unit tests! Just had one comment unrelated to these changes.

packages/react/src/primitives/Text/__test__/Text.test.tsx
reesscot: I can do this as a separate PR

Issue #710, opened 2021-11-15
Flex & Grid don't accept ref prop while View does


dreamorosi:
This issue is related to `@aws-amplify/ui-react` next.

As shown in the image below, trying to pass a `ref` to a `<Flex />` component, the same happens with the `<Grid />` one. 
![image](https://user-images.githubusercontent.com/7353869/141821659-bf63a3d2-0264-4906-bde8-8ed1295d64ba.png)

On the other hand, the `<View />` component accepts it.

reesscot:
Hi @dreamorosi, thanks for opening this ticket, it confirms that this is a needed feature. Currently only View and Button support the ref prop. I'm currently in the process of adding support for the ref prop on the other primitives. Stay tuned!

PR #720, opened 2021-11-16
ForwardRef support - Flex and Grid

reesscot:
*Issue:*
Closes https://github.com/aws-amplify/amplify-ui/issues/710

*Description of changes:*
This PR adds forward ref support for `Flex` and `Grid` primitives by wrapping them in `React.forwardRef`. I'm limiting the number of primitives changed per PR to make them easier to review.

```jsx
  
  // ref.current.getBoundingClientRect()
  
  return (
    <Flex ref={ref} />
  );
};
```

### DIFF
--- a/.changeset/heavy-bulldogs-sit.md
+++ b/.changeset/heavy-bulldogs-sit.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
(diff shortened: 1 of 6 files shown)


dbanksdesign (approved):
:shipit: 

you the real mvp

PR #721, opened 2021-11-16
ForwardRef support - Simple Inputs

reesscot:
*Description of changes:*
This PR adds forward ref support for the simple input related primitives (ones that directly render View) by wrapping them in `React.forwardRef`. I'm limiting the number of primitives changed per PR to make them easier to review.

```jsx
  
  // ref.current.focus()
  
  return (
    <Input ref={ref} />
  );
};
```

### DIFF
--- a/packages/react/src/primitives/FieldGroupIcon/FieldGroupIcon.tsx
+++ b/packages/react/src/primitives/FieldGroupIcon/FieldGroupIcon.tsx
@@ -1,19 +1,27 @@
+import * as React from 'react';
 import classNames from 'classnames';
 
 import { ComponentClassNames } from '../shared/constants';
-import { FieldGroupIconProps, Primitive } from '../types';
+import { FieldGroupIconProps, PrimitiveWithForwardRef } from '../types';
 import { View } from '../View';
 
-export const FieldGroupIcon: Primitive<FieldGroupIconProps, 'button'> = ({
-  className,
-  children,
-  isVisible = true,
-  excludeFromTabOrder = false,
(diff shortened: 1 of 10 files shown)


packages/react/src/primitives/FieldGroupIcon/FieldGroupIcon.tsx
```diff
@@ -1,19 +1,27 @@
...
-export const FieldGroupIcon: Primitive<FieldGroupIconProps, 'button'> = ({
-  className,
-  children,
-  isVisible = true,
-  excludeFromTabOrder = false,
-  ...rest
-}) => {
+const FieldGroupIconInner: PrimitiveWithForwardRef<
+  FieldGroupIconProps,
+  'button' | 'div'
```
dbanksdesign, line 10: What's the reason to change `'button'` to `'button' | 'div'`?

dbanksdesign (approved):
LGTM! One question about a change in the types, nothing blocking.

packages/react/src/primitives/FieldGroupIcon/FieldGroupIcon.tsx
reesscot: This one was tricky to type because the `FieldGroupIcon` can either be a `div` in the case of an non-interactive icon, or a `button` in the case of a clickable icon. Example here:
https://ui.docs.amplify.aws/components/textfield#inner-components-icons

Screen Shot 2021-11-15 at 11 12 15 PM

PR #731, opened 2021-11-16
ForwardRef support - textarea, checkbox, radio

reesscot:
*Description of changes:*
This PR adds forward ref support for the `TextArea`, `Checkbox`, `CheckboxField`, `Radio` and `RadioGroupField` primitives (ones that directly render View) by wrapping them in `React.forwardRef`. I'm limiting the number of primitives changed per PR to make them easier to review.

```jsx
  
  // Below returns INPUT because the ref is forwarded down to DOM element
  // (CheckboxField => Checkbox => View => <input>)
  // ref.current.nodeName 
  
  return (
    <CheckboxField ref={ref} />
  );
};
```

### DIFF
--- a/packages/react/src/primitives/Checkbox/Checkbox.tsx
+++ b/packages/react/src/primitives/Checkbox/Checkbox.tsx
@@ -1,3 +1,4 @@
+import * as React from 'react';
 import classNames from 'classnames';
 
 import { useCheckbox } from './useCheckbox';
@@ -7,24 +8,27 @@ import { Input } from '../Input';
 import { Text } from '../Text';
 import { VisuallyHidden } from '../VisuallyHidden';
 import { CheckboxProps } from '../types/checkbox';
-import { Primitive } from '../types/view';
+import { PrimitiveWithForwardRef } from '../types/view';
 import { splitPrimitiveProps } from '../shared/styleUtils';
 import { ComponentClassNames } from '../shared/constants';
 import { useTestId } from '../utils/testUtils';
 
-export const Checkbox: Primitive<CheckboxProps, 'input'> = ({
-  checked,
-  className,
-  defaultChecked,
-  hasError,
-  isDisabled,
-  label,
-  labelHidden,
-  onChange: onChangeProp,
-  size,
-  testId,
-  ..._rest
-}) => {
+const CheckboxInner: PrimitiveWithForwardRef<CheckboxProps, 'input'> = (
+  {
+    checked,
+    className,
+    defaultChecked,
+    hasError,
+    isDisabled,
+    label,
+    labelHidden,
+    onChange: onChangeProp,
+    size,
+    testId,
+    ..._rest
+  },
+  ref
+) => {
   const { baseStyleProps, flexContainerStyleProps, rest } =
     splitPrimitiveProps(_rest);
 
@@ -58,6 +62,7 @@ export const Checkbox: Primitive<CheckboxProps, 'input'> = ({
           onBlur={onBlur}
           onChange={onChange}
           onFocus={onFocus}
+          ref={ref}
           type="checkbox"
           {...rest}
         />
@@ -93,5 +98,7 @@ export const Checkbox: Primitive<CheckboxProps, 'input'> = ({
     </Flex>
   );
 };
+
+export const Checkbox = React.forwardRef(CheckboxInner);
 
 Checkbox.displayName = 'Checkbox';
--- a/packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx
+++ b/packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx
@@ -1,37 +1,43 @@
+import * as React from 'react';
 import classNames from 'classnames';
-import { useMemo } from 'react';
 
 import { RadioGroupContext, RadioGroupContextType } from './context';
 import { FieldErrorMessage, FieldDescription } from '../Field';
 import { Flex } from '../Flex';
 import { Label } from '../Label';
-import { RadioGroupFieldProps, Primitive } from '../types';
+import { RadioGroupFieldProps, PrimitiveWithForwardRef } from '../types';
 import { ComponentClassNames } from '../shared/constants';
 import { useStableId } from '../shared/utils';
 
 // Note: RadioGroupField doesn't extend the JSX.IntrinsicElements<'input'> types (instead extending 'typeof Flex')
 // because all rest props are passed to Flex container
-export const RadioGroupField: Primitive<RadioGroupFieldProps, typeof Flex> = ({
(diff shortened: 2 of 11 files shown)


dbanksdesign (approved):
LGTM! :shipit:

packages/react/src/primitives/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    expect(textarea.dataset['variation']).toBe('quiet');
+  });
+
+  it('can set defaultValue (uncontrolled)', async () => {
+    render(<TextArea defaultValue="test" />);
+
+    const textarea = (await screen.findByRole(
+      'textbox'
+    )) as HTMLTextAreaElement;
+    expect(textarea.value).toBe('test');
```
zchenwei, line 65:
```suggestion
    expect(textarea).toHaveValue('test');
```

packages/react/src/primitives/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  it('can set value (controlled component)', async () => {
+    // onChange added to silence console error
+    render(<TextArea value="test" onChange={() => {}} />);
+
+    const textarea = (await screen.findByRole(
+      'textbox'
+    )) as HTMLTextAreaElement;
+    expect(textarea.value).toBe('test');
```
zchenwei, line 75:
```suggestion
    expect(textarea).toHaveValue('test');
```

packages/react/src/primitives/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    expect(textarea).toHaveAttribute('readonly');
+    expect(textarea).toHaveAttribute('required');
+  });
+
+  it('should set size and variation data attributes', async () => {
+    render(<TextArea size="small" variation="quiet" />);
+
+    const textarea = await screen.findByRole('textbox');
+    expect(textarea.dataset['size']).toBe('small');
+    expect(textarea.dataset['variation']).toBe('quiet');
```
zchenwei, line 56:
```suggestion
    expect(textarea).toHaveAttribute('data-size', 'small');
    expect(textarea).toHaveAttribute('data-variation', 'quiet');
```

packages/react/src/primitives/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    expect(ref.current.nodeName).toBe('TEXTAREA');
+  });
+
+  it('should render the state attributes', async () => {
+    render(<TextArea size="small" hasError isDisabled isReadOnly isRequired />);
+
+    const textarea = await screen.findByRole('textbox');
+    expect(textarea).toHaveAttribute('disabled');
+    expect(textarea).toHaveAttribute('readonly');
+    expect(textarea).toHaveAttribute('required');
```
zchenwei, line 48: nit: Do you mind updating those to test against their real boolean values? This makes me feel a bit nervous anyway

packages/react/src/primitives/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    )) as HTMLTextAreaElement;
+    expect(textarea.value).toBe('test');
+  });
+
+  it('show add aria-invalid attribute to textarea when hasError', async () => {
+    render(<TextArea id="testField" hasError={true} />);
+    const textarea = (await screen.findByRole(
+      'textbox'
+    )) as HTMLTextAreaElement;
+    expect(textarea).toHaveAttribute('aria-invalid');
```
zchenwei, line 83:
```suggestion
    expect(textarea).toHaveAttribute('aria-invalid', 'true');
```

zchenwei (approved):
LGTM! No blockers that I see

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -7,24 +8,27 @@ import { Input } from '../Input';
...
-  hasError,
-  isDisabled,
-  label,
-  labelHidden,
-  onChange: onChangeProp,
-  size,
-  testId,
-  ..._rest
-}) => {
+const CheckboxInner: PrimitiveWithForwardRef<CheckboxProps, 'input'> = (
```
hvergara, line 16: non-related: if Checkbox is only being used internally, should be a Primitive? Maybe should be a simple React.FC component instead.

packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx
```diff
@@ -1,37 +1,43 @@
...
+    name,
+    size,
+    value,
+    ...rest
+  },
+  ref
+) => {
   const fieldId = useStableId(id);
-  const radioGroupContextValue: RadioGroupContextType = useMemo(
+  const radioGroupContextValue: RadioGroupContextType = React.useMemo(
```
hvergara, line 40: do we need to import the entire `React` package just to use `useMemo` and `forwardRef`? I'm ok with using this pattern on tests, but IMO components should only import what's going to be used.

packages/react/src/primitives/TextArea/__tests__/TextArea.test.tsx
hvergara: +1

packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx
reesscot: I prefer the explicitness of `React.useMemo` over `useMemo`. 🤷‍♂️ I think it makes it easier at a glance to know the function came from React. I also prefer the React namespace import over explicit imports because it means we don't have to continually update the import line whenever I use a new hook from React. I doubt if there is going to be bundle difference between the two import options, so it's mostly stylistic afaik.

packages/react/src/primitives/TextArea/__tests__/TextArea.test.tsx
reesscot: Sure, but when true, the underlying DOM attribute will be an empty string, not a boolean. Updated to specifically target empty string.

packages/react/src/primitives/Checkbox/Checkbox.tsx
reesscot: We still want to expose the intrinsic attributes from the underlying DOM element, so keeping it as Primitive seems better.

hvergara (approved):

PR #732, opened 2021-11-16
added build tests for docs

ErikCH:
*Issue #, if available:*
Add tests if docs fail to build


.github/workflows/tests.yml
```diff
@@ -116,3 +116,25 @@ jobs:
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: lts/*
+          cache: 'yarn'
+
+      - name: Install packages
+        run: yarn --frozen-lockfile
+
+      - name: Build docs package
+        run: yarn workspace docs build
```
ericclemmons, line 140: `yarn docs build` also works, just like `yarn vue build`

dbanksdesign:
that build error is super not helpful. Also should we *not* be using `useLayoutEffect` because it is clogging up the logs with warnings?

ericclemmons:
@dbanksdesign The top part seems to be the culprit:

```
./src/pages/components/togglebutton/examples.tsx
Attempted import error: 'IconFormatColorFill' is not exported from '@aws-amplify/ui-react' (imported as 'IconFormatColorFill').
info  - Collecting page data...

./src/pages/index.page.tsx
Attempted import error: 'IconCopy' is not exported from '@aws-amplify/ui-react' (imported as 'IconCopy').

./src/pages/components/togglebutton/examples.tsx
Attempted import error: 'IconFormatColorFill' is not exported from '@aws-amplify/ui-react' (imported as 'IconFormatColorFill').

./src/pages/index.page.tsx
Attempted import error: 'IconCopy' is not exported from '@aws-amplify/ui-react' (imported as 'IconCopy').

```

As for `useLayoutEffect`, I recall using it because a component had to wait for the DOM to update before firing the effect.

We can stub it via https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85#gistcomment-3937704 or revisit the usage to clean up the noise.

ericclemmons (approved):

PR #734, opened 2021-11-16
Fixing icon issues with docs build and react build

dbanksdesign:
*Issue #, if available:*

*Description of changes:*

* No longer running build:icons as part of react build since it will override prettier's changes
* Fixing broken icons in docs, should unblock #732

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -26,7 +26,7 @@
   ],
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "yarn run build:icons && yarn run build:ts && yarn run build:catalog",
+    "build": "yarn run build:ts && yarn run build:catalog",
     "build:ts": "tsup --minify --keep-names",
     "build:catalog": "ts-node scripts/generatePrimitivesCatalog.ts",
     "dev": "tsup --watch src",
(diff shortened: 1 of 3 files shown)


ericclemmons (approved):

packages/react/package.json
```diff
@@ -26,7 +26,7 @@
   ],
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "yarn run build:icons && yarn run build:ts && yarn run build:catalog",
+    "build": "yarn run build:ts && yarn run build:catalog",
```
ericclemmons, line 29: When will icons be built as part of CI or local development on a clean repo?

reesscot (approved):

packages/react/package.json
dbanksdesign: Right now it is generating files that are checked into git so we only need to run it when we need the icons to change. A dev won't need to run this during local development. If we wanted, we could have it directly generate finalized build artifacts instead, then we would add this back.

ericclemmons: Ooh! Good call.  Glad to have this removed then 👍

Issue #714, opened 2021-11-15
Authenticator@next: When loginMechanism set to email, the forget password placeholder should be email.


artidata:
Currently the placeholder is written as username, though putting down email address will still progress. 
https://user-images.githubusercontent.com/20266094/141839509-0e93e49b-3742-43d5-948d-be42a66fab9a.mp4

ErikCH:
Hi @artidata ! I'm working on a fix for this now, so it will show correctly, instead of just `username` for all types of inputs.

ericclemmons:
Fixed via #738 🎉

PR #738, opened 2021-11-17
Fix forgot password placeholder

ErikCH:
*Issue #, if available:*
#714 

*Description of changes:*
Updated placeholder label so it shows the correct `email/phone number/username` label on the reset password page. 

image

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/reset-password/reset-password.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/reset-password/reset-password.component.ts
@@ -1,20 +1,27 @@
-import { Component, HostBinding, Input } from '@angular/core';
+import { Component, HostBinding, Input, OnInit } from '@angular/core';
 import { AuthenticatorService } from '../../../../services/authenticator.service';
-import { translate } from '@aws-amplify/ui';
+import { getAliasInfoFromContext, translate } from '@aws-amplify/ui';
 
 @Component({
   selector: 'amplify-reset-password',
   templateUrl: './reset-password.component.html',
 })
-export class ResetPasswordComponent {
+export class ResetPasswordComponent implements OnInit {
   @HostBinding('attr.data-amplify-authenticator-resetPassword') dataAttr = '';
   @Input() public headerText = translate('Reset your password');
 
   // translated texts
   public sendCodeText = translate('Send Code');
   public backToSignInText = translate('Back to Sign In');
+  public labelText = translate<string>('Username');
 
   constructor(public authenticator: AuthenticatorService) {}
+
+  ngOnInit(): void {
+    const { authState } = this.authenticator;
+    const { label } = getAliasInfoFromContext(authState.context);
+    this.labelText = `Enter your ${label.toLowerCase()}`;
+  }
 
   public get context() {
     const { updateForm, toSignIn, submitForm, error } = this.authenticator;
--- a/packages/ui/src/i18n/dictionaries/en.ts
+++ b/packages/ui/src/i18n/dictionaries/en.ts
@@ -16,6 +16,8 @@ export const enDict = {
   Email: 'Email',
   'Enter your code': 'Enter your code',
   'Enter your username': 'Enter your username',
+  'Enter your phone number': 'Enter your phone number',
+  'Enter your email': 'Enter your email',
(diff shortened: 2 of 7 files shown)


packages/vue/src/components/reset-password.vue
```diff
@@ -1,3 +1,64 @@
...
+import { useAuthenticator } from '../composables/useAuth';
+
+const attrs = useAttrs();
+const emit = defineEmits(['resetPasswordSubmit', 'backToSignInClicked']);
+
+const { state, send } = useAuthenticator();
+const { error, isPending } = toRefs(useAuthenticator());
+
+const { label } = getAliasInfoFromContext(state.context);
+const labelText = `Enter your ${label.toLowerCase()}`;
```
ErikCH, line 18: I updated this to use `useAuthenticator` instead of `useAuth`.

packages/ui/src/i18n/dictionaries/en.ts
```diff
@@ -16,6 +16,8 @@ export const enDict = {
   Email: 'Email',
   'Enter your code': 'Enter your code',
   'Enter your username': 'Enter your username',
+  'Enter your phone number': 'Enter your phone number',
+  'Enter your email': 'Enter your email',
```
ErikCH, line 20: We don't have translations for this for other languages yet...

packages/vue/src/composables/useAuth.ts
```diff
@@ -15,20 +15,23 @@ export const useAuth = (serv?: InterpretService) => {
...
-      for (const key of Object.keys(facade)) {
-        //@ts-ignore
-        useAuthenticatorValue[key] = facadeValues[key];
-      }
-      useAuthenticatorValue.send = send;
-      useAuthenticatorValue.state = state;
-    });
+  createValues();
+  watchEffect(() => {
+    createValues();
```
ErikCH, line 20: I might refactor this again in the future, but this seems to work better, then surrounding in an an onMounted...

packages/ui/src/i18n/dictionaries/en.ts
wlee221: I think that's fine, we need to rely on contributors for translations. We can start asking Discord community soon!

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/reset-password/reset-password.component.ts
```diff
@@ -1,21 +1,28 @@
...
   public sendCodeText = translate('Send Code');
   public backToSignInText = translate('Back to Sign In');
+  public labelText = translate<string>('Username');
 
   constructor(public authenticator: AuthenticatorService) {}
 
+  ngOnInit(): void {
+    const { authState } = this.authenticator;
+    const { label } = getAliasInfoFromContext(authState.context);
+    this.labelText = `Enter your ${label.toLowerCase()}`;
```
wlee221, line 23: Just double checking, should the labelText be just 

```
"Username" | "Email" | "Phone"
```

or 

```ts
`Enter your {$label.toLowerCase)`
```?

I believe the original issue was that the label showed "username". My worry with this approach is that "`Enter your ${label.toLowerCase()}`" would be inconsistent with other auth field labels we have.

ErikCH: This seems to be an inconsistency between Vue, React and Angular. React and Vue had 'Enter your username', while Angular had just 'Username'. So I fixed Angular so it matched Vue and React.

packages/ui/src/i18n/dictionaries/en.ts
ErikCH: 👍

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/reset-password/reset-password.component.ts
ErikCH: I think the slight inconsistency between the different fields is fine. I like 'Enter your...' text.

wlee221: Ah got it! In that case I'm good.

wlee221 (approved):

PR #741, opened 2021-11-17
updating alert style

dbanksdesign:
*Issue #, if available:*

*Description of changes:*

* Added font color tokens to match badge tokens
* Added some BEM classnames to alert innards 
* Added `color: inherit` so the dismiss button would look a bit cleaner
* Added an icon size token and removed the icon size prop (since we don't support size on `<Icon>` anymore)
* Removed `headingLevel` because it didn't really make sense semantically

https://user-images.githubusercontent.com/321279/142162014-6eae0ea2-0d74-4199-9e7a-3061319d0521.mp4

### DIFF
--- a/packages/react/src/primitives/Alert/Alert.tsx
+++ b/packages/react/src/primitives/Alert/Alert.tsx
@@ -20,7 +20,6 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
   gap,
   hasIcon = true,
   heading,
-  iconSize,
   isDismissible = false,
   justifyContent = 'space-between',
   onDismiss,
@@ -64,7 +63,7 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
             className={ComponentClassNames.AlertDismiss}
             onClick={dismissAlert}
           >
-            <IconClose size={iconSize} />
+            <IconClose />
           </Button>
         )}
       </Flex>
--- a/packages/react/src/primitives/Alert/Alert.tsx
+++ b/packages/react/src/primitives/Alert/Alert.tsx
@@ -1,10 +1,11 @@
-import { useState, useCallback } from 'react';
-import { ComponentClassNames } from '../shared/constants';
+import * as React from 'react';
 import classNames from 'classnames';
+
+import { ComponentClassNames } from '../shared/constants';
 import { AlertProps, Primitive } from '../types';
 import { View } from '../View';
 import { Flex } from '../Flex';
-import { Heading } from '../Heading';
+import { Text } from '../Text';
 import { Button } from '../Button';
 import { AlertIcon } from './AlertIcon';
 import { IconClose } from '../Icon';
@@ -19,7 +20,6 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
   gap,
   hasIcon = true,
   heading,
-  headingLevel,
   iconSize,
   isDismissible = false,
   justifyContent = 'space-between',
@@ -28,9 +28,9 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
   wrap,
   ...rest
 }) => {
-  const [dismissed, setDismissed] = useState<boolean>(false);
+  const [dismissed, setDismissed] = React.useState<boolean>(false);
 
-  const dismissAlert = useCallback(() => {
+  const dismissAlert = React.useCallback(() => {
     setDismissed(!dismissed);
 
     if (isFunction(onDismiss)) {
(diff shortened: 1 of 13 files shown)


packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -51,22 +51,19 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
         wrap={wrap}
         {...rest}
       >
-        <Flex alignItems="center">
```
dbanksdesign, line 54: Removed this extra `Flex` since it was unnecessary.

packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -51,22 +51,19 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
...
-                {heading}
-              </Heading>
-            )}
-            {children}
-          </View>
-        </Flex>
+        {hasIcon && <AlertIcon variation={variation} />}
+        <View role="alert" flex="1">
+          {heading && (
+            <View className={ComponentClassNames.AlertHeading}>{heading}</View>
```
dbanksdesign, line 57: heading is now just a `View` with a className because semantically it does not need a specific heading level. An alert can appear anywhere in the document.

packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -51,22 +51,19 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
...
+          )}
+          <View className={ComponentClassNames.AlertBody}>{children}</View>
```
dbanksdesign, line 59: Added a wrapper around children so it could be targeted with styles if needed.

packages/react/src/primitives/Alert/AlertIcon.tsx
```diff
@@ -1,26 +1,23 @@
 import * as React from 'react';
 
-import { AlertVariations, IconSize } from '../types';
+import { ComponentClassNames } from '../shared/constants';
+import { AlertVariations } from '../types';
 import { IconInfo, IconError, IconWarning, IconCheckCircle } from '../Icon';
 
 interface AlertIconProps {
   variation: AlertVariations;
-  iconSize: IconSize;
```
dbanksdesign, line 8: Removed this since Icon no longer supports it.

packages/ui/src/theme/css/component/heading.scss
```diff
@@ -1,6 +1,5 @@
 .amplify-heading {
   color: var(--amplify-components-heading-color);
-  display: block;
```
dbanksdesign, line 3: Random dupe style, figured might as well fix.

packages/ui/src/theme/tokens/components/alert.js
```diff
@@ -1,23 +1,37 @@
 module.exports = {
   // Default styles
+  color: { value: '{colors.font.primary.value}' },
   backgroundColor: { value: '{colors.background.tertiary.value}' },
-  paddingVertical: { value: '{space.small.value}' },
-  paddingHorizontal: { value: '{space.medium.value}' },
+  paddingBlock: { value: '{space.small.value}' },
```
dbanksdesign, line 5: Changing these to logical names: https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block

packages/react/src/primitives/Alert/Alert.tsx
reesscot: 💯

reesscot (approved):
LGTM 🚢

packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -51,22 +51,19 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
...
+        </View>
         {isDismissible && (
-          <Button variation="link" onClick={dismissAlert}>
+          <Button
+            variation="link"
+            className={ComponentClassNames.AlertDismiss}
+            onClick={dismissAlert}
+          >
             <IconClose size={iconSize} />
```
zchenwei, line 67: looks like `size` prop could be removed now?

packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -19,7 +20,6 @@ export const Alert: Primitive<AlertProps, typeof Flex> = ({
   gap,
   hasIcon = true,
   heading,
-  headingLevel,
   iconSize,
```
zchenwei, line 23: `iconSize` could be removed?

packages/react/src/primitives/Alert/__tests__/Alert.test.tsx
```diff
@@ -44,10 +44,13 @@ describe('Alert: ', () => {
...
-    const alert = await screen.findByText('Test heading');
-    expect(alert.nodeName).toBe('H6');
+    const alertHeading = await screen.findByText('Test heading');
+    expect(alertHeading.nodeName).toBe('DIV');
     expect(
-      alert.parentElement.parentElement.parentElement.classList.contains(
+      alertHeading.classList.contains(ComponentClassNames.AlertHeading)
+    ).toBe(true);
+    expect(
+      alertHeading.parentElement.parentElement.classList.contains(
```
zchenwei, line 53: nit: These tests look brittle. We could find some time to refactor it by querying elements in a better manner

packages/ui/src/theme/css/component/alert.scss
```diff
@@ -1,25 +1,45 @@
...
   }
 }
 
-.amplify-alert-heading {
+.amplify-alert__icon {
+  font-size: var(--amplify-components-alert-icon-size);
+}
+
+.amplify-alert__heading {
+  color: inherit;
```
zchenwei, line 32: I was under the impression that `color` prop is inherited by default. It seems that we have set `color` on `Heading` and was wondering if we should remove that and let the color inherit itself on `Heading`

zchenwei (commented):
Left a few comments. I like these improvement! 👍

zchenwei (approved):
No blockers that I see

packages/ui/src/theme/css/component/alert.scss
dbanksdesign: You are correct, I actually added this in before I changed the heading from a `<Heading>` to a `<View>`. This style was needed if we used a Heading, but not if we used a View.

PR #742, opened 2021-11-17
chore(angular): make authenticator service variable public

wlee221:
*Issue #, if available:*

*Description of changes:* Currently, `private authenticator: AuthenticatorService` is marked as `private` in `custom-slots` example. This should be marked as public because they are being used in the templates. Technically the samples still worked because transpiled JS doesn't actually have private variables (yet), but it's a good practice.

### DIFF
--- a/examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.ts
+++ b/examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.ts
@@ -9,7 +9,7 @@ import awsExports from '@environments/auth-with-federated/src/aws-exports';
   templateUrl: 'custom-slots.component.html',
 })
 export class CustomSlotsComponent {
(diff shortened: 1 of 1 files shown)


examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.ts
```diff
@@ -9,7 +9,7 @@ import awsExports from '@environments/auth-with-federated/src/aws-exports';
   templateUrl: 'custom-slots.component.html',
 })
 export class CustomSlotsComponent {
-  constructor(private authenticator: AuthenticatorService) {
```
ErikCH, line 12: Good catch! You are right should be public!

ErikCH (approved):
Looks good!

Milan-Shah (approved):

ericclemmons (approved):

PR #750, opened 2021-11-17
Remove export of POC primitives

reesscot:
*Description of changes:*
This PR removes the unused POC primitives from the public API.


dbanksdesign (approved):
LGTM! :shipit:

packages/react/src/primitives/index.tsx
```diff
@@ -4,46 +4,6 @@ export * from './hooks';
...
-
-export function Footer(props) {
-  return <footer data-amplify-footer="" {...props} />;
-}
-
-export function Main(props) {
-  return <main data-amplify-main="" {...props} />;
-}
-
 export function Form(props) {
```
zchenwei, line 7: Why do we still keep this?

reesscot: Because it's still being used in many of the Authenticator components. I'd prefer to remove it in a separate PR.

```bash
> src/components/Authenticator/SignUp/SignUp.tsx:4:23: error: No matching export in "src/index.tsx" for import "Form"
    4 │ import { Button, Flex, Form, View } from '../../..';
      ╵                        ~~~~
etc about 10 more times
```

zchenwei (approved):
LGTM

ericclemmons (approved):
Thanks for cleaning this up!

PR #665, opened 2021-11-10
Document how to access and manage `authState`

wlee221:
*Issue #, if available:* Depends on #637, closes #254, closes #664, related to #640.

*Description of changes:* Basically documenting how to do https://docs.amplify.aws/ui/auth/authenticator/q/framework/react/#recommended-usage on `@next`. In draft.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #753, opened 2021-11-17
feat(angular): expose `authenticator.*` from slot contexts.

wlee221:
*Issue #, if available:*

*Description of changes:*

- [x] Expose `authenticator.*` context from each slot

Now, all slots will emit `useAuthenticator` facade helpers from each slot. That way, we don't have to decide which contexts are needed and which ones are needed for each slot.

- [x] Expose "default" context from each slot

Previously, you had to use 

```html
<ng-template amplifySlot="sign-in-footer" let-toResetPassword="toResetPassword">
  <button (click)="toResetPassword()">Forgot Password?</button>
</ng-template>
```

syntax to use context facades. Now slots provide shorthand for customers to use the top level `authenticator` directly through `let-authenticator`:

```html
<ng-template amplifySlot="sign-in-footer" let-authenticator>
  <button (click)="authenticator.toResetPassword()">Forgot Password?</button>
</ng-template>
```

- [ ] TODO: Look into if this context can be strongly typed.


packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -151,6 +152,19 @@ export class AuthenticatorService implements OnDestroy {
...
+  /** @deprecated For internal use only */
+  public get slotContext() {
+    const slotContext = {
+      ...this._facade,
+      ...this._sendEventAliases,
+    };
+
+    return {
+      ...slotContext,
+      $implicit: { ...slotContext },
```
wlee221, line 164: This `$implicit` is what allows for `let-authenticator` shorthand syntax.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-verify-user/amplify-confirm-verify-user.component.html
```diff
@@ -6,7 +6,7 @@
       data-amplify-fieldset
       [disabled]="authenticator.isPending"
     >
-      <amplify-slot name="confirm-verify-user-header">
+      <amplify-slot name="confirm-verify-user-header" [context]="context">
```
ErikCH, line 9: How does Angular make it so deeply nested slots like this, is available to the `authenticator` component? In vue this slot would only be available in the confirm verify user component.

examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.html
```diff
@@ -18,7 +18,7 @@
     </h3>
   </ng-template>
 
-  <ng-template amplifySlot="sign-in-footer">
+  <ng-template amplifySlot="sign-in-footer" let-authenticator>
```
ErikCH, line 21: I like this, but this doesn't match the dx of React and Vue. At least for the examples. In React and Vue we receive the `useAuthenticator/authenticator` function by importing it in, but maybe it's OK doing it either way? 

Is it better dx to expose it through `let-*` or importing the service in is the question? Or maybe both is fine? @ericclemmons and I were discussing this yesterday. I don't have a strong opinion either way. If this is the way we want to go though, should I then expose authenticator through Vue's scoped slots too...

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-verify-user/amplify-confirm-verify-user.component.html
wlee221: Good question, I used `CustomComponentsService` to transfer those slots into deeply nested slots.


1. All `amplifySlot` templates passed into top-level `<amplify-authenticator` are provided to `CustomComponentsService`

https://github.com/aws-amplify/amplify-ui/blob/ca1b0b8bb686eaf61e5ba5899b5f3e2e19c73ad9/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts#L74-L76

2. `amplify-slot` component, wherever it's placed, uses that service to fetch the `ng-template` and renders it:

https://github.com/aws-amplify/amplify-ui/blob/ca1b0b8bb686eaf61e5ba5899b5f3e2e19c73ad9/packages/angular/projects/ui-angular/src/lib/utilities/amplify-slot/amplify-slot.component.ts#L25-L33

so this is agnostic of where it's placed.

examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.html
wlee221: Good point. For Angular, I do really enjoy this DX, though I see that other framework used `useAuthenticator` syntax. Maybe it's okay for Angular as it's more template-oriented than Vue? 

My idea is that template variables should be first class citizen, then AuthenticatorServices.

wlee221: For now, I think I want to leave it at `let-authenticator`. After #665 I can look to document `AuthenticatorService` better.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-verify-user/amplify-confirm-verify-user.component.html
ErikCH: 👍

examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.html
ErikCH: OK, 👍

examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.ts
```diff
@@ -9,7 +9,7 @@ import awsExports from '@environments/auth-with-federated/src/aws-exports';
   templateUrl: 'custom-slots.component.html',
 })
 export class CustomSlotsComponent {
-  constructor() {
+  constructor(private authenticator: AuthenticatorService) {
```
ErikCH, line 12: Want this back to public?

ErikCH: I'll approve and you can make that change

ErikCH (approved):
👍 after fix from private/public

examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.ts
wlee221: Oh right, good catch!

wlee221:
```suggestion
  constructor(public authenticator: AuthenticatorService) {
```

PR #755, opened 2021-11-17
Remove POC Form primitive

reesscot:
*Description of changes:*
Removes the `Form` primitive and replaces it with `form` in all the places it was used in Authenticator.

TODO:

- [x] Test locally in next-example apps for style regressions

### DIFF
--- a/packages/react/src/helpers/utils.ts
+++ b/packages/react/src/helpers/utils.ts
@@ -1,7 +1,14 @@
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
 
-export const isInputEventTarget = (
+export const isInputOrSelectElement = (
   target: unknown
-): target is HTMLInputElement => {
+): target is HTMLInputElement | HTMLSelectElement => {
+  return (
+    (target as HTMLInputElement)?.nodeName === 'INPUT' ||
+    (target as HTMLSelectElement)?.nodeName === 'SELECT'
+  );
+};
+
+export const isInputElement = (target: unknown): target is HTMLInputElement => {
   return (target as HTMLInputElement)?.nodeName === 'INPUT';
 };
--- a/packages/react/src/helpers/utils.ts
+++ b/packages/react/src/helpers/utils.ts
@@ -1,5 +1,7 @@
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
 
-export const isInputTarget = (target: unknown): target is HTMLInputElement => {
+export const isInputEventTarget = (
+  target: unknown
+): target is HTMLInputElement => {
   return (target as HTMLInputElement)?.nodeName === 'INPUT';
 };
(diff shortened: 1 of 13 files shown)


packages/react/src/helpers/utils.ts
```diff
@@ -1 +1,5 @@
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
+
+export const isInputTarget = (target: unknown): target is HTMLInputElement => {
```
hvergara, line 3: you are going to hate me, but: can we rename it to `isInputEventTarget`? current name seems too generic IMO

hvergara (changes requested):

packages/react/src/helpers/utils.ts
reesscot: No problem, updated

packages/react/src/helpers/utils.ts
```diff
@@ -1 +1,7 @@
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
+
+export const isInputEventTarget = (
+  target: unknown
+): target is HTMLInputElement => {
+  return (target as HTMLInputElement)?.nodeName === 'INPUT';
+};
```
ericclemmons, line 7: Was there a bug or type issue?  I'm curious what you saw that we may have missed so far...

ericclemmons (approved):
Thanks for doing this!

Nothing blocking, but curious what prompted the creation of `isInputEventTarget`

packages/react/src/helpers/utils.ts
reesscot: Yes, changing from `Form` to `form` exposed that the `handleChange` function had the wrong type on its `event` parameter, which caused type errors on `form.onChange` and when trying to de-structure the `event.target`. The form onchange event doesn't know what type of element fired the change event, so the `isInputEventTarget` typeguard ensures it was an input element.

```diff
-  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+ const handleChange = (event: React.FormEvent<HTMLFormElement>) => {
```

packages/react/src/helpers/utils.ts
```diff
@@ -1 +1,14 @@
...
+): target is HTMLInputElement | HTMLSelectElement => {
+  return (
+    (target as HTMLInputElement)?.nodeName === 'INPUT' ||
+    (target as HTMLSelectElement)?.nodeName === 'SELECT'
+  );
+};
+
+export const isInputElement = (target: unknown): target is HTMLInputElement => {
+  return (target as HTMLInputElement)?.nodeName === 'INPUT';
+};
```
slaymance, line 14: Is it worthwhile to break out both input and select checks into separate utils? Could we use them in the future?

```typescript
export const isInputElement = (target: unknown): target is HTMLInputElement =>
  (target as HTMLInputElement)?.nodeName === 'INPUT';

export const isSelectElement = (target: unknown): target is HTMLSelectElement =>
  (target as HTMLSelectElement)?.nodeName === 'SELECT';

export const isInputOrSelectElement = overSome([isInputElement, isSelectElement]);
```

slaymance (approved):
Looks good to me!

packages/react/src/helpers/utils.ts
reesscot: Split them into separate functions

hvergara (approved):

PR #759, opened 2021-11-17
Updating button style and removing docs overrides

dbanksdesign:
*Issue #, if available:*

*Description of changes:*

* Removing docs styling overrides of button so using the button in an app will look the same as in the docs
* Added some more button theme design tokens for users to target
* Updated the border color of the default button to match field controls
* Added box shadows to focus states of all variations of button
* Updated the docs a bit, still need to move examples to examples folder though

CleanShot 2021-11-17 at 14 31 45@2x

https://user-images.githubusercontent.com/321279/142293721-1a3714a8-d14d-4c85-896c-3f752f783628.mp4


docs/src/pages/components/button/demo.tsx
```diff
@@ -1,114 +1,166 @@
...
+      </SelectField>
+
+      <SelectField
+        name="Size"
+        id="size"
+        label="Size"
+        value={props.size}
+        onChange={(event) => props.setSize(event.target.value as ButtonSizes)}
+      >
+        <option value="">default</option>
```
sammartinez, line 60: Just caught this from the video, so not sure if we needed to Uppercase for consistency 

```suggestion
        <option value="">Default</option>
```

ericclemmons (approved):
This looks much more consistent!

docs/src/pages/components/button/demo.tsx
```diff
@@ -1,114 +1,166 @@
...
+  SelectField,
+  TextField,
+} from '@aws-amplify/ui-react';
+
+import { Demo } from '@/components/Demo';
+
+const propsToCode = (props) => {
+  return (
+    `<Button` +
+    (props.disabled ? `\n  isDisabled={${props.disabled}}` : '') +
```
reesscot, line 19: Will we need to keep this in sync with [152-163](https://github.com/aws-amplify/amplify-ui/pull/759/files#diff-f85e0a9930f64b27b62d72172a82b9d915145c2ac8db04e94c178e0664e3cc8cR152-R163) below? Feels a bit awkward

docs/src/pages/components/button/demo.tsx
```diff
@@ -1,114 +1,166 @@
...
+      <SwitchField
+        label="isLoading"
+        defaultChecked={props.loading}
+        labelPosition="end"
+        onChange={(event) => {
+          props.setLoading(event.target.checked as ButtonProps['isLoading']);
+        }}
+      />
+
+      {/* <TextField
```
reesscot, line 94: Should we remove this commented out code?

reesscot (approved):
LGTM, just a few comments

docs/src/styles/docs/example.scss
```diff
@@ -28,7 +28,11 @@
 .docs-component-demo {
   width: 100%;
   margin-bottom: var(--amplify-space-large);
-
+  
+  & pre[class*=language-] {
+    max-height: none;
```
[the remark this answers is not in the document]
dbanksdesign: Yup this will only affect the code snippet in the demo block at the top of primitive pages, but there might be a better solution. The issue was they weren't stretching:

CleanShot 2021-11-18 at 09 33 14@2x

I could solve that another way though

docs/src/pages/components/button/demo.tsx
dbanksdesign: Yea I wonder if there is a better way to do this...

zchenwei (approved):
LGTM

PR #760, opened 2021-11-17
Docs: Update navigation so query params are passed over

ErikCH:
*Description of changes:*

When navigating the docs site if you change to `Vue` or `Angular` the site forgets these selections when moving around the site.

My fix will always pass the current platform query params to every `NavLink`.

- [x] Update `NavLink` to accept query
- [x] Fix bug when clicking external link that wasn't opening.

https://user-images.githubusercontent.com/65630/142316030-66397f57-4441-44b1-996e-16f594686ba8.mp4

### DIFF
--- a/docs/src/components/Layout/Header.tsx
+++ b/docs/src/components/Layout/Header.tsx
@@ -20,19 +20,19 @@ import { Logo } from '@/components/Logo';
 import { FrameworkChooser } from './FrameworkChooser';
 
 const NavLink = ({ href, children, isExternal = false }) => {
-  const router = useRouter();
-  const isCurrent = router.pathname.startsWith(href);
+  const { pathname, query } = useRouter();
+  const isCurrent = pathname.startsWith(href);
   const className = `docs-nav-link ${isCurrent ? 'current' : ''}`;
 
   if (isExternal) {
     return (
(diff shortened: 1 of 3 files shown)


docs/src/components/Layout/Header.tsx
```diff
@@ -20,19 +20,19 @@ import { Logo } from '@/components/Logo';
 import { FrameworkChooser } from './FrameworkChooser';
 
 const NavLink = ({ href, children, isExternal = false }) => {
-  const router = useRouter();
-  const isCurrent = router.pathname.startsWith(href);
+  const { pathname, query } = useRouter();
+  const isCurrent = pathname.startsWith(href);
```
dbanksdesign, line 24: Could we have a check here to make sure `href` isn't just "/" so the logo doesn't have the current styling for every page since every pathname starts with "/"? Or maybe there is a better way?

dbanksdesign (commented):
Just one minor thing about the logo in the header

docs/src/components/Layout/Header.tsx
ErikCH: OK! Yes, I'll add that.

dbanksdesign (approved):
LGTM! :shipit:

PR #761, opened 2021-11-17
ForwardRef support for TextField

reesscot:
*Description of changes:*
This PR adds forward ref support for the `TextField` primitive by wrapping it in `React.forwardRef`. It also required some TS magic to infer the correct HTML type based on the `isMultiline` prop. 

```jsx
  
  // Below returns INPUT because the ref is forwarded down to DOM element (TextField => Input => <input>)
  // ref.current.nodeName
  
  return (
    <TextField label="name" ref={ref} />
  );
};
```

After this PR is merged I will be able to add forwardRef support to `PasswordField` and other `TextField` based primitives.


hvergara (approved):
LGTM

packages/react/src/primitives/TextField/TextField.tsx
```diff
@@ -110,4 +108,16 @@ export const TextField = <Multiline extends boolean>(
   );
 };
 
-TextField.displayName = 'TextField';
+export const TextField = React.forwardRef(TextFieldPrimitive) as <
```
zchenwei, line 111: This seems dissatisfying but don't have a good a idea on my mind

zchenwei (approved):
LGTM 🚢

packages/react/src/primitives/TextField/TextField.tsx
reesscot: Yeah, I tried a number of different solutions, but the other options created new problems. This was the only option that fully worked. Perhaps we can revisit it later.

reesscot: For posterity, here is one of the other solutions I tried, which involves redeclaring the forwardRef function definition.

```jsx
// Redeclare `React.forwardRef` to support higher order function type inference
// which only works with plain function types (default definition )
// see: https://fettblog.eu/typescript-react-generic-forward-refs/
//
// Note: Without this only shared props between InputProps and TextAreaProps will be available, a
// and any unique props such as `resize` on TextArea or HTMl attributes will be missing from the type.
declare module 'react' {
  function forwardRef<T, P = {}>(
    render: (props: P, ref: React.Ref<T>) => React.ReactElement | null
  ): (props: P & React.RefAttributes<T>) => React.ReactElement | null;
}
```

The problem with this approach is that it affected other modules causing them to not be able to have a `displayName`. 

Ref: https://fettblog.eu/typescript-react-generic-forward-refs/

PR #766, opened 2021-11-18
Fix primitives catalog generator

hvergara:
*Description of changes:*
Improves component props extraction algorithm, including `React.forwardRef` and generic function types (e.g `Collection` / `TextField`)


Milan-Shah (approved):

ericclemmons (approved):
This looks good!  But since it was a regression, we may want to consider a test to validate the result of this.

I don't know _exactly_ broke so I'm guessing a bit here :)

PR #770, opened 2021-11-18
Add missing useState import

ericclemmons:
*Issue #, if available:*

https://ui.docs.amplify.aws/components/grid fails with an error because `useState` is not defined.

*Description of changes:*


reesscot (approved):

reesscot:
My only question is how did this ever work before?

PR #774, opened 2021-11-18
Remove `usePagination` and `SharedText` from exports

reesscot:
*Description of changes:*
Removes internal `usePagination` hook and i18n `SharedText` from exports.

Note: This removes the `amplify-ui/packages/react/src/primitives/hooks.ts` file which only exported internal `usePagination` hook, but doesn't touch the `packages/react/src/hooks/index.ts` file which exports external hooks.


hvergara (approved):

ericclemmons:
This change broke the docs build:
> ./src/pages/HomePrimitivePreview.tsx
Attempted import error: 'usePagination' is not exported from '@aws-amplify/ui-react' (imported as 'usePagination').

See: https://github.com/aws-amplify/amplify-ui/runs/4255160369?check_suite_focus=true

PR #778, opened 2021-11-18
Add bottom-padding to sign-in and sign-up footer

wlee221:
*Issue #, if available:*

*Description of changes:* Adding default padding-bottom to `sign-in-footer`.

Also adjusted `data-amplify-*` structure slightly. Now the routed screen is labeled `data-amplify-router`. Each screen has `data-amplify-header` and `data-amplify-footer` that has default styles. 

Overriding those headers will also remove those styles.

## Vue

Vue1

Vue2

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.html
@@ -1,9 +1,7 @@
 <div data-amplify-authenticator [attr.data-variation]="variation">
   <div data-amplify-container>
-    <div data-amplify-header>
-      <amplify-slot name="header" [context]="context"></amplify-slot>
-    </div>
-    <div data-amplify-body>
+    <amplify-slot name="header" [context]="context"></amplify-slot>
+    <div data-amplify-router>
       <amplify-tabs
         (tabChange)="onTabChange()"
         *ngIf="route === 'signIn' || route === 'signUp'"
@@ -103,9 +101,7 @@
       </amplify-slot>
     </div>
 
-    <div data-amplify-footer>
-      <amplify-slot name="footer" [context]="context"></amplify-slot>
-    </div>
+    <amplify-slot name="footer" [context]="context"></amplify-slot>
   </div>
 </div>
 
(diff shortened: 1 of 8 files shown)


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.html
```diff
@@ -1,9 +1,7 @@
 <div data-amplify-authenticator [attr.data-variation]="variation">
   <div data-amplify-container>
-    <div data-amplify-header>
-      <amplify-slot name="header" [context]="context"></amplify-slot>
-    </div>
-    <div data-amplify-body>
+    <amplify-slot name="header" [context]="context"></amplify-slot>
+    <div data-amplify-router>
```
wlee221, line 4: No longer shipping default styles as discussed

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-in/sign-in.component.html
```diff
@@ -27,15 +27,17 @@
...
-      amplify-button
-      fontWeight="normal"
-      size="small"
-      variation="link"
-      fullWidth="true"
-      (click)="authenticator.toResetPassword()"
-    >
-      {{ forgotPasswordText }}
-    </button>
+    <div data-amplify-footer>
```
wlee221, line 30: We use default styles under the slot, so that these opinionated slots are gone if customers override it.

One cool thing they can do is to put `data-amplify-footer`, and they'll get back the styles!

packages/react/src/components/Authenticator/Router/index.tsx
```diff
@@ -49,7 +49,7 @@ export function Router({
         <View data-amplify-container="">
           <Header />
 
-          <View data-amplify-body="">
+          <View data-amplify-router="">
```
wlee221, line 52: Changed naming for clarity/declarativeness

ericclemmons (approved):

reesscot:
nit: what do you think of reducing the spacing above and below Forgot Password / Reset Password to match the padding above and below the `or` divider? It still feels like about 10px too much space (this isn't blocking)

reesscot (approved):

wlee221:
> nit: what do you think of reducing the spacing above and below Forgot Password / Reset Password to match the padding above and below the or divider? It still feels like about 10px too much space (this isn't blocking)

Agreed, let me add an issue for that, we have an inconsistent spacing general.

PR #782, opened 2021-11-18
ci: Enable publish to @latest

wlee221:
*Issue #, if available:*

*Description of changes:* Uncomments publish scripts for @latest 😄 

### DIFF
--- a/.github/workflows/publish-latest.yml
+++ b/.github/workflows/publish-latest.yml
@@ -25,7 +25,6 @@ jobs:
         id: changesets
         uses: changesets/action@b98cec97583b917ff1dc6179dd4d230d3e439894
         with:
-          version: yarn version:latest
           publish: yarn release:latest
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish-latest.yml
+++ b/.github/workflows/publish-latest.yml
@@ -25,7 +25,7 @@ jobs:
         id: changesets
         uses: changesets/action@b98cec97583b917ff1dc6179dd4d230d3e439894
         with:
-          publish: yarn release:latest
+          publish: yarn publish:latest
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-latest.yml
+++ b/.github/workflows/publish-latest.yml
@@ -21,9 +21,9 @@ jobs:
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@b98cec97583b917ff1dc6179dd4d230d3e439894
-        # with:
-        # version: yarn version:latest # commented out until we're ready for GA release
-        # publish: yarn release:latest # commented out until we're ready for GA release
+        with:
+          version: yarn version:latest
+          publish: yarn release:latest
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # NPM_TOKEN: ${{ secrets.NPM_TOKEN }} # commented out until we're ready for GA release
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-latest.yml
+++ b/.github/workflows/publish-latest.yml
@@ -11,6 +11,9 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2
+        with:
+          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
+          fetch-depth: 0
       - name: Setup Node.js LTS
         uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,10 +17,10 @@
     "build:icons": "yarn react build:icons",
     "test": "yarn react test && yarn e2e test",
     "prepare": "husky install",
-    "version:next": "yarn changeset version --snapshot next-$(git rev-parse --short=7 HEAD) && yarn build",
+    "version:next": "yarn changeset version --snapshot next-$(git rev-parse --short=7 HEAD) && yarn angular build",
```
ericclemmons, line 20: This improves build times because everything should already be built, `changeset version` updates versions, but `yarn angular build` _rebuilds_ with the correct version?

.github/workflows/publish-latest.yml
```diff
@@ -11,6 +11,9 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2
+        with:
+          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
+          fetch-depth: 0
```
ericclemmons, line 16: This is interesting.  I'm trying to understand what behavior prompted this.

My understand was that the release PR (https://github.com/aws-amplify/amplify-ui/pull/596) only needs `HEAD` since `.changesets` has all the files there and it's able to calculate it without a huge history.

Is there a comment or reference here? It seems like an easy thing to miss.

Also, is changeset re-calculating versions from the entire history? Or just from `HEAD`? 🤔

.github/workflows/publish-latest.yml
```diff
@@ -21,9 +24,9 @@ jobs:
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@b98cec97583b917ff1dc6179dd4d230d3e439894
-        # with:
-        # version: yarn version:latest # commented out until we're ready for GA release
-        # publish: yarn release:latest # commented out until we're ready for GA release
+        with:
+          version: yarn version:latest
```
ericclemmons, line 28: If this PR is merged, does this version to latest or publish to NPM? Or is this passing the commands for that PR to run once it's merged?

https://github.com/aws-amplify/amplify-ui/pull/596

ericclemmons (commented):
Comments to aid my understanding and really put my fears at ease :D

package.json
```diff
@@ -17,10 +17,10 @@
...
     "publish:next": "yarn changeset publish --tag next",
-    "version:latest": "echo \"not implemented until stable release\"",
-    "publish:latest": "echo \"not implemented until stable release\""
+    "version:latest": "yarn changeset version && yarn angular build",
```
wlee221, line 22: Actually, `&& yarn angular build` is not actually necessary, because `publish` action runs on the main branch after the new versions have been merged. So the first build (chained off `yarn install`) would have already captured the new version.
```suggestion
    "version:latest": "yarn changeset version",
```

package.json
wlee221: See my comment. below, it's actually unneeded. But yes, that was my original intention because other packages were already built from `yarn --frozen-lock`.

.github/workflows/publish-latest.yml
wlee221: To be perfectly honest, I just took this from https://github.com/changesets/action#with-publishing. And this actions haven't had problem so far with generating changelog on #596, so I'm not sure this is necessary.

.github/workflows/publish-latest.yml
wlee221: Good question, this was just to pass the command that `@changesets/action` can use... but this is not actually neede reading through their README: https://github.com/changesets/action#with-version-script

> If the version script is present, this action will run that script instead of changeset version, so please make sure that your script calls changeset version at some point.

I'll remove it from `yml`, but keep the version command in `package.json` just in case we need to publish manually.

package.json
wlee221: With discussion on https://github.com/aws-amplify/amplify-ui/pull/782#discussion_r752555234, I'll keep this command as-is for manual deploy purpose only.

`@changesets/action` will use its native versioning logic.

.github/workflows/publish-latest.yml
ericclemmons: Let's keep it since it's consistent with https://github.com/changesets/action#with-publishing.

.github/workflows/publish-latest.yml
```diff
@@ -21,9 +24,8 @@ jobs:
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@b98cec97583b917ff1dc6179dd4d230d3e439894
-        # with:
-        # version: yarn version:latest # commented out until we're ready for GA release
-        # publish: yarn release:latest # commented out until we're ready for GA release
+        with:
+          publish: yarn release:latest
```
ericclemmons, line 28: Sorry, I don't see a `release:latest` command?

Only `version:latest` and `publish:latest`

wlee221: Oh god, you're right. I need to take a breath 😄 
```suggestion
          publish: yarn publish:latest
```

ericclemmons (approved):

Milan-Shah (approved):

PR #783, opened 2021-11-18
React documentation prefers Authenticator over withAuthenticator

### LINKED ISSUE #640: Examples on new <Authenticator/> instead of the HOC withAuthenticator, migration docs.
As much as I like the new design and its extensive configuration, I will prefer the latest stable release approach with `AuthState` and `onAuthUIStateChange` as seen in this docs: https://docs.amplify.aws/ui/auth/authenticator/q/framework/react/#recommended-usage. 

Using the past  `<AmplifyAuthenticator/>` is more component composition friendly compared to the new HOC `withAuthenticator`. I  guess the new UI is not backward compatible as I have tried to simply replace `<AmplifyAuthenticator/>` with the new `<Authenticator/>`, but it doesn't work. Will you support the Authstate approach or withAuthenticator is the only future?

ericclemmons:
*Issue #, if available:* Closes #640

> Using the past <AmplifyAuthenticator/> is more component composition friendly compared to the new HOC withAuthenticator. I guess the new UI is not backward compatible as I have tried to simply replace <AmplifyAuthenticator/> with the new <Authenticator/>, but it doesn't work.

This will also make way for use cases like https://codesandbox.io/s/change-your-layout-rh4z9?file=/src/App.js with `useAuthenticator`.

### DIFF
--- a/docs/src/pages/components/authenticator/LabelsAndTextDemo.tsx
+++ b/docs/src/pages/components/authenticator/LabelsAndTextDemo.tsx
@@ -73,7 +73,7 @@ function Screen({ Component }: ScreenProps) {
 export function LabelsAndTextDemo({ Component }: ScreenProps) {
   return (
     <Authenticator.Provider>
-      <View data-amplify-authenticator="" data-variaton="modal">
+      <View data-amplify-authenticator="">
         <View data-amplify-container="">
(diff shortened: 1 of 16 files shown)


reesscot (approved):
LGTM

PR #785, opened 2021-11-18
Move hooks to internal module

hvergara:
*Description of changes:*
Move special hooks into a separate module: `internal`

`import { useAuth } from '@aws-amplify/ui-react` => `import { useAuth } from '@aws-amplify/ui-react/internal`


wlee221 (approved):
LGTM, I'm good to merge this once docs test passes

packages/react/src/hooks/useAuth.ts
```diff
@@ -15,28 +15,6 @@ export interface UseAuthResult {
   fetch?: () => void;
 }
 
-/**
- * React hook for Amplify Auth.
```
reesscot, line 19: Any reason for removing the useage comment? Seems useful.

reesscot (approved):
💯

packages/react/src/hooks/useAuth.ts
hvergara: API is not public

PR #786, opened 2021-11-18
Revert "Remove `usePagination` and `SharedText` from exports"

ericclemmons:
Reverts aws-amplify/amplify-ui#774


zchenwei (approved):
LGTM

packages/react/__tests__/exports.ts
```diff
@@ -1373,6 +1373,7 @@ describe('@aws-amplify/ui-react', () => {
           "ScrollView",
           "SearchField",
           "SelectField",
+          "SharedText",
```
ericclemmons, line 1376: @reesscot This should still be removed, right?

zchenwei: yea, that should be removed for sure

reesscot: Yup, I can add the removal back in followup pr

slaymance (approved):
Verified locally that this fixes the docs build issue.

PR #812, opened 2021-11-19
Use Text primitive for Button loading text

hvergara:
*Description of changes:*
Replace `<span>` component with a `<Text as="span">` primitive

### DIFF
--- a/packages/react/src/primitives/Button/Button.tsx
+++ b/packages/react/src/primitives/Button/Button.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 
 import { ComponentClassNames } from '../shared/constants';
 import { ButtonProps, PrimitiveWithForwardRef } from '../types';
+import { Text } from '../Text';
 import { View } from '../View';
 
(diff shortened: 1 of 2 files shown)


ericclemmons (approved):

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -37,7 +38,11 @@ const ButtonInner: PrimitiveWithForwardRef<ButtonProps, 'button'> = (
       type={type}
       {...rest}
     >
-      {isLoading && loadingText ? <span>{loadingText}</span> : children}
+      {isLoading && loadingText ? (
+        <Text as="span">{loadingText}</Text>
```
reesscot, line 42: Hehe, `Text` didn't even exist when this was written.

reesscot (approved):
:shipit:

packages/react/src/primitives/Button/Button.tsx
zchenwei: I was wondering if this `span` is really required. Looks like not styles applied on it

zchenwei (approved):

PR #819, opened 2021-11-21
Fixed 'RESET_PASSWORD' translation

strugman:
*Issue #, if available:*
None, I think

*Description of changes:*
I believe there's a copy & paste error here: `RESET_PASSWORD: 'Send code'` has a value of a next field.


wlee221 (approved):
LGTM, thanks!

packages/ui/src/i18n/translations.ts
```diff
@@ -68,7 +68,7 @@ export const DefaultTexts = {
   PROFILE: 'Profile',
   RESEND_CODE: 'Resend Code',
   RESET_PASSWORD_HEADING: 'Reset your password',
-  RESET_PASSWORD: 'Send Code',
```
ericclemmons: @wlee221 By changing `Send Code` to `Reset Password`, would that affect other translations like:

https://github.com/aws-amplify/amplify-ui/blob/9fd90d45b01a38ee6d78f0cb67238b210750115c/packages/ui/src/i18n/dictionaries/authenticator/fr.ts#L47

I agree that `RESET_PASSWORD` isn't aligned with the usage – the flaw with `DefaultTexts` is that it's too close to the translation vs. the _role_. For example, we don't differentiate text for actions vs. headings vs. labels vs. placeholders.  A `Create Account` tab has a different purpose than a `Create Account` button.

wlee221: To answer your first question -- no, it'll not affect other translations because as we already have an entry for 'Send Code' here https://github.com/aws-amplify/amplify-ui/blob/b9c4103a5364a15db7a94c3c2b2f61354c14e3a1/packages/ui/src/i18n/translations.ts#L72. I think this was more of a typo as @strugman mentioned.

wlee221: I'm not sure if we really need to distinguish roles though, because to `I18n`, strings with the same keys are translated the same way regardless of the roles. They can't really distinguish the roles, so I'm wondering if we should even just have list of translatable keys instead of this middle ground solution we have now.

PR #825, opened 2021-11-22
Refactoring i18n package to accommodate new UI modules in cleaner and moduler way

ppraja22:
*Adding new modules in i18n package seems cumbersome*

*Made changes in the existing i18n package to have translation files at folder level in-place of having it at parent (dictionaries) level. Ultimately for new UI modules , we will create new folder under parent(dictionaries) folder and put all the related translation files and default text to that folder*

_Current :-_
```
--dictionaries
   |
   -- All Translation file
```

_Post this PR :-_
```
--dictionaries
   |
   --authenticator
      |
      --all translation files and default texts related to authenticator
   --new XYZ UI module
      |
      --all translation files and default texts related to this new module
```

Tested by changing the translation values in existing component as well by adding and changing the values in new UI module's components (which is what is being targeted to accommodate from this PR)"


packages/ui/src/i18n/dictionaries/index.ts
```diff
@@ -1,7 +1,15 @@
...
-export { enDict } from './en';
-export { esDict } from './es';
-export { frDict } from './fr';
-export { itDict } from './it';
-export { jaDict } from './ja';
-export { zhDict } from './zh';
+import * as authenticatorDict from './authenticator';
+
+//merge all the new module translations in respective locale constants
+export const deDict = { ...authenticatorDict.deDict };
```
ericclemmons, line 4: Just adding this for my own understanding: this change is enabling this in the future:

```suggestion
export const deDict = {
  ...authenticatorDict.deDict,
  ...someOtherDict.deDict
};
```

packages/ui/src/i18n/translations.ts
```diff
@@ -33,63 +34,7 @@ import {
...
-  SIGNING_IN_BUTTON: 'Signing in',
-  SKIP: 'Skip',
-  SUBMIT: 'Submit',
-  SUBMITTING: 'Submitting',
-  VERIFY_CONTACT: 'Verify Contact',
-  VERIFY_HEADING: 'Account recovery requires verified contact information',
-  VERIFY: 'Verify',
-  WEBSITE: 'Website',
-} as const;
+export const DefaultTexts = { ...defaultTexts } as const;
```
ericclemmons, line 37: @wlee221 Are `DefaultTexts` an API we want exposed for customers?  I noticed that it was already exported, but we don't document it:
> https://ui.docs.amplify.aws/components/authenticator#internationalization-i18n

ericclemmons (approved):
This looks good, just making sure I didn't misunderstand anything!

packages/ui/src/i18n/translations.ts
wlee221: No, i don't think we found a strong reason to expose it yet. I found it more intuitive to just write the translatable string directly than having to look through these `DefaultTexts` map.

wlee221 (approved):

Issue #821, opened 2021-11-21
New Authenticator component does not support dialCode attribute?


gomesp:
I used to be able to define the "dialCode" in the previous ```AmplifyAuthenticator``` react component (via ```AmplifySignUp/formFields``` attribute). 

It seems that now I need to create a completely custom component if I wanted the same behaviour (i.e. using the ```PhoneNumberField``` component).

Is it possible to add the dialCode attribute as a value of the ```signUpAttributes``` instead to keep the implementations simple?

ericclemmons:
@gomesp You're looking to have it default to a different value (e.g. `+55`) rather than `+1` correct?

We just need to expose it with something like:

```js
<Authenticator
  initialFormValues={{
    countryCode: '+55',
    email: 'my@example.com'
  }}
>
```

The alternative is exposing a `SignUp.PhoneNumber` slot that could be replaced with https://ui.docs.amplify.aws/components/phonenumberfield#usage:

```js
<PhoneNumberField defaultCountryCode="+55" label="Phone Number" />;
```

Is either of these what you're looking for?

gomesp:
Thanks @ericclemmons, yeah, I was hoping there would be a way to setup the default country code as in your first example. 

If I understand the [documentation correctly](https://ui.docs.amplify.aws/components/authenticator#sign-up-fields), I could add the PhoneNumberField as a slot, but I thought that would be a small enough improvement to the out-of-the-box experience that would help users avoid having to customising the component if they don't need to.

ericclemmons:
@gomesp We haven't created a slot just for the phone number for exactly the reason you described.

Providing initial form values would be best, so #826 will help resolve this need for all field values 🙏

gomesp:
Thanks @ericclemmons, I'll subscribe to the PR and see how it progresses.

PR #826, opened 2021-11-22
Handle phone number parsing in separate xstate action

slaymance:
*Issue #, if available:* [Asana](https://app.asana.com/0/1200334286823427/1201022646170709)

*Description of changes:* This pull request refactors how `country_code` is handled by xstate when `phone_number` is a login mechanism.

An action called `parsePhoneNumber` is added which combines `country_code` and `phone_number` into a single value, parses the value, deletes `country_code` from `formValues` and updates `formValues` with the new phone number value. The action is run upon entry to the state where `signIn` or `signUp` form submission happens.

This addresses the phone number value as a `loginMechanism` for both `signIn` and `signUp`, as well as when phone number is a sign up attribute.


ericclemmons:
@slaymance This is looking good!  Can you add an example here or within the docs for how the customer in #821 would set the initial value?

ericclemmons:
I totally forgot about this PR!  As part of https://github.com/aws-amplify/amplify-ui/pull/911 I actually _removed_ `UsernameAlias` 😱

ericclemmons:
Is it possible for the https://ui.docs.amplify.aws/components/phonenumberfield to combine `country_code` + `phone` into `phone_number` instead?

packages/vue/src/components/user-name-alias.vue
```diff
@@ -146,12 +150,28 @@ if (userNameAlias) {
...
+
+  fullPhoneNumber.value[target.name] = target.value;
+
+  const { country_code, [name]: phoneNumber = '' } = fullPhoneNumber.value;
+
+  updateForm({
+    name,
+    value: country_code + phoneNumber,
+  });
+};
```
ErikCH, line 176: This looks good, but when I was testing the `signUpAttributes` with `phone_number` it was causing cognito to return back `Invalid phone number format.`.

It doesn't send over the area code in that instance.

image

ErikCH: It works fine if the phone_number is the primary login mechanism though.

slaymance: Ah this makes sense since we are using the `PhoneNumberField` primitive when phone number is a sign up attribute and not the `UserNameAlias` component. I just realized our e2e tests don't actually test signing up with attributes. I'll look into addressing this problem when phone is an attribute.

PR #829, opened 2021-11-22
Fix redirect from signIn to confirmSignUp

### LINKED ISSUE #818: Authenticator: Username cannot be empty error message when refreshing screen after signup
Steps to reproduce error below:

1. signup with username/password
2. refresh screen/browser
3. attempt to login and redirect to Confirm Sign Up input
4. input correct auth code into input
5. hit enter

Screen Shot 2021-11-20 at 7 27 08 PM

I am not sure of any workarounds so wanted to report this corner case scenario.

wlee221:
*Issue #, if available:* Fixes #818

*Description of changes:* Transition from `signIn` to `confirmSignUp` wasn't working correctly because `setCredentials` assumes `formValues[primaryAlias]` contains the username. This is not true for `signIn` because `signIn` always has its `primaryAlias` input `name` set to `username`. 

This PR handles `signIn` case and adds a test for it.

### DIFF
--- a/packages/ui/src/machines/authenticator/actions.ts
+++ b/packages/ui/src/machines/authenticator/actions.ts
@@ -66,7 +66,8 @@ export const setConfirmSignUpIntent = assign({
 export const setCredentials = assign({
   authAttributes: (context: SignInContext | SignUpContext, _) => {
     const [primaryAlias] = context.loginMechanisms;
-    const username = context.formValues[primaryAlias];
+    const username =
+      context.formValues[primaryAlias] ?? context.formValues['username'];
     const password = context.formValues?.password;
 
     return { username, password };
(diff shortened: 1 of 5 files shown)


packages/ui/src/machines/authenticator/actions.ts
```diff
@@ -66,7 +66,8 @@ export const setConfirmSignUpIntent = assign({
 export const setCredentials = assign({
   authAttributes: (context: SignInContext | SignUpContext, _) => {
     const [primaryAlias] = context.loginMechanisms;
-    const username = context.formValues[primaryAlias];
+    const username =
+      context.formValues[primaryAlias] ?? context.formValues['username'];
```
wlee221, line 70: The actual fix is here. Ideally, we would enforce better input `name` conventions but that's a bigger refactor.

ErikCH (approved):
All tests pass I'm good

PR #839, opened 2021-11-22
ForwardRef for PhoneNumberField

reesscot:
*Description of changes:*
This PR adds forward ref support for the `PhoneNumberField` primitive by wrapping it in `React.forwardRef`. In this case we're supporting both the standard `ref` for access to the `input` field, and `countryCodeRef` for access to the `select` field.

```jsx
  const countryCodeRef = React.useRef(null);

  
  // Below returns INPUT because the ref is forwarded down to DOM element (PhoneNumberField => TextField => Input => <input>)
  // ref.current.nodeName

  // Below returns SELECT because the ref is forwarded down to DOM element (PhoneNumberField => SelectField => Input => <input>)
  // ref.current.nodeName
  
  return (
    <PhoneNumberField label="name" ref={ref} countryCodeRef={countryCodeRef}/>
  );
};
```

### DIFF
--- a/.changeset/violet-bees-draw.md
+++ b/.changeset/violet-bees-draw.md
@@ -1,0 +1,6 @@
+---
+"docs": patch
+"@aws-amplify/ui-react": patch
+---
+
+ForwardRef for PhoneNumberField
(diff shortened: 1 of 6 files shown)


hvergara (approved):
LGTM :)

PR #841, opened 2021-11-22
Update PasswordField to support showPasswordButton ref

reesscot:
*Description of changes:*
This PR adds forward ref support for the `ShowPasswordButton` in the `PasswordField` primitive by offering a `showPasswordRef` prop.

```jsx
export const RefExample = () => {
  const inputRef = React.useRef(null);
  const showPasswordRef = React.useRef(null);

  const onShowPasswordClick = React.useCallback(() => {
    inputRef.current.focus();
  }, []);

  React.useEffect(() => {
    if (showPasswordRef && showPasswordRef.current) {
      console.count('useeffect');
      showPasswordRef.current.addEventListener(
        'click',
        onShowPasswordClick,
        false
      );
      return () => {
        showPasswordRef.current.removeEventListener(
          'click',
          onShowPasswordClick,
          false
        );
      };
    }
  }, [onShowPasswordClick]);

  return (
    <PasswordField
      ref={inputRef}
      showPasswordRef={showPasswordRef}
      label="Password"
    />
  );
};
```

### DIFF
--- a/packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
+++ b/packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
@@ -14,7 +14,6 @@ const ShowPasswordButtonPrimitive: PrimitiveWithForwardRef<
 > = ({ fieldType, size, ...rest }, ref) => {
   return (
     <Button
-      {...rest}
       ariaLabel={
         fieldType === 'password'
           ? ariaLabelText.showPassword
@@ -23,6 +22,7 @@ const ShowPasswordButtonPrimitive: PrimitiveWithForwardRef<
       className={ComponentClassNames.FieldShowPassword}
       ref={ref}
       size={size}
+      {...rest}
     >
       {fieldType === 'password' ? (
         <IconVisibility size={size} />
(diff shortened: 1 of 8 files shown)


packages/react/src/primitives/types/passwordField.ts
```diff
@@ -13,11 +14,12 @@ export interface PasswordFieldProps extends TextInputFieldProps {
    * @default "current-password"
    */
   autoComplete?: 'new-password' | 'current-password' | string;
+
+  showPasswordRef?: React.Ref<HTMLButtonElement>;
```
hvergara, line 18:
```suggestion
  showPasswordButtonRef?: React.Ref<HTMLButtonElement>;
```

packages/react/src/primitives/PasswordField/ShowPasswordButton.tsx
```diff
@@ -1,28 +1,28 @@
...
-      size={size}
+      {...rest}
       ariaLabel={
         fieldType === 'password'
           ? ariaLabelText.showPassword
           : ariaLabelText.hidePassword
       }
+      className={ComponentClassNames.FieldShowPassword}
+      ref={ref}
+      size={size}
```
hvergara, line 25: IMO `...rest` should be at the bottom of the JSX prop list to allow developers to override `className`, `size` and `ariaLabel`

zchenwei (approved):
LGTM

PR #851, opened 2021-11-23
ForwardRef support for SearchField

reesscot:
*Description of changes:*
This PR adds forward ref support for the `SearchField` primitive by wrapping it in `React.forwardRef`. In this case we're supporting both the standard `ref` for access to the `input` field, and `searchButtonRef` for access to the `button` elements.

```jsx
  const inputRef = React.useRef(null);
  const searchButtonRef = React.useRef(null);
  
  return (
    <SearchField label="search" 
      ref={inputRef}
      searchButtonRef={searchButtonRef}
      />
  );
};
```

### DIFF
--- a/.changeset/tidy-buckets-build.md
+++ b/.changeset/tidy-buckets-build.md
@@ -1,0 +1,6 @@
+---
+"docs": patch
+"@aws-amplify/ui-react": patch
+---
+
+ForwardRef support for SearchField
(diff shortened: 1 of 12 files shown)


dbanksdesign (approved):
LGTM! :shipit:

dbanksdesign (approved):
LGTM! x2

PR #855, opened 2021-11-23
ci: Enable tests to run on forks and action-created PR

### LINKED ISSUE #845: CI: tests do not run on forks or "Version Packages" PR
`Tests` workflow do not run on forks or "Version Packages" PR because...

1. Workflows triggered from forks do not have access to environment secrets for security reasons.
2. Workflows triggered from PRs [do not run](https://github.com/peter-evans/create-pull-request/issues/48) when a PR is created by Github actions.

Reading through https://securitylab.github.com/research/github-actions-preventing-pwn-requests/, we can leverage label-triggered e2e test that acts as "approve and run" button.

wlee221:
*Issue #, if available:* Closes #845

### Description of changes 

This enables **maintainers** to run tests on forks and PRs created by `github-actions` like #838. Tests on forks couldn't run because environment secrets aren't passed for security reasons.

To enable tests on these cases, `tests.yml` trigger is updated to `pull_request_target`. It is granted a read/write repository token and can access secrets even when it is triggered from a fork, so we need to be very careful about the tokens/secrets we expose.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ has practical advices. This PR uses label-based approach -- maintainer first checks that code changes do not tamper secrets/tokens, then they can add `run-tests` to trigger the tests. Tests will run only when the label was *added* and not on subsequent commits to prevent malicious commits after the label was added.

### Additional security hardening:

- [x] `GITHUB_TOKEN` should only have write access to `pull-requests`
- [x] Workflow will automatically remove `run-tests` label after the workflow was triggered.
- [x] #856 

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,11 +7,37 @@ on:
   push:
     branches: [main]
 
-  pull_request:
+  pull_request_target:
     branches: [main]
+    types: [opened, synchronize, labeled]
+
+permissions: read-all
 
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    # We run tests only if it's:
+    #   1) push to main, or
+    #   2) pull request not from a fork (ie. internal PR), or
+    #   3) pull request from a fork (ie. external PR) that was added "run-tests" label
+    if: |
+      github.event_name == 'push' ||
+      (github.event.pull_request.head.repo.full_name == github.repository) || 
+      (github.event.action == 'labeled' && github.event.label.name == 'run-tests')
+    steps:
+      # This would be a good place to run and cache `yarn`, `yarn build`, etc
+      - name: Remove run-tests label, if applicable
+        if: always() && github.event.label.name == 'run-tests'
+        uses: actions/github-script@0.3.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const { issue: { number: issue_number }, repo: { owner, repo } } = context;
+            const label = 'run-tests';
+            github.issues.removeLabel({ owner, repo, issue_number, name: label });
+
   unit:
+    needs: setup
     runs-on: ubuntu-latest
     env:
       NODE_ENV: test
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -7,11 +7,39 @@ on:
...
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    # We run tests only if it's:
+    #   1) push to main, or
+    #   2) pull request not from a fork (ie. internal PR), or
+    #   3) pull request from a fork (ie. external PR) that was added "run-tests" label
+    if: |
+      github.event_name == 'push' ||
+      (github.event.pull_request.head.repo.full_name == github.repository) ||
```
wlee221, line 27: e.g. https://github.com/wlee221/changeset-poc/pull/29

.github/workflows/tests.yml
```diff
@@ -7,11 +7,39 @@ on:
...
+  setup:
+    runs-on: ubuntu-latest
+    # We run tests only if it's:
+    #   1) push to main, or
+    #   2) pull request not from a fork (ie. internal PR), or
+    #   3) pull request from a fork (ie. external PR) that was added "run-tests" label
+    if: |
+      github.event_name == 'push' ||
+      (github.event.pull_request.head.repo.full_name == github.repository) || 
+      (github.event.action == 'labeled' && github.event.label.name == 'run-tests')
```
wlee221, line 28: e.g. https://github.com/wlee221/changeset-poc/pull/30

.github/workflows/tests.yml
```diff
@@ -126,6 +162,10 @@ jobs:
     steps:
       - name: Checkout Amplify UI
         uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event_name == 'push' && github.ref || github.event.pull_request.head.ref }}
```
ericclemmons, line 166: Can you add a comment on this logic?

It reads like `github.ref` is correct when we push, but PRs rely on the `head.ref` on what to test?

ericclemmons (approved):
This looks great!  Let's have one last internal conversation just so there are no surprises. 🙏

.github/workflows/tests.yml
wlee221: Good point! Yes, that's it. `github.ref` on `pull_request_target` always points to the target branch (`main`), so we need to make it a bit more explicit.

wlee221: https://github.com/aws-amplify/amplify-ui/pull/855/commits/76552aa0891bcba68de38fea6cf63af158866b23

Issue #767, opened 2021-11-18
[Feedback] Authenticator@next Confirm password, arguably a bad UX.


artidata:
In my opinion, the confirm password whether in the sign up form or reset password form of the authenticator is a bad UX. There are 2 things that I don't like. 

First, The error message is shown on the **first** character typed. Users definitely know that it doesn't match, why prompt them?   

![Screenshot 2021-11-18 10 18 34 AM](https://user-images.githubusercontent.com/20266094/142365272-bb58eddd-79c3-4757-9f95-f095066d1079.png). 

There are better ways to show the error. Facebook's report password only report such error when the number of retyped characters is larger. I personally show the error when users press the submit button. Watch here: 

https://user-images.githubusercontent.com/20266094/142366396-aca4de0a-8f77-4d8d-b73e-e6fd9e1c9097.mp4

Second, I don't really like the phrase _'Your passwords must match'_. Facebook's reset password use the phrase _'Passwords do not match'_ . I personally use 'Passwords need to match'. However, this comes down to personal preference.

ericclemmons:
I agree – this experience can be optimized in a non-breaking way 👍

slaymance:
Another idea if we decide we want to keep password validation prior to submitting the form: we could validate on a blur event for the Confirm Password field. From a user's standpoint, the error message would only display after they have moved focus away from the Confirm Password field.

Maybe even blur event handling + showing error message when confirm length exceeds initial new password length.

wlee221:
One more thing, we need to make those validation messages customizable too. Agreed with @slaymance to use `blur` event to start password validations.

Milan-Shah:
@artidata We have improved our confirm password error validation experience. You can try it out in our latest version @aws-amplify/ui-angular@2.0.3. Let us know if you have any more suggestions. We are here to help :)

PR #858, opened 2021-11-23
Improve Confirm Password Error Validation

ErikCH:
*Issue #, if available:*

#767 

*Description of changes:*
Improved UX for confirm. Now it shows the error on blur, or when the value is over 6 characters

Validation occurs on blur (Less then six characters)

https://user-images.githubusercontent.com/65630/143136838-ce186970-354d-425d-97ef-5572343ee2ad.mp4

Validation occurs after six or more characters are type

https://user-images.githubusercontent.com/65630/143136867-92ed44e0-af20-40f8-ba6f-d9c2d6d2c841.mp4

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
@@ -24,6 +24,7 @@
     [disabled]="disabled"
     [labelHidden]="labelHidden"
     [autocomplete]="inferAutocomplete()"
+    (onBlur)="onBlur($event)"
   ></amplify-password-field>
 
   <amplify-text-field
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
@@ -24,7 +24,7 @@
     [disabled]="disabled"
     [labelHidden]="labelHidden"
     [autocomplete]="inferAutocomplete()"
-    (onBlur)="onBlur($event)"
+    (setBlur)="onBlur($event)"
   ></amplify-password-field>
 
   <amplify-text-field
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
@@ -65,6 +65,12 @@ export class FormFieldComponent implements OnInit {
     return validationError[this.name];
   }
 
+  public onBlur($event: Event) {
+    let { name } = <HTMLInputElement>$event.target;
+
+    this.authenticator.updateForm({ name: `touched_${name}`, value: true });
+  }
+
   inferLabel(): string {
     const label = this.label || this.attributeMap[this.name]?.label;
     return translate<string>(label);
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
@@ -68,7 +68,7 @@ export class FormFieldComponent implements OnInit {
   public onBlur($event: Event) {
     let { name } = <HTMLInputElement>$event.target;
 
-    this.authenticator.updateForm({ name: `touched_${name}`, value: true });
+    this.authenticator.updateBlur({ name });
   }
 
   inferLabel(): string {
--- a/packages/angular/projects/ui-angular/src/lib/primitives/password-field/password-field.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/primitives/password-field/password-field.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input } from '@angular/core';
+import { Component, EventEmitter, Input, Output } from '@angular/core';
 import { translate } from '@aws-amplify/ui';
 import { nanoid } from 'nanoid';
 
@@ -16,6 +16,7 @@ export class PasswordFieldComponent {
   @Input() placeholder = '';
   @Input() required = true;
   @Input() labelHidden = false;
+  @Output() onBlur = new EventEmitter<Event>();
 
   public type: 'text' | 'password' = 'password';
 
--- a/packages/angular/projects/ui-angular/src/lib/primitives/password-field/password-field.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/primitives/password-field/password-field.component.ts
@@ -16,7 +16,7 @@ export class PasswordFieldComponent {
   @Input() placeholder = '';
   @Input() required = true;
   @Input() labelHidden = false;
-  @Output() onBlur = new EventEmitter<Event>();
+  @Output() setBlur = new EventEmitter<Event>();
 
   public type: 'text' | 'password' = 'password';
 
--- a/packages/e2e/cypress/integration/common/sign-in.ts
+++ b/packages/e2e/cypress/integration/common/sign-in.ts
@@ -33,5 +33,5 @@ When(
 );
 
 When('I type my password', () => {
-  cy.findInputField('Password').type(Cypress.env('VALID_PASSWORD'));
+  cy.findInputField('Password').type(Cypress.env('VALID_PASSWORD')).blur();
 });
--- a/packages/e2e/cypress/integration/common/sign-in.ts
+++ b/packages/e2e/cypress/integration/common/sign-in.ts
@@ -33,5 +33,5 @@ When(
 );
 
 When('I type my password', () => {
-  cy.findInputField('Password').type(Cypress.env('VALID_PASSWORD')).blur();
+  cy.findInputField('Password').type(Cypress.env('VALID_PASSWORD'));
 });
--- a/packages/e2e/cypress/integration/common/sign-up.ts
+++ b/packages/e2e/cypress/integration/common/sign-up.ts
@@ -9,7 +9,8 @@ Given('I click {string}', (text: string) => {
 When('I confirm my password', () => {
   cy.findInputField('Confirm Password')
     .type(Cypress.env('VALID_PASSWORD'))
-    .blur();
+    .blur()
+    .wait(100);
 });
 
 Then('I see {string} as an input field', (name: string) => {
--- a/packages/e2e/cypress/integration/common/sign-up.ts
+++ b/packages/e2e/cypress/integration/common/sign-up.ts
@@ -7,7 +7,9 @@ Given('I click {string}', (text: string) => {
 });
 
 When('I confirm my password', () => {
-  cy.findInputField('Confirm Password').type(Cypress.env('VALID_PASSWORD'));
+  cy.findInputField('Confirm Password')
+    .type(Cypress.env('VALID_PASSWORD'))
+    .blur();
 });
 
 Then('I see {string} as an input field', (name: string) => {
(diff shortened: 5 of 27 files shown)


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
```diff
@@ -65,6 +65,12 @@ export class FormFieldComponent implements OnInit {
     return validationError[this.name];
   }
 
+  public onBlur($event: Event) {
+    let { name } = <HTMLInputElement>$event.target;
+
+    this.authenticator.updateForm({ name: `touched_${name}`, value: true });
```
ErikCH, line 71: Now we can track the touched elements in the form with the name `touched_password` or `touched_confirm_password`

packages/angular/projects/ui-angular/src/lib/primitives/password-field/password-field.component.ts
```diff
@@ -16,6 +16,7 @@ export class PasswordFieldComponent {
   @Input() placeholder = '';
   @Input() required = true;
   @Input() labelHidden = false;
+  @Output() onBlur = new EventEmitter<Event>();
```
ErikCH, line 19: This could have been an `Input()` or `Output()`. I felt Output worked here so I left it

packages/e2e/cypress/integration/common/sign-in.ts
```diff
@@ -33,5 +33,5 @@ When(
 );
 
 When('I type my password', () => {
-  cy.findInputField('Password').type(Cypress.env('VALID_PASSWORD'));
+  cy.findInputField('Password').type(Cypress.env('VALID_PASSWORD')).blur();
```
ErikCH, line 36: Tests were failing unless I added `blur` here.

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -14,13 +14,26 @@ export const defaultServices = {
...
       return null;
-    } else if (password && confirm_password && password !== confirm_password) {
-      // only return an error if both fields have text entered and the passwords do not match
+    } else if (
+      password &&
+      confirm_password &&
+      password !== confirm_password &&
+      ((touched_confirm_password && touched_password) ||
+        (password.length >= 6 && confirm_password.length >= 6))
+    ) {
```
ErikCH, line 33: Ideally the length should come from the zero config. However, six characters is the smallest length you can set in Cognito. In another issue I want this password validator expanded to check all the password settings

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
```diff
@@ -24,6 +24,7 @@
     [disabled]="disabled"
     [labelHidden]="labelHidden"
     [autocomplete]="inferAutocomplete()"
+    (setBlur)="onBlur($event)"
```
ErikCH, line 27: Angular linter complained here if I called this `blur` or `onblur`. So I changed the name to `setBlur`

ErikCH:
The only UX I don't love here is if the user types in a password that's less then six characters long, and it matches the confirm password. An error won't display until they click somewhere else on the form. So they could possibly click the `Create Account` but and then they'll receive the error. I think this is fine until we look at adding in proper password validation from Zero config.

packages/e2e/cypress/integration/common/sign-up.ts
```diff
@@ -7,7 +7,10 @@ Given('I click {string}', (text: string) => {
 });
 
 When('I confirm my password', () => {
-  cy.findInputField('Confirm Password').type(Cypress.env('VALID_PASSWORD'));
+  cy.findInputField('Confirm Password')
+    .type(Cypress.env('VALID_PASSWORD'))
+    .blur()
+    .wait(100);
 });
```
ErikCH, line 14: There is something odd timing wise, that I cannot duplicate in real life only in cypress. Where if you send the blur at the exact same time as you click the acknowledgement button it causes the validation not to fire and the button won't light up. That's why I added a wait in here. Not sure if there is a better way of handling the blur...

ericclemmons (commented):
I wanted to review this before the holidays because forms have been on my mind, particularly because of React Native support.

I think this PR has shown where we've outgrown our simplistic `formData` structure and need meta data to support that complexity like https://formik.org/docs/overview#the-gist does:

- Validate entire form on submit
- Validate single/multiple fields (e.g. confirm password) on `blur`
- Validate single/multiple fields on `change`
- Accessing `touched.*` meta data
- Accessing `errors.*` meta data

My only _actual_ hesitation with this PR is that `touched_*` properties become a public API that's accessible via `formData` when customizing a service.  But also knowing that form validation is tricky & we'll want to track `touched` for most fields to perform more targeted validation.

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -14,13 +14,26 @@ export const defaultServices = {
...
+      // Only return an error if both fields have text entered,
+      // the passwords do not match, and the fields have been
+      // touched or the password and confirm password is longer then or equal to 6.
```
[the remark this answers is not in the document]
ErikCH: Correct, so if they don't type anything in the second field, the error will not show. But if they tried to submit an error would show. 

I could add a check if they typed something in the password filed `abcd` and then clicked the `confirm_pasword` so at least it's touched, then clicked elsewhere it would show an error.

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
```diff
@@ -23,6 +23,11 @@ export function FormFields() {
   // Only 1 is supported, so `['email', 'phone_number']` will only show `email`
   const loginMechanism = fieldNames.shift() as LoginMechanism;
 
+  const handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    const { name } = event.target;
+    updateForm({ name: `touched_${name}`, value: true });
```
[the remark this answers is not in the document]
ErikCH: That's a good idea. I'll dig into this more. I tried this at first adding another event but I was receiving some errors, I'll chat with @wlee221  on it too.

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -14,13 +14,26 @@ export const defaultServices = {
   // Validation hooks for overriding
   async validateCustomSignUp(formData): Promise<ValidatorResult> {},
   async validateConfirmPassword<Validator>(formData): Promise<ValidatorResult> {
-    const { password, confirm_password } = formData;
+    const {
+      password,
+      confirm_password,
+      touched_confirm_password,
+      touched_password,
```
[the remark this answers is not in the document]
ErikCH: Yeah that would make sense. Perhaps, we could add this in a future release? Seems like a big task

packages/ui/src/machines/authenticator/defaultServices.ts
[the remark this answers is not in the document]
ErikCH: OK Pushed up another check for that scenario

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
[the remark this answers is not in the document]
ErikCH: Alright, I added a new `BLUR` change event and a `handleBlur` action. Also added in checks for `confirmResetPassword` and `forceNewPassword` so both of those states can also check the password validation with `blur` for all frameworks.

packages/ui/src/helpers/auth.ts
```diff
@@ -199,6 +199,7 @@ export const getSendEventAliases = (send: Sender<AuthEvent>) => {
     signOut: sendToMachine('SIGN_OUT'),
     submitForm: sendToMachine('SUBMIT'),
     updateForm: sendToMachine('CHANGE'),
+    updateBlur: sendToMachine('BLUR'),
```
ErikCH, line 202: New facade for updateBlur

packages/ui/src/machines/authenticator/actions.ts
```diff
@@ -110,3 +110,13 @@ export const handleInput = assign({
...
   },
 });
+
+export const handleBlur = assign({
+  formValues: (context, event: AuthEvent) => {
+    const { name } = event.data;
+    return {
+      ...context['formValues'],
+      [`touched_${name}`]: true,
+    };
```
ErikCH, line 120: Might as well put this logic here. But to @ericclemmons maybe something more like `formik` is needed in the future

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -14,13 +14,35 @@ export const defaultServices = {
...
-    } else if (password && confirm_password && password !== confirm_password) {
-      // only return an error if both fields have text entered and the passwords do not match
+    } else if (
+      (password || confirm_password) &&
+      touched_password &&
+      touched_confirm_password &&
+      password !== confirm_password
+    ) {
+      return {
+        confirm_password: 'Your passwords must match',
```
ErikCH, line 34: New check if they've entered text in either `password` or `confirm_password` and they've touched both, and the passwords don't match, then show error.

wlee221 (commented):
Thanks for the update, this is a big change! Sorry my review was late. Got few comments around tracking blurred inptus.

packages/ui/src/machines/authenticator/defaultServices.ts
ErikCH: Yeah, let me see if I can do that!

packages/e2e/cypress/integration/common/sign-up.ts
ErikCH: 👍

packages/ui/src/machines/authenticator/defaultServices.ts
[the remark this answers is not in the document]
ErikCH: What if I create a new `formControls` context, that will have all the `touched.*` values, at least for right now just for `password` and `confirm_password`. We can build on this to add in more form validation for the future, and it will remove the values out of `formValues` into their own separate object.  I'll push that up, so I can get your feedback.

packages/ui/src/machines/authenticator/defaultServices.ts
ErikCH: Pushed

packages/ui/src/machines/authenticator/defaultServices.ts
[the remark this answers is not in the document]
ErikCH: Pushed in 032b48b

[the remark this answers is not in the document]
wlee221: Awesome! One last nit -- what do you think about renaming `formControls` to just `touched` for now, and let the keys be `${name}` instead of `touched_${name}`? Then when we need other `formControls`, we can expose them as separate facade. I think this separation of form facades will improve DX than having a general `formControls`, on the long run. What do you think @ericclemmons @ErikCH?

This will need some modification to `runValidators` shape though... 😓  but I think now is a good time to make changes, before we document them!

wlee221 (commented):
Apologies my reviews are getting late. Had one last nit about `formControls` 🙏

packages/ui/src/machines/authenticator/defaultServices.ts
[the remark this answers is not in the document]
ErikCH: Sure! I updated the name to `touched` and update the validators 15c8b02 and 86533e8

wlee221 (approved):
LGTM, thanks @ErikCH!!

PR #861, opened 2021-11-23
Forward ref support for TableRow, Tabs, ToggleButton & ToggleButtonGroup

zchenwei:
*Description of changes:*

This PR will be adding Forward ref support for `TableRow`, `Tabs`, `ToggleButton` & `ToggleButtonGroup`

### DIFF
--- a/packages/react/src/primitives/Tabs/Tabs.tsx
+++ b/packages/react/src/primitives/Tabs/Tabs.tsx
@@ -1,16 +1,16 @@
-import * as React from 'react';
+import classNames from 'classnames';
 import {
   Root,
   List,
   Trigger as RadixTab,
   Content as Panel,
 } from '@radix-ui/react-tabs';
-import { TabsProps, TabItemProps } from '../types';
+import * as React from 'react';
+
 import { Flex } from '../Flex';
+import { View } from '../View';
+import { TabsProps, TabItemProps, PrimitiveWithForwardRef } from '../types';
 import { ComponentClassNames } from '../shared/constants';
-import classNames from 'classnames';
-import { convertStylePropsToStyleObj, prefixer } from '../shared/styleUtils';
-import { Primitive } from '../types';
 
 const isTabsType = (child: any): child is React.Component<TabItemProps> => {
   return (
@@ -21,29 +21,26 @@ const isTabsType = (child: any): child is React.Component<TabItemProps> => {
   );
 };
 
-export const Tabs: Primitive<TabsProps, typeof Flex> = ({
-  alignContent,
-  alignItems,
-  ariaLabel,
-  children,
-  className,
-  defaultIndex = 0,
-  currentIndex,
-  onChange,
-  indicatorPosition,
-  direction,
-  gap = '0',
-  spacing,
-  justifyContent,
-  wrap,
-  ...rest
-}: TabsProps) => {
+const TabsPrimitive: PrimitiveWithForwardRef<TabsProps, typeof Flex> = (
+  {
+    ariaLabel,
+    children,
+    className,
+    defaultIndex = 0,
+    currentIndex,
+    onChange,
+    indicatorPosition,
+    spacing,
+    ...rest
+  }: TabsProps,
+  ref
+) => {
   const tabs = React.Children.map(children, (child) => {
     if (!isTabsType(child)) {
       console.warn(
         'Amplify UI: <Tabs> component only accepts <TabItem> as children.'
       );
-      return null;
+      return {};
     }
 
     return child.props;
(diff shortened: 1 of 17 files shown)


packages/react/src/primitives/View/View.tsx
```diff
@@ -29,11 +26,8 @@ const ViewInner = <Element extends ElementType = 'div'>(
     as,
     {
       'aria-label': ariaLabel,
-      className,
       'data-testid': testId,
       disabled: isDisabled,
-      id,
-      role,
```
zchenwei, line 36: This is a bug. It basically provides `undefined` props and if we render like this `<View as={RadixTab}  />` but not supplying `role` and `id`, they will be overridden and clear out, probably through `{...rest}` internally.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -21,29 +21,26 @@ const isTabsType = (child: any): child is React.Component<TabItemProps> => {
...
+  {
+    ariaLabel,
+    children,
+    className,
+    defaultIndex = 0,
+    currentIndex,
+    onChange,
+    indicatorPosition,
+    spacing,
+    ...rest
```
reesscot, line 34: Thanks for cleaning this up

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -21,29 +21,26 @@ const isTabsType = (child: any): child is React.Component<TabItemProps> => {
...
+  }: TabsProps,
+  ref
+) => {
   const tabs = React.Children.map(children, (child) => {
     if (!isTabsType(child)) {
       console.warn(
         'Amplify UI: <Tabs> component only accepts <TabItem> as children.'
       );
-      return null;
+      return {};
```
zchenwei, line 43: `null` will throw error on mapping operation.

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -58,33 +55,26 @@ export const Tabs: Primitive<TabsProps, typeof Flex> = ({
...
-            >
-              {title}
-            </RadixTab>
-          ))}
+          {React.Children.map(children, (child, index) => {
+            if (!isTabsType(child)) {
+              return null;
+            }
+
+            return React.cloneElement(child, {
```
zchenwei, line 72: By cloning, we are able to clone the `ref` passed along. I don't find a way to reference `ref` directly from `child`

packages/react/src/primitives/View/View.tsx
reesscot: So, `className`, `role` and `id` will be passed through the `{...nonStyleProps}` correct?

reesscot (commented):
Looks good, just want to confirm the change to View doesn't cause any regressions.

packages/react/src/primitives/View/View.tsx
zchenwei: Yes, exactly. But if we list it explicitly and not specify values for them, then it pass in `undefined`

packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -96,7 +86,22 @@ export const Tabs: Primitive<TabsProps, typeof Flex> = ({
   );
 };
 
-export const TabItem: React.FC<TabItemProps> = () => <></>;
+const TabItemPrimitvie: PrimitiveWithForwardRef<TabItemProps, 'div'> = (
```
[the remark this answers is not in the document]
zchenwei: Nice catch!

packages/react/src/primitives/View/View.tsx
reesscot: Yeah, I've seen a related issue with the Radix `Dropdown` primitives we use in `Menu` where a passed `className` prop overrode the internal `className` and would break the component behavior. It looks like we have a test around `id`, and `role` but not for `className`. Would you mind adding a `className` test to `View.test.tsx`?

ericclemmons (approved):

reesscot:
It looks like we have a View tests around `id`, and `role` but not for `className`. Would you mind adding a `className` test to View.test.tsx?

reesscot (changes requested):
It looks like we have a View tests around id, and role but not for className. Would you mind adding a className test to View.test.tsx?

packages/react/src/primitives/shared/constants.ts
```diff
@@ -107,6 +107,5 @@ export enum ComponentClassNames {
   TextField = 'amplify-textfield',
   ToggleButton = 'amplify-togglebutton',
   ToggleButtonGroup = 'amplify-togglebuttongroup',
-  View = 'amplify-view',
```
zchenwei, line 110: Looks like this is not in use anywhere? Could we remove it?

slaymance (approved):

reesscot (approved):
LGTM, thanks for updating the tests.

PR #883, opened 2021-11-26
Add useTheme unit tests

hvergara:
*Description of changes:*
Add unit tests for `useTheme` React hook. Also makes it more resilient, not assuming that `React.useContext` will return an expected value.

### DIFF
--- a/packages/react/src/hooks/__tests__/useTheme.test.ts
+++ b/packages/react/src/hooks/__tests__/useTheme.test.ts
@@ -1,0 +1,46 @@
+import { createTheme, defaultTheme } from '@aws-amplify/ui';
+import * as React from 'react';
+import { useTheme } from '../useTheme';
+
+jest.mock('react');
+
+describe('useTheme', () => {
+  afterEach(() => jest.clearAllMocks());
+
+  it('should return a theme object', () => {
+    const customTheme = createTheme({
+      name: 'my-theme',
+      tokens: {
+        colors: {
+          font: {
+            primary: { value: 'red' },
+          },
+        },
+      },
+    });
+
+    (React.useContext as jest.Mock).mockReturnValue({ theme: customTheme });
+
+    const theme = useTheme();
+
+    expect(theme).toBe(customTheme);
+  });
+
+  it('should return a default theme if context is not available', () => {
+    (React.useContext as jest.Mock).mockReturnValue(undefined);
+
+    const theme = useTheme();
+
+    expect(theme).toBe(defaultTheme);
+  });
+
+  it('should return a default theme if context.theme is undefined', () => {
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/__tests__/useTheme.test.ts
```diff
@@ -0,0 +1,46 @@
...
+          font: {
+            primary: { value: 'red' },
+          },
+        },
+      },
+    });
+
+    (React.useContext as jest.Mock).mockReturnValue({ theme: customTheme });
+
+    const theme = useTheme();
```
ayush987goyal, line 24: It would be better to use the `renderHook` helper from `@testing-library/react-hooks` to test custom hooks. We are already doing that in other hooks test cases.

packages/react/src/hooks/useTheme.ts
```diff
@@ -1,10 +1,14 @@
...
-  const { theme } = useContext(AmplifyContext);
+  const context = React.useContext(AmplifyContext);
 
-  return theme;
+  if (typeof context === 'undefined' || typeof context.theme === 'undefined') {
+    return defaultTheme;
+  }
+
+  return context.theme;
 };
```
ayush987goyal, line 14: Since we are already changing some aspects of the hook, do you think it is better to use the existing `useAmplify` hook here instead of using `useContext` with the `AmplifyContext`? That was using the context would be at just one place. Once we do that, we can move the nullish check of context in that hook.

packages/react/src/hooks/__tests__/useTheme.test.ts
hvergara: good idea! will refactor to use `renderHook`

packages/react/src/hooks/useTheme.ts
hvergara: agreed, will also be refactored to have a single source of truth

packages/react/src/hooks/__tests__/useTheme.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { createTheme, defaultTheme, WebTheme } from '@aws-amplify/ui';
+import { renderHook } from '@testing-library/react-hooks';
+import { AmplifyProvider } from '../../components/AmplifyProvider';
+import { useTheme } from '../useTheme';
+
+const serializeTheme = (theme: WebTheme) => JSON.stringify(theme);
```
hvergara, line 6: have to serialize themes before comparing them to test equality: https://github.com/facebook/jest/issues/8475

reesscot (approved):
LGTM

ayush987goyal (approved):

PR #911, opened 2021-11-30
Login with Email + Phone Number 

ericclemmons:
*Issue #, if available:* https://app.asana.com/0/1201410500481323/1201377903144607

*Description of changes:*

- [x] Add `adminui-auth-with-email-and-phone environment`

	This only has a single `loginMechansisms` value of `email`, despite there being two selected.

- [x] Add `cli-auth-with-email-and-phone environment`

	**Unlike the Admin UI**, this only has two `loginMechansisms` values with `email` and `phone_number`.

- [x] When `loginMechanisms` has more than one option, show a toggle:
	
	> Screen Shot 2021-11-30 at 1 39 23 PM

- [ ] Forms should **always** have `username` as a field for Sign In or Sign Up

- [ ] `country_code` plus `phone` should be computed into `phone_number` in the UI
- [ ] `submitForm` should accept `{ [attribute]: value, ... }` from `FormData(form)` to catch hidden fields such as `username`


packages/vue/src/components/sign-in.vue
```diff
@@ -24,20 +30,18 @@ const forgotYourPasswordLink = computed(() =>
...
 
-const { state, send } = useAuth();
-const actorState = computed(() =>
-  getActorState(state.value)
-) as ComputedRef<SignInState>;
+const { error, isPending, state, send, updateForm } = useAuthenticator();
+const { label, type } = getAliasInfoFromContext(state.context);
+const { formValues } = getActorContext(state) as SignInContext;
+const phoneNumber = computed(
+  () => `${formValues.country_code ?? ''}${formValues.phone ?? ''}`
```
ericclemmons, line 37: @ErikCH 

This isn't updating the `phoneNumber` in the view on change.

I suspect it's because the `formValues` from the machine aren't reactive

packages/vue/src/components/sign-in.vue
```diff
@@ -22,22 +28,27 @@ const forgotYourPasswordLink = computed(() =>
 );
 
 const signInButtonText = computed(() => translate('Sign in'));
-const signIngButtonText = computed(() => translate('Signing in'));
 
-const { state, send } = useAuth();
-const actorState = computed(() =>
-  getActorState(state.value)
-) as ComputedRef<SignInState>;
+const { error, isPending, state, send, updateForm } = useAuthenticator();
```
ericclemmons: @ErikCH I'm struggling with the reactivity of `error` and `isPending` for the following examples:

- http://localhost:3000/ui/components/authenticator/sign-in-with-phone
- http://localhost:3000/ui/components/authenticator/sign-in-with-email
- http://localhost:3000/ui/components/authenticator/sign-in-with-username

ErikCH: 👀

ErikCH: Ok, I fixed reactivity and also made some tweaks to sign-in

ericclemmons: Thank you @ErikCH !

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -399,10 +407,9 @@ export const signInActor = createMachine<SignInContext, AuthEvent>(
     },
     services: {
       async signIn(context) {
-        const source = context.formValues;
-        const { country_code, username, password } = source;
+        const { username, password } = context.formValues;
```
wlee221, line 410: Should this be 
```suggestion
        const { username, password } = event.data ?? context.formValues;
```
now that we're submitting form with data:

https://github.com/aws-amplify/amplify-ui/blob/9a10937632d79043ab79f62b5f932262bf7f7926/packages/react/src/components/Authenticator/SignIn/SignIn.tsx#L49-L55

wlee221: Ahh scratch that, https://github.com/aws-amplify/amplify-ui/blob/9a10937632d79043ab79f62b5f932262bf7f7926/packages/ui/src/machines/authenticator/actions.ts#L116-L119 resolves it

wlee221:
@ericclemmons Weird, I think all three frameworks are failing on the same tests. I could repro this on Angular end, and the error was 

```
Cannot read properties of undefined (reading 'filter')
    at getAliasInfoFromContext (index.js:1676)
    at SignUpFormFieldsComponent.getAliasInfo (sign-up-form-fields.component.ts:33)
    at SignUpFormFieldsComponent.ngOnInit (sign-up-form-fields.component.ts:27)
```

on this line:

https://github.com/aws-amplify/amplify-ui/blob/f1c2d82db913be13425e8b4476983d7cf6f55c40/packages/ui/src/helpers/auth.ts#L134

wlee221:
I believe the additional React error is due to `Authenticator does not have a default implementation for username. Customize Authenticator.SignUp.FormFields to add your own.` whenever `loginMechanism` is username.

I'll finish up Angular implementation meanwhile.

wlee221:
I got all the Angular tests passing! What's remaining on React/Vue side are:
- Vue: Remove side effect on https://github.com/aws-amplify/amplify-ui/blob/bd713016dbd5260ec4587b728d6083bd3444b40b/packages/vue/src/components/alias-control.vue#L47-L54
- Vue: Remove `usre-name-alias-component` usage in signUp 
- React: ensure `AttributeField` support `username` too (or handle it separately)

ericclemmons:
@wlee221 I'll make these updates.  I didn't think that `AttributeField` would include `username` since that's not an actual sign-up attribute – this PR mostly makes `username` a computed value or an explicit text field.

But I'll double-check to see if there's some behavior I messed up on...

ericclemmons:
Ahhh, my mistake – what was once working is failing now!

packages/vue/src/components/authenticator-sign-up-form-fields.vue
```diff
@@ -57,11 +57,14 @@ const loginMechanism = fieldNames.shift() as LoginMechanism;
 const actorContext = computed(() => {
   return getActorContext(state.value);
 }) as ComputedRef<SignInContext>;
-const formValues = actorContext.value?.formValues as AuthFormData;
+const formValues = computed(
+  () => actorContext.value?.formValues as AuthFormData
+);
+actorContext.value.formValues;
 const loginMechanismValue = computed(() => {
```
ErikCH: Updated reactivity, made this computed

packages/vue/src/components/sign-up.vue
```diff
@@ -24,18 +24,21 @@ const onInput = (e: Event): void => {
...
+    submit(e);
   }
 };
 
-const submit = (): void => {
-  props.submitForm();
+const submit = (e: Event): void => {
+  const formData = new FormData(<HTMLFormElement>e.target);
+  const json = Object.fromEntries(formData);
+  props.submitForm(json);
```
ErikCH, line 41: Now it's submitting the json info. Not sure if handleInput is needed or not

packages/vue/src/components/authenticator-sign-up-form-fields.vue
```diff
@@ -57,11 +57,14 @@ const loginMechanism = fieldNames.shift() as LoginMechanism;
 const actorContext = computed(() => {
   return getActorContext(state.value);
 }) as ComputedRef<SignInContext>;
-const formValues = actorContext.value?.formValues as AuthFormData;
+const formValues = computed(
+  () => actorContext.value?.formValues as AuthFormData
+);
+actorContext.value.formValues;
```
ericclemmons, line 63: Why is `actorContext.value.formValues` accessed here, since it doesn't look like it does anything?

Is it here for the Proxy to get accessed and trigger an initial computation?

packages/vue/src/components/sign-up.vue
ericclemmons: `handleInput` is still necessary for dynamically computing the `username` from the `country_code` + `phone` and persisting it in the form for submission.

packages/vue/src/components/authenticator-sign-up-form-fields.vue
ErikCH: oops, I'll delete that. Not sure how that got in there.

ericclemmons:
I'm extremely frustrated with this PR at this point.  It seems each small change is breaking another part and GitHub is slow just to evaluate this.

So, I'm instead thinking of closing this to land much smaller PRs:

- [ ] Land the environments separately:
	- [ ] CLI
	- [ ] Admin UI
- [ ] Make the change to the state machine separately and **only update frameworks to pass again** without "correcting" any implementations.

	What I mean by this is computing the `username` and `phone_number` values.

- [ ] Land the feature https://app.asana.com/0/1201410500481323/1201377903144607 separately for
	- [ ] React
	- [ ] Angular
	- [ ] Vue

PR #912, opened 2021-11-30
Fix height jumps between screens

wlee221:
*Issue #, if available:*

*Description of changes:* This top-aligns Authenticator contents so that tabs and screens do not jump up and down on transitions. 

I added a new `<div data-amplify-router-content>` to contain the screen elements and enable scrolling if needed. 

Authenticator AVO

^ Please click for better resolution!

This was necessary to ensure that tab headers don't scroll with rest of the content. This distinction is only needed for `signIn` and `signUp`, and other screens will have `data-amplify-router-content` and `data-amplify-router` coincide in the same div.

- [x] CSS
- [x] React
- [x] Vue
- [x] Angular

https://user-images.githubusercontent.com/43682783/144142048-5e998ae1-4676-449b-942e-a3e68d979c2e.mp4


examples/next/pages/ui/components/authenticator/modal/index.page.tsx
```diff
@@ -8,7 +8,28 @@ Amplify.configure(awsExports);
...
+        'nickname',
+        'phone_number',
+        'picture',
+        'preferred_username',
+        'profile',
+        'updated_at',
+        'website',
+        'zoneinfo',
+      ]}
+    >
```
wlee221, line 32: Gonna remove this later, just testing really tall screen.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.html
```diff
@@ -1,22 +1,33 @@
 <div data-amplify-authenticator [attr.data-variation]="variation">
   <div data-amplify-container>
     <amplify-slot name="header" [context]="context"></amplify-slot>
-    <div data-amplify-router>
+    <div
+      data-amplify-router
+      [attr.data-amplify-router-content]="hasTabs() ? undefined : ''"
```
wlee221, line 6: For screens with tabs, the actual scrollable content (ie. `data-amplify-router-content`) is the `tab-item` inside this container. For tab-less screens, this distinction is unneeded and `data-amplify-router-content` is safely placed on this container. See attached picture in the PR.

packages/angular/projects/ui-angular/src/lib/primitives/tab-item/tab-item.component.ts
```diff
@@ -10,4 +10,5 @@ export class TabItemComponent {
   @Input() id: string;
   @Input() labelledById: string;
   @Input() tabIndex: number;
+  @HostBinding('style.display') display = 'block'; // emulate div behavior
```
wlee221, line 13: Angular nuance here, `tab-item` wrapper needs this set so that it takes the proper height set by CSS.

packages/vue/src/components/authenticator.vue
```diff
@@ -221,54 +228,56 @@ watch(
...
-          @sign-in-submit="onSignInSubmitI"
-          ref="signInComponent"
-        >
-          <template #signInSlotI>
-            <slot name="sign-in"></slot>
-          </template>
-
-          <template
-            #form="{ info, onSignInSubmit, onForgotPasswordClicked, onInput }"
+        <div v-if="hasTabs()" data-amplify-router-content>
```
wlee221, line 231: GitHub is showing a lot of spacing changes below... but the only code change is adding this `div` around it.

.changeset/green-buttons-hear.md
```diff
@@ -0,0 +1,8 @@
+---
+"@aws-amplify/ui-react": patch
+"@aws-amplify/ui": patch
+"@aws-amplify/ui-vue": patch
+"@aws-amplify/ui-angular": patch
+---
+
+Fix tab jumps between screens and tabs
```
wlee221, line 8:
```suggestion
Fix height jumps between screens and tabs
```

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.html
```diff
@@ -1,22 +1,33 @@
...
+          data-amplify-router-content
+        >
           <!-- signIn component -->
           <amplify-slot
             name="sign-in"
             [context]="context"
             *ngIf="route === 'signIn'"
           >
-            <amplify-sign-in></amplify-sign-in>
+            <amplify-sign-in data-amplify-router-content></amplify-sign-in>
```
wlee221, line 23: Oops, that's a mistake

packages/vue/src/components/authenticator.vue
```diff
@@ -193,6 +193,12 @@ watch(
     signOut.value = s;
   }
 );
+
+const hasTabs = () => {
+  return (
+    actorState.value?.matches('signIn') || actorState.value?.matches('signUp')
+  );
+};
```
ErikCH, line 201: This works, but it's more performant if we switch this to a `computed` property. It won't run nearly as often, then a method since `computed` properties are cached. 

```suggestion
const hasTabs = computed(() => {
  return (
    actorState.value?.matches('signIn') || actorState.value?.matches('signUp')
  );
});
```

Make sure to then change the `hasTabs()` to `hasTabs` without the parenthesis. 

Later on, I should add in `useAuthenticator` and grab the data from `route` instead of `actorState`. But I can do that later.

wlee221: Ah, TIL! Thanks, will adjust.

wlee221: Done! 54197f3

ErikCH (approved):
LGTM

ericclemmons:
So, this is a ridiculously well-documented PR.

Do you think it would be a low lift for additional styles like https://css-tricks.com/scroll-shadows-with-javascript/?

The reason being, on sufficiently long sign-up forms, a shadow would make it clear when there's overflow (so end-users don't miss the submit button or off-screen fields)

packages/ui/src/theme/css/component/authenticator.scss
```diff
@@ -17,6 +17,16 @@
 
   /* Having a z-index at least "wins" by default */
   z-index: 1;
+
+  /* top-align at the top 20% of the screen */
+  padding-top: 20vh;
```
reesscot, line 22: add `box-sizing: border-box;` as well, otherwise this is adding 20vh to the total element height, making it taller than the viewport.



Screen Shot 2021-12-01 at 11 10 20 AM

packages/ui/src/theme/css/component/authenticator.scss
```diff
@@ -17,6 +17,16 @@
...
+  align-content: flex-start; // top align
+}
+
+[data-amplify-authenticator][data-variation='modal']
+  [data-amplify-router-content] {
+  max-height: 60vh;
```
reesscot, line 28: What if we make this full screen height on mobile, that way they don't need to scroll unnecessarily. Note that we need to beware of the `100vh` [issues on mobile ](https://css-tricks.com/css-fix-for-100vh-in-mobile-webkit/)if we do this.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.html
reesscot: How will this work for customers who hide the tabs with CSS?

reesscot (commented):
Looking good, just had some question on the mobile experience.

packages/ui/src/theme/css/component/authenticator.scss
wlee221: Good catch, adjusting:
```suggestion
  box-sizing: border-box; // don't add padding to total height
  padding-top: 20vh;
```

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.html
wlee221: `data-amplify-router` and `data-amplify-router-content` styles do not affect tab headers, so hiding tabs with CSS is unaffected by this change.

packages/ui/src/theme/css/component/authenticator.scss
wlee221: Done here! 06d96fca2d3e0a33e98de9c865ef14a0b49ac7de

Android:

Screen Shot 2021-12-01 at 1 07 29 PM

iOS (flex gap is a different problem, it's only [supported on safari 14.8+](https://caniuse.com/flexbox-gap)):

image

reesscot (approved):
LGTM

wlee221:
> Do you think it would be a low lift for additional styles like https://css-tricks.com/scroll-shadows-with-javascript/?

@ericclemmons Done! https://github.com/aws-amplify/amplify-ui/pull/912/commits/85187402c014622cc6e102c8ecc8fe46187d8914




https://user-images.githubusercontent.com/43682783/144321714-b4790a16-b885-446f-b7d6-8d3d45c8a163.mp4

PR #920, opened 2021-12-01
Adding font docs

dbanksdesign:
*Issue #, if available:*

*Description of changes:*

* Adding instructions on installation page to add the default font with multiple options
* Updated the default theme font tokens to include the different names for the Inter variable font because some installation methods use different font names 'InterVariable' v. 'Inter Var' v. 'Inter' 
* Updated the doc site CSS to use the Inter font directly from the creator's site so we can get special open type features. 

CleanShot 2021-12-01 at 11 34 42@2x


docs/src/styles/docs/base.scss
```diff
@@ -1,10 +1,8 @@
-[data-amplify-theme] {
-  font-family: var(--amplify-fonts-default-static);
-}
-@supports (font-variation-settings: normal) {
-  [data-amplify-theme] {
-    font-family: var(--amplify-fonts-default-variable);
-  }
+@import url('https://rsms.me/inter/inter.css');
```
reesscot, line 1: What do you think of hosting this file in our repo so we don't take an external load-time dependency? My concern is if this file moves or downloading it is slow, it might block first render?

reesscot (approved):
LGTM, just one non-blocking comment

docs/src/styles/docs/base.scss
dbanksdesign: Yea that is probably a better idea. Let me do that real quick

dbanksdesign: @reesscot added the font into the docs site and tested it locally

hvergara (approved):

reesscot (approved):

PR #921, opened 2021-12-01
Docs: update sandpack to not autorun on homepage

reesscot:
*Description of changes:*
This PR fixes an issue where Firefox locks up on homepage when running the demo (sandpack). The fix is to disable `autorun` so that it only runs when clicking `run`. This is a stopgap measure, as clicking `run` in Firefox will still lock up the browser, so a follow up fix is needed for a complete solution. However, this will make the homepage useable in Firefox.

_Before in Firefox (Initial page load):_

_After in Firefox (Initial page load):_


hvergara (approved):

dbanksdesign (approved):
LGTM! :shipit:

PR #948, opened 2021-12-06
Documentation updates (reorg styles and fix Text styles)

reesscot:
*Description of changes:*
This PR makes the following changes:
1. Move all primitive documentation styles to `primitives` folder
2. Cleanup Text CSS styling (was broken, using old `amplify-ui` namespace)


docs/src/pages/components/text/react.mdx
```diff
@@ -97,19 +97,10 @@ import '@aws-amplify/ui-react/styles.css';
 
 <Example>
   <Text className="primary-styled-variable-text" variation="primary">
-    This is my styled text
+    {'This is my styled text'}
```
reesscot, line 100: This fixes an console error `p tag cannot be nested inside p tag`.

docs/src/pages/components/text/react.mdx
```diff
@@ -97,19 +97,10 @@ import '@aws-amplify/ui-react/styles.css';
...
   </Text>
 </Example>
 
-To replace the Text styling, unset it:
```
reesscot, line 104: Removing this guidance because using `all: unset` doesn't have the effect that most developers would expect, and would cause more harm than good.

docs/src/pages/components/text/react.mdx
dbanksdesign: Interesting..

dbanksdesign (approved):
LGTM! :shipit:

PR #396, opened 2021-09-17
Attempt to cache deps

ericclemmons:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #951, opened 2021-12-07
ci: cache artifacts and remove repeated steps

### LINKED ISSUE #927: CI is running too slow

### Please describe your bug.

CI is running slower than expected! It's currently running at 35~40 min. This can be shortened by:

- [ ] caching build artifacts
- [ ] remove duplicate/repeated builds
- [ ] run `yarn environements pull` in parallel
- [ ] Running e2e and docs test in parallel

Other CI enhancement ideas include:

- [ ] cancel concurrent ci runs in pull request
- [ ] #880 
- [ ] Conditional tests

_No response_

wlee221:
*Issue #, if available:* Fixes #927

*Description of changes:* This uses [`actions/cache`](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows) to shorten the CI runtime to reuse artifacts and dependencies.

## Cache strategy

Cache is identified by a unique key. With `hashFiles` and other identifiers, we can construct keys to explicitly define the criteria for cache hit. For example, I use 

```
key: ${{ runner.os }}-cypress-${{ hashFiles('yarn.lock') }}
``` 
to cache cypress runner. This ensures that we reuse the same cypress file between workflows as long as they have the same `yarn.lock` file. If `yarn.lock` changes, which means cypress runner have possible changes, we rebuild and cache the new cypress runner.

## Minor improvements

This PR also separates `build` from `install` to allow for finer optimizations. We were rebuilding ui packages multiple times due to this.

## For reviewers

For reviewing, please see `pull_request` runs instead of `pull_request_target`. `pull_request_target` runs against the workflow on `main`, which isn't up-to-date with these changes and fail. Here's the latest run: https://github.com/aws-amplify/amplify-ui/actions/runs/1560372622

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,15 +60,18 @@ jobs:
         run: yarn --frozen-lockfile
       - name: build @aws-amplify/ui package
         run: yarn ui build
+      - name: Cache cypress runner
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/Cypress
+          key: ${{ runner.os }}-cypress-${{ hashFiles('yarn.lock') }}
       - name: Cache build
         uses: actions/cache@v2
-        id: build-cache
         with:
           path: |
             ./node_modules
-            **/node_modules
             ./packages/ui/dist
-            ~/.cache/Cypress
+            **/node_modules
           key: ${{ runner.os }}-yarn-${{ steps.get-ref.outputs.commit }}
   unit:
     needs: setup
@@ -172,6 +175,13 @@ jobs:
           node-version: lts/*
           cache: 'yarn'
 
+      - name: Restore cypress runner Cache
+        uses: actions/cache@v2
+        id: restore-cypress-cache
+        with:
+          path: ~/.cache/Cypress
+          key: ${{ runner.os }}-cypress-${{ hashFiles('yarn.lock') }}
+
       - name: Restore Cache
         uses: actions/cache@v2
         id: restore-cache
@@ -184,7 +194,7 @@ jobs:
           key: ${{ runner.os }}-yarn-${{ needs.setup.outputs.commit }}
 
       - name: Install packages
-        if: steps.restore-cache.outputs.cache-hit != 'true'
+        if: steps.restore-cache.outputs.cache-hit != 'true' || steps.restore-cypress-cache.outputs.cache-hit != 'true'
         run: yarn --frozen-lockfile
 
       - name: Build @aws-amplify/ui package
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
         id: get-ref
         run: |
           echo "::set-output name=key::${{ github.event_name == 'push' && github.ref || github.event.pull_request.head.ref }}"
-          echo "::set-output name=commit::$(git rev-parse --short "$GITHUB_SHA")
+          echo "::set-output name=commit::$(git rev-parse --short "$GITHUB_SHA")"
       - name: Checkout Amplify UI
         uses: actions/checkout@v2
         with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,6 +66,7 @@ jobs:
             ./node_modules
             **/node_modules
             **/dist
+            ~/.cache/Cypress
           key: ${{ runner.os }}-yarn-${{ steps.get-ref.outputs.commit }}
   unit:
     needs: setup
@@ -104,6 +105,7 @@ jobs:
             ./node_modules
             **/node_modules
             **/dist
+            ~/.cache/Cypress
           key: ${{ runner.os }}-yarn-${{ needs.setup.outputs.commit }}
 
       - name: Install packages
@@ -170,6 +172,7 @@ jobs:
             ./node_modules
             **/node_modules
             **/dist
+            ~/.cache/Cypress
           key: ${{ runner.os }}-yarn-${{ needs.setup.outputs.commit }}
 
       - name: Install packages
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,9 +113,6 @@ jobs:
       - name: Lint packages
         run: yarn workspace ${{ matrix.package }} lint
 
-      - name: Build ${{ matrix.package }} package
-        run: yarn workspace ${{ matrix.package }} build
-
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -37,7 +40,34 @@ jobs:
...
+      - name: Install packages
+        run: yarn --frozen-lockfile
+      - name: Cache build
+        uses: actions/cache@v2
+        id: build-cache
+        with:
+          path: |
+            ./node_modules
+            **/node_modules
+            **/dist
```
wlee221, line 68: dist doesn't really have to be cached here, especially now that we are looking to do conditional testing. I'll look into `caching` it from `unit`.

.github/workflows/tests.yml
```diff
@@ -37,7 +40,34 @@ jobs:
...
+        with:
+          node-version: lts/*
+          cache: 'yarn'
+      - name: Get ref and commit
+        id: get-ref
+        run: |
+          echo "::set-output name=key::${{ github.event_name == 'push' && github.ref || github.event.pull_request.head.ref }}"
+          echo "::set-output name=commit::$(git rev-parse --short "$GITHUB_SHA")"
+      - name: Install packages
+        run: yarn --frozen-lockfile
```
wlee221, line 60: We should look into separating `build` from `install`, at least from CI side. This will make more rooms for optimization.

ericclemmons: Feel free to actually remove it here – my previous attempts separated them, too: #396 

We'll just need to update the `CONTRIBUTING.md` and maybe the `.husky` hook to build after `git pull`.

.github/workflows/tests.yml
```diff
@@ -37,7 +40,43 @@ jobs:
...
+      - name: Get ref and commit
+        id: get-ref
+        run: |
+          echo "::set-output name=key::${{ github.event_name == 'push' && github.ref || github.event.pull_request.head.ref }}"
+          echo "::set-output name=commit::$(git rev-parse --short "$GITHUB_SHA")"
+      - name: Install packages
+        run: yarn --frozen-lockfile --prefer-offline
+      - name: Build ui package
+        run: yarn ui build
+      - name: Cache cypress runner
```
wlee221, line 63: Instead of caching everything at once, we cache each artifact. This allows us to fine-tune cache keys based on use cases.

.github/workflows/tests.yml
```diff
@@ -37,7 +40,43 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/Cypress
+          key: ${{ runner.os }}-cypress-${{ hashFiles('yarn.lock') }}
```
wlee221, line 67: Only reuse cypress runner if `yarn.lock` changes

.github/workflows/tests.yml
```diff
@@ -125,25 +195,76 @@ jobs:
...
+        if: steps.restore-cache.outputs.cache-hit != 'true' || steps.restore-cypress-cache.outputs.cache-hit != 'true'
+        run: yarn --frozen-lockfile --prefer-offline
+
+      - name: Build @aws-amplify/ui package
+        if: steps.restore-ui-cache.outputs.cache-hit != 'true'
+        run: yarn ui build
+
+      - name: Build ${{ matrix.package }} package
+        if: steps.restore-package-cache.outputs.cache-hit != 'true'
+        run: yarn ${{ matrix.package }} build
```
wlee221, line 238: We build `@aws-amplify/ui` as a fallback if we didn't hit the cache.

.github/workflows/tests.yml
```diff
@@ -125,25 +195,76 @@ jobs:
...
+      - name: Get CLI versions
+        id: cli-version
+        run: echo "::set-output name=version::$(amplify --version)"
+
+      - name: Create or restore environments cache
+        id: environments-cache
+        uses: actions/cache@v2
+        with:
+          path: environments/**/aws-exports.js
+          key: ${{ runner.os }}-environments-${{ steps.cli-version.outputs.version }}-${{ hashFiles('environments/**/amplify/**') }}
```
wlee221, line 252: We reuse environments unless 1) CLI version and 2) `environments/**` contents change

.github/workflows/tests.yml
```diff
@@ -37,7 +40,43 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: ./packages/ui/dist
+          key: ${{ runner.os }}-ui-${{ steps.get-ref.outputs.commit }}
+      - name: Cache node_modules
+        uses: actions/cache@v2
+        with:
+          path: |
+            ./node_modules
+            **/node_modules
```
wlee221, line 78: This would be the first cache to remove if we're facing any cache misses. `node_modules` is large at ~600mb, and `yarn` is already being cached, so this doesn't save that much time for its storage cost.

For reference, based on https://github.com/aws-amplify/amplify-ui/actions/runs/1560169048, caching `node_modules` saved 47 seconds. If we assume we implement conditional testing and only run unit testing, that goes down to only 7 seconds. I'll revisit this when we do conditional testing.

wlee221: `actions/cache` is very generous though and provides 10GB of cache storage. https://github.com/actions/cache

wlee221: Actually, screw it, I'll get rid of it. This is really a duplciate of `yarn cache`, I'm down to have conditional testing do more work than this tiny optimization.

wlee221: (sry for this long thread, I was going back and forth)

slaymance (approved):
This looks great!

package.json
```diff
@@ -1,7 +1,6 @@
 {
   "private": true,
   "scripts": {
-    "install": "yarn build",
```
ericclemmons, line 4: With this change, what do you think about adding `yarn build` to our `post-merge` hook?

https://github.com/aws-amplify/amplify-ui/blob/5e575bdf522eacd0a5cfa0c32be6da52246e8984/.husky/post-merge#L5-L6

When switching branches, it can be risky to not rebuild the packages. (For example, if you're working with `yarn react dev`, but forget to `yarn ui build` after switching)

ericclemmons (approved):
Let's give this a go!

I think there's a small DX improvement for local contributors, but definitely not blocking!

package.json
wlee221: That's a good catch, let's do that.

PR #953, opened 2021-12-07
Refocus `SearchField` input field on clear button click

### LINKED ISSUE #848: feature request: `SearchField` should re-focus input field on clear
For accessibility/usability, `SearchField` should re-focus the Input Field when user clicks on "clear" button. This would be a more user friendly flow that allows users to immediately re-enter their search query without focusing the field again with their mouse. This behavior already occurs when users hit the `Esc` key, but doesn't currently occur if they use the clear button with their mouse

Current behavior:

Desired behavior:
Clicking the clear button re-focuses input field.

reesscot:
*Issue #, if available:*
Closes #848 

*Description of changes:*
This PR adds behavior of refocusing the search input field when a user clicks the clear button so that a user can reenter their search query.
 

### DIFF
--- a/packages/react/src/primitives/SearchField/useSearchField.tsx
+++ b/packages/react/src/primitives/SearchField/useSearchField.tsx
@@ -1,0 +1,75 @@
+import * as React from 'react';
+
+import { isFunction } from '../shared/utils';
+import { SearchFieldProps } from '../types';
+import { useComposeRefsCallback } from '../../hooks/useComposeRefsCallback';
+
+const ESCAPE_KEY = 'Escape';
+const ENTER_KEY = 'Enter';
+const DEFAULT_KEYS = [ESCAPE_KEY, ENTER_KEY];
+
+export const useSearchField = ({
+  onSubmit,
+  onClear,
+  externalRef,
+}: Partial<
+  SearchFieldProps & { externalRef: React.ForwardedRef<HTMLInputElement> }
+>) => {
+  const [value, setValue] = React.useState<string>('');
+  const internalRef = React.useRef<HTMLInputElement>(null);
+  const composeRefs = useComposeRefsCallback({ externalRef, internalRef });
+
+  const onClearHandler = React.useCallback(() => {
+    setValue('');
+    internalRef?.current?.focus();
+    if (isFunction(onClear)) {
+      onClear();
+    }
+  }, [setValue, onClear]);
+
+  const onSubmitHandler = React.useCallback(
+    (value: string) => {
+      if (isFunction(onSubmit)) {
+        onSubmit(value);
+      }
+    },
+    [onSubmit]
+  );
+
+  const onKeyDown = React.useCallback(
+    (event) => {
+      const key = event.key;
+
+      if (DEFAULT_KEYS.includes(key)) {
+        event.preventDefault();
+      }
+
+      if (key === ESCAPE_KEY) {
+        onClearHandler();
+      } else if (key === ENTER_KEY) {
+        onSubmitHandler(value);
+      }
+    },
+    [value, onClearHandler, onSubmitHandler]
+  );
+
+  const onInput = React.useCallback(
+    (event) => {
+      setValue(event.target.value);
+    },
+    [setValue]
+  );
+
+  const onClick = React.useCallback(() => {
+    onSubmitHandler(value);
+  }, [onSubmitHandler, value]);
+
+  return {
+    value,
+    onClearHandler,
+    onKeyDown,
+    onInput,
+    onClick,
+    composeRefs,
+  };
+};
--- a/packages/react/src/primitives/SearchField/useSearchField.tsx
+++ b/packages/react/src/primitives/SearchField/useSearchField.tsx
@@ -1,23 +1,21 @@
 import * as React from 'react';
 
 import { isFunction } from '../shared/utils';
-import { SearchFieldProps } from '../types';
+import { SearchFieldProps, UseSearchFieldProps } from '../types';
 import { useComposeRefsCallback } from '../../hooks/useComposeRefsCallback';
 
 const ESCAPE_KEY = 'Escape';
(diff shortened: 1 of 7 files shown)


packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
+
+import { isFunction } from '../shared/utils';
+import { SearchFieldProps } from '../types';
+import { useComposeRefsCallback } from '../../hooks/useComposeRefsCallback';
+
+const ESCAPE_KEY = 'Escape';
+const ENTER_KEY = 'Enter';
+const DEFAULT_KEYS = [ESCAPE_KEY, ENTER_KEY];
```
hvergara, line 9: suggestion: convert it into a `Set`. Lookups are faster than an array (`Set.prototype.has` v/s `Array.prototype.includes`)

```suggestion
const DEFAULT_KEYS = new Set([ESCAPE_KEY, ENTER_KEY]);
```

packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const useSearchField = ({
+  onSubmit,
+  onClear,
+  externalRef,
+}: Partial<
+  SearchFieldProps & { externalRef: React.ForwardedRef<HTMLInputElement> }
```
hvergara, line 16: nitpick: can you create a separate type for hook props? (e.g `UseSearchFieldProps`)

```ts
export type UseSearchFieldProps = Partial<SearchFieldProps> & {
  externalRef?: React.ForwardedRef<HTMLInputElement>
};
```

packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
...
+>) => {
+  const [value, setValue] = React.useState<string>('');
+  const internalRef = React.useRef<HTMLInputElement>(null);
+  const composeRefs = useComposeRefsCallback({ externalRef, internalRef });
```
hvergara, line 20:
```suggestion
  const composedRefs = useComposeRefsCallback({ externalRef, internalRef });
```

packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
...
+const DEFAULT_KEYS = [ESCAPE_KEY, ENTER_KEY];
+
+export const useSearchField = ({
+  onSubmit,
+  onClear,
+  externalRef,
+}: Partial<
+  SearchFieldProps & { externalRef: React.ForwardedRef<HTMLInputElement> }
+>) => {
+  const [value, setValue] = React.useState<string>('');
```
hvergara, line 18: do we need to set it as an empty string? I wonder if `undefined` or `null` is a better default value here.

packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const internalRef = React.useRef<HTMLInputElement>(null);
+  const composeRefs = useComposeRefsCallback({ externalRef, internalRef });
+
+  const onClearHandler = React.useCallback(() => {
+    setValue('');
+    internalRef?.current?.focus();
```
hvergara, line 24: I'd prefer to be safe and check if `internalRef.current` is not undefined

```suggestion
    if (internalRef && internalRef.current) {
      internalRef.current.focus()
    }
```

packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
...
+      if (isFunction(onSubmit)) {
+        onSubmit(value);
+      }
+    },
+    [onSubmit]
+  );
+
+  const onKeyDown = React.useCallback(
+    (event) => {
+      const key = event.key;
```
hvergara, line 41:
```suggestion
      const { key } = event;
```

packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+      if (DEFAULT_KEYS.includes(key)) {
```
hvergara, line 43: if you use a `Set` instead, this would be:

```suggestion
      if (DEFAULT_KEYS.has(key)) {
```

packages/react/src/primitives/SearchField/useSearchField.tsx
```diff
@@ -0,0 +1,75 @@
...
+        event.preventDefault();
```
hvergara, line 44: can we early return after this? the rest of the code is unreachable after this condition

hvergara (approved):

packages/react/src/primitives/SearchField/useSearchField.tsx
reesscot: I believe `undefined` will result in the classic React error "input changed from controlled to uncontrolled", so it's better to make it an empty string.

hvergara: 👍🏼

packages/react/src/primitives/SearchField/useSearchField.tsx
reesscot: Fixed

packages/react/src/primitives/SearchField/useSearchField.tsx
reesscot: Fixed

packages/react/src/primitives/SearchField/useSearchField.tsx
reesscot: I think the optional chaining should be sufficient here

PR #956, opened 2021-12-08
Add eslint React rules of hooks

reesscot:
*Description of changes:*
This PR adds the eslint plugin `eslint-plugin-react-hooks` to the React repo and sets it up to run on the primitives folder. I've also fixed all of the error and warning messages coming from the plugin.


packages/react/src/primitives/Collection/Collection.tsx
```diff
@@ -51,6 +51,7 @@ export const Collection = <Item,>({
 }: CollectionProps<Item>): JSX.Element => {
   const [searchText, setSearchText] = useState<string>();
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   const onSearch = useCallback(debounce(setSearchText, TYPEAHEAD_DELAY_MS), [
```
reesscot, line 55: @hvergara Do you have any concerns with ignoring the rules of hooks error message on this code? Judging by the code, I don't think there is a risk of stale closures here, given it only uses the `setSearchText` function, and `TYPEAHEAD_DELAY_MS` is a constant.
Screen Shot 2021-12-07 at 2 37 31 PM

packages/react/src/primitives/Pagination/PaginationItem.tsx
```diff
@@ -17,11 +17,20 @@ export const PaginationItem: React.FC<PaginationItemProps> = ({
   ariaLabel,
   ...rest
 }) => {
+  const onChange = React.useCallback(() => {
```
reesscot, line 20: The error here was that useCallback was being called inside of a conditional, so I've moved them out of the `switch`.

packages/react/src/primitives/Pagination/useRange.ts
```diff
@@ -13,17 +13,17 @@ export const ELLIPSIS = '...';
...
-  totalPages: number,
-  siblingCount = 1
+  currentPageParam: number,
+  totalPagesParam: number,
+  siblingCountParam = 1
 ): (string | number)[] => {
   const range = useMemo(() => {
     // The current page should not be less than 1
-    currentPage = Math.max(currentPage, 1);
+    const currentPage = Math.max(currentPageParam, 1);
```
reesscot, line 22: Error fixed here was:

Assignments to the `currentPage` variable from inside React Hook useMemo will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useMemo.

Solved by creating a local variable `currentPage`.

packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
```diff
@@ -6,30 +6,32 @@ import { ComponentClassNames } from '../shared/constants';
...
-const CountryCodeSelectPrimitive: Primitive<CountryCodeSelectProps, 'select'> =
-  ({ className, ...props }, ref) => {
-    const countryCodeOptions = React.useMemo(
-      () =>
-        countryDialCodes.map((dialCode) => (
-          <option key={dialCode} value={dialCode}>
-            {dialCode}
-          </option>
-        )),
-      [countryDialCodes]
```
reesscot, line 18: Since `countryDialCodes` is a contant, there's no need to list it as a dependency.

packages/react/src/primitives/SwitchField/useSwitch.ts
```diff
@@ -17,7 +17,7 @@ export const useSwitch = (props) => {
         setIsOn(event.target.checked);
       }
     },
-    [onChange, isChecked]
+    [onChange, isControlled, isDisabled]
```
reesscot, line 20: Uneeded dependency `isChecked`, and missing `isControlled` and `isDisabled`

packages/react/package.json
```diff
@@ -35,12 +35,13 @@
     "build:ts": "tsup --minify --keep-names",
     "build:catalog": "ts-node scripts/generatePrimitivesCatalog.ts",
     "dev": "tsup --watch src",
+    "eslint:primitives": "npx eslint ./src/primitives --ext .js,.ts,.tsx",
```
ayush987goyal, line 38: Why do we want to lint only primitives? I think it would be an overall benefit to run rules of hook (and other lint rules) and all the components. Thoughts?

packages/react/src/primitives/SwitchField/useSwitch.ts
hvergara: good catch, eslint! :D

packages/react/src/primitives/Collection/Collection.tsx
hvergara: I'm good

packages/react/package.json
hvergara, line 38: no need to prepend `npx` here

hvergara (changes requested):

packages/react/package.json
reesscot: Yeah, I'm starting with the primitives since I knew how to fix those errors. Will follow up with a draft PR to run eslint on all the files so I can get help on some of the Authenticator lint errors.

reesscot: Fixed

hvergara (approved):

PR #960, opened 2021-12-08
Adjust getting started with react example to use createTheme

interdigitize:
*Issue #, if available:*

*Description of changes:*
On https://ui.docs.amplify.aws/theming
under "To extend or override a token in the default theme with createTheme" the code example in the documentation doesn't include createTheme. It is currently a ts js hybrid with no opening paren. 

This PR adjusts the example so that it works as expected if copied. 

### DIFF
--- a/docs/src/pages/theming/getting-started.react.mdx
+++ b/docs/src/pages/theming/getting-started.react.mdx
@@ -22,7 +22,7 @@ To extend or override a token in the default theme with `createTheme`
 import { AmplifyProvider, Theme } from '@aws-amplify/ui-react';
 
 // Step 1: Create a new Theme with your custom values
-const theme: Theme = {
+const theme = createTheme({
   name: 'my-theme',
   tokens: {
     colors: {


reesscot:
Hi @interdigitize, thank you for opening this issue. I did some looking into the theming docs and realized that passing a custom theme object to `createTheme` is now unecessary because we already run `createTheme` in the [AmplifyProvider here](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/components/AmplifyProvider/index.tsx#L22). I've submitted a [PR](https://github.com/aws-amplify/amplify-ui/pull/1039) to update the documentation to clarify this. 

However, I was wondering if you'd like to adjust this PR (or open separate PR) to fix the other problem you mentioned with the odd TypeScript types in `jsx` bug. It would be nice to have separate tabs for JavaScript and Typescript examples. See https://github.com/aws-amplify/amplify-ui/blame/main/docs/src/pages/components/authenticator/quick-start.react.mdx for an example of importing and using the `Tabs` component in the docs. We could move the current typescript code to a "TypeScript" tab and write Javascript code in a "JavaScript" tab that removes the `Theme` type import and type annotation.

interdigitize:
Hi @reesscot, I'll take a look at the tabs and either update this or open a separate PR—depending on what adjustments are required and if it still makes sense in the context of this PR.

interdigitize:
Hi @reesscot, here is the new PR:  https://github.com/aws-amplify/amplify-ui/pull/1069

PR #963, opened 2021-12-08
docs: custom security headers

reesscot:
*Description of changes:*
Add security headers recommended by [snyk.io](https://snyk.io/test/website-scanner/?test=211208_AiDcC6_122c7eecaf52dc2e36ee6791479515b6) to the docs site. Will not merge until fully tested in the PR preview.

1. Added [customHTTP.yml](https://docs.aws.amazon.com/amplify/latest/userguide/custom-headers.html#setting-custom-headers) with all security headers except CSP
2. Updated `_document.page.tsx` to add a `meta` tag for CSP with different rules for dev and prod modes
3. Update `next.config.yml` to add the same security headers as `customHTTP.yml` in dev mode (won't apply to production since we deploy static site files).

Note: I'm using a meta tag for the CSP rules instead of a security header set in `customHTTP` because it allows us to only allow the hashed inline script that is built by NextJS and no other inline scripts. See https://developers.google.com/web/fundamentals/security/csp#the_meta_tag

_Request headers:_
RequestSecurityHeaders

_CSP meta tag_
Screen Shot 2021-12-08 at 2 07 30 PM

Webpage Test is giving us an A for security now (was F before due to lack of Security Headers)
https://webpagetest.org/result/211208_BiDcS7_e4309a6474f3ae7e4df2e30ac41b25db/


customHttp.yml
```diff
@@ -0,0 +1,14 @@
+customHeaders:
```
ericclemmons, line 1:
```suggestion
# Custom headers when serving docs through Amplify Hosting
#
# See: https://docs.aws.amazon.com/amplify/latest/userguide/custom-headers.html#setting-custom-headers
customHeaders:
```

ericclemmons (approved):
Let's see what shows up in the PR preview 🙏

customHttp.yml
reesscot: Added

docs/src/pages/_document.page.tsx
```diff
@@ -1,28 +1,39 @@
...
 
 const favicon =
   process.env.NODE_ENV === 'development'
     ? '/svg/favicon-dev.svg'
     : '/svg/favicon.svg';
 
+const cspHashOf = (text) => {
+  const hash = crypto.createHash('sha256');
+  hash.update(text);
+  return `'sha256-${hash.digest('base64')}'`;
```
hvergara, line 12: IMO this function should not include the single quotes in the returned value

docs/src/pages/_document.page.tsx
```diff
@@ -1,28 +1,39 @@
...
+};
```
hvergara, line 13:
```suggestion
};

```

docs/src/pages/_document.page.tsx
```diff
@@ -1,28 +1,39 @@
...
 class MyDocument extends Document {
   render() {
+    // See: https://github.com/vercel/next.js/blob/master/examples/with-strict-csp/pages/_document.js
+    let csp = `default-src 'self';
```
hvergara, line 17: can we put this logic into a separate function? I feel it adds unnecessary noise to the render function

docs/next.config.js
```diff
@@ -19,6 +19,37 @@ module.exports = withNextPluginPreval({
     ignoreBuildErrors: true,
   },
 
+  async headers() {
+    return [
+      {
+        // Apply these headers to all routes in your application.
+        source: '/(.*)',
+        headers: [
```
hvergara, line 27: as `yaml` is already a dependency of some packages, can we read these values from the `customHttp.yml` file?

docs/src/pages/_document.page.tsx
```diff
@@ -1,28 +1,39 @@
...
+              NextScript.getInlineScriptSource(this.props)
+            )}`;
+    }
+
     return (
       <Html>
         <Head>
+          <meta httpEquiv="Content-Security-Policy" content={csp} />
           <link rel="icon" type="image/svg+xml" href={favicon} />
-          {/* Adding custom variable fonts from google */}
```
hvergara, line 14: why are we removing these?

docs/src/pages/_document.page.tsx
reesscot: Sure, will update

docs/src/pages/_document.page.tsx
reesscot: We are using the self-hosted Inter font, so this was adding extra calls to the page that aren't necessary.

Milan-Shah (approved):
LGTM!

docs/next.config.js
reesscot: Is this blocking, since this only relates to the dev build? Would prefer to do as follow up item since this is a security issue

hvergara: sure! np

hvergara (approved):

PR #895, opened 2021-11-26
Convert docs from SSR to SSG

ericclemmons:
### Before

	rm -rf docs/.next && time yarn docs build
	✨  Done in 61.51s.
	yarn docs build  218.09s user 36.18s system 412% cpu 1:01.65 total

### After

	✨  Done in 48.16s.
	yarn docs build  141.65s user 44.45s system 385% cpu 48.301 total

*Description of changes:*

- [x] Upgrade `next` from `11.0.1` to `12.0.4`
	
	We couldn't do this before because we had to share the same version of Next.js as https://github.com/aws-amplify/docs

- [x] Opt-in to new Rust/SWC compiler for speed
- [x] Remove `I18n` (that wasn't being used yet), but requires SSR
- [x] Add `next lint` (via `yarn next lint`) to catch any obvious errors
- [x] Fix lint errors (mostly missing `key`)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #965, opened 2021-12-09
Update nextjs to 12.0.7

reesscot:
*Description of changes:*
Update NextJS version to `^12.0.7`. This is to address a security vulnerability in NextJS < 12.0.5. This security vulnerability shouldn't affect us in production (only development) since we are not running NextJS SSR in production, only as SSG.


docs/package.json
```diff
@@ -27,7 +27,7 @@
     "gray-matter": "^4.0.3",
     "lodash": "^4.17.21",
     "mdx-prism": "^0.3.3",
-    "next": "12",
+    "next": "latest",
```
wlee221, line 30: Should we set this to `^12.0.0` so it doesn't (one day) bump up to v13 that Amplify Console wouldn't support?

yarn.lock
```diff
@@ -15884,7 +15893,7 @@ next-tick@~1.0.0:
   resolved "https://registry.yarnpkg.com/next-tick/-/next-tick-1.0.0.tgz#ca86d1fe8828169b0120208e3dc8424b9db8342c"
   integrity sha1-yobR/ogoFpsBICCOPchCS524NCw=
 
-next@12, next@^11.0.1:
+next@^11.0.1, next@latest:
   version "11.1.2"
```
wlee221, line 15897: Looks like the resolved version is still pointing to 11.1.2. Should we update https://github.com/aws-amplify/amplify-ui/blob/fe5f2d88cba29b5eff8ca99df8c10c7b1c5f9e1b/package.json#L39 as well?

And if we go from 11 to 12, do we need to make adjustment to Amplify Console to reflect the update? I think this is more of a question to @ErikCH or @ericclemmons who has worked on hosting on console side.

docs/package.json
reesscot: Updated to ^12.0.7 since the latest has the security fixes in it.

yarn.lock
reesscot: My understanding was that we had already upgraded to 12 in https://github.com/aws-amplify/amplify-ui/pull/895. I don't think anything is currently using 11, so I updated this to `^12.0.7`. Let me know if I'm still misunderstanding. 

It may also makes sense to remove this resolutions entry, as I don't know what it's used for anymore. I'm guessing it was necessary before when we had to match the main docs site next js version. @ericclemmons @wlee221 thoughts?

reesscot: Hmm, but on `main` I'm seeing  `"version": "11.1.2",` in node_modules. Is that because of this resolutions entry?

wlee221: Yeah, I think it makes sense to remove that resolution entry in favor of explicit dependencies on `docs/` and `examples/next`.

package.json
```diff
@@ -36,7 +36,7 @@
     "fs-extra": "^10.0.0",
     "jest": "^26.6.3",
     "json-schema": "^0.4.0",
-    "next": "^11.0.1",
+    "next": "^12.0.7",
```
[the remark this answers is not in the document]
wlee221: Let's get rid of resolutions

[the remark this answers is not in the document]
ericclemmons: Yes, if `next` is explicitly defined in the other packages then `resolutions` is redundant.

I'm happy to get this resolved ASAP, since we can always adjust later.

reesscot:
Closing in favor of #966

PR #968, opened 2021-12-09
Adding analytics to docs site

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Adding in the tracking code from the docs repo with some tweaks. I can see the analytics requests going out for each page visit:

CleanShot 2021-12-13 at 16 01 50@2x


docs/src/pages/_app.page.tsx
```diff
@@ -19,6 +20,9 @@ function MyApp({ Component, pageProps }) {
   const [colorMode, setColorMode] = React.useState<ColorMode>('system');
   const [themeOverride, setThemeOverride] = React.useState('');
 
+  configure();
```
dbanksdesign, line 23: This has a check to see if it is configured yet, so running it for each page render is ok?

docs/src/pages/_document.page.tsx
```diff
@@ -25,16 +25,20 @@ const getCSPContent = (context: Readonly<HtmlProps>) => {
       style-src 'self' 'unsafe-inline';
       font-src 'self' data:;
       frame-src *.codesandbox.io;
-      script-src 'unsafe-eval' 'self' '${cspInlineScriptHash}'
+      img-src 'self' cm.everesttech.net amazonwebservices.d2.sc.omtrdc.net;
+      connect-src 'self' *.shortbread.aws.dev dpm.demdex.net;
+      script-src 'unsafe-eval' 'self' '${cspInlineScriptHash}' a0.awsstatic.com;
```
dbanksdesign, line 30: These are all what the tracking stuff needs. I figured these out by building, seeing an error, adding the domain to the proper place, repeat.

dbanksdesign (commented):
Take a look around.

docs/src/utils/track.ts
```diff
@@ -0,0 +1,161 @@
+// https://github.com/aws-amplify/docs/blob/main/src/utils/track.ts
+let configured = false;
+
+declare global {
+  interface Window {
+    s: any;
```
hvergara, line 6: can we use a different variable name? `s` seems to be a little bit confusing

hvergara (approved):

docs/src/pages/_document.page.tsx
```diff
@@ -25,16 +25,20 @@ const getCSPContent = (context: Readonly<HtmlProps>) => {
       style-src 'self' 'unsafe-inline';
       font-src 'self' data:;
       frame-src *.codesandbox.io;
-      script-src 'unsafe-eval' 'self' '${cspInlineScriptHash}'
+      img-src 'self' cm.everesttech.net amazonwebservices.d2.sc.omtrdc.net;
```
reesscot, line 28: Do you want to add `dpm.demdex.net`; here as well?
```suggestion
      img-src 'self' cm.everesttech.net amazonwebservices.d2.sc.omtrdc.net dpm.demdex.net;
```

docs/src/utils/track.ts
reesscot: Pretty sure `s` is what Adobe analytics requires...

reesscot (approved):

docs/src/pages/_app.page.tsx
```diff
@@ -19,6 +20,9 @@ function MyApp({ Component, pageProps }) {
...
+  trackPageVisit();
```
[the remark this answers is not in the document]
reesscot: Should we wrap this in a conditional to only run in prod mode?

docs/src/utils/track.ts
dbanksdesign: Yup that variable is not set by us

Issue #974, opened 2021-12-10
Add new documentation for Validate/modify formData on SignUp 


ErikCH:
### On which framework/platform would you like to see this feature implemented?

React, Angular, Vue

### Which UI component is this feature-request for?

Authenticator

### Please describe your feature-request in detail.

Need documentation for feature #836 

### Please describe a solution you'd like.

Add documentation to docs site

### We love contributors! Is this something you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

ErikCH:
This has been merged!

PR #975, opened 2021-12-10
Validate/modify formData on SignUp 

ErikCH:
*Issue #, if available:*
#836

*Description of changes:*
Added additional ways to override common submissions for the following:

handleSignUp: SignUp
handleSignIn: Sign In
handleConfirmSignIn: Confirm Sign In
handleConfirmSignUp: Confirm Sign Up
handleForgotPasswordSubmit: Forgot Password Submit
handleForgotPassword: Forgot Password

I'll add docs for this feature in #974

### DIFF
--- a/examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
+++ b/examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
@@ -1,5 +1,5 @@
 import { Component } from '@angular/core';
-import { Amplify } from 'aws-amplify';
+import { Amplify, Auth } from 'aws-amplify';
 import awsExports from './aws-exports';
 
 @Component({
@@ -10,4 +10,20 @@ export class SignUpWithEmailComponent {
   constructor() {
     Amplify.configure(awsExports);
   }
+
+  services = {
+    async handleSignUp(formData: Record<string, any>) {
+      const { username, password, attributes } = formData;
+      // custom username
+      const newAttributes = {
+        ...attributes,
+        email: attributes.email.toLowerCase(),
+      };
+      return Auth.signUp({
+        username: username.toLowerCase(),
+        password,
+        attributes: newAttributes,
+      });
+    },
+  };
 }
--- a/examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
+++ b/examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
@@ -13,16 +13,14 @@ export class SignUpWithEmailComponent {
 
   services = {
     async handleSignUp(formData: Record<string, any>) {
-      const { username, password, attributes } = formData;
+      let { username, password, attributes } = formData;
       // custom username
-      const newAttributes = {
-        ...attributes,
-        email: attributes.email.toLowerCase(),
-      };
+      username = username.toLowerCase();
+      attributes.email = attributes.email.toLowerCase();
       return Auth.signUp({
-        username: username.toLowerCase(),
+        username,
         password,
-        attributes: newAttributes,
+        attributes,
       });
     },
   };
--- a/packages/e2e/cypress/integration/common/shared.ts
+++ b/packages/e2e/cypress/integration/common/shared.ts
@@ -31,9 +31,13 @@ Given(
       throw error;
     }
 
-    cy.intercept(routeMatcher, { fixture });
+    cy.intercept(routeMatcher, { fixture }).as('route');
   }
 );
+
+Given('I verify the body has {string} included', (value: string) => {
+  cy.wait('@route').its('request.body.Username').should('include', value);
+});
 
 Given(
   'I intercept {string} with error fixture {string}',
@@ -79,6 +83,13 @@ When('I type an invalid password', () => {
 When('I type a new {string}', (field: string) => {
   cy.findInputField(field).typeAliasWithStatus(field, `${Date.now()}`);
 });
+
+When(
+  'I type a new {string} with value {string}',
+  (field: string, value: string) => {
+    cy.findInputField(field).type(value);
+  }
+);
 
 When('I click the {string} tab', (label: string) => {
   cy.findByRole('tab', {
--- a/packages/e2e/features/ui/components/authenticator/sign-up-with-email.feature
+++ b/packages/e2e/features/ui/components/authenticator/sign-up-with-email.feature
@@ -12,6 +12,16 @@ Feature: Sign Up with Email
     And I don't see "Phone Number" as an input field
 
   @angular @react @vue  
+  Scenario: Sign up with a new email & password and lowercase the email 
+    Given I intercept '{ "headers": { "X-Amz-Target": "AWSCognitoIdentityProviderService.SignUp" } }' with fixture "sign-up-with-email"
+    When I type a new "email" with value "TEST@example.com"
+    And I type my password
+    And I confirm my password
+    And I click the "Create Account" button
+    And I verify the body has "test@example.com" included
+    Then I see "Confirmation Code"
+
+@angular @react @vue  
   Scenario: Sign up with a new email & password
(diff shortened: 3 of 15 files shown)


packages/e2e/features/ui/components/authenticator/sign-up-with-email.feature
```diff
@@ -12,6 +12,16 @@ Feature: Sign Up with Email
...
 
   @angular @react @vue  
+  Scenario: Sign up with a new email & password and lowercase the email 
+    Given I intercept '{ "headers": { "X-Amz-Target": "AWSCognitoIdentityProviderService.SignUp" } }' with fixture "sign-up-with-email"
+    When I type a new "email" with value "TEST@example.com"
+    And I type my password
+    And I confirm my password
+    And I click the "Create Account" button
+    And I verify the body has "test@example.com" included
+    Then I see "Confirmation Code"
```
ErikCH, line 22: This verifies that the username is lowercased

packages/ui/src/machines/authenticator/actors/resetPassword.ts
```diff
@@ -17,183 +16,191 @@ import {
...
-            entry: [sendUpdate(), 'setUsername', 'clearError'],
-            invoke: {
-              src: 'resetPassword',
-              onDone: {
-                target: '#resetPasswordActor.confirmResetPassword',
+export type ResetPasswordMachineOptions = {
+  services?: Partial<typeof defaultServices>;
+};
+
+export function resetPasswordActor({ services }: ResetPasswordMachineOptions) {
```
ErikCH, line 23: Only change here is adding in the `services`

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -11,6 +11,53 @@ export const defaultServices = {
     return Auth.currentAuthenticatedUser();
   },
 
+  async handleSignUp(formData): Promise<any> {
+    return Auth.signUp(formData);
+  },
```
ericclemmons, line 16: I knew eventually we'd have to solve some of this tech debt :(

This is similar the original design for what `signUp` (and `signIn`, etc.) should look like for customers to replace:

```
async signUp({ username, password, attributes }) {
  return Auth.signUp({ username, password, attributes });
}
```

What we're missing is for this transform to happen _before_ calling the `signUp` or `signIn` services.

Even https://github.com/aws-amplify/amplify-ui/pull/911/files#diff-7746662de6b4ab9d0242a764c3f8ee8b1c40f29db8f3480989d14f8370d5bdd8R276 is too complex – these `services` should be mostly a pass-through for customers to mutate the data as you're doing here.

My recommendation is that `invoke: { src: 'signUp' }` instead passes the normalized `data`:
> https://xstate.js.org/docs/guides/communication.html#the-invoke-property

I see why you had to introduce a new service like `handle*` – we let our services grow too complex rather than being 1-line passthroughs to `Auth.*` methods.

wlee221: Fyi I've looked into using `invoke` with `data`: this option is only applicable when you're invoking a machine. 

> `data` - (optional, used only when invoking machines) 
  [adsf](https://xstate.js.org/docs/guides/communication.html#the-invoke-property)

ericclemmons: Whoops!  Well, through whatever means, what I was hoping for was:

- Our `signIn` / `signUp` services are what customers override
- These services share type definitions with the underlying `Auth.*` call
- These services, when replaced, **don't** lose functionality/behavior. (`signUp` today has an allowlist)

When typing this out, I had a realization – **just because they pass `signUp` doesn't mean that replaces the entire `services.signUp` method**.

ErikCH: @ericclemmons What sort of ways should we accomplish this without having the whole service overwrite `services.signUp` ?

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -31,10 +31,14 @@ Given(
...
 
-    cy.intercept(routeMatcher, { fixture });
+    cy.intercept(routeMatcher, { fixture }).as('route');
   }
 );
 
+Given('I verify the body has {string} included', (value: string) => {
+  cy.wait('@route').its('request.body.Username').should('include', value);
+});
+
```
ErikCH, line 41: `@route` is given with the previous intercept

examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
```diff
@@ -10,4 +10,20 @@ export class SignUpWithEmailComponent {
   constructor() {
     Amplify.configure(awsExports);
   }
+
+  services = {
+    async handleSignUp(formData: Record<string, any>) {
```
ericclemmons, line 15: Is there a better Type or Interface that customers can depend on for this `formData`?

We know that this will be something like `LoginMechanism | SignUpAttributes`.  Or because this can be _anything_, they should use `Record`?

examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
```diff
@@ -10,4 +10,20 @@ export class SignUpWithEmailComponent {
...
+      // custom username
+      const newAttributes = {
+        ...attributes,
+        email: attributes.email.toLowerCase(),
+      };
+      return Auth.signUp({
+        username: username.toLowerCase(),
+        password,
+        attributes: newAttributes,
+      });
```
ericclemmons, line 26: Would mutations make this example seem a bit clearer (mainly so that that `Auth.signUp` signature is straightforward)?

```suggestion
      let { username, password, attributes } = formData;
      
      username = username.toLowerCase();
      attributes.email = attributes.email.toLowerCase()
      
      return Auth.signUp({ username, password, attributes });
```

Or would something like this pose a risk because we're not explicitly immutable within the machine?

packages/ui/src/machines/authenticator/signUp.ts
```diff
@@ -261,7 +261,7 @@ export function createSignUpMachine({ services }: SignUpMachineOptions) {
             get(user, 'username') || get(authAttributes, 'username');
           const { password } = authAttributes;
 
-          await Auth.confirmSignUp(username, code);
+          await services.handleConfirmSignUp({ username, code });
```
ericclemmons, line 264: Ahh, I see the mistake I made – custom handlers like `validateSignUp` were being passed as `services` **but aren't actually XState compatible services** with `(context, event)` signatures:
> https://xstate.js.org/docs/guides/communication.html#invoking-promises

Maybe what we should've done is have 1st class properties like https://github.com/NearHuscarl/flutter_login#installation:

```js
<Authenticator
  initialState="signUp"
  handleSignUp={customSignUp}
  // or
  onSignUp={customSignUp}
  validateSignUp={customValidator)
/>
```

The other problem with our XState design is that we have to pass `services` around to create new machines.  Whereas technically these may make more sense on `context` since they're not actual `services`.

Technically to customers `services` is just a wrapper like `components` around common pieces – we just have to recognize internally that some things are actual `services` vs. custom handler functions.

ericclemmons (approved):
Thanks for your patience with me @ErikCH!

This is a great PR 💪 

If there are any further changes, we can address those in #974.

Also, this PR has shown me where I was confused with XState `services` vs. non-XState handlers.

examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
ErikCH: Yeah, I could make this change.

packages/ui/src/machines/authenticator/signUp.ts
ErikCH: Yeah, that could be something we do. But I thought we didn't want to pollute the authenticator with more props?

examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts
ErikCH: Let me think about this, maybe we can.

Issue #984, opened 2021-12-13
Add Sitemap for Docs website


hvergara:
We need to create a static sitemap to help crawlers to index our docs website

https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap

PR #990, opened 2021-12-14
Add sitemap to Docs website

hvergara:
*Issue #, if available:*
#984 

*Description of changes:*
Add a sitemap to docs website using `next-sitemap` library


docs/package.json
```diff
@@ -6,6 +6,7 @@
     "dev": "next dev -p 5000",
     "build": "next build && next export",
     "lint": "next lint",
+    "postbuild": "next-sitemap",
```
ericclemmons, line 9: The build is failing likely because this isn't being executed with the correct `$PATH`:

```suggestion
    "postbuild": "yarn next-sitemap",
```

hvergara: oof, added now. thanks!

docs/package.json
```diff
@@ -29,6 +30,7 @@
     "mdx-prism": "^0.3.3",
     "next": "12",
     "next-plugin-preval": "^1.2.1",
+    "next-sitemap": "^1.6.203",
```
dbanksdesign, line 33: Do we need to declare this dependency is we are using `npx` in the script? Would `npx` always pull the latest version and not use the version defined here?

dbanksdesign (commented):
Just one question about `npx` and listing it as a dependency. Other than that looks good

docs/package.json
zchenwei: +1

reesscot: My understanding of `npx` was that it would first look locally to see if the package exists in `node_modules` and then only download from npm if it's not present.

reesscot (approved):
LGTM

PR #992, opened 2021-12-14
CI: Get ref before checkout

wlee221:
*Issue #, if available:*

*Description of changes:* tests/setup was checking out the wrong `ref` because `${{ steps.get-ref.outputs.ref }}` were undefined at that step. Reordering steps to correct it.


ErikCH (approved):
LGTM

Issue #995, opened 2021-12-15
Not asking for TOTP setup after FORCE_CHANGE_PASSWORD when no phone number available


AntonBarwald:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Vue

### Which UI component?

Authenticator

### Please describe your bug.

1. Use a Cognito User Pool with MFA required (with both MFA method, Authenticator apps and SMS message)
2. Create a user with email only
3. Signing in with a user with a status of FORCE_CHANGE_PASSWORD, redirected to the change password screen. 
4. Logged in state despite no MFA used (that is required on the Cognito pool)
5. Login again user are presented with the view to setup a TOTP device.

### What's the expected behaviour?

1. Signing in with a user with a status of FORCE_CHANGE_PASSWORD, redirected to the change password screen. 
2. When password is updated, user are presented with the view to setup a TOTP device.

### Help us reproduce the bug!

1. Use a Cognito User Pool with MFA required
2. Create a user with email only
3. Login using the Amplify based on the steps in "Please describe your bug." - section

### Additional information and screenshots

@aws-amplify/ui@3.0.5
@aws-amplify/ui-vue@2.0.5

Relates to https://github.com/aws-amplify/amplify-ui/pull/986

ErikCH:
Hi @AntonBarwald !

Thanks for adding this in!

In #986 I added in the functionality for SMS, but not SETUP TOTP. I went ahead and just pushed up a PR for that. We should get that in shortly!

ErikCH:
This has been merged but not released yet. Since it's with the `ui` library, I believe a new install should work. Otherwise, we'll have a new release on this soon.

PR #997, opened 2021-12-15
Force Change Password should also check for SETUP TOTP

ErikCH:
*Issue #, if available:*
#995 

*Description of changes:*
Updated code so it also checks for SETUP TOTP after force change password, not just SMS MFA

### DIFF
--- a/packages/ui/src/machines/authenticator/actors/signIn.ts
+++ b/packages/ui/src/machines/authenticator/actors/signIn.ts
@@ -231,7 +231,11 @@ export function signInActor({ services }: SignInMachineOptions) {
                         actions: ['setUser', 'setChallengeName'],
                         target: '#signInActor.confirmSignIn',
                       },
-
+                      {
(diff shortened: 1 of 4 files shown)


packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -231,7 +231,11 @@ export function signInActor({ services }: SignInMachineOptions) {
                         actions: ['setUser', 'setChallengeName'],
                         target: '#signInActor.confirmSignIn',
                       },
-
+                      {
+                        cond: 'shouldSetupTOTP',
+                        actions: ['setUser', 'setChallengeName'],
+                        target: '#signInActor.setupTOTP',
+                      },
```
ErikCH, line 238: we should also check for SetupTOTP not just SMS

wlee221 (approved):
:shipit:

PR #999, opened 2021-12-15
Docs fix: add missing dep @aws-amplify/ui-react required by Webpack 5

reesscot:
*Description of changes:*
This fixes an issue where the Amplify Hosting build is failing when using Webpack 5 because the docs/package.json doesn't explicitly declare the `@aws-amplify/ui-react` dependency.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -10,6 +10,7 @@
     "test": "$_ run build"
   },
   "dependencies": {
+    "@aws-amplify/ui-react": "^2.1.5",
     "@aws-amplify/ui-react-v1": "npm:@aws-amplify/ui-react",
     "@codesandbox/sandpack-react": "0.1.9",
     "@cucumber/gherkin": "^19.0.3",
--- a/amplify.yml
+++ b/amplify.yml
@@ -7,7 +7,7 @@ applications:
             - nvm install 14
             - nvm use 14
             - node -v
-            - (cd .. && yarn install)
+            - (cd .. && yarn install && yarn build)
         build:
           commands:
             - nvm install 14
@@ -17,7 +17,7 @@ applications:
       artifacts:
         baseDirectory: out
         files:
-          - '**/*'
+          - "**/*"
       cache:
         paths:
           - node_modules/**/*


docs/package.json
```diff
@@ -10,6 +10,7 @@
     "test": "$_ run build"
   },
   "dependencies": {
+    "@aws-amplify/ui-react": "^2.1.5",
```
ericclemmons, line 13: @wlee221 I'm curious if Changesets will automatically update this dependency 🤔 

Alternatively, this could be

```suggestion
    "@aws-amplify/ui-react": "latest",
```

or

```suggestion
    "@aws-amplify/ui-react": "*",
```

Whichever works for both local development (by not actually installing the NPM dependency) and Amplify Hosting.

ericclemmons (approved):
Approved, contingent on https://pr-999.d3inl0muob87le.amplifyapp.com/ successfully deploying.

And it looks like this works fine with `yarn workspaces` because `yarn.lock` wasn't affected.

ericclemmons:
One reason why the build could still be failing is that **packages aren't being built first!**

https://github.com/aws-amplify/amplify-ui/blob/382e24fa6e68dc2527ea2ee614e69e6d2ef29c2d/amplify.yml#L10

We need an additional `yarn build` since #927 removed this by default from `yarn install`.

docs/package.json
wlee221: Changesets will! It keeps track of cross dependencies. We can do a sanity check by 

1) add changeset to bump versions:

```
cp .github/changeset-presets/bump-versions.md .changeset
```

2) version packages

```
yarn changeset version
```

3) Check that docs' `@aws-amplify/ui-react` version is bumped

wlee221 (approved):
LGTM, thanks for the catch @reesscot

docs/package.json
wlee221: I'll do that while we wait for preview deployment.

wlee221: Mocked version bump here: https://github.com/aws-amplify/amplify-ui/compare/add-missing-react-dependency...add-missing-react-dependency-test-version-bump. Looks good!

wlee221: Also confirmed that `node_modules/@aws-amplify/ui-react` symlinks to local package.

ericclemmons: Thanks @wlee221 !  Let's keep this as-is then 👍

PR #1006, opened 2021-12-16
Added docs to override calls

ErikCH:
*Issue #, if available:*
#974

*Description of changes:*
Added docs for React/Vue/Angular to override function calls

### DIFF
--- a/docs/src/pages/components/authenticator/index.page.mdx
+++ b/docs/src/pages/components/authenticator/index.page.mdx
@@ -236,6 +236,8 @@ For your [configured social providers](https://docs.amplify.aws/lib/auth/social/
   <Authenticator socialProviders={['amazon', 'apple', 'facebook', 'google']} />
 </Example>
 
+_[Step by step video](https://youtu.be/8KwZNn56F78) on setting up social providers._
+
 ### `variation`
 
 The Authenticator has multiple variations to meet the needs of your application.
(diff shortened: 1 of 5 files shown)


docs/src/pages/components/authenticator/index.page.mdx
```diff
@@ -236,6 +236,8 @@ For your [configured social providers](https://docs.amplify.aws/lib/auth/social/
   <Authenticator socialProviders={['amazon', 'apple', 'facebook', 'google']} />
 </Example>
 
+_[Step by step video](https://youtu.be/8KwZNn56F78) on setting up social providers._
```
reesscot, line 239: Should we embed the video rather than linking off to it?

ErikCH: I just want it linked right now instead of embedding it. That way it doesn't distract from the rest of the help articles.

docs/src/pages/components/authenticator/index.page.mdx
```diff
@@ -371,6 +373,29 @@ The following example customizes these slots with:
...
+### Override Function Calls
+
+You can override the call to `signUp`, `signIn`, `confirmSignIn`, `confirmSignUp`, `forgotPassword` and `forgotPasswordSubmit` functions.
+To override a call you must create a new `services` object with an `async` `handle*` function that returns an `aws-amplify` `Auth.*` promise.
+
+The service object must then be passed into the `authenticator` component as a `services` prop. For example, let's imagine you'd like to lowercase the `username` and the `email` attributes during `signUp`.
+This would be overriden like so:
+
+<Fragment>
+  {({ platform }) => import(`./overrides/username.${platform}.mdx`)}
```
reesscot, line 385: Question on this example. It's not obvious to me from just reading the example where the `Auth.signUp` function is coming from. Should we include the `import {Auth} from 'aws-amplify'` in this example?

ErikCH: Yeah, that's a good idea, I'll add that.

ErikCH: Ok that's back in, I'm just showing the whole file, like we do in Angular and Vue.

reesscot (approved):
LGTM

PR #1007, opened 2021-12-16
Don't use "username" as input type attribute for Reset Password

rvmourik:
*Issue #, if available:*

*Description of changes:*
The reset password screen has a HTML input field for entering the username (or other). The Label Text is dynamic and coming from the getAliasFromContext Helper (Never the less the label represents the type rather than an actual label).

As far as I could see the alias type can be **username**, **email** or **tel**, the last two are valid HTML input type's according to the docs:

https://www.w3schools.com/html/html_form_input_types.asp

So I added a check for checking if the type equals "username", if this is the case using a `input[type=text]` seems more appropriate.


reesscot:
Thank you for your contribution. This code has since been refactored. Feel free to open a new PR if there is still an issue that needs to be addressed.

PR #969, opened 2021-12-09
Pin @radix-ui/react-id dependency to 0.1.1

hvergara:
*Issue #, if available:*
#967 
https://github.com/radix-ui/primitives/issues/1026

*Description of changes:*
One of our dependencies (`@radix-ui/react-id`) stopped exporting the `IdProvider` symbol in version `0.1.2` causing errors on customers trying to build their projects. This PR locks the dependency verison to `0.1.1` until we have a fix for this issue.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1010, opened 2021-12-17
Remove IdProvider and port useId from radix

reesscot:
*Description of changes:*
This PR removes the `IdProvider` and `@radix-ui/react-id` direct dependency from the repo.  In addition the `useId` function from `@radix-ui/react-id` has been ported over to `useStableId` to prevent changes to radix unexpectedly breaking our components. I've also bumped up the Radix dependencies used by Tabs, Expander, Slider, and Menu primitives to the latest versions, since the newer versions don't require the `IdProvider`.

Updating the `@radix-ui/react-tabs` dependency from 0.1.1 to 0.1.2 also caused the Tabs DOM element to change from DIV to BUTTON, which is more accessible. Updated tests and types accordingly.

See: 
https://github.com/aws-amplify/amplify-ui/pull/969
https://github.com/radix-ui/primitives/issues/1026

TextField with auto-generated id
Screen Shot 2022-01-05 at 2 37 33 PM

SliderField with auto-generated id
Screen Shot 2022-01-05 at 11 55 27 AM

Expander with auto-generated id
Screen Shot 2022-01-05 at 11 49 52 AM

_TODO_

- [x] Smoke test Tabs, Expander, Slider, and Menu components


packages/react/src/primitives/utils/useStableId.ts
```diff
@@ -0,0 +1,31 @@
+// Adapted from https://github.com/radix-ui/primitives/blob/main/packages/react/id/src/id.tsx#L8
+import * as React from 'react';
+import { useLayoutEffect } from './useLayoutEffect';
+
+// Prefixed autogenerated id created by useStableId
+export const AUTO_GENERATED_ID_PREFIX = 'amplify-id';
+
+export type UseStableId = (id?: string) => string;
```
hvergara, line 8: do we need to export it? there's no other references to `UseStableId` type on this PR

packages/react/src/primitives/utils/useLayoutEffect.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as React from 'react';
+
+/**
+ * On the server, React emits a warning when calling `useLayoutEffect`.
+ * This is because neither `useLayoutEffect` nor `useEffect` run on the server.
+ * We use this safe version which suppresses the warning by replacing it with a noop on the server.
+ *
+ * See: https://reactjs.org/docs/hooks-reference.html#uselayouteffect
+ */
+const useLayoutEffect = Boolean(globalThis?.document)
```
hvergara, line 11: For SSR, we should call `useEffect` instead of a `noop` function. This is what some other popular libraries do:

https://github.com/reduxjs/react-redux/blob/d16262582b2eeb62c05313fca3eb59dc0b395955/src/components/connectAdvanced.js#L40
https://github.com/atlassian/react-beautiful-dnd/blob/master/src/view/use-isomorphic-layout-effect.js

packages/react/src/primitives/utils/useStableId.ts
```diff
@@ -0,0 +1,31 @@
...
+const useReactId = (React as any).useId || (() => undefined);
+let count = 0;
+
+/**
+ * Create a uuid to use with amplify fields unless
+ * an id is provided
+ * @param id user specified id
+ * @returns string
+ */
+export const useStableId: UseStableId = (id) => {
```
hvergara, line 21: if `id` is provided, can we return it asap?

packages/react/src/primitives/utils/useStableId.ts
```diff
@@ -0,0 +1,31 @@
...
+import { useLayoutEffect } from './useLayoutEffect';
+
+// Prefixed autogenerated id created by useStableId
+export const AUTO_GENERATED_ID_PREFIX = 'amplify-id';
+
+export type UseStableId = (id?: string) => string;
+
+// Create a local version of React.useId which will reference React.useId for React 18
+// and fallback to noop for React 17 and below
+const useReactId = (React as any).useId || (() => undefined);
```
hvergara, line 12: New React `useId` function always return a `string` > https://github.com/facebook/react/blob/4729ff6d1f191902897927ff4ecd3d1f390177fa/packages/react-reconciler/src/ReactFiberHooks.new.js#L2035

hvergara, line 12: Also, what about this? We can keep type safety if `useId` is present.
```suggestion
const useReactId = 'useId' in React ? React.useId : (() => undefined);
```

packages/react/src/primitives/utils/useStableId.ts
reesscot: It's used as the type of the `useStableId` function. My understanding is that if you export a function that uses a type, you must also export the type. Is that not the case?

packages/react/src/primitives/utils/useStableId.ts
reesscot: Looks great, will update.

packages/react/src/primitives/utils/useStableId.ts
reesscot: We can't early return before the hooks or we won't be following the rules of hooks. I don't think we can return it any earlier.

packages/react/src/primitives/utils/useLayoutEffect.ts
reesscot: I've seen that before too, but recently I've been wondering if it's a valid pattern. [useEffect is a noop on the server](https://github.com/reduxjs/react-redux/blob/d16262582b2eeb62c05313fca3eb59dc0b395955/src/components/connectAdvanced.js#L36) anyway, so why not be more declarative with a noop function instead? I've also seen warnings about `useEffect` when it's called on the server side, so this reducing unnecessary noise.

packages/react/src/primitives/utils/useStableId.ts
reesscot: Not sure if you're trying to suggest a change here. The reason the noop function returns `undefined` is to make sure the `id` is `undefined` on SSR and first render to prevent hydration mismatch issues.

reesscot: So, since we're still using React 17, and it doesn't have useId yet, and this causes a type error. I think I'd prefer to typecast as opposed to disabling type checking on this line. I think the risk is pretty low in this case. What do you think?

packages/react/src/primitives/utils/useStableId.ts
hvergara: is not mandatory, but if you feel it's safer to export it, I'm good with that :)

packages/react/src/primitives/utils/useLayoutEffect.ts
hvergara: cool, i'm fine with keeping it as noop

packages/react/src/primitives/utils/useStableId.ts
hvergara: aaaah, got it, thanks for the clarification

packages/react/src/primitives/utils/useStableId.ts
hvergara: right, resolving this comment

packages/react/src/primitives/utils/useStableId.ts
hvergara: I'm not a fan of ignoring types, but if there's no way to do it, i'm good

hvergara (approved):
LGTM

packages/react/src/primitives/SliderField/SliderField.tsx
```diff
@@ -6,12 +6,13 @@ import { ComponentClassNames } from '../shared/constants';
 import { FieldDescription, FieldErrorMessage } from '../Field';
 import { FieldGroup } from '../FieldGroup';
 import { Flex } from '../Flex';
-import { isFunction, useStableId } from '../shared/utils';
```
dbanksdesign, line 9: Non-blocking: should we just have one `utils` instead of `shared/utils` and `utils/`?

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { IdProvider } from '@radix-ui/react-id';
```
dbanksdesign, line 2: Removing code feels good 😄

dbanksdesign (approved):
Just 1 non-blocking comment. LGTM!

packages/react/src/primitives/SliderField/SliderField.tsx
reesscot: My goal is to move all the utils functions out of shared and to a `utils/` folder. I can do that in a follow up PR

Issue #1008, opened 2021-12-16
Amplify - Authentication / Cognito - React - Create User - Invalid phone number format


rdick:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### Please describe your bug.

![image](https://user-images.githubusercontent.com/62073529/146445852-bee37788-300d-427f-8ad6-f2847defb66f.png)

When I try to add a new user with AWS Cognito / Amplify portal it says the phone number is  an "Invalid phone number format". This is a bug. 

When I switch the country code to different value eg. +34 (instead of +1) and use the same phone number it it works

### What's the expected behaviour?

I should be able to use the +1 country code and my phone number (7804397870) and it should allow me to add a user

### Help us reproduce the bug!

![image](https://user-images.githubusercontent.com/62073529/146445852-bee37788-300d-427f-8ad6-f2847defb66f.png)
Please see picture. This is how the bug was created

### Code Snippet

I just followed the website code and completed the following

import Amplify from 'aws-amplify';
import awsconfig from './aws-exports';
import { Authenticator } from '@aws-amplify/ui-react';
import '@aws-amplify/ui-react/styles.css';

Amplify.configure(awsconfig);

.... much lower down in react component

			<Authenticator>
				{content}
			</Authenticator>

### Additional information and screenshots

![image](https://user-images.githubusercontent.com/62073529/146445852-bee37788-300d-427f-8ad6-f2847defb66f.png)
See picture

rdick:
Hi Scott! Thanks for looking into this.

It should be noted that this is NOT a problem with other country codes.

PR #1014, opened 2021-12-17
Authenticator: fix issue where phone_number validation fails

reesscot:
*Issue #, if available:*
Closes https://github.com/aws-amplify/amplify-ui/issues/1008

*Description of changes:*

This fixes an issue where phone number validation fails for country code `+1` when there are other form validation errors such as a short password. This causes the `phone_number` value to be mutated, and `N` number of `+1`'s to be appended (where N is number of times the form has been submitted):
```
UserAttributes: { phone_number: "+1+1+12223334444"}
Username: "+1+1+12223334444
```

The fix ensures that `formValues.phone_number` is not mutated on each `signUp` call.

### DIFF
--- a/packages/ui/src/machines/authenticator/signUp.ts
+++ b/packages/ui/src/machines/authenticator/signUp.ts
@@ -281,16 +281,22 @@ export function createSignUpMachine({ services }: SignUpMachineOptions) {
           const { formValues, loginMechanisms } = context;
           const [primaryAlias] = loginMechanisms ?? ['username'];
 
+          let phoneNumberWithCountryCode;
           if (formValues.phone_number) {
-            formValues.phone_number =
+            phoneNumberWithCountryCode =
               `${formValues.country_code}${formValues.phone_number}`.replace(
                 /[^A-Z0-9+]/gi,
                 ''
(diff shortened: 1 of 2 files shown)


ericclemmons (approved):

packages/ui/src/machines/authenticator/signUp.ts
```diff
@@ -281,16 +281,22 @@ export function createSignUpMachine({ services }: SignUpMachineOptions) {
           const { formValues, loginMechanisms } = context;
           const [primaryAlias] = loginMechanisms ?? ['username'];
 
+          let phoneNumberWithCountryCode;
           if (formValues.phone_number) {
-            formValues.phone_number =
```
ericclemmons, line 285: > The fix ensures that formValues.phone_number is not mutated on each signUp call.

You said it! Mutations are a huge risk!

I'm redoing #911, but that fixes `phone_number` to *always be a computed value in the form* so that this logic doesn't need to live in the machine.

ericclemmons (approved):

zchenwei (approved):
LGTM

Issue #1020, opened 2021-12-20
Failed to resolve '@aws-amplify/ui/styles.css' when using Parcel or Vite


julienben:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### Please describe your bug.

Hello! I'm trying to switch from Webpack to Vite or Parcel and having an issue in both when trying to `import '@aws-amplify/ui-react/styles.css'`.

I get the following error (in Parcel):

```
🚨 @parcel/core: Failed to resolve '@aws-amplify/ui/styles.css' from './node_modules/@aws-amplify/ui-react/dist/index.css'

/Users/myname/code/reponame/node_modules/@aws-amplify/ui-react/dist/index.css:1:9
> 1 | @import"@aws-amplify/ui/styles.css";
>   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  2 | 

🚨 @parcel/resolver-default: Cannot load file './@aws-amplify/ui/styles.css' in './node_modules/@aws-amplify/ui-react/dist'.
```

This line works (only in Parcel): `import '@aws-amplify/ui-react/node_modules/@aws-amplify/ui/dist/styles.css';`

Any idea why this might be happening? Thanks!!

### What's the expected behaviour?

The CSS should load with the snippet provided in the documentation.

### Help us reproduce the bug!

1. Create a React app with Vite or Parcel as the bundler
2. Try importing `@aws-amplify/ui-react/styles.css`
3. Get error message ;)

### Code Snippet

```typescript
// Put your code below this line.

```

reesscot:
Hi @julienben, thank for reporting this issue. I was able to reproduce this error using parcel, and it's now fixed on the main branch. It will be generally available when the next release gose out. In the meantime you can test out the fix by installing the change with the following pre-release tag:

`npm i @aws-amplify/ui-react@0.0.0-next-59d8b8e-20211120234739`

[...]

Issue #1005, opened 2021-12-16
Can't resolve './@aws-amplify/ui/styles.css'


nextdooroldwang:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### Please describe your bug.

```
 error  in ./node_modules/@aws-amplify/ui-react/dist/index.css

Module build failed (from ./node_modules/@umijs/bundler-webpack/lib/webpack/plugins/mini-css-extract-plugin/src/loader.js):
ModuleNotFoundError: Module not found: Error: Can't resolve './@aws-amplify/ui/styles.css' in '/node_modules/@aws-amplify/ui-react/dist'
```

### What's the expected behaviour?

Compiled successfully

### Help us reproduce the bug!

1.yarn add aws-amplify @aws-amplify/ui-react
2.import '@aws-amplify/ui-react/styles.css';
3.yarn start

**I opend the CSS Modules in my program。**

### Code Snippet

```typescript
import { Amplify } from 'aws-amplify';
import { Authenticator } from '@aws-amplify/ui-react';

import '@aws-amplify/ui-react/styles.css';
import awsExports from './aws-exports';
import styles from './index.less';

Amplify.configure(awsExports);

export default function Layout() {
    return (
      <Authenticator>
        {({ signOut, user }) => (
          <main>
            <h1>Hello {user.username}</h1>
            <div onClick={signOut}>Sign out</div>
          </main>
        )}
      </Authenticator>
    );
}

```

### Additional information and screenshots

    "@aws-amplify/ui-react": "^2.1.5",
    "aws-amplify": "^4.3.11",

Mini Showcase Repository(REQUIRED)
https://github.com/nextdooroldwang/umi-sample.git

ericclemmons:
So it looks like `@aws-amplify/ui-react/styles.css` is correctly resolving, but the `@import"@aws-amplify/ui/styles.css"` isn't:

> https://unpkg.com/browse/@aws-amplify/ui-react@2.1.5/dist/index.css

[...]

PR #1024, opened 2021-12-20
Build `ui` styles into `ui-react` package

reesscot:
*Issue #, if available:*
Fixes: https://github.com/aws-amplify/amplify-ui/issues/1020 and https://github.com/aws-amplify/amplify-ui/issues/1005

*Description of changes:*
Customers are experiencing an issue with the [parcel bundler](https://parceljs.org/recipes/react/) where the CSS statement `@import '@aws-amplify/ui/styles.css';` pointing to `ui` package cannot be resolved via `node_modules`. This PR fixes this by building the styles from `ui` into `ui-react` using `tsup`.

Testing TODO:

- [x] Test in ViteJS (Scott)
- [x] Test in Parcel (Scott)
- [x] Test in NextJS (Scott)
- [x] Test in CRA (Scott)
- [x] Test in UMJS (Erik)

### DIFF
--- a/packages/react/src/styles.css
+++ b/packages/react/src/styles.css
@@ -1,1 +1,1 @@
-@import '@aws-amplify/ui/styles.css';
+@import '../../ui/dist/styles.css';
--- a/packages/react/src/styles.ts
+++ b/packages/react/src/styles.ts
@@ -1,0 +1,1 @@
+import './styles.css';
--- a/.changeset/hungry-forks-play.md
+++ b/.changeset/hungry-forks-play.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+Build ui styles into ui-react package to fix [parcel](https://parceljs.org/) and [UMIJS](https://umijs.org/) apps.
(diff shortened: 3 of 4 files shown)


ErikCH (approved):
👍

packages/react/src/styles.ts
```diff
@@ -0,0 +1 @@
+import './styles.css';
```
dbanksdesign, line 1: OMG this is way cleaner than writing a build script to copy the file over. Bravo.

packages/react/src/styles.css
```diff
@@ -1 +1 @@
-@import '@aws-amplify/ui/styles.css';
+@import '../../ui/dist/styles.css';
```
dbanksdesign, line 1: If we are using tsup for the build script, do we still need to change this line? Just curious

dbanksdesign (approved):
LGTM! Just one question for my own ignorance.

packages/react/src/styles.css
reesscot: Unfortunately, no. If I don't change the path to point to the relative file, `tsup` will just copy it the import statement:
```
@import"@aws-amplify/ui/styles.css";
```
So, for `tsup` to copy the file, we must use the relative path.

PR #1025, opened 2021-12-20
Fix cats on 404

dbanksdesign:
*Issue #, if available:*

*Description of changes:* Fixing an issue with loading a cat photo on the 404 page. Now we are grabbing a random cat photo on the client side only so there isn't a flash of content and making sure we have a cat before rendering the image. 


reesscot (approved):
:shipit:

PR #1039, opened 2021-12-21
Theming docs cleanup - remove `createtheme` references

reesscot:
*Description of changes:*
There are multiple unnecessary references to `createTheme` in the [Theme docs ](https://ui.docs.amplify.aws/theming). The `AmplifyProvider` already runs `createTheme` to extend custom themes from the default theme, so it's redundant for customers to do this as well.


dbanksdesign (approved):
LGTM! :shipit:

Issue #1048, opened 2021-12-27
Text component doesn't show newlines


hvergara:
If a newline is present inside of a `<Text>` component, it should be rendered correctly. Customers can disable this behavior resetting the `white-space` CSS property for specific Text component instances.

PR #1049, opened 2021-12-27
Add white-space (pre-wrap) behavior for Text components

hvergara:
*Issue #, if available:*
#1048 

*Description of changes:*
Add a `pre-wrap` word-wrap behavior for `<Text>` components. Note: this only applies when the component is rendered as a HTML `p` tag.

### DIFF
--- a/packages/ui/src/theme/css/component/text.scss
+++ b/packages/ui/src/theme/css/component/text.scss
@@ -10,6 +10,10 @@
     display: inline;
   }
 
+  @at-root p#{&} {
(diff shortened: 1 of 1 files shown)


ErikCH (approved):
👍

wlee221 (commented):
For my understanding, why just for `p`?

reesscot:
What problem does this PR solve? I don't think this isn the default browser behavior of a `p` tag, so it seems like it will be unexpected for many `Text` component users.

Issue #1050, opened 2021-12-27
Vue: Improve Memory Allocation with useAuthenticator


ErikCH:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Vue

### Which UI component?

Authenticator

### How is your app built?

All

### Please describe your bug.

The `useAuthenticator` hook is relying on the `watchEffect` and the `reactive` variables in Vue. This works, however when reactive variables are run outside of a component sometimes they may cause memory problems, or a memory leak, since they are not always disposed off.

A solution to this is to use the [effectScope](https://v3.vuejs.org/api/effect-scope.html) API in vue. Or use the `vueuse` library `createSharedComposable` which handles memory when creating shared composables.

### What's the expected behaviour?

Memory is disposed of properly

### Help us reproduce the bug!

Using console.log in `useAuthenticator` in the `watchEffect` shows the effect being run many times. 

### Code Snippet

```typescript
// Put your code below this line.

```

ErikCH:
Merged

PR #1052, opened 2021-12-27
Vue: Fix memory allocation for useAuthenticator composable

ErikCH:
*Issue #, if available:*
#1050 

*Description of changes:*
Added wrapper to handle memory better when reusing the `useAuthenticator` composable in multiple vue instances. 

### DIFF
--- a/packages/vue/src/composables/useAuth.ts
+++ b/packages/vue/src/composables/useAuth.ts
@@ -3,6 +3,7 @@ import { ref, reactive, Ref, watchEffect } from 'vue';
 import { getServiceFacade } from '@aws-amplify/ui';
 import { facade } from './useUtils';
 import { InterpretService } from '@/components';
+import { createSharedComposable } from '@vueuse/core';
 
 const service = ref() as Ref<InterpretService>;
 let useAuthenticatorValue = reactive({ ...facade, send: '', state: '' }) as any;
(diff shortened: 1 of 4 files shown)


packages/vue/src/composables/useAuth.ts
```diff
@@ -3,6 +3,7 @@ import { ref, reactive, Ref, watchEffect } from 'vue';
 import { getServiceFacade } from '@aws-amplify/ui';
 import { facade } from './useUtils';
 import { InterpretService } from '@/components';
+import { createSharedComposable } from '@vueuse/core';
```
wlee221, line 6: That's an awesome library 👀

wlee221 (approved):

packages/vue/src/composables/useAuth.ts
ErikCH: It is! Lot's of great utilities in it!

reesscot:
Mind adding a little more detail in the description as to how this improves memory useage? Link to blog article or something?

ErikCH:
@reesscot I showed a pattern of using external composables on my YouTube channel and a commenter mentioned that I should look into how memory is being affected, since he's seen memory leaks with composables that use reactive properties.

After further research I learned about [effectScope](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md) and how it helps solve this problem by disposing `computed` and `watch` elements outside of a Vue instance.

I then looked at how the popular [VueUse](https://github.com/vueuse/vueuse) library was creating it's composables. It was using `useEffect` in several places. I then saw it had it's own `createSharedComposable` utility that wrapped itself around `effectScope`.

After adding this into my code base, it seemed that the `watchEffect` in `useAuthenticator` was being called a lot less than it did before.

Issue #1059, opened 2021-12-29
Responsive breakpoints return `base` between breakpoint configs 


reesscot:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

All

### Please describe your bug.

When refreshing the app a few pixels below a given breakpoint config, say the `medium` breakpoint (767px, 1px below the `medium` breakpoint), the `base` styling is erroneously being returned instead of the `small` breakpoint.

### What's the expected behaviour?

When rendering an app at 1px below a given breakpoint, the lower breakpoint styling config should be rendered.

### Help us reproduce the bug!

1. Go to https://ui.docs.amplify.aws/theming/responsive#object-syntax
2. Open devtools and trigger responsive mode
3. Change screensize to 767px and refresh the page
4. Notice that the background color is `base="red"` and not `small: 'orange',` like it should be

### Additional information and screenshots

![Screen Shot 2021-12-29 at 10 48 24 AM](https://user-images.githubusercontent.com/6165315/147697123-786e7836-cf86-4e61-a546-f7009a0a5d38.png)

ayush987goyal:
> few pixels below a given breakpoint config

The few pixels below is precisely upto 15px below the breakpoint config.

For example, the `xl` breakpoint is at `1280px` . So for screen widths b/w `1265px - 1279px`  (14px diff), the breakpoint is returned as base.

PR #1060, opened 2021-12-29
Update media query logic to return correct breakpoint

reesscot:
*Issue #, if available:*
Closes https://github.com/aws-amplify/amplify-ui/issues/1059

*Description of changes:*
Because `em`'s are calculated based on the base font size, there was a 14px gap below breakpoints where the default breakpoint of `base` is returned. This fixes the issue using `calc()` to calculate the max-width as the upper breakpoint minus `1px`. This allows us to continue using `em` breakpoints while returning the correct breakpoint styling for 1-14px below a given breakpoint.

**Before fix**
At 767px (1px below medium 768px), background color rendering **incorrectly** as `base: "red"`.

Screen Shot 2021-12-29 at 10 48 24 AM

**After fix**
At 767px (1px below medium 768px), background color rendering **correctly** as `small: "orange"`
Screen Shot 2021-12-29 at 12 03 42 PM

### DIFF
--- a/.changeset/dull-turkeys-juggle.md
+++ b/.changeset/dull-turkeys-juggle.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+Fix media query logic to return correct breakpoint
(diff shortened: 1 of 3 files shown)


hvergara (approved):
LGTM!

ErikCH (approved):
👍

wlee221 (approved):
Til!

Issue #1062, opened 2021-12-29
Importing @aws-amplify/ui-react/styles.css will override global font-family


lacion:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

### Please describe your bug.

Importing `@aws-amplify/ui-react/styles.css` needs to be done before any other custom CSS otherwise html -> font-family will be overridden and custom fonts defined in the Theme will never take effect

### What's the expected behaviour?

Creating a custom theme with a font definition should change the UI font and not conflict with my own app CSS

```javascript
const myTheme = {
  name: 'my-theme',
  tokens: {
    fonts: {
      default: {
        variable: { value: 'Raleway' },
        static: { value: 'Raleway' },
      },
    },
  },
};
```

### Help us reproduce the bug!

Simply import `@aws-amplify/ui-react/styles.css` after any other CSS and Amplify UI CSS html {font-family} CSS definition will be the one use globally

### Code Snippet

```javascript
const myTheme = {
  name: 'my-theme',
  tokens: {
    fonts: {
      default: {
        variable: { value: 'Raleway' },
        static: { value: 'Raleway' },
      },
    },
  },
};
```

PR #1065, opened 2021-12-30
Move font-family rule to [data-amplify-theme]

reesscot:
*Issue #, if available:*
Fixes https://github.com/aws-amplify/amplify-ui/issues/1062

*Description of changes:*
This fixes an issue where defining a fonts token in custom theme didn't override the default Inter font. This was occurring because the `font-family: var(--amplify-fonts-default-static)` was only defined at the root. The `AmplifyProvider` converts custom theme tokens to CSS variables at the [data-amplify-theme] node and below. Since the rule was at the root node level, `--amplify-fonts-default-static` still had the original value of `Inter`, so the custom font token (defined at the [data-amplify-theme] node) wouldn't be read.

Solution:
Moving it to the [data-amplify-theme] ensures that the custom font token CSS variable is picked up correctly.
Screen Shot 2021-12-29 at 5 05 11 PM

Screen Shot 2021-12-29 at 5 13 09 PM

*Testing*
Manually tested using my [demo app in Webpack 5](https://github.com/reesscot/amplify-ui-build-systems/tree/css-import-order/react/webpack5) and confirmed that Open Sans is being rendered.

### DIFF
--- a/.changeset/fuzzy-gifts-punch.md
+++ b/.changeset/fuzzy-gifts-punch.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui": patch
+---
+
+Fix issue where custom theme fonts are ignored
--- a/packages/ui/src/theme/css/styles.scss
+++ b/packages/ui/src/theme/css/styles.scss
@@ -4,11 +4,11 @@
 
 @import 'dist/theme';
 
-html {
+[data-amplify-theme] {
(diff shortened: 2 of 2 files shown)


zchenwei (approved):
LGTM!

wlee221 (approved):

dbanksdesign (approved):

PR #1066, opened 2021-12-31
chore: removing unnecessary conditional check in useTheme

zchenwei:
*Description of changes:*

The conditional check inside `useTheme` is redundant for the below 2 reasons.

1. We created the context with `defaultTheme` as the `defaultValue`. The `defaultValue`  of a context will be used by React when a component does not have a matching Provider above it in the tree. 
```
export const AmplifyContext = React.createContext<AmplifyContextType>({
  theme: defaultTheme,
});
```

2. Inside `AmplifyProvider`, `theme` prop defaults to be `defaultTheme` if it is not supplied.
```
export function AmplifyProvider({
  children,
  colorMode,
  theme = defaultTheme,
}: AmplifyProviderProps) {
  const webTheme = createTheme(theme);
.....
```

Tests:

The existing test cases have cover all possible scenarios, for which the conditional check aims. By removing the check, tests still pass successfully.

Screen Shot 2021-12-30 at 5 02 45 PM

### DIFF
--- a/packages/react/src/hooks/useAmplify.ts
+++ b/packages/react/src/hooks/useAmplify.ts
@@ -3,15 +3,13 @@ import * as React from 'react';
 import { Theme } from '@aws-amplify/ui';
 
 import { AmplifyContext } from '../components/AmplifyProvider/AmplifyContext';
-import { getThemeFromContext } from './useTheme';
 
 interface UseAmplifyOutput {
   theme: Theme;
 }
 
 export function useAmplify(): UseAmplifyOutput {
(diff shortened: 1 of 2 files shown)


slaymance (approved):
Nice, this is good cleanup!

zchenwei:
Had a discussion with @hvergara, for some reason, context cannot resolved correctly when components are not wrapped within an `AmplifyProvider`. Getting this merged in will require a full e2e test with Studio. Let's hold it off and revisit it anytime later since the cleanup is not a must

PR #1082, opened 2022-01-04
Angular + Vue: Fix Fonts

ErikCH:
*Description of changes:*
After we moved the fonts to `[data-amplify-theme]` in #1065 this caused the fonts in Vue and Angular to default back to Times New Roman.

To fix this, I added back these fonts only for Vue and Angular. When Vue and Angular implement their own theming/ component system, this can be deleted. 

Before:
image

After:

image

### DIFF
--- a/.changeset/orange-toys-remain.md
+++ b/.changeset/orange-toys-remain.md
@@ -1,0 +1,6 @@
+---
+'@aws-amplify/ui-vue': patch
+'@aws-amplify/ui-angular': patch
+---
+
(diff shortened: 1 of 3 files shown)


zchenwei (approved):
LGTM! Mind adding some comments above there to ensure that we will not lose the context when revisiting?

ErikCH:
> LGTM! Mind adding some comments above there to ensure that we will not lose the context when revisiting?

Sure!

wlee221 (approved):
Thanks!

Issue #881, opened 2021-11-25
Setup TOTP UI - how to read QR code when it's on your phone's screen


ClaudioVR:
Hi

I'm working on a Vue app which uses amplify-ui's authenticator component with MFA activated. We have run into a bit of a QR conundrum. 

The QR code displayed in the setup totp component works fine if the user is working on a desktop and uses their phone to take a picture of the QR code.

However, if the user is working on their phone, they are suddenly faced with a QR code on their own screen. Although not impossible to do, reading the QR code in this situation can become tedious and frustrating. 

Could a solution be to add the SETUP CODE as a string beneath the QR code. On mobile, the user would now be able to copy the code and enter it manually into their authentication app???

As users will mostly only have to scan the QR code once when creating an account, I realise one solution is to have users create an account whilst on a desktop machine....

Is this even a problem? It's my first time using mfa on an application. 
Please do let me know if there's already a practical solution to this out there. 

Many thanks in advance.

ErikCH:
Hi @ClaudioVR ! Thanks for the request. 

This could be a handy addition. I'll mark this as a future feature request.

PR #1083, opened 2022-01-04
Updated to QR code page

ErikCH:
*Issue #, if available:*
#881 

*Description of changes:*
Updated QR code page so users can copy the secret key, instead of using QR code.

https://user-images.githubusercontent.com/65630/148132938-3be26b59-df36-4996-af81-4fa9135ae769.mp4

### DIFF
--- a/packages/ui/src/theme/css/component/authenticator.scss
+++ b/packages/ui/src/theme/css/component/authenticator.scss
@@ -139,3 +139,32 @@
     padding: 0px 1.5rem;
   }
 }
+[data-amplify-copy] {
+  display: flex;
+  font-size: 0.688rem;
+  gap: 1rem;
+  justify-content: center;
+  align-items: center;
+
+  @media (max-width: 30rem) {
+    word-break: break-all;
+  }
+}
+
+[data-amplify-copy-svg] {
+  cursor: pointer;
+  position: relative;
+
+  &:hover [data-amplify-copy-tooltip] {
+    visibility: visible;
+    font-size: 0.55rem;
+  }
+}
+
+[data-amplify-copy-tooltip] {
+  visibility: hidden;
+  position: absolute;
+  top: -1rem;
+  left: -0.1rem;
+  color: var(--amplify-colors-teal-100);
+}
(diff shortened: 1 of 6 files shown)


packages/vue/src/components/setup-totp.vue
```diff
@@ -1,3 +1,98 @@
...
+  qrCodeImageSource: '',
+  isLoading: true,
+});
+let secretKey = ref('');
+let copyTextLabel = ref(translate('COPY'));
+
+function copyText() {
+  navigator.clipboard.writeText(secretKey.value);
+  copyTextLabel.value = translate('COPIED');
+}
```
ErikCH, line 28: This was the only change, just moved script setup to top.

packages/ui/src/theme/css/component/authenticator.scss
```diff
@@ -139,3 +139,32 @@
...
+[data-amplify-copy] {
+  display: flex;
+  font-size: 0.688rem;
+  gap: 1rem;
+  justify-content: center;
+  align-items: center;
+
+  @media (max-width: 30rem) {
+    word-break: break-all;
+  }
```
wlee221, line 151: What do you think about making `[data-amplify-copy]` `hidden` if not on mobile?

ErikCH: I dunno, I was just using a QR code the other day, and I appreciated how the secret key was also displayed. I don't think it has to be just for mobile.

wlee221: That's fair! My only concern is that end-users won't know what to do with that secret key, but customers can use header slots to add directions.

wlee221 (approved):
🚢

packages/ui/src/theme/css/component/authenticator.scss
ErikCH: Yeah, I was debating back and forth if I should add in some instructions on how to add 2FA to Authy, or Google Authenticator, but I didn't want to add more requirements unless a customer wanted too.

And they can add in their own with a slot, so you're right.

PR #1085, opened 2022-01-05
Setup continuous deployment to `@next` on `ui-svelte/main`

wlee221:
*Issue #, if available:*

*Description of changes:* Sets up continuous deployment to `@next` on `ui-svelte/main`

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -16,6 +16,9 @@
     "next-example",
     "vue-example",
     "amplify-ui-angular-mono",
-    "e2e"
+    "e2e",
+    "@aws-amplify/ui-angular",
+    "@aws-amplify/ui-react",
+    "@aws-amplify/ui-vue"
   ]
 }
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,10 +5,10 @@ name: Tests
 
 on:
   push:
-    branches: [main]
+    branches: [ui-svelte/main]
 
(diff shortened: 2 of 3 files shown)


.changeset/config.json
```diff
@@ -16,6 +16,9 @@
     "next-example",
     "vue-example",
     "amplify-ui-angular-mono",
-    "e2e"
+    "e2e",
+    "@aws-amplify/ui-angular",
+    "@aws-amplify/ui-react",
+    "@aws-amplify/ui-vue"
```
wlee221, line 22: Ignoring these packages so that we only publish `@aws-amplify/ui-svelte` from this branch.

.github/workflows/tests.yml
```diff
@@ -5,10 +5,10 @@ name: Tests
 
 on:
   push:
-    branches: [main]
+    branches: [ui-svelte/main]
 
-  pull_request_target:
-    branches: [main]
+  pull_request:
```
wlee221, line 10: I changed this to `pull_request` so that we can have faster feedback loop on adding svelte e2e tests.

(`pull_request_target` always runs tests against `main`, which can be a pain if you're modifying `tests.yml` on a different branch)

wlee221:
Closing this as `workflow_run` only runs on `default` branch. https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run

PR #1089, opened 2022-01-05
Checkout the head_branch when publish `@next`

wlee221:
*Issue #, if available:*

*Description of changes:* With `ui-svelte/main` branch set up, this workflow now needs to checkout the head branch that triggered the workflow, instead of assuming it to be `main`.


ErikCH (approved):
Good catch!

PR #1097, opened 2022-01-06
Put yarn docs start into background process

reesscot:
*Description of changes:*
We added new steps to tests github action to [run e2e tests on the docs site](https://github.com/aws-amplify/amplify-ui/blob/main/.github/workflows/tests.yml#L338-L342). I couldn't get the new steps to run on a separate branch, so it was merged to main without full testing. I'm seeing that the ["Start docs site" step is stuck](https://github.com/aws-amplify/amplify-ui/runs/4721645361?check_suite_focus=true). Based on the [other steps](https://github.com/aws-amplify/amplify-ui/blob/main/.github/workflows/tests.yml#L265), it looks like `yarn docs start` command is not exiting and should be put in the background using `&` to allow for the `npx wait-on` command to run.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -336,7 +336,7 @@ jobs:
         run: yarn docs build
 
       - name: Start docs site
-        run: yarn docs start && npx wait-on -c waitOnConfig.json -t 20000 http-get://localhost:3000
+        run: yarn docs start & npx wait-on -c waitOnConfig.json -t 20000 http-get://localhost:3000
 
       - name: Run E2E tests against docs
         run: yarn workspace e2e test:theme


wlee221 (approved):
Thanks for a quick fix! FYI to test workflow changes on your branch, you can temporarily change the workflow trigger from `pull_request_target` to `pull_request`. Example: https://github.com/aws-amplify/amplify-ui/pull/951/commits/50df11c76d9712e65e690b7f0a3b3c724c226715.

For context, `pull_request` runs against your commit where as `pull_request_target` runs against `main`.  We've been using the latter because latter can be run on forks with proper environment variables. https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target

reesscot:
> Thanks for a quick fix! FYI to test workflow changes on your branch, you can temporarily change the workflow trigger from `pull_request_target` to `pull_request`. Example: [50df11c](https://github.com/aws-amplify/amplify-ui/commit/50df11c76d9712e65e690b7f0a3b3c724c226715).
> 
> For context, `pull_request` runs against your commit where as `pull_request_target` runs against `main`. We've been using the latter because latter can be run on forks with proper environment variables. https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target

Thanks, this is good to know!

Issue #1038, opened 2021-12-21
Reduct `ui-react` bundle size - remove autoprefixer


reesscot:
Remove the `autoprefixer` dependency to reduce the bundle size.

PR #1098, opened 2022-01-06
Remove autoprefixer to reduce `ui-react` package bundle size

reesscot:
*Issue #, if available:*
Closes https://github.com/aws-amplify/amplify-ui/issues/1038

*Description of changes:*
This PR removes the `autoprefixer` and `postcss-js` runtime dependencies from the `ui-react` package. This should increase performance on prop styles as well as reduce the package size.

*PR Rationale*
Vendor prefixing has become much less common today than it was in the past. New CSS features are usually released [behind a feature flag these days](https://css-tricks.com/is-vendor-prefixing-dead/). Additionally, the [BaseStyleProps](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/shared/styleUtils.ts#L191-L240) we support are very well supported and do not require prefixes. Given this information it's better to remove it and reduce the runtime performance hit.


dbanksdesign (approved):
LGTM! Do you know how much weight we will save with this?

zchenwei (approved):
LGTM! Thanks for making this improvement

reesscot:
> LGTM! Do you know how much weight we will save with this?

@dbanksdesign I think we should be able to shave off 19.7% or ~447kB. I say "I think" because I'm not seeing the changes affect the yarn.lock file. We seem to have a circular dependency through our [ui-react-v1 dependency](https://github.com/aws-amplify/amplify-ui/blob/main/yarn.lock#L455-L474). I'm thinking that once this change is published, the yarn lock reference pointing to npm should update. If not, I'll look into it more.
Screen Shot 2022-01-06 at 2 40 12 PM

reesscot:
Update on results after release:

[2.1.6](https://bundlephobia.com/package/@aws-amplify/ui-react@2.1.6) - Included `autoprefixer` and dependencies
Screen Shot 2022-01-24 at 2 49 58 PM
2.2MB Minified, 425 kB Minified and GZipped 

[2.1.7](https://bundlephobia.com/package/@aws-amplify/ui-react@2.1.7)
Screen Shot 2022-01-24 at 2 49 51 PM
1.4MB Minified, 318 kB Minified and GZipped 

### Bottom line package size reduction from 2.1.6 to 2.1.7:
**36% reduction of minified size**
**25% reduction of  minified and GZipped size**

PR #1102, opened 2022-01-06
Update @aws-amplify/ui-svelte@next so it installs correctly

ErikCH:
*Description of changes:*
`@aws-amplify/ui-svelte@next` gave errors when running. 

This fixes those errors so Svelte sees the folder correctly. 

### DIFF
--- a/packages/svelte/package.json
+++ b/packages/svelte/package.json
@@ -1,57 +1,63 @@
 {
-	"name": "@aws-amplify/ui-svelte",
-	"version": "0.0.1",
-	"type": "module",
-	"main": "dist/index.js",
-	"svelte": "index.js",
-	"exports": {
-		".": {
-			"import": "./dist/index.js"
-		},
-		"./styles.css": "./dist/styles.css"
-	},
-	"browser": {
-		"./styles.css": "./dist/styles.css"
-	},
-	"scripts": {
-		"dev": "svelte-kit dev",
-		"build": "svelte-kit build",
-		"postbuild": "svelte-kit package",
-		"package": "svelte-kit package",
-		"preview": "svelte-kit preview",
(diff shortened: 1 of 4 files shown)


packages/svelte/package.json
```diff
@@ -1,57 +1,63 @@
...
+    "./styles.css": "./dist/styles.css"
+  },
+  "files": [
+    "dist",
+    "LICENSE"
+  ],
+  "scripts": {
+    "dev": "svelte-kit dev",
+    "build": "svelte-kit build",
+    "postbuild": "svelte-kit package && shx cp ../ui/dist/styles.css dist/styles.css",
```
ErikCH, line 24: For now I'm just using a library called shx (cross platform) to copy the styles to the dist folder. Later, I'll see if it's possible with a vite/rollup plugin. For some reason SvelteKit won't recognize the styles import otherwise, from `ui`.

packages/svelte/src/lib/components/authStore.ts
```diff
@@ -8,16 +8,17 @@ import { interpret } from 'xstate';
 
 import { writable, get } from 'svelte/store';
 
-let _facade = writable<ReturnType<typeof getServiceContextFacade>>(null);
+let _facade = writable(null);
```
ErikCH, line 11: I'm not sure why but setting this as `<typeof getServiceContextFacade>` is making Svelte remove the `getServiceContextFacade` as if it's a type during the build. This is causing a run time error because Svelte can't find `getServiceContextFacade`. By removing this type, it fixes the error.

wlee221: interesting... 🤯

wlee221 (approved):
:shipit:

Issue #1095, opened 2022-01-05
bug: phone number field border alignment


reesscot:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Other

### Which UI component?

Other

### How is your app built?

All

### Please describe your bug.

Borders of the country selector and input box don't align in docs:
![Screen Shot 2022-01-05 at 2 53 42 PM](https://user-images.githubusercontent.com/6165315/148301572-d5c8451b-bb93-49c0-8f2d-e4bbbdfe4732.png)

### What's the expected behaviour?

Border outlines should align

### Help us reproduce the bug!

Visit: https://ui.docs.amplify.aws/components/phonenumberfield#demo

PR #1103, opened 2022-01-06
fix: fixing unaligned border outlines

zchenwei:
*Issue #, if available:*
#1095 

*Description of changes:*

Fixing `PhoneNumberField` styling to get the height of `CountryCodeSelect` and `TextField` matching

Screen Shot 2022-01-06 at 12 42 39 PM

### DIFF
--- a/packages/ui/src/theme/css/component/countryCodeSelect.scss
+++ b/packages/ui/src/theme/css/component/countryCodeSelect.scss
@@ -1,0 +1,3 @@
+.amplify-countrycodeselect {
+  height: var(--amplify-components-countrycodeselect-height);
+}
--- a/packages/ui/src/theme/tokens/components/countryCodeSelect.js
+++ b/packages/ui/src/theme/tokens/components/countryCodeSelect.js
@@ -1,0 +1,5 @@
+module.exports = {
+  height: {
+    value: '100%',
+  },
+};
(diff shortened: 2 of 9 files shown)


packages/ui/src/theme/tokens/components/countryCodeSelect.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
reesscot, line 1: This doesn't feel like it should be in tokens since there's no countryCodeSelect primitive. Can we move this to the phoneNumberField.js tokens file?

packages/ui/src/theme/css/component/countryCodeSelect.scss
```diff
@@ -0,0 +1,3 @@
+.amplify-countrycodeselect {
```
reesscot, line 1: Same comment as below, this feels like it should be in the phoneNumberField.scss file, since that's the primitive?

packages/ui/src/theme/tokens/components/countryCodeSelect.js
reesscot: Ah, nevermind, I didn't realize that CountryCodeSelect existed.

reesscot (approved):
Looks good, just needs a changeset

PR #1114, opened 2022-01-07
Use explicit aws-amplify imports for `ui`, `ui-react`, `ui-vue` and `ui-angular` packages

reesscot:
_Hat tip to @johnpc, this is mostly a clone of https://github.com/aws-amplify/amplify-ui/pull/580 which had too many merge conflicts to resolve, so I started over with this PR and tested how this could solve some of our Node.js global variable problems._

*Description of changes:*
Removes the dependency on the umbrella `aws-amplify` package and instead maps to @aws-amplify/core and @aws-amplify/auth packages. The benefit of this change is to prevent importing other unused packages which contain NodeJS globals. This makes it simpler to use our react package with Vite/ESBuild, requiring customers only to polyfill the `global` variable due to `@aws-amplify/auth` but not `process` or `exports` (you can see that I've removed them from docs as part of this PR).

```diff
    <script>
      window.global = window;
-      window.process = {
-        env: { DEBUG: undefined },
-      };
-      var exports = {};
    </script>
```

*Sample Vite app BEFORE this change with only `global` polyfill:*
Screen Shot 2022-01-07 at 2 08 00 PM

*Sample Vite app AFTER this change with only `global` polyfill:*
Screen Shot 2022-01-07 at 2 29 10 PM

*TODO:*
- [ ] Test Authenticator 


dbanksdesign (approved):
LGTM!

oh i like it

packages/ui/package.json
```diff
@@ -54,6 +54,7 @@
     "type-fest": "^2.3.4"
   },
   "peerDependencies": {
-    "aws-amplify": "3.x.x || 4.x.x"
```
reesscot, line 57: Should we stick with `aws-amplify`, or ask customers to install `@aws-amplify/core` and `@aws-amplify/auth` instead? @dbanksdesign @hvergara

dbanksdesign: If we declare the peer dependencies like this, a customer could still install `aws-amplify` and that would satisfy these peer requirements right?

docs/src/pages/getting-started/installation/troubleshooting.react.mdx
```diff
@@ -57,10 +57,6 @@ When working with a [Vite](https://vitejs.dev) project you must make a few modif
 ...
   <script>
     window.global = window;
-    window.process = {
```
reesscot, line 60: I may need to do this later after release, otherwise it will be updated as soon as I merge to master. But I wanted to show that it can be removed after this fix!

packages/ui/package.json
reesscot: Yes, when I tested this locally in my dev env I didn't specifically download these two dependencies, and already had `aws-amplify` installed. However, maybe legacy imports still needs the `aws-amplify` peer dependency?

zchenwei: Maybe `@aws-amplify/ui-react-legacy` still need `aws-amplify` peer dependency since it lives in react package?

reesscot:
It's a bit unclear what the issue is here, but currently we are thinking that the explicit imports doesn't work correctly with ESBuild, and is causing there to be two Amplify instances. It's know that when Amplify is initialized twice that Amplify apps don't work correctly, and in this cause it prevents the aws-exports from being configured. We need to do further investigation as to why this only happens with the Vite build.

johnpc (approved):

Issue #1111, opened 2022-01-07
Documentation: Theming CSS Broken Link


Jesmaster:
The "alternative styling with a styling libraries link" is broken on https://ui.docs.amplify.aws/theming#css

PR #1116, opened 2022-01-08
build: moving docs back to hybrid site to support redirects

zchenwei:
*Issue #, if available:*
#1111 

Some links on docs site that require redirects do not work as expect because we are deploying our docs site with SSG.

*Description of changes:*

Moving docs site back to SSR

### DIFF
--- a/docs/next.config.js
+++ b/docs/next.config.js
@@ -51,7 +51,7 @@ module.exports = withNextPluginPreval({
   },
 
   // These redirects are because of the IA change from previous docs
-  redirects() {
+  async redirects() {
     return [
       // Normalizing URLs
       // these need to come before the generic redirects
--- a/amplify.yml
+++ b/amplify.yml
@@ -15,9 +15,9 @@ applications:
             - node -v
             - yarn build
       artifacts:
-        baseDirectory: out
+        baseDirectory: .next
         files:
-          - "**/*"
+          - '**/*'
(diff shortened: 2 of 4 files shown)


reesscot:
@zchenwei I don't think docs changes should have a changeset since we don't publish `docs` as a package.

reesscot (changes requested):
Please check the PR preview, it doesn't seem to be working right now.

zchenwei:
> g



> Please check the PR preview, it doesn't seem to be working right now.

It should be working now. It was stuck last Friday, the build succeeded though

dbanksdesign (approved):
LGTM! :shipit:

docs/next.config.js
```diff
@@ -51,7 +51,7 @@ module.exports = withNextPluginPreval({
   },
 
   // These redirects are because of the IA change from previous docs
-  redirects() {
+  async redirects() {
```
wlee221, line 54: For my understanding, why `async`?

zchenwei: `Next` config requires it to be async anyway and we didn't configure it correctly in the first place :( <- https://nextjs.org/docs/api-reference/next.config.js/redirects. But yea, `Next` doesn't explain much about why it should be async.

reesscot (approved):
LGTM!

Issue #1033, opened 2021-12-21
Tabs doesn't handle null children


dbanksdesign:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Any

### Please describe your bug.

Using a ternary operator that returns null instead of a `TabItem` component inside the `Tabs` component throws an error. 

<img width="732" alt="CleanShot 2021-12-21 at 10 02 33@2x" src="https://user-images.githubusercontent.com/321279/146977261-3a1582fe-ee26-4514-9965-2df4ed1f2f4c.png">

<img width="571" alt="CleanShot 2021-12-21 at 10 02 06@2x" src="https://user-images.githubusercontent.com/321279/146977283-6f321ac4-9af3-418b-8069-b3139bbda2a9.png">

### What's the expected behaviour?

It should not throw the error and only show 1 tab

### Help us reproduce the bug!

Use the code and see the error :) 

### Code Snippet

```jsx
<Tabs>
  <TabItem title="Tab 1">
    <View>{/* */}</View>
  </TabItem>
  {themeControls ? <TabItem title="Tab 2">{/* */}</TabItem> : null}
</Tabs>
```

PR #1117, opened 2022-01-10
handle null children in tabs

jacoblogan:
*Issue #1033

Add null check to tabs primitive to handle null children correctly

### DIFF
--- a/packages/react/src/primitives/Tabs/__tests__/Tabs.test.tsx
+++ b/packages/react/src/primitives/Tabs/__tests__/Tabs.test.tsx
@@ -42,6 +42,32 @@ describe('Tabs: ', () => {
     expect(ref.current.nodeName).toBe('DIV');
   });
 
+  it('should skip over null children', async () => {
+    render(
+      <Tabs testId="tabsTest">
+        <TabItem title="Tab 1">Tab 1</TabItem>
+        {null}
+      </Tabs>
+    );
+    const tabs = await screen.findByTestId('tabsTest');
+    expect(tabs.children.length).toEqual(1);
+  });
+
+  it('should log a warning for null children', async () => {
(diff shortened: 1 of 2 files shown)


packages/react/src/primitives/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -42,6 +42,32 @@ describe('Tabs: ', () => {
...
+      <Tabs testId="tabsTest">
+        <TabItem title="Tab 1">Tab 1</TabItem>
+        {null}
+      </Tabs>
+    );
+    const tabs = await screen.findByTestId('tabsTest');
+    expect(tabs.children.length).toEqual(1);
+  });
+
+  it('should log a warning for null children', async () => {
```
dbanksdesign, line 56: Should we log a warning for this case? A common conditional pattern I see in react is

```jsx
{someCondition ? <Component /> : null}
```

This would show a warning for this case right?

dbanksdesign (commented):
LGTM! Just one non-blocking question for the console warning message

packages/react/src/primitives/Tabs/__tests__/Tabs.test.tsx
zchenwei: Yea, this would confuse users if they try to do conditional rendering

jacoblogan: yes that is a good point, I've corrected that so `null` will no longer produce the warning and added another test to ensure that previous cases that were providing that warning are still doing so

reesscot (approved):
LGTM

zchenwei (approved):
LGTM

PR #1142, opened 2022-01-13
Add explicit state for initializing authMachine

wlee221:
*Issue #, if available:*

*Description of changes:* This PR refactors `authMachine` to have an explicit `INIT` step that takes in all `authenticatorMachineProps` and applies it to its `context`. 

### Context

Previously, machine props need to be passed in *before* the machine was created:

https://github.com/aws-amplify/amplify-ui/blob/933d08dd9fccd1e313d4377e2f72969c4a0e025f/packages/react/src/components/Authenticator/Provider/index.tsx#L28-L32

### Problem

This behavior is susceptible to timing issues -- machine can be only created after `authenticator` is mounted with proper props (and Provider on React). For example, if customer tries to use `useAuthenticator` React hook outside `<Authenticator/ >`, their `useAuthenticator` runs before `Authenticator` is mounted before `machineProps` are passed in. This results in `useAuthenticator` trying to use `useMachine` before `machine` is ready to be created and fails. 

### Proposed Solution

The easiest way IMO to resolve it is to adjust `authMachine` so that it doesn't have to wait for `AuthenticatorMachineProps`. That way, machine can be flexibly created any timing as long as it is followed up with `INIT` step.

image

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
+++ b/packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
@@ -37,17 +37,20 @@ export class AuthenticatorService implements OnDestroy {
     signUpAttributes,
     socialProviders,
   }: AuthenticatorMachineOptions) {
-    const machine = createAuthenticatorMachine({
-      initialState,
-      loginMechanisms,
-      services,
-      signUpAttributes,
-      socialProviders,
-    });
+    const machine = createAuthenticatorMachine();
 
-    const authService = interpret(machine, {
-      devTools: process.env.NODE_ENV === 'development',
-    }).start();
+    const authService = interpret(machine).start();
+
+    authService.send({
+      type: 'INIT',
+      data: {
+        initialState,
+        loginMechanisms,
+        socialProviders,
+        signUpAttributes,
+        services,
+      },
+    });
 
     this._subscription = authService.subscribe((state) => {
       this._authState = state;
--- a/packages/react/src/components/Authenticator/Provider/index.tsx
+++ b/packages/react/src/components/Authenticator/Provider/index.tsx
@@ -23,19 +23,24 @@ const useAuthenticatorValue = ({
   signUpAttributes,
   services,
 }: ProviderProps) => {
-  const [state, send] = useMachine(
-    () =>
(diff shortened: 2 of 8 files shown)


packages/react/src/components/Authenticator/Provider/index.tsx
```diff
@@ -23,19 +23,20 @@ const useAuthenticatorValue = ({
   signUpAttributes,
   services,
 }: ProviderProps) => {
-  const [state, send] = useMachine(
-    () =>
-      createAuthenticatorMachine({
+  const [state, send] = useMachine(() => createAuthenticatorMachine());
```
wlee221, line 26: In later PR, I'll adjust this so that `machine` can be created without explicit `<Authenticator.Provider>` around package users' app. This will allow `useAuthenticator` to be used outside `<Authenticator >` internals and still have access to global machine.

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -9,50 +9,54 @@ import { createSignUpMachine } from './signUp';
...
+              actions: 'configure',
+            },
+          },
+        },
+        setup: {
           invoke: [
             {
               // TODO Wait for Auth to be configured
-              src: 'getCurrentUser',
+              src: (context, _) => context.services.getCurrentUser(),
```
wlee221, line 40: `services` are now stored through `context`.

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -37,18 +37,21 @@ export class AuthenticatorService implements OnDestroy {
     signUpAttributes,
     socialProviders,
   }: AuthenticatorMachineOptions) {
-    const machine = createAuthenticatorMachine({
-      initialState,
-      loginMechanisms,
-      services,
-      signUpAttributes,
-      socialProviders,
+    const machine = createAuthenticatorMachine();
```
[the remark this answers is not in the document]
ErikCH: 👍

ErikCH (approved):
This looks good. I just tested it out manually as well and didn't see any issues.

Issue #1092, opened 2022-01-05
box-sizing styling affects non-amplify elements


carlton-bennett:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Create React App

### Please describe your bug.

[`This rule`](https://github.com/aws-amplify/amplify-ui/blob/1972bcd10bd702419494728e2debabf21bc3983b/packages/ui/src/theme/css/styles.scss#L22) in the CSS for amplify-ui affects other non-amplify elements on the page.

### What's the expected behaviour?

The styles included with amplify-ui should be scoped such that they don't affect non-amplify elements on the page.

### Help us reproduce the bug!

Create a new app using the code snippet below. The `box-sizing` of the blue section is set to `content-box` in the CSS. The rule from amplify-ui overrides this to be `border-box`. Clicking the toggle button will toggle an inline style which sets it back to `content-box`. The div changes size to illustrate the real-world impact of this rule.

### Code Snippet

```css
// index.css
section {
  box-sizing: content-box;
  background-color: blue;
  width: 500px;
  height: 100px;
  padding: 50px;
}
```

```typescript
// index.tsx
import { AmplifyProvider } from '@aws-amplify/ui-react';
import React, { useState } from 'react';
import ReactDOM from 'react-dom';

import '@aws-amplify/ui-react/dist/index.css';
import './index.css';


function App() {
    const [ force, setForce ] = useState(false);

    return (
        <AmplifyProvider>
            <section
                style={!force ? undefined : {
                    boxSizing: 'content-box'
                }}
            />

            <button onClick={() => setForce(!force)}>
                toggle
            </button>
        </AmplifyProvider>
    );
}

ReactDOM.render(
    <App />,
    document.getElementById('root')
);

```

reesscot:
Hi @carlton-bennett,
The specificity of the border-box rule is `0 1 0 0`, so you should be able to override it with a class:

```css
// index.css
.my-section {
  box-sizing: content-box;
}
```

```jsx
 <AmplifyProvider>
            <section
[...]

PR #1151, opened 2022-01-14
fix: change box-sizing to have less specificity

reesscot:
*Issue #, if available:*
Closes https://github.com/aws-amplify/amplify-ui/issues/1092

*Description of changes:*
This PR fixes an issue where a customer was using a library which injected its CSS too early to override ours. Our `box-sizing` CSS rule had a specificity of `0 1 0`. We primarily need this level of specificity for the `amplify*` classes, but can reduce the specificity to`0 0 0` for other elements. This means any element used in an Amplify UI app will have a default `box-sizing` of `border-box`, but can easily override it with any level of specificity. This should allow other library UI components to be used inside of Amplify UI components such as `Flex` without changing their `box-sizing` styling. 

### DIFF
--- a/docs/src/styles/docs/base.scss
+++ b/docs/src/styles/docs/base.scss
@@ -38,8 +38,6 @@ html {
 
 // Remove this when this gets fixed in the UI package
 * {
-  box-sizing: border-box;
-
   /* Selection */
   &::-moz-selection {
     background: var(--amplify-colors-overlay-10);
@@ -111,7 +109,9 @@ code {
   background-color: var(--amplify-colors-background-secondary);
   font-size: 0.875em;
   border-radius: 0.3em;
-  font-family: "Source Code Pro", "Fira Code", "Fira Mono", ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
+  font-family: 'Source Code Pro', 'Fira Code', 'Fira Mono', ui-monospace,
+    SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New',
+    monospace;
   direction: ltr;
   text-align: left;
   white-space: pre;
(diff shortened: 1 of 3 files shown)


docs/src/styles/docs/base.scss
```diff
@@ -38,8 +38,6 @@ html {
 
 // Remove this when this gets fixed in the UI package
 * {
-  box-sizing: border-box;
```
reesscot, line 41: This was only affecting the docs site. Removing as it's redundant with the new CSS rule below.

zchenwei (approved):
LGTM! :shipit:

packages/ui/src/theme/css/styles.scss
```diff
@@ -13,14 +13,15 @@
   }
 }
 
-[class*='amplify'] {
-  all: unset; /* protect against external styles */
+// Note: This rule can be easily overwritten when
+// needed due to very low specificity of 0 0 0
+* {
```
dbanksdesign, line 18: Do we still need this? Taking a look at the original issue, https://github.com/aws-amplify/amplify-ui/issues/1092, it sounds like we don't want our styles seeping into other non-amplify ui components. Could this just be removed as all our components have classnames that start with 'amplify'?

dbanksdesign (commented):
One question about the inclusion of the * selector

packages/ui/src/theme/css/styles.scss
reesscot: If we remove this, then we will be changing the default `box-sizing` behavior of the library for all of the children elements inside of `amplify` class components. For example, flex children which are not amplify components would have a different `box-sizing` behavior than amplify components.

  

**Without** `* { box-sizing: border-box; }`

```jsx

<Flex>

<div>1</div> {/* content-box */}

<Button>2</Button> {/* border-box */}

</Flex>

```

  

**With** `* { box-sizing: border-box; }`

```jsx

<Flex>

<div>1</div> {/* border-box */}

<Button>2</Button> {/* border-box */}

</Flex>

```

  

Leaving it off seems like it would lead to unexpected behavior for customers. What do you think @dbanksdesign

dbanksdesign (approved):
LGTM!

PR #1159, opened 2022-01-17
Fix secret key in QR code being empty string

### LINKED ISSUE #1158: QR Code is missing secret key on TOTP setup page

React

### Which UI component?

Authenticator

### How is your app built?

Create react app (react-scripts)

### Please describe your bug.

The the TOTP setup page, the QR code is invalid. The contents of the QR code are 
This bug was introduced in #1083 

Specifically, the reason for this bug is that `secretKey` is now being stored using `React.useState`, and so `secretKey` is referenced before the component has rerendered, and therefore `totpCode` still has its old value.

https://github.com/aws-amplify/amplify-ui/blob/f888639c7fe2464b689dc6c7663cbd9b2a5312ae/packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx#L32-L35

A solution would be to change these lines to
```typescript
      const newSecretKey = await Auth.setupTOTP(user);
      setSecretKey(newSecretKey);
      const issuer = 'AWSCognito';
      const totpCode = `otpauth://totp/${issuer}:${user.username}?secret=${newSecretKey}&issuer=${issuer}`;
      const qrCodeImageSource = await QRCode.toDataURL(totpCode);
```

The contents of the QR code should be `otpauth://totp/${issuer}:${user.username}?secret=${this.secretKey}&issuer=${issuer}`

In reality, `this.secretKey` is an empty string.

Start any authentication flow with requires one time passwords. Try scanning the QR code with an Authenticator app. Your Authenticator app will say it is invalid. Try parsing the QR code with any generic QR code scanner (e.g. Google Lens). You'll see that the string is something like `otpauth://totp/AWSCognito:abcder?secret=&issuer=AWSCognito`

### Code Snippet

_No response_

_No response_

bwindsor:
Fixes #1158 

As described in the issue, a recent change to the code led to the QR code for TOTP being incorrectly generated, such that the `secretKey` encoded in the QR code was actually just an empty string. This is because the `secretKey` is being stored in a `React.useState` variable, which is not updated until the component rerenders, which is after the string to be used for the QR code is generated.

### DIFF
--- a/.changeset/five-seas-flash.md
+++ b/.changeset/five-seas-flash.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ui-react': patch
+---
+
+Fix TOTP QR being incorrectly generated, as it was using an empty string as the secretKey
--- a/packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
+++ b/packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
@@ -29,11 +29,12 @@ export const SetupTOTP = (): JSX.Element => {
 
   const generateQRCode = async (user): Promise<void> => {
     try {
-      setSecretKey(await Auth.setupTOTP(user));
+      const newSecretKey = await Auth.setupTOTP(user);
       const issuer = 'AWSCognito';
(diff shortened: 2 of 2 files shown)


bwindsor:
Superseded by #1163

PR #1162, opened 2022-01-18
Attempt #2: Add explicit state for initializing authMachine (#1142)

wlee221:
Descriptions: We previously had to revert #1142 because it was causing docs failure to https://ui.docs.amplify.aws/components/authenticator#labels--text. 

This PR adds a fix for that failure by making sure machine is fully initialized before the component renders: 2669bf6.

For additional context, see the original PR #1142.

### DIFF
--- a/docs/src/pages/components/authenticator/LabelsAndTextDemo.tsx
+++ b/docs/src/pages/components/authenticator/LabelsAndTextDemo.tsx
@@ -71,15 +71,28 @@ function Screen({ Component }: ScreenProps) {
 }
 
 export function LabelsAndTextDemo({ Component }: ScreenProps) {
+  const OnMachineInit = ({ children }) => {
+    /**
+     * This waits for Authenticator machine to init before its inner components
+     * start consuming machine context.
+     */
+    const { route } = useAuthenticator();
+    if (!route || route === 'idle' || route === 'setup') return null;
+
+    return <>{children}</>;
+  };
+
   return (
     <Authenticator.Provider>
-      <View data-amplify-authenticator="">
-        <View data-amplify-container="">
-          <View data-amplify-body>
-            <Screen Component={Component} />
+      <OnMachineInit>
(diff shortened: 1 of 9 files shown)


docs/src/pages/components/authenticator/LabelsAndTextDemo.tsx
```diff
@@ -71,15 +71,28 @@ function Screen({ Component }: ScreenProps) {
...
+  const OnMachineInit = ({ children }) => {
+    /**
+     * This waits for Authenticator machine to init before its inner components
+     * start consuming machine context.
+     */
+    const { route } = useAuthenticator();
+    if (!route || route === 'idle' || route === 'setup') return null;
+
+    return <>{children}</>;
+  };
```
wlee221, line 83: This is the only additional change since #1142.

Labels and Text demo was using `Authenticator.<SubComponent>` directly, which relied on `authMachine` to be already initialized and ready. With #1142, `authMachine` isn't fully initiated when this demo renders, which caused a runtime error. This component goes around it by rendering demo only if the machine is set up.

wlee221: https://user-images.githubusercontent.com/43682783/150034299-bbdc5c5e-178a-4c86-9349-163a3080a292.mp4

ErikCH: Good enough for me!

ErikCH: And there is no where else we use these subcomponents?

ErikCH (approved):
Looks good, just had that one question

docs/src/pages/components/authenticator/LabelsAndTextDemo.tsx
wlee221: Yep, this is the only place i could find it used 👍

PR #1164, opened 2022-01-18
Removed hard-coded styles from Authenticator Sign Up button

### LINKED ISSUE #1133: Bug: Authenticator Create Account Button Element Styles
The Authenticator "Create Account" button gets the following element styles

`    border-radius: 0px;
    font-weight: normal;`

Very easy to replicate, simply place an Authenticator component, you will see the Sign In button has no hard styles on the element. Click the Create Account tab and look at the button and the font is no longer bold and the border radius is gone. There should be no styles added on the element level to the button as these are overriding the default theme button styles.

You can see this in the documentation page as well:
https://ui.docs.amplify.aws/components/authenticator#initialstate

Jesmaster:
Fixes #1133

### DIFF
--- a/.changeset/six-dancers-call.md
+++ b/.changeset/six-dancers-call.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+Removed hard-coded styles from Authenticator Sign Up button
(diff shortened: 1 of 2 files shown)


zchenwei (approved):
LGTM

PR #1165, opened 2022-01-18
Pin all @aws-amplify/ui-react dependencies

joebuono:
*Issue #, if available:* Part of CoE

*Description of changes:*

In order to prevent new dependency versions from breaking our library, we need to pin all our dependencies to specific versions. 

This PR pins all `package.json` 3rd-party dependencies to specific versions. 


zchenwei (approved):
LGTM

hvergara (approved):
Looks good! thanks for doing this <3

Issue #1130, opened 2022-01-12
Instructions for using Authenticator on logged out pages / components?


paulsingh:
### On which framework/platform would you like to see this feature implemented?

Vue

### Which UI component is this feature-request for?

Authenticator

### Please describe your feature-request in detail.

The [Quick Start instructions](https://ui.docs.amplify.aws/components/authenticator?platform=vue#quick-start) are great... if you want the entire app locked down. Beyond that... it's not immediately clear to me (and likely others, too) how I could exclude certain pages from Authenticator. (eg, Home, About, etc.)

In the meantime, I've managed to get far enough along with the following (but it doesn't seem like it would be appropriate to use this in a real-world app):

```
const statePayload = useAuthenticator().state; 
console.log("statePayload = " + JSON.stringify(statePayload, null, 2))
// user = statePayload.context.user
// email = statePayload.context.user.attributes.email
```

### Please describe a solution you'd like.

If someone could write up a little more documentation on how to use Authenticator inside apps that have a need for both authenticated _and_ unauthenticated pages, I'd be happy to try things out and document it further for others. 

### We love contributors! Is this something you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

ErikCH:
HI @paulsingh !

That's a good idea! We definitely don't have much documentation on how to handle apps with multiple routes and pages.

Until we have the docs up you could try a couple things. I'm going to assume you are using vue-router and you have multiple pages in your app.

First go into your router file and add new meta tags for every route you'd like to require Auth.  Something like this.

```
const routes = [
  {
    path: '/posts',
    component: PostsLayout,
    children: [
      {
        path: 'new',
        component: PostsNew,
        // only authenticated users can create posts
        meta: { requiresAuth: true }
      },
      {
        path: ':id',
        component: PostsDetail,
        // anybody can read a post
        meta: { requiresAuth: false }
      }
    ]
  }
]
```
[(From the official docs)
](https://next.router.vuejs.org/guide/advanced/meta.html#route-meta-fields)

Then you can setup route guards. Like this

```
router.beforeEach(async (to, from, next) => {
  const requiresAuth = to.matched.some((record) => record.meta.requiresAuth);
  const isAuthenticated = await Auth.currentUserInfo();

 if (requiresAuth && !isAuthenticated) {
    next("/login");
  } else {
    next();
  }
});
```

[...]

PR #1075, opened 2022-01-04
Document `useAuthenticator` hook

wlee221:
*Issue #, if available:* Depends on #637, closes #254, closes #664, close #980, related to #640.

*Description of changes:* 

- [x] Angular `useAuthenticator` example
- [x] Angular `useAuthenticator` documentation
- [x] Vue `useAuthenticator` example
- [x] Vue `useAuthenticator` documentation
- [x] React `useAuthenticator` example
- [x] React `useAuthenticator` documentation

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #1058, opened 2021-12-29
Support for unprotected routes


Shomari:
### On which framework/platform would you like to see this feature implemented?

React

### Which UI component is this feature-request for?

Authenticator

### Please describe your feature-request in detail.

It would be nice to be able to simply pass in a list of routes that should be unprotected.  These routes would not be redirected to the login flow and could proceed as normal.  Currently, you have to come up with work around for routing if you want a home page to be unprotected but all other to be protected.  

This is harder with the new version of the UI as you don't have the same access to authState as you did in the previous version where you could follow the docs https://docs.amplify.aws/ui/auth/authenticator/q/framework/react/#basic-usage to get the result you were looking for. 

### Please describe a solution you'd like.

Pass in a list of routes to Authenticator that would not be redirected

### We love contributors! Is this something you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

Milan-Shah:
@Shomari Thank you for creating a feature-request. We are currently working on our useAuthenticatorHook Milestone which includes passing in a list of routes to Authenticator that would not be redirected. We recently merged a PR to enable useAuthenticator usage outside of <Authenticator /> component. Please see here[https://github.com/aws-amplify/amplify-ui/releases/tag/%40aws-amplify%2Fui-react%402.2.0]. Let us know if you have any questions or feedback :)

Issue #664, opened 2021-11-10
docs: Document facades we provide 


wlee221:
Related: #254

We now provide public hooks/services for Authenticator: #532 (React), #588 (Vue), and #627 (Angular). We should document facades we provide. 

We'll need to be careful to focus on actual use cases though, so that it's not just a giant table.

### TODOs

- [x] Create examples matching https://codesandbox.io/s/change-your-layout-rh4z9?file=/src/Login.js
- [x] Document `Authenticator.Provider` for React
- [x] Document `useAuthenticator` hooks & services

ericclemmons:
https://codesandbox.io/s/change-your-layout-rh4z9?file=/src/App.js is a great example for React.

wlee221:
It's been quite long since I last tackled this. 

I think it'd be good to be a little verbose in documentation and provide `tabs` of facadades we provide:

```
tab 1 (context) - user, route, error, isPending, codeDeliveryDetails
tab 2 (transitions) - to* hooks and `signOut`
tab 3 (forms) - updateForm, submitForm
```

PR #1168, opened 2022-01-19
Enable `useAuthenticator` usage outside <Authenticator />

wlee221:
*Issue #, if available:* #637, #254, #664, #640.

*Description of changes:* This PR enables `useAuthenticator` hook to be used outside `Authenticator`. Also adds `/useAuthenticator` example and e2e tests that'll supplement docs PR #1075.

## Use Case

Main use case is custom routing (#1058), which is not possible today because Authenticator assumes SPA usage and requires Authenticated content to be nested inside `Authenticator`. #1130 puts it well:

> The [Quick Start instructions](https://ui.docs.amplify.aws/components/authenticator?platform=vue#quick-start) are great... if you want the entire app locked down. Beyond that... it's not immediately clear to me (and likely others, too) how I could exclude certain pages from Authenticator. (eg, Home, About, etc.).

## Problem

Internally, we used `useAuthenticator` and `Authenticator.Provider` to share `auth` contexts. To use `useAuthenticator` outside `Authenticator`, we falsely went around this issue by wrapping `<Authenticator.Provider />` around their app:

```tsx
(1) <Authenticator.Provider> 
(2)   <App>
(3)     <Authenticator> 
(4)       <Authenticator.Provider>
(5)         <Authenticator.SignIn />
(6)        ...
```

Problem here was that two providers create two different xstate machines. `useAuthenticator` from `<App />` (2) returned false data because it is referencing the wrong provider (1) instead of (4).

### Wasted re-renders

Another React nuance we overlooked is that `useAuthenticator` changes on every xstate transitions, including form input events! 

This is not only computationally expensive but also can trigger repeated mounting of `<Authenticator />` because `MyApp` (1) encompasses it. This opens up crucial bugs like form losing its values on re-render. [xstate docs](https://xstate.js.org/docs/recipes/react.html#global-state-react-context) explains better than I can:

> React context can be a tricky tool to work with - if you pass in values which change too often, it can result in re-renders all the way down the tree. 

### Sync Provider Context Value 

Now `Authenticator.Provider` detects whether there's another `Provider` above it. If so, it'll just pass that parent value. This resolves multiple xstate machine instantiations.

### Serve static value from `Authenticator.Provider`

Now `Authenticator.Provider` only maintains a static reference to `authMachine` service to prevent re-renders, as suggested from [xstate docs](https://xstate.js.org/docs/recipes/react.html#global-state-react-context). `useAuthenticator` instead listens to state changes and provides respective auth context.

### Split `Slot` logic to internal hook

We've been using `useAuthenticator` to share custom components for Authenticator. This is only needed internally, is static, and is not needed externally. This has been split to `useCustomComponents` hook to focus `Authenticator.Provider` on just `authMachine` contexts. Splitting up complex context value is encouraged to reduce wasted re-renders: https://github.com/facebook/react/issues/15156.

### Add optimization mechanism to useAuthenticator

Even with static served `Authenticator.Provider`, `useAuthenticator` still changes frequently on every state change. This PR adds `useAuthenticatorRoute` and `useAuthenticatorUser` that only triggers re-render whenever `route` or `user` change respectively. This provides developers control over when they want auth updates.

Under the hood, this is done by `selector` optional parameter to base `useAuthenticator`. `selector` is a function that accepts current facade values and returns desired value(s) that should trigger re-render. 

This uses `xstate`'s [`useSelector`](https://xstate.js.org/docs/packages/xstate-react/#useselector-actor-selector-compare-getsnapshot), which provides just what we need: 

>  This hook will only cause a rerender if the selected value changes, ...


packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -1,33 +1,68 @@
...
+
+  // Helper component that sends init event to the parent provider
+  function InitMachine({ children, ...machineProps }) {
+    const { _send, route } = useAuthenticator();
+    useEffect(() => {
+      if (route === 'idle') {
+        _send({
+          type: 'INIT',
+          data: machineProps,
+        });
```
wlee221, line 49: This is a follow up from #1162. We're setting `INIT` event after we get the necessary `machineProps` in `Authenticator`.

packages/react/src/components/Authenticator/ConfirmSignUp/ConfirmSignUp.tsx
```diff
@@ -12,18 +13,20 @@ import {
...
-        Header = ConfirmSignUp.Header,
-        Footer = ConfirmSignUp.Footer,
-      },
-    },
     isPending,
     resendCode,
     submitForm,
     updateForm,
     codeDeliveryDetails: { DeliveryMedium, Destination } = {},
   } = useAuthenticator();
```
wlee221, line 21: `useAuthenticator` now returns only contexts related to `auth`!

packages/react/src/components/Authenticator/Router/index.tsx
```diff
@@ -1,19 +1,19 @@
...
+import { useCustomComponents } from '../hooks/useCustomComponents';
 import { ConfirmResetPassword, ResetPassword } from '../ResetPassword';
 import { SetupTOTP } from '../SetupTOTP';
 import { SignInSignUpTabs } from '../shared';
 import { ConfirmVerifyUser, VerifyUser } from '../VerifyUser';
 
 export type RouterProps = {
   className?: string;
-  children: ({
+  children?: ({
```
ErikCH, line 16: Glad you made this optional. Always bothered me you had to add the children using `authenticator`

packages/e2e/features/ui/components/authenticator/useAuthenticator.feature
```diff
@@ -0,0 +1,27 @@
...
+  outside the Authenticator component. React and Vue provides it through `useAuthenticator` 
+  hook and composable respectively, and Angular provides it through `AuthenticatorService` 
+  service. They can be used to access current authState (routes), authenticated user, etc.
+
+  See https://ui.docs.amplify.aws/components/authenticator#headless for details.
+
+  Background:
+    Given I'm running the example "/ui/components/authenticator/useAuthenticator"
+
+  @angular @react @vue @todo-angular @todo-vue
```
ErikCH, line 13: I can try to make sure Vue works in the future here, with this new useAuthenticator feature.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    if (!selector) return false;
+
+    // convert xstate state to facade that selector understands
+    const prevFacade = getFacade(prevState);
+    const nextFacade = getFacade(nextState);
+
+    return isEqual(selector(prevFacade), selector(nextFacade));
+  };
+
+  const state = useSelector(service, (state) => state, comparator);
```
ErikCH, line 96: Why isn't the selector passed in from the `useAuthenticator` used here? Like this

`const state = useSelector(service, selector, comparator)`

packages/e2e/features/ui/components/authenticator/useAuthenticator.feature
wlee221: Yep, I'll do one for Angular too. Let's do that on separate PRs though, in case there are `useAuthenticator` gotchas in other frameworks. (and tbh I'm scared I broke something this PR and have to revert it)

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: Ah good question, I'll add a code comment for this as well.

Whenever you pass a specific `selector` to `useSelector`, the hook will return *only* that selected value. e.g. `useSelector(service, (state) => state.context.[...].user, comparator)` will only return `user` and not other contexts like `route` or `toResetPassword`. 

I wanted `useAuthenticator` to provide a consistent set of values, but this is up to discussions.

dbanksdesign (commented):
This might be a bit out of my depth, but overall it looks good to me. Very good explanations throughout! Will this PR include updates to documentation too? I like what you have for the changeset if you want to repurpose that content.

wlee221:
@dbanksdesign, yep documentation will be handled from #1075. I'll merge them together after both are ready/approved 👍

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: ---
synced with @ErikCH offline. We'll adjust `useAuthenticatorUser` to just return the user. Same with other `authenticator` helpers.

.changeset/aero-stingrays-raise.md
```diff
@@ -0,0 +1,19 @@
+---
+"@aws-amplify/ui-react": patch
```
wlee221, line 2: This should be minor

```suggestion
"@aws-amplify/ui-react": minor
```

.changeset/weak-stingrays-raise.md
```diff
@@ -0,0 +1,39 @@
+---
+"@aws-amplify/ui-react": patch
```
wlee221, line 2: This should be minor

```suggestion
"@aws-amplify/ui-react": minor
```

wlee221:
my css file is here as well: https://codesandbox.io/s/legacy-picker-customization-jq7wz?file=/src/App.js

ErikCH (approved):
Looks good!

packages/ui/src/helpers/auth.ts
```diff
@@ -216,6 +216,8 @@ export const getServiceContextFacade = (state: AuthMachineState) => {
   const user = state.context?.user;
   const actorState = getActorState(state);
   const actorContext = getActorContext(state) as ActorContextWithForms;
+
+  // contexts that do not need actors
```
wlee221, line 220:
```suggestion
```

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  const state = useSelector(service, xstateSelector, comparator);
+
+  /**
+   * If selector was passed into the hook, we return just the selected value.
+   *
+   * Otherwise if developer `useAuthenticator()` without selector, then we return
+   * back the whole facade.
+   */
```
wlee221, line 136:
```suggestion
```

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  );
+
+  const getFacade = (state: AuthMachineState) => {
+    return { ...sendAliases, ...getServiceContextFacade(state) };
+  };
+
+  /**
+   * Selects which value to return from `useAuthenticator`. If selector is not
+   * provided, then we return the whole state back.
+   */
```
wlee221, line 97:
```suggestion
  /**
   * For `useSelector`'s selector argument, we just return back the `state`.
   * The reason is that whenever you select a specific value of the state, the
   * hook will return *only* that selected value instead of the whole `state`.
   *
   * To provide a consistent set of facade, we let the `selector` trivially return
   * itself and let comparator decide when to re-render.
   */
```

PR #1174, opened 2022-01-21
No lock tests

jacoblogan:
small change so that github action tests will ignore the yarn lockfile and test will run with newly resolved dependencies similar to what users will experience.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,7 +55,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: lts/*
-          cache: 'yarn'
+          cache: "yarn"
       - name: Install packages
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -124,7 +124,7 @@ jobs:
 
       - name: Install packages
         if: steps.restore-cache.outputs.cache-hit != 'true'
-        run: yarn --frozen-lockfile --prefer-offline  --prefer-offline
+        run: yarn --no-lockfile --prefer-offline  --prefer-offline
```
zchenwei, line 127: Mind removing a duplicate `--prefer-offline`?

zchenwei (approved):
LGTM! No blockers that I see

wlee221 (approved):

Issue #1155, opened 2022-01-16
Users get stuck after AdminCreateUser API called with a missing required attribute


ThomasRooney:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

### Please describe your bug.

Given a cognito userpool that has a required attribute "preferred_username"
And a user is invited to that userpool via a `cognito.adminCreateUser` API call lacking this `preferred_username` required attribute (in this case, the API call is made in a lambda upon another user inviting them)
The user is created successfully, and the admin invitation message is sent (default: `Your username is {####} and temporary password is {####}`.

When the user attempts to login through the `Authenticator` component, after the user enters username/password, they enter a 
 screen for changing their password for login.
 
Upon attempting to login, the error message `Invalid attributes given, preferred_username is missing` is provided.

However, given the user has no place to enter the required attribute `preferred_username`, they are now stuck until manual resolution by an administrator.

### What's the expected behaviour?

In the Cognito Hosted UI, and the prior amplify authenticator UI, the user is prompted to enter their `preferred_username`, similar to the signup flow.
Upon entering this attribute and confirming the password change, the attribute is expected to make its way to Cognito.

### Help us reproduce the bug!

Version: "@aws-amplify/ui-react": "2.1.8",
Also in version: "2.0.0" (I upgraded to 2.1.8 to validate it wasn't already solved)

Available workaround: creating the user with the full set of attributes. However this isn't ideal in many scenarios.

Network Request:

POST https://cognito-idp.eu-west-2.amazonaws.com/

```
{
    "ChallengeName": "NEW_PASSWORD_REQUIRED",
    "ClientId": "<REDACTED>",
    "ChallengeResponses": {
        "NEW_PASSWORD": "<REDACTED>",
        "USERNAME": "<REDACTED>"
    },
    "Session": "<REDACTED>"
}
```

Network Response:

`{"__type":"InvalidParameterException","message":"Invalid attributes given, preferred_username is missing"}`

### Code Snippet

[...]

PR #1176, opened 2022-01-21
Force New Password - Add Support For Required Attributes

ErikCH:
*Issue #, if available:*
#1155 

*Description of changes:*
If an Admin creates a user manually, the first time a user logs in they are brought to the `Force New Password` page. In previous versions of the Authenticator this would display the password, and confirm password as well as any required attributes the user needed. 

Without these required attributes the sign in cannot continue and the user will see an error.

This PR adds in a check to see if the attributes are needed, and adds them. 

We also will use the same logic we used in the sign up page. Only common attributes that are defined internally will display. If a user has a custom attribute, they can override with slots the `Force New Password Fields` to add it in.
 
This has been added to the docs as well.

image


packages/e2e/cypress/fixtures/force-change-password-with-attributes.json
```diff
@@ -0,0 +1,8 @@
+{
+  "ChallengeName": "NEW_PASSWORD_REQUIRED",
+  "ChallengeParameters": {
+    "requiredAttributes": "[\"userAttributes.gender\",\"userAttributes.given_name\",\"userAttributes.nickname\",\"userAttributes.middle_name\",\"userAttributes.name\",\"userAttributes.zoneinfo\",\"userAttributes.profile\",\"userAttributes.family_name\",\"userAttributes.phone_number\",\"userAttributes.website\",\"userAttributes.preferred_username\",\"userAttributes.locale\",\"userAttributes.picture\",\"userAttributes.address\"]",
+    "userAttributes": "{\"email_verified\":\"true\",\"phone_number_verified\":\"true\",\"phone_number\":\"+1••••••4444\",\"email\":\"••••••+FORCE_CHANGE_PASSWORD@••••••.com\"}"
+  },
+  "Session": "••••••"
+}
```
ErikCH, line 8: Mocked out the response we see when a new password is required with attributes.

packages/react/src/components/Authenticator/ForceNewPassword/FormFields.tsx
```diff
@@ -0,0 +1,179 @@
...
+} from '@aws-amplify/ui';
+
+import { useAuthenticator } from '..';
+import { PhoneNumberField, TextField } from '../../..';
+
+export function FormFields() {
+  const { _state } = useAuthenticator();
+  const { country_code } = getActorContext(_state) as SignUpContext;
+
+  const actorState = getActorState(_state) as SignInState;
```
ErikCH, line 17: These new components are modified heavily from their source `Sign Up Form Fields`

examples/angular/src/pages/ui/components/authenticator/sign-up-with-attributes/sign-up-with-attributes.component.html
```diff
@@ -1,4 +1,15 @@
...
+    <amplify-force-new-password-form-fields></amplify-force-new-password-form-fields>
+    <amplify-text-field
+      label="Zone Info"
+      id="12233"
+      autocomplete="false"
+      placeholder="Zone Info"
+      name="zoneinfo"
+      type="text"
+    ></amplify-text-field>
+  </ng-template>
```
wlee221, line 12: Apologies for the late review. QQ, is there a zero-config value we can use?

ErikCH: There are a handful of attributes that we decided that we aren't going to have default mappings for because they are app specific, including zoneinfo. As mentioned here https://ui.docs.amplify.aws/components/authenticator#signupattributes

The `zoneInfo` is in the zero config, you are right. We just don't have a mapping for it.

wlee221: Ah gotcha, I was lost yesterday xD

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -69,17 +71,29 @@ export function signInActor({ services }: SignInMachineOptions) {
                 onDone: [
                   {
                     cond: 'shouldSetupTOTP',
-                    actions: ['setUser', 'setChallengeName'],
+                    actions: [
+                      'setUser',
+                      'setChallengeName',
+                      'setRequiredAttributes',
```
wlee221, line 77: For my understanding -- can there be `setRequiredAttributes` for `setupTOTP` and and `shouldConfirmSignIn`. If not, should we only call that action for `forceChangePassword`?

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -439,9 +472,22 @@ export function signInActor({ services }: SignInMachineOptions) {
...
+            phone_number,
+            country_code,
+            ...rest
+          } = formValues;
+
+          let phoneNumberWithCountryCode;
+          if (phone_number) {
+            phoneNumberWithCountryCode =
+              `${country_code}${phone_number}`.replace(/[^A-Z0-9+]/gi, '');
+            rest = { ...rest, phone_number: phoneNumberWithCountryCode };
```
wlee221, line 487: Doesn't have to be on this PR, but should we use @slaymance's `parsePhoneNumber` here from #826?

wlee221 (approved):
LGTM, got two non-blocking questions

packages/ui/src/machines/authenticator/actors/signIn.ts
ErikCH: Yeah, right now we assume it only works for `forceChangePassword` I can delete those.  Maybe in the future we'll find a scenario we'll need it for setupTOTP and shouldConfirmSignIn, but not right now

wlee221 (approved):
:shipit:

PR #1178, opened 2022-01-21
Docs update - SwitchField

zchenwei:
*Issue #, if available:*

#941 

*Description of changes:*

This PR will be updating `SwitchField` docs.


docs/src/pages/components/switchfield/SwitchPropControls.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <SwitchField
+          name="isChecked"
+          onChange={() => {
+            setIsChecked(!isChecked);
+          }}
+          label="isChecked"
+        />
+        <Button
+          variation="link"
+          onClick={() => setIsChecked(undefined)}
```
joebuono, line 150: This button doesn't seem to do anything when I click it. It should control whether it's switched on or off, correct?

joebuono (commented):
Nice job! Just a few thoughts:

**General comments**:
- Let’s be consistent about referring to the primitive as `SwitchField`, not `switchField`. 
- Also, let’s be consistent about referring to the `SwitchField` as either a primitive or a component (it’s a primitive, correct?)

**Props order**:
Is there any pattern to the order in which the props are organized in the Usage section? Here’s a suggested order:

Since `label` is a required prop, perhaps the label related props can go immediately after the accessibility section:
- label
- labelPosition
- isLabelHidden

Name/value come as a pair and are somewhat related to `label`, so they could go next:
- name
- value

Checked-related props:
- isChecked
- defaultChecked

Style-related props:
- thumbColor
- trackColor
- trackCheckedColor
- size

Callback
- onChange

**Nit**: The description for `isChecked` is a little verbose. Perhaps we only need the first sentence?

docs/src/pages/components/switchfield/SwitchPropControls.tsx
zchenwei: That button is used to clear the controlled state set by `isChecked`. When it works as a controlled component, you cannot change its on/off state by simply click on it

zchenwei:
> Nice job! Just a few thoughts:
> 
> **General comments**:
> 
> * Let’s be consistent about referring to the primitive as `SwitchField`, not `switchField`.
> * Also, let’s be consistent about referring to the `SwitchField` as either a primitive or a component (it’s a primitive, correct?)
> 
> **Props order**: Is there any pattern to the order in which the props are organized in the Usage section? Here’s a suggested order:
> 
> Since `label` is a required prop, perhaps the label related props can go immediately after the accessibility section:
> 
> * label
> * labelPosition
> * isLabelHidden
> 
> Name/value come as a pair and are somewhat related to `label`, so they could go next:
> 
> * name
> * value
> 
> Checked-related props:
> 
> * isChecked
> * defaultChecked
> 
> Style-related props:
> 
> * thumbColor
> * trackColor
> * trackCheckedColor
> * size
> 
> Callback
> 
> * onChange
> 
> **Nit**: The description for `isChecked` is a little verbose. Perhaps we only need the first sentence?

For sure, I like to be consistent on referring it as `SwitchField` and primitive.

Also, thanks for bringing up the props order for discussion. I think this could also apply to the code, currently we organized these props in the code alphabetically, which is not that straightforward to look for related props or those in pair, and it could be error-prone when we need to delete or add a group of props, while organize them in group can solve this issues. Any thoughts? @reesscot @dbanksdesign

joebuono (approved):
LGTM! Let's remember to consider re-ordering the props at a later time.

PR #1181, opened 2022-01-23
Fixed setting secret key value when generating TOTP QR code

sgpopov:
*Issue #, if available:*

*Description of changes:*
The value for the secret key was incorrectly set when generating the QR code for the TOTP, thus resulting in an invalid QR being produced.

It seems that this bug was introduced by #1083  and currently, the link looks like this:

```
otpauth://totp/AWSCognito:test@google.com?secret=[object Object]&issuer=AWSCognito
```

### DIFF
--- a/packages/vue/src/components/setup-totp.vue
+++ b/packages/vue/src/components/setup-totp.vue
@@ -38,7 +38,7 @@ onMounted(async () => {
   try {
     secretKey.value = await Auth.setupTOTP(user);
     const issuer = 'AWSCognito';
-    const totpCode = `otpauth://totp/${issuer}:${user.username}?secret=${secretKey}&issuer=${issuer}`;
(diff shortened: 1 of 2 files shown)


packages/vue/src/components/setup-totp.vue
```diff
@@ -38,7 +38,7 @@ onMounted(async () => {
   try {
     secretKey.value = await Auth.setupTOTP(user);
     const issuer = 'AWSCognito';
-    const totpCode = `otpauth://totp/${issuer}:${user.username}?secret=${secretKey}&issuer=${issuer}`;
+    const totpCode = `otpauth://totp/${issuer}:${user.username}?secret=${secretKey.value}&issuer=${issuer}`;
```
ErikCH, line 41: Looks good! 👍

PR #1186, opened 2022-01-24
fix: remove docs legacy ui-react dep

reesscot:
*Description of changes:*
This fixes an issue in yarn.lock where the `ui-react-v1` was oddly pointing to v2.  ~This is fixed by pinning to the same legacy version we use in the [react/package.json](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/package.json#L53).~ 

*Update*
Digging a bit more I realized that the ui-react-legacy depency is not being used anywhere in the docs, and would be available through the `ui-react` dependency if needed in the future, so I'm removing it from the docs dependencies. 


joebuono (approved):
Good catch!

ErikCH (approved):
Looks good. That should pin it correctly

zchenwei (approved):
LGTM!

wlee221:
@reesscot Ah I know exactly what this error is! I found this while working with @calebpollman on setting up his ui environment. 

It's related to latest version of Cypress + cucumber. Can you look to apply the fix documented here? https://github.com/TheBrainFamily/cypress-cucumber-preprocessor/pull/658

PR #1187, opened 2022-01-25
Docs update - Heading

zchenwei:
*Issue #, if available:*
#933 

*Description of changes:*
This PR will be updating `Heading` docs

### DIFF
--- a/docs/src/data/links.ts
+++ b/docs/src/data/links.ts
@@ -21,7 +21,7 @@ export const baseComponents: ComponentNavItem[] = [
   {
     href: '/components/heading',
     label: 'Heading',
-    body: ``,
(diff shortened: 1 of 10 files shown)


joebuono (commented):
Good job!

When I pulled this PR and ran it locally, I noticed that in the main demo, the default heading size is 6, which is correct. However, when I selected the 'Select heading level' option, then the heading was no longer bold. Perhaps the easiest way to fix this is to remove the placeholder option 'Select heading level' from the main demo?

joebuono (approved):

zchenwei:
> Good job!
> 
> When I pulled this PR and ran it locally, I noticed that in the main demo, the default heading size is 6, which is correct. However, when I selected the 'Select heading level' option, then the heading was no longer bold. Perhaps the easiest way to fix this is to remove the placeholder option 'Select heading level' from the main demo?

Good callout, let me fix it!

PR #1189, opened 2022-01-25
Docs update - ScrollView

zchenwei:
*Issue #, if available:*
#933 

*Description of changes:*
This PR will be updating `ScrollView` docs


docs/src/data/links.ts
```diff
@@ -41,7 +41,7 @@ export const baseComponents: ComponentNavItem[] = [
   {
     href: '/components/scrollview',
     label: 'Scrollview',
```
joebuono, line 43: nit: 'ScrollView'

joebuono (approved):
LGTM!

docs/src/data/links.ts
zchenwei: Good catch

Issue #1131, opened 2022-01-12
Documentation: SelectField missing name


Jesmaster:
The documentation for the SelectField is missing the "name" prop. I tried it out in code and it works fine.

https://ui.docs.amplify.aws/components/selectfield

Jesmaster:
Also "name" and "value" are missing from the props list on the CheckboxField docs.

zchenwei:
> Also "name" and "value" are missing from the props list on the CheckboxField docs.

I think the main reason we didn't put these props are because they are not visible from page, which might cause confusion for some customers

Jesmaster:
Yeah that makes sense, maybe a note somewhere that all of the normal html properties are available would be good, need to assume it right now.

jacoblogan:
A section has been added to the docs component pages describing this behavior

PR #1192, opened 2022-01-26
Docs update - Image

zchenwei:
*Issue #, if available:*
#933 

*Description of changes:*
This PR will be updating `Image` docs


joebuono (commented):
Nice!

Regarding the main demo, I'm wondering if we need to include all the possible props for Image because it's a bit overwhelming at first glance. 

For example, I think we can remove `sizes` and `srcSet` because those props are addressed in the 'Responsive Image optimization' section, and it's confusing to know what to put into those fields in the demo. Same goes for `objectFit` and `objectPosition`. 

Other props we can probably remove from the demo are `borderRadius`, `boxShadow`, `color`,  `maxHeight`, `minHeight`, `maxWidth`, and `minWidth` because their effects aren't obvious in the demo. 

Maybe in a section below the demo we can include a list of all possible props (see the [Grid docs](https://ui.docs.amplify.aws/components/grid?platform=react#mapping-grid-css-properties-to-grid-props) as an example of listing props. 

Thoughts?

zchenwei:
> Nice!
> 
> Regarding the main demo, I'm wondering if we need to include all the possible props for Image because it's a bit overwhelming at first glance.
> 
> For example, I think we can remove `sizes` and `srcSet` because those props are addressed in the 'Responsive Image optimization' section, and it's confusing to know what to put into those fields in the demo. Same goes for `objectFit` and `objectPosition`.
> 
> Other props we can probably remove from the demo are `borderRadius`, `boxShadow`, `color`, `maxHeight`, `minHeight`, `maxWidth`, and `minWidth` because their effects aren't obvious in the demo.
> 
> Maybe in a section below the demo we can include a list of all possible props (see the [Grid docs](https://ui.docs.amplify.aws/components/grid?platform=react#mapping-grid-css-properties-to-grid-props) as an example of listing props.
> 
> Thoughts?

I think your point is valid and I do consider removing some props because of their invisible effects. But we have customers just want to see all props on demo, for example #1131 , but yea I think including a prop description section is a good point and on our radar #931, but before we start tackling that, I think we had better keep all there for customer reference

joebuono (approved):

docs/src/pages/components/image/examples/ImageObjectFitandPositionExample.tsx
```diff
@@ -0,0 +1,15 @@
+import { Image } from '@aws-amplify/ui-react';
```
wlee221, line 1: It seems we have duplicate files with different casing: 

```
$ git clone git@github.com:aws-amplify/amplify-ui.git

warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'docs/src/pages/components/image/examples/ImageObjectFitAndPositionExample.tsx'
  'docs/src/pages/components/image/examples/ImageObjectFitandPositionExample.tsx'
```

Can you look to remove one of the duplicates?

PR #1209, opened 2022-01-28
Add useAuthSignOutAction hook

reesscot:
Adds an internal hook to trigger an Auth Sign Out action. 

*Example use*:
```jsx

export const SignOutButton = () => {

  const authSignOutAction = useAuthSignOutAction({ global: true });

  return (
    <Button onClick={authSignOutAction}>
      Sign out
    </Button>
  );
};
```

*Example of reloading the page in response to action completion:*
```js
Hub.listen("ui-actions", (capsule) => {
  if (capsule.payload.event === "AuthSignOut_Finished") {
    window.location.reload();
  }
});
```

*Todo*

- [x] Add unit tests

*Description of changes:*


packages/react/src/hooks/actions/useAuthSignOutAction.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface UseAuthSignOutAction {
+  (opts?: SignOutOpts): () => Promise<void>;
+}
+
+export const useAuthSignOutAction: UseAuthSignOutAction =
+  (opts) => async () => {
+    try {
+      Hub.dispatch(HUB_AUTH_SIGNOUT_NAMESPACE, {
+        event: ACTION_EVENT_STARTED,
```
alharris-at, line 19: Don't you want the opts also attached to this as well? We should probably indicate whether or not it was global in the callback hooks.

alharris-at (commented):
Overall LGTM, probably want to add unit tests as well?

packages/react/src/hooks/actions/useAuthSignOutAction.tsx
reesscot: Good idea, will update

alharris-at (approved):

packages/react/src/hooks/actions/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const ACTIONS_CHANNEL = 'ui-actions';
+export const ACTION_AUTH_SIGNOUT = 'AuthSignOut';
+
+export const ACTION_STARTED = 'Started';
+export const ACTION_FINISHED = 'Finished';
+
+export const ACTION_AUTH_SIGNOUT_STARTED = `${ACTION_AUTH_SIGNOUT}.${ACTION_STARTED}`;
```
zchenwei, line 7: nit: The real string `AuthSignOut.Started` may look a bit weird. It seems `AuthSignOut_Started` has a good looking IMO

packages/react/src/internal.tsx
```diff
@@ -1,6 +1,7 @@
 export * from './hooks/useAuth';
 export * from './hooks/useDataStore';
 export * from './hooks/useStorageURL';
+export * from './hooks/actions/useAuthSignOutAction';
```
zchenwei, line 4: Should we start using explicit export?

packages/react/src/hooks/actions/constants.ts
reesscot: Yeah, that makes sense to me. Will update

packages/react/src/internal.tsx
reesscot: Yeah, that's probably safer. Will update

packages/react/src/hooks/actions/__tests__/useAuthSignOutAction.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    await authSignOutAction();
+    expect(signOutSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('should call Hub with started and finished events', async () => {
+    const authSignOutAction = useAuthSignOutAction();
+
+    await authSignOutAction();
+    expect(hubDispatchSpy).toHaveBeenCalledTimes(2);
+    expect(hubDispatchSpy).toHaveBeenCalledWith('ui-actions', {
```
zchenwei, line 29: nit: `ui-action` could be replaced with `ACTIONS_CHANNEL`

zchenwei (approved):
LGTM! Just one non-blocking comment

PR #1213, opened 2022-01-29
feat: adding useNavigateAction hook

zchenwei:
*Description of changes:*

This PR will be adding an internal hook to trigger a `Navigate` action.

*Example use*
```jsx
export const NavigateActionExample = () => {

  const navigateToAmazonHomepageButtonAction = useNavigateAction({
    type: 'url',
    url: 'https://www.amazon.com/',
  });

  return (
    <Button onClick={navigateToAmazonHomepageButtonAction}>
      Redirect To Amazon Homepage
    </Button>
  );
};
```

*Todo*

- [x] Add unit tests


packages/react/src/hooks/actions/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const HUB_AUTH_SIGNOUT_NAMESPACE = 'AuthSignOut';
+export const HUB_NAVIGATE_NAMESPACE = 'navigate';
```
zchenwei, line 2: I was wondering which case we should use for the channel name. Would like to be consistent with the existing ones we have, like:
```
core
auth
api
analytics
interactions
pubsub
storage
xr
datastore
```

reesscot: Yeah, I'm thinking we should add a new channel name `actions`

zchenwei: Having `actions` channel and then differentiate actions through `event`? Like `XXX_Started` and `XXX_Finished`

packages/react/src/hooks/__tests__/useNavigateAction.test.ts
```diff
@@ -0,0 +1,105 @@
+import { Hub } from '@aws-amplify/core';
+
+import {
+  ACTION_EVENT_STARTED,
+  ACTION_EVENT_FINISHED,
+  HUB_NAVIGATE_NAMESPACE,
+} from '../actions/constants';
```
reesscot, line 7: Could just be `./constants` right?

packages/react/src/hooks/actions/useNavigateAction.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Hub } from '@aws-amplify/core';
+
+import {
+  ACTIONS_CHANNEL,
+  ACTION_NAVIGATE_FINISHED,
+  ACTION_AUTH_SIGNOUT_STARTED,
+  ACTION_NAVIGATE_STARTED,
+} from './constants';
+
+type NavigateType = 'url' | 'anchor' | 'reload';
```
alharris-at, line 11: Could you export this type as well, just in case anybody wants to accept a `UseNavigateActionProps` externally?

packages/react/src/hooks/actions/useNavigateAction.ts
```diff
@@ -0,0 +1,61 @@
...
+    default:
+      run = () => {
+        console.warn(
+          'Please provide a valid navigate type. Available types are "url", "anchor" and "reload".'
+        );
+      };
+  }
+
+  const navigateAction = () => {
+    Hub.dispatch(ACTIONS_CHANNEL, { event: ACTION_NAVIGATE_STARTED });
```
alharris-at, line 55: Same comment as in scott's PR, but could you include the props in this event?

packages/react/src/hooks/actions/useNavigateAction.ts
zchenwei: Oh yeah, that's a good point. Will do

packages/react/src/hooks/actions/useNavigateAction.ts
zchenwei: For sure, will update

packages/react/src/hooks/__tests__/useNavigateAction.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    expect(Hub.dispatch).toHaveBeenCalledWith(ACTIONS_CHANNEL, {
+      event: ACTION_NAVIGATE_STARTED,
+    });
+    expect(Hub.dispatch).toHaveBeenLastCalledWith(ACTIONS_CHANNEL, {
+      event: ACTION_NAVIGATE_FINISHED,
+    });
+  };
+
+  it('Should call window.open', () => {
+    windowSpy = jest.spyOn(window, 'open');
```
reesscot, line 33: Nice, super clean!

packages/react/src/hooks/actions/useNavigateAction.ts
reesscot, line 55: Should include the props / options here in the `data` property

packages/react/src/hooks/actions/useNavigateAction.ts
```diff
@@ -0,0 +1,61 @@
...
+    run();
+    Hub.dispatch(ACTIONS_CHANNEL, { event: ACTION_NAVIGATE_FINISHED });
```
reesscot, line 57: Same comment as above

packages/react/src/hooks/actions/useNavigateAction.ts
```diff
@@ -0,0 +1,61 @@
...
+  anchor?: string;
+
+  target?: React.HTMLAttributeAnchorTarget;
+}
+
+export const defaultTarget = '_self';
+
+export const useNavigateAction = (props: UseNavigateActionProps) => {
+  const { type, url, anchor, target } = props;
+  let run: NavigateRun;
```
reesscot, line 29: Should this be wrapped in useCallback since hooks are run on every render and we don't want to recreate the `run` function this unless the props change?

packages/react/src/hooks/actions/useNavigateAction.ts
```diff
@@ -0,0 +1,61 @@
...
+  url?: string;
+
+  anchor?: string;
+
+  target?: React.HTMLAttributeAnchorTarget;
+}
+
+export const defaultTarget = '_self';
+
+export const useNavigateAction = (props: UseNavigateActionProps) => {
```
reesscot, line 27: Nit: could call this `options` since `props` feels more like a react component.

reesscot (changes requested):
Looking good, just a few comments

packages/react/src/hooks/actions/useNavigateAction.ts
zchenwei: Thats a good point. Will update

packages/react/src/internal.tsx
```diff
@@ -1,6 +1,7 @@
 export * from './hooks/useAuth';
 export * from './hooks/useDataStore';
 export * from './hooks/useStorageURL';
+export { useNavigateAction } from './hooks/actions/useNavigateAction';
```
reesscot, line 4: Should we also export `UseNavigateActionOptions`?

reesscot (approved):
One final comment, otherwise LGTM!

packages/react/src/internal.tsx
zchenwei: Yea, that would be great. Will update!

Issue #1223, opened 2022-01-31
The AmplifyAuthenticatorModule for Angular throws StateMachine errors


stocaaro:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Angular

### Which UI component?

Authenticator

### How is your app built?

Angular

### Please describe your bug.

When using the `AmplifyAuthenticatorModule` from `@aws-amplify/ui-angular`, I see the error:
```
Error: node_modules/xstate/lib/types.d.ts:567:18 - error TS2430: Interface 'StateMachine<TContext, TStateSchema, TEvent, TTypestate, TAction, TServiceMap, TResolvedTypesMeta>' incorrectly extends interface 'StateNode<TContext, TStateSchema, TEvent, TTypestate, TServiceMap, TResolvedTypesMeta>'.
``` 

### What's the expected behaviour?

I expected the UI component to render for my application.

### Help us reproduce the bug!

Following the [Getting started with Angular steps](https://docs.amplify.aws/start/getting-started/setup/q/integration/angular/) I ran:
- `npx -p @angular/cli ng new amplify-app` (Y, CSS)
- `cd amplify-app`
- Add the suggested content to the `src/polyfills.ts` file
- `amplify init` (\n, Y,AWS profile, default)
- Here's where I think I screwed up
  - I used `npm install --save aws-amplify @aws-amplify/ui-angular`
  - When I should have run `npm install --save aws-amplify @aws-amplify/ui-angular@1.x.x`
  - Resulting in errors about `AmplifyUIAngularModule` missing
  - **Following these steps exactly result in using @aws-amplify/ui-angular 1.0.35, which works with the steps in this flow.**
- From here, I moved over to the [ui docs guide](https://ui.docs.amplify.aws/getting-started/installation?platform=angular)
- Adding `AmplifyAuthenticatorModule` to `app.module.ts` as described  causes the `StateMachine` error.
- I tested turning all strict settings in `tsconfig.json` off and the same error occurr.

### Code Snippet

```typescript
[...]

PR #1224, opened 2022-02-01
Pin xstate versions

wlee221:
*Issue #, if available:*

*Description of changes:* This pins all `xstate` version to last known working version in `yarn.lock`. This fixes Typescript error found in https://github.com/aws-amplify/amplify-ui/issues/1223, which actually happened to React/Vue as well as Angular. 

Not sure why this wasn't caught in `no-lock` test we have, we shoudl look into it.

### DIFF
--- a/packages/e2e/cypress/plugins/index.js
+++ b/packages/e2e/cypress/plugins/index.js
@@ -8,7 +8,7 @@
 // ***********************************************************
 
 const cucumber = require('cypress-cucumber-preprocessor').default;
-const resolve = require('resolve');
+const path = require('path');
 require('dotenv-safe').config({
   allowEmptyValues: true,
 });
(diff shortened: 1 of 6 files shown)


Milan-Shah (approved):

packages/e2e/cypress/plugins/index.js
```diff
@@ -8,7 +8,7 @@
 // ***********************************************************
 
 const cucumber = require('cypress-cucumber-preprocessor').default;
-const resolve = require('resolve');
+const path = require('path');
```
wlee221, line 11: This was a known bug whenever `yarn.lock` was updated -- https://github.com/TheBrainFamily/cypress-cucumber-preprocessor/pull/658/files

PR #1227, opened 2022-02-01
Remove --prefer-offline tag

wlee221:
*Issue #, if available:*

*Description of changes:* `--prefer-offline` will prefer whatever's on cache over whatever's on latest. This means that CI can possibly install versions that are different from what customers would install, based on what's on `yarn cache`.

For example, if we depend on `A: ^1.0.0`, CI cache has `A: 1.0.5`, and the latest is `A: 1.0.10`:
- CI will install `1.0.5` because it's on cache
- Customers will install `1.10.0` because it's the latest. 

Removing this flag to enforce consistent testing.

### DIFF
--- a/.github/workflows/publish-latest.yml
+++ b/.github/workflows/publish-latest.yml
@@ -20,7 +20,7 @@ jobs:
           node-version: lts/*
           cache: 'yarn'
       - name: Install and build packages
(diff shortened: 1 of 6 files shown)


jacoblogan (approved):
LGTM

zchenwei (approved):

PR #1228, opened 2022-02-01
feat: adding useStateMutationHook

zchenwei:
*Description of changes:*

This PR will be adding `useStateMutationAction`

*Example use*

```jsx
import { Flex, TextField } from '@aws-amplify/ui-react';

export const Example = () => {
  const [userName, setUserName] = useStateMutationAction('example');

  const handleChange = (event) => {
     setUserName(event.currentTarget.value);
  }

  return (
    <Flex>
      <TextField label="Username" value={userName} onChange={handleChange} />
    </Flex>
  );
}
```
*Todo*

- [x] Add unit tests


packages/react/src/hooks/actions/useStateMutationAction.ts
```diff
@@ -0,0 +1,38 @@
...
+
+type UseStateMutationAction = [string, (event: StateChangeEvent) => void];
+
+export const useStateMutationAction = (
+  initialState: string
+): UseStateMutationAction => {
+  const [state, setState] = React.useState(initialState);
+
+  const setNewState = React.useCallback((event: StateChangeEvent) => {
+    const newState = event.currentTarget.value;
```
zchenwei, line 22: I was wondering how general the use cases of `useStateMutationAction` would be. From [the golden file](https://code.amazon.com/packages/AmplifyStudioCommon/blobs/d76a84faef43f6ad8ea0b36031d6a932864b8f66/--/component_golden_files/react/form.tsx#L19), it is limited to only be used on form fields, which means the mutable state will always be a `string`.

zchenwei: Had a discussion with @reesscot, we should have codegen to handle the event part: 
```ts
const [userNameTextFieldValue, setUserNameTextFieldValue] = useStateMutationAction('vizsla');
const handleOnChange = (e) => {
 const value = e.target.value; // may need some kind of check if value exists
 setUserNameTextFieldValue(value)
}
```

Will update the golden file

alharris-at: Discussed offline, but I think we should just have you accept a value T as input, and effectively just wrap the setState hook w/ the hub actions. This way it's easier for customers to use this in the same way they use `useState` today.

packages/react/src/hooks/actions/useStateMutationAction.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    Hub.dispatch(ACTIONS_CHANNEL, {
+      event: ACTION_STATE_MUTATION_STARTED,
+      data: newState,
```
alharris-at, line 26: Do you want to embed oldState and newState in the events?

zchenwei: That sounds good!

alharris-at (approved):

packages/react/src/hooks/actions/useStateMutationAction.ts
```diff
@@ -0,0 +1,37 @@
...
+type UseStateMutationAction<T> = [T, (newState: T) => void];
+
+export const useStateMutationAction = <T>(
+  initialState: T
+): UseStateMutationAction<T> => {
+  const [state, setState] = React.useState(initialState);
+
+  const setNewState = React.useCallback(
+    (newState: T) => {
+      const oldState = state;
```
reesscot, line 19: Would prefer to call this `prevState` rather than `oldState`, which is more in line with React documentation.
```suggestion
      const prevState = state;
```
https://reactjs.org/docs/hooks-reference.html

packages/react/src/hooks/actions/useStateMutationAction.ts
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { Hub } from '@aws-amplify/core';
+
+import {
+  ACTIONS_CHANNEL,
+  ACTION_STATE_MUTATION_FINISHED,
+  ACTION_STATE_MUTATION_STARTED,
+} from './constants';
+
+type UseStateMutationAction<T> = [T, (newState: T) => void];
```
reesscot, line 10: I'm not a big fan of single letter Type variables. Please give T a name, like `StateType`.

reesscot (changes requested):
Looks great, just a few requests.

packages/react/src/hooks/actions/useStateMutationAction.ts
zchenwei: It seems we normally would use a single letter for generic type, which I do see the value by differentiating it from other interfaces or types when navigating through the code

packages/react/src/hooks/actions/useStateMutationAction.ts
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { Hub } from '@aws-amplify/core';
```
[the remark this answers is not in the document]
reesscot: On that note, can you please also fix the import on useNavigateAction to import from `aws-amplify`?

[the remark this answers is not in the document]
zchenwei: I have another PR where I am moving the useNav test file to the correct folder. Will include the update there <- #1239

packages/react/src/hooks/actions/useStateMutationAction.ts
reesscot: It's standard for generics in the Typescript guide, but I think it's bad practice because it doesn't say anything about what Type is supposed to be passed in. The `<>` indicate that it's a generic, I don't think we have to be vague about what the type should be. 
See: https://twitter.com/kentcdodds/status/1083831867294441472
https://fettblog.eu/tidy-typescript-name-your-generics/

zchenwei: Updated!

reesscot (approved):
:shipit:

Issue #1222, opened 2022-01-31
Typing in Search component of Collection breaks 


reesscot:
If a the `items` prop passed to Collection contains a `null` value, typing in the search will throw an error.

alharris-at:
Minimal Reproduction using the examples provided in docs https://ui.docs.amplify.aws/components/collection

### Create Test App

```sh
npx create-react-app bug-repro
cd bug-repro
npm install @aws-amplify/ui-react
cat > src/App.js <<- EOM
import {
  Collection,
  Card,
  Heading,
  Text,
  AmplifyProvider,
  defaultTheme
} from '@aws-amplify/ui-react';
import '@aws-amplify/ui-react/styles.css';

export default function App() {
  const visitNewZealand = [
    {
      title: 'Fiordland National Park',
      description: 'This national park includes the famous fjords of Milford, Dusky and Doubtful Sounds.',
    },
    {
      title: 'Bay of Islands, North Island',
      description: 'Three hours north of Auckland, this area features over 144 islands to explore.',
    },
    {
      title: 'Queenstown, South Island',
      description: null,
    },
  ];
  return (
    <AmplifyProvider theme={defaultTheme}>
      <Collection
        type='list'
        items={visitNewZealand}
        gap='1.5rem'
        isSearchable
      >
        {(item, index) => (
          <Card key={index} padding='1rem'>
            <Heading level={4}>{item.title}</Heading>
            <Text>{item.description}</Text>
          </Card>
        )}
      </Collection>
    </AmplifyProvider>
  );
}
EOM
npm start
```

### Reproduce Bug in UI

* Navigate to localhost:3000
* Enter search text into collection search field.

alharris-at:
Because data synced from datastore (a common usecase) allows for nulls, specifically in the `_deleted` field if the item hasn't been soft-deleted. This is likely that folks will run into this.

reesscot:
Thanks for the details @alharris-at!

PR #1243, opened 2022-02-03
fix: fixing null value error in Collection search

zchenwei:
*Issue #, if available:*
#1222 

Fixing error in Collection search when `items` prop contains `null` value


zchenwei:
> Any idea why yarn.lock file changed?

It was changed after pulling down from main

reesscot (approved):
LGTM

PR #1251, opened 2022-02-03
Pin all dependencies

joebuono:
*Issue #, if available:*

*Description of changes:*

This PR pins all dependencies in the `ui`, `angular`, and `vue` repos based on the `yarn.lock` file. 

This is in addition to the previously pinned `react` dependencies. 

### DIFF
--- a/.changeset/shiny-badgers-film.md
+++ b/.changeset/shiny-badgers-film.md
@@ -1,0 +1,7 @@
+---
+"@aws-amplify/ui": patch
+"@aws-amplify/ui-vue": patch
+"@aws-amplify/ui-angular": patch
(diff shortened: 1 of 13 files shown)


.changeset/shiny-badgers-film.md
```diff
@@ -0,0 +1,8 @@
+---
+"amplify-ui-angular-mono": patch
```
wlee221, line 2: Can you remove `amplify-ui-angular-mono`? This is a private dev folder, but changeset adds it by default 😓

wlee221 (approved):
Looks good, let's wait for test results

PR #1253, opened 2022-02-03
feat: add data store action hooks

reesscot:
Adds internal hooks to execute DataStore Create, Update, and Delete actions.

*Create example use*:
```tsx
export const CreateCustomerButton = () => {
  const createCustomerAction = useDataStoreCreateAction({
    model: Customer,
    fields: {
      firstName: 'Din',
      lastName: 'Djarin',
    },
  });

  return (
    <Button onClick={createCustomerAction}>
      Create
    </Button>
  );
};
```

*Update example*
```tsx
export const UpdateCustomerButton = () => {
  const { overrides, ...rest } = props;

  const updateCustomerAction = useDataStoreUpdateAction({
    model: Customer,
    id: 'd9887268-47dd-4899-9568-db5809218751',
    fields: {
      firstName: 'Din',
      lastName: 'Djarin',
    },
  });

  return (
    <Button onClick={updateCustomerAction}>
      Update
    </Button>
  );
};
```

*Delete example*
```tsx
export const DeleteCustomerButton = () => {

  const deleteCustomerAction = useDataStoreDeleteAction({
    model: Customer,
    id: 'd9887268-47dd-4899-9568-db5809218751',
  });

  return (
    <Button onClick={deleteCustomerAction}>
      Delete
    </Button>
  );
};
```

*Todo*

- [x] Add unit tests


packages/react/src/hooks/actions/useDataStoreUpdate.ts
```diff
@@ -0,0 +1,71 @@
...
+> {
+  model: PersistentModelConstructor<Model>;
+  id: string;
+  fields: UseDataStoreUpdateActionFields<Model>;
+}
+
+type UseDataStoreUpdateActionFields<Model> = ModelInit<
+  Model,
+  { readOnlyFields: 'createdAt' | 'updatedAt' }
+>;
```
zchenwei, line 29:
```suggestion
export interface UseDataStoreUpdateActionOptions<
  Model extends PersistentModel
> {
  model: PersistentModelConstructor<Model>;
  id: string;
  fields: ModelInit<Model, { readOnlyFields: 'createdAt' | 'updatedAt' }>;
}
```

packages/react/src/hooks/actions/useDataStoreUpdate.ts
```diff
@@ -0,0 +1,71 @@
...
+        data: { fields, id },
+        message: ACTION_DATASTORE_UPDATE_STARTED_MESSAGE,
+      });
+
+      const original = await DataStore.query(model, id);
+      // If query by id doesn't return an item,
+      // original will be undefined
+      // so we'll throw a helpful message.
+      if (!original) {
+        throw new Error(`${DATASTORE_QUERY_BY_ID_ERROR}: ${id}`);
```
zchenwei, line 50: Should we throw an error for a query error? Maybe `console.error` would be more gentle? That way users don't need to add a `try...catch` block around it

zchenwei (commented):
Looking good! Left 2 comments for discussion.

packages/react/src/hooks/actions/useDataStoreUpdate.ts
reesscot: Good point, these can definitely be merged. Will update

packages/react/src/hooks/actions/useDataStoreUpdate.ts
reesscot: This will be caught by the surrounding [try/catch](https://github.com/aws-amplify/amplify-ui/pull/1253/files/f9236b03bcc9c6787f5edf216152745c6203bcbe#diff-fba6e7642d5c788d3b2c19b66c54d008f395dfe4f554d13fec405e7fa658ae48R38) and will be reported in a hub event (we won't expect the consumer to catch the error). The main reason I throw is because if original is underfined, then [Datastore.save](https://github.com/aws-amplify/amplify-ui/pull/1253/files/f9236b03bcc9c6787f5edf216152745c6203bcbe#diff-fba6e7642d5c788d3b2c19b66c54d008f395dfe4f554d13fec405e7fa658ae48R53) throws with a cryptic error that doesn't really identify the problem, which is that the original item wasn't found, so it can't be copied and updated.

zchenwei: Oh, I miss that. My bad but yea that makes sense

zchenwei (approved):

packages/react/src/hooks/actions/useDataStoreUpdate.ts
reesscot: Fixed

packages/react/src/internal.tsx
```diff
@@ -11,6 +11,13 @@ export {
 } from './hooks/actions/useNavigateAction';
```
0618, line 11: Just curious, why is this file `.tsx`?

alharris-at (approved):

packages/react/src/internal.tsx
reesscot: Good point, should be a ts file. Will update.

Issue #1244, opened 2022-02-03
Menu Component Incorrect Font Family


bestickley:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

N/A

### Please describe your bug.

The font-family of the buttons within the Menu component do not use Inter.

### What's the expected behaviour?

font-family of buttons in Menu component will be Inter by default.

### Help us reproduce the bug!

Docs for Menu component: https://ui.docs.amplify.aws/components/menu

### Code Snippet

N/A

### Additional information and screenshots

![image](https://user-images.githubusercontent.com/35735118/152347817-733253ea-79bf-4e0d-8112-f6c14e7c1f1b.png)

reesscot:
Thanks for reporting this issue @bestickley!

jacoblogan:
@bestickley thanks again for your report, we really appreciate it.  A PR has been merged in recently to resolve this issue, can you verify that it is working as you expect it?

PR #1259, opened 2022-02-04
update all theming to happen on a root level

jacoblogan:
*Issue #1244

*Description of changes: Update theming to be applied at the root level, this is address an issue that was seen with the menu content which is created outside the scope of the current theme declaration.


packages/ui/src/theme/createTheme.ts
```diff
@@ -78,7 +78,7 @@ export function createTheme(
   // flattenProperties is another internal Style Dictionary function
   // that creates an array of all tokens.
   let cssText =
-    `[data-amplify-theme="${name}"] {\n` +
+    `:root, [data-amplify-theme="${name}"] {\n` +
```
reesscot, line 81: If we define this at the root, shouldn't we _only_ define it at the root (potentially removing `[data-amplify-theme="${name}"]` ? It seems like having it defined at both locations will cause issues... Thoughts, @dbanksdesign.

dbanksdesign: CSS variables are inherited, so in the CSSOM the :root set of variables will get overridden by the [data-amplify-theme] at that point. Since both sets are the same it shouldn't really do anything... unless someone overrode a variable at :root it would get overridden. If we do remove the [data-amplify-theme] completely we need to rethink how we handle the theme overrides, because right now those happen on that element...

dbanksdesign: Keeping [data-amplify-theme] selector would make sure overrides like dark mode work, however because the original issue (themes not working on Portal elements like Menu), overrides would not work those elements...

dbanksdesign: We could change how we do overrides by injecting data attributes on the html element directly...

reesscot: Yeah, I was thinking we would do all overrides at the html element level so that we can affect Portal elements.

dbanksdesign: @jacoblogan does that make sense? We'll need to update the AmplifyProvider code for that

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -21,16 +21,23 @@ export function AmplifyProvider({
...
     <AmplifyContext.Provider
       value={{
         theme: webTheme,
       }}
     >
       <IdProvider>
-        <div data-amplify-theme={name} data-amplify-color-mode={colorMode}>
-          {children}
-        </div>
+        <div>{children}</div>
```
dbanksdesign, line 40: Do we need the `<div>` here? Not blocking, just thinking out loud.

dbanksdesign (approved):
LGTM! Just one non-blocking comment.

dbanksdesign:
Oh wait it looks like the CSS changes got lost somewhere...

packages/react/src/components/AmplifyProvider/index.tsx
zchenwei: +1

PR #1262, opened 2022-02-04
Add `yarn setup` and `yarn refresh` monorepo script

wlee221:
*Issue #, if available:*

*Description of changes:* Now you can run `yarn setup` at monorepo, that does `yarn && yarn build`. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
   "private": true,
   "scripts": {
     "build": "turbo run build --include-dependencies --no-deps --scope=*@aws-amplify/*",
-    "dev": "yarn install && yarn build",
     "clean": "turbo run clean && rm -rf node_modules/.cache",
+    "setup": "yarn install && yarn build",
     "angular": "yarn workspace @aws-amplify/ui-angular",
     "react": "yarn workspace @aws-amplify/ui-react",
     "ui": "yarn workspace @aws-amplify/ui",
--- a/package.json
+++ b/package.json
@@ -1,8 +1,9 @@
 {
   "private": true,
   "scripts": {
-    "clean": "turbo run clean && rm -rf node_modules/.cache",
     "build": "turbo run build --include-dependencies --no-deps --scope=*@aws-amplify/*",
+    "dev": "yarn install && yarn build",
+    "clean": "turbo run clean && rm -rf node_modules/.cache",
     "angular": "yarn workspace @aws-amplify/ui-angular",
     "react": "yarn workspace @aws-amplify/ui-react",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,8 +1,9 @@
 {
   "private": true,
   "scripts": {
-    "clean": "turbo run clean && rm -rf node_modules/.cache",
     "build": "turbo run build --include-dependencies --no-deps --scope=*@aws-amplify/*",
+    "dev": "yarn install && yarn build",
+    "clean": "turbo run clean && rm -rf node_modules/.cache",
```
wlee221, line 6:
```suggestion
    "setup": "turbo run build --include-dependencies --no-deps --scope=*@aws-amplify/*",
    "clean": "turbo run clean && rm -rf node_modules/.cache",
    "dev": "yarn install && yarn build",
```

package.json
```diff
@@ -1,8 +1,9 @@
 {
   "private": true,
   "scripts": {
-    "clean": "turbo run clean && rm -rf node_modules/.cache",
     "build": "turbo run build --include-dependencies --no-deps --scope=*@aws-amplify/*",
+    "clean": "turbo run clean && rm -rf node_modules/.cache",
+    "setup": "yarn install && yarn build",
```
wlee221, line 6: This is a fairly simple script, so I'm open to discussion around whether we actually need it. 

Motivation for me was:
- I liked the previous DX of following up `yarn` with `yarn build` (which was removed to prevent repeated `yarn build`s in CI/local).
- This can help new devs (?) -- question mark because `yarn && yarn build` is a common DX.

dbanksdesign: I like it. People can still use `yarn && yarn build` if they want, but this just makes it easier

reesscot: I do this all the time, so I like it. Can we add it to the contributing as well though? I'd also like a reset that does `yarn clean && yarn install && yarn build`

ErikCH: 👍

dbanksdesign (approved):

package.json
wlee221: I like `yarn reset`! (or `yarn refresh` so that it conveys it's reinstalling stuffs) I'll do that here.

For contributing, I'll create a separate PR because there are other long needed changes, so I'll centralize those in a separate one.

wlee221: [7d4a624](https://github.com/aws-amplify/amplify-ui/pull/1262/commits/7d4a62471f4660853443ec10d45bfc6afaf7819e)

PR #1272, opened 2022-02-08
Update action hub events

reesscot:
*Description of changes:*
Update action hook Hub events to use `ui` channel and improve event names.


alharris-at (approved):

0618 (commented):
Interesting, why do we need to rename `packages/react/src/internal.ts` back to `.tsx`?

0618:
Would you provide some context why we need to change the action hub event to use `ui` channel?

reesscot:
> Interesting, why do we need to rename `packages/react/src/internal.ts` back to `.tsx` after #1253 ?

It broke the build after renaming it to `ts`. The internal.js file was no longer being built to the `dist` folder. I didn't have time to dig into it further so I reverted that change.

reesscot:
> Would you provide some context why we need to change the action hub event to use `ui` channel?

We had a long discussion offline about the correct channel to use, and landed on `ui` being the best channel for any future `ui` action events. `ui` is also more consistently named with the already existing categories:
https://docs.amplify.aws/lib/utilities/hub/q/platform/js/#channels

0618 (approved):

packages/react/src/hooks/actions/__tests__/useAuthSignOutAction.test.ts
```diff
@@ -2,10 +2,12 @@ import { useAuthSignOutAction } from '../useAuthSignOutAction';
 
 import { Auth, Hub } from 'aws-amplify';
 import {
-  ACTIONS_CHANNEL,
   ACTION_AUTH_SIGNOUT_FINISHED,
   ACTION_AUTH_SIGNOUT_STARTED,
+  EVENT_ACTION_AUTH_SIGNOUT,
+  UI_CHANNEL,
 } from '../constants';
+import { AMPLIFY_SYMBOL } from '../../../helpers/constants';
```
dbanksdesign, line 10: Should `AMPLIFY_SYMBOL` just be in `constants` rather than `helpers/constants`? It would make these imports a bit cleaner throughout.

packages/react/src/helpers/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const AMPLIFY_SYMBOL = (typeof Symbol !== 'undefined' &&
```
dbanksdesign, line 1: Might be helpful to add a comment why we have this/what this is for. I kinda get it, but I also kinda don't.

packages/react/src/hooks/actions/constants.ts
```diff
@@ -1,40 +1,52 @@
...
-export const ACTION_DATASTORE_DELETE_FINISHED_ERRORS_MESSAGE = `${ACTION_DATASTORE_DELETE} action ${ACTION_FINISHED} with errors`;
-
-export const ACTION_DATASTORE_UPDATE_STARTED = `${ACTION_DATASTORE_UPDATE}_${ACTION_STARTED}`;
-export const ACTION_DATASTORE_UPDATE_FINISHED = `${ACTION_DATASTORE_UPDATE}_${ACTION_FINISHED}`;
-export const ACTION_DATASTORE_UPDATE_STARTED_MESSAGE = `${ACTION_DATASTORE_UPDATE} action ${ACTION_STARTED}`;
-export const ACTION_DATASTORE_UPDATE_FINISHED_MESSAGE = `${ACTION_DATASTORE_UPDATE} action ${ACTION_FINISHED}`;
-export const ACTION_DATASTORE_UPDATE_FINISHED_ERRORS_MESSAGE = `${ACTION_DATASTORE_UPDATE} action ${ACTION_FINISHED} with errors`;
+/**
+ * UI Actions use the `ui` channel
+ * Format for `ui` channel events is EVENT_TYPE:CATEGORY:NAME:STATUS
```
dbanksdesign, line 3: This comment is helpful.

yarn.lock
```diff
@@ -4450,6 +4450,11 @@
   resolved "https://registry.yarnpkg.com/@next/env/-/env-11.1.3.tgz#dc698e00259242012955e43a40788fcf21ba9e37"
   integrity sha512-5+vaeooJuWmICSlmVaAC8KG3O8hwKasACVfkHj58xQuCB5SW0TKW3hWxgxkBuefMBn1nM0yEVPKokXCsYjBtng==
 
+"@next/env@11.1.4":
```
dbanksdesign, line 4453: Why are these yarn lock changes in here? It doesn't look like any dependencies changed?

dbanksdesign (approved):
A few non-blocking things. LGTM

yarn.lock
reesscot: hmm, not sure. I'll remove the yarn.lock file changes for now

reesscot: I reset hard against main and nothing changed. I think this is showing because main has a new version of `yarn.lock` and the `actions` branch I'm merging into doesn't have them. I'll cleanup the actions branch before we do a merge to `main`.

packages/react/src/hooks/actions/__tests__/useAuthSignOutAction.test.ts
reesscot: that's a good idea, I'll clean this up in follow up pr

PR #1278, opened 2022-02-09
Dedupe `aws-amplify` versions

reesscot:
*Description of changes:*
This PR fixes an issue where we were pulling in two separate versions of the `aws-amplify` dependencies, 4.3.14 and 4.3.8. Now we will only have the latest version of `aws-amplify` in the node_modules folder.

Note: I found this bug when making an examples app of DataStore hooks. Having multiple versions of `@aws-amplify/datastore` in `node_modules` causes DataStore not to fully initialize, leaving the `DataStore.save` call to throw the following cryptic error message:
```
Cannot read properties of undefined (reading 'version')
```

### DIFF
--- a/examples/vue/package.json
+++ b/examples/vue/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@aws-amplify/ui-vue": "^2.1.1",
-    "aws-amplify": "^4.1.3",
+    "aws-amplify": "latest",
     "vue": "^3.0.5",
     "vue-router": "4"
   },
--- a/canary/angular/angularcli/package.json
+++ b/canary/angular/angularcli/package.json
@@ -20,7 +20,7 @@
     "@angular/platform-browser-dynamic": "~11.2.14",
     "@angular/router": "~11.2.14",
     "@aws-amplify/ui-angular": "^2.0.11",
-    "aws-amplify": "^4.3.13",
+    "aws-amplify": "latest",
     "rxjs": "~6.6.0",
     "tslib": "^2.0.0",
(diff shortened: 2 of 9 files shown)


dbanksdesign (approved):
LGTM! :shipit:

joebuono (approved):

examples/vue/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@aws-amplify/ui-vue": "^2.1.0",
-    "aws-amplify": "^4.1.3",
+    "aws-amplify": "latest",
```
zchenwei, line 14: Should we do the same for other packages in the repo? They are in `^4.3.13`

package.json
```diff
@@ -51,6 +51,7 @@
     "packages/angular/projects/ui-angular"
   ],
   "resolutions": {
+    "aws-amplify": "latest",
```
zchenwei, line 54: I would prefer update this under each package itself. IMO, `resolutions` is the last resort to pin transitive dependencies

zchenwei (commented):
Nice find! Left my thoughts for discussion.

examples/vue/package.json
wlee221: +1

reesscot: Yes, I will update this PR to include all packages.

joebuono (approved):
LGTM!

jacoblogan (approved):
LGTM

zchenwei (approved):
LGTM!

PR #1297, opened 2022-02-11
Updating line-heights to be consistent and adding some simple reset

dbanksdesign:
#### Description of changes

Made line heights for controls (fields and buttons) consistent.

* Removed overrides in the docs messing up the default theme
* Updated the default theme line heights
* Added some simple CSS resets and global styles

Ran the docs site locally:

https://user-images.githubusercontent.com/321279/153547457-ab57099f-bc03-48d6-a4ea-fdc532f3c042.mp4

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/ui/src/theme/css/styles.scss
+++ b/packages/ui/src/theme/css/styles.scss
@@ -44,15 +44,6 @@ select {
   html:focus-within {
    scroll-behavior: auto;
   }
-  
-  *,
-  *::before,
-  *::after {
-    animation-duration: 0.01ms !important;
-    animation-iteration-count: 1 !important;
-    transition-duration: 0.01ms !important;
-    scroll-behavior: auto !important;
-  }
 }
 
 [class*='amplify'] {
--- a/packages/ui/src/theme/css/styles.scss
+++ b/packages/ui/src/theme/css/styles.scss
@@ -4,19 +4,55 @@
 
 @import 'dist/theme';
 
-[data-amplify-theme] {
+// Keeping these selector low (001) so it can be easily overridden
+html {
   font-family: var(--amplify-fonts-default-static);
 }
+
 @supports (font-variation-settings: normal) {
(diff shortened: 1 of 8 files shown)


packages/ui/src/theme/css/styles.scss
```diff
@@ -4,21 +4,57 @@
...
+
+@media (prefers-reduced-motion: reduce) {
+  html:focus-within {
+   scroll-behavior: auto;
+  }
+  
+  *,
+  *::before,
+  *::after {
+    animation-duration: 0.01ms !important;
```
0618, line 51: Are we should we have to use `!important` here? If possible, we may prefer to save it as the last resource. https://css-tricks.com/when-using-important-is-the-right-choice/

This case feels like we need to use `!important`, but I'm not sure, so just to double-check 😁

joebuono (approved):
LGTM!

packages/ui/src/theme/css/styles.scss
dbanksdesign: Yea I should have added a comment here, we can maybe remove those styles. I got them from a CSS reset: https://piccalil.li/blog/a-modern-css-reset/ it does force our customers to make inclusive apps (respecting reduced motion settings), but maybe it is a little heavy-handed

packages/ui/src/theme/css/styles.scss
```diff
@@ -4,21 +4,48 @@
...
 // needed due to very low specificity of 0 0 0
 * {
   box-sizing: border-box;
 }
 
+input,
+button,
+textarea,
+select {
+  font: inherit;
```
zchenwei, line 40: Shouldn't this be inherited by default? Do we set anything that affect those elements?

zchenwei (approved):
LGTM!

PR #1302, opened 2022-02-11
update SwitchField to allow updates on controlled component

jacoblogan:
#### Description of changes
Update `SwitchField` logic to allow modifications of the checked state through the `onChange` handler.  Also updated the documentation to reflect this change and display an example of the change.

Additional tests were added to capture the changed behavior and manually validated the behavior through the documentation site.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

switch

### DIFF
--- a/.changeset/fifty-keys-move.md
+++ b/.changeset/fifty-keys-move.md
@@ -1,0 +1,6 @@
+---
+"docs": patch
+"@aws-amplify/ui-react": patch
+---
+
+update SwitchField to allow user updates on controlled component
--- a/.changeset/fifty-keys-move.md
+++ b/.changeset/fifty-keys-move.md
@@ -1,5 +1,4 @@
 ---
-"docs": patch
 "@aws-amplify/ui-react": patch
 ---
 
--- a/docs/src/pages/components/switchfield/react.mdx
+++ b/docs/src/pages/components/switchfield/react.mdx
@@ -45,7 +45,9 @@ The most basic usage simply includes a SwitchField component passing in a requir
 
 ### isChecked
 
-The `isChecked` property is a `boolean` value that defines if the switch is on or off. This is a boolean property and being set to either true or false will change this into a controlled switchField which can only be switched on or off using this controlled value.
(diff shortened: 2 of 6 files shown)


docs/src/pages/components/switchfield/react.mdx
```diff
@@ -45,7 +45,9 @@ The most basic usage simply includes a SwitchField component passing in a requir
 
 ### isChecked
 
-The `isChecked` property is a `boolean` value that defines if the switch is on or off. This is a boolean property and being set to either true or false will change this into a controlled switchField which can only be switched on or off using this controlled value.
+The `isChecked` property is a `boolean` value that defines if the switch is on or off. This is a boolean property and being set to either true or false will change this into a controlled switchField.
+If you would like to be able to toggle a controlled `SwitchField` through the click event of the `SwitchField` component, you will need to pass in an `onChange` event function which updates the controlled value.
+This is demonstrated in the example below.
```
joebuono, line 50: Perhaps we can rephrase the above to avoid being redundant? Suggestion: 

The `isChecked` prop is a `boolean` value that defines if the SwitchField is on or off. 

If you would like to be able to toggle a controlled `SwitchField`, you can pass in an `onChange` event handler to update the controlled value.

joebuono (commented):
Looking good. Just a reminder that we don't want to add changesets for docs updates because it messes up our release pipeline 👍

docs/src/pages/components/switchfield/react.mdx
dbanksdesign: What if we had a section called 'Controlled component' or something, similar to what we do here: https://ui.docs.amplify.aws/components/togglebutton#controlled-component and talk about `isChecked` and `onChange` in the same paragraph.

dbanksdesign (approved):

.changeset/fifty-keys-move.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
```
dbanksdesign, line 2: Could you remove this line? we don't need to add docs to the changeset

PR #1321, opened 2022-02-15
chore: using aria-describedby on field controls

zchenwei:
#### Description of changes

This PR will update all the field controls to use `aria-describedby` as part of our accessibility goal.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


dbanksdesign (approved):
LGTM! :shipit:

packages/react/src/primitives/StepperField/StepperField.tsx
```diff
@@ -103,7 +105,7 @@ const StepperFieldPrimitive: Primitive<StepperFieldProps, 'input'> = (
             onClick={handleDecrease}
             size={size}
           >
-            <IconRemove data-testid={DECREASE_ICON} size={size} />
```
reesscot, line 106: Should we remove the size prop from line 40 as well?

zchenwei: Good catch! Will do

PR #1331, opened 2022-02-15
feat: adding Loader to loading state Button

zchenwei:
#### Description of changes
 
This PR will add a `Loader` to `Button` in loading state.

Also, updated the primitive preview on home page to keep `Loader` aligned with other primitive horizontally

| before | after |
| ----- | ----- |
| 154147515-369b9565-a6e7-4b22-9ed9-e6601218fb84 | Screen Shot 2022-02-15 at 12 28 46 PM |

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


0618 (approved):

0618:
LGTM. 

Just a small comment. If the spinner is not related to the alignment, I would separate them into 2 PRs, because if somehow have to revert one of them, we don't have to loose the other one.

zchenwei:
> LGTM.
> 
> Just a small comment. If the spinner is not related to the alignment, I would separate them into 2 PRs, because if somehow have to revert one of them, we don't have to loose the other one.

Oh yea, here it is just a one liner . Since I changed the `Loader` size tokens so need to check back the doc site to see if its preview layout get affected, kinda related <- https://github.com/aws-amplify/amplify-ui/pull/1331/files#diff-04693c5b8676a0fcb2cf25ccbdbe05eb973223d6911760b8034b8bd79e605639R95

packages/ui/src/theme/css/component/button.scss
```diff
@@ -133,29 +125,27 @@
...
       border-color: var(
-        --amplify-components-button-primary-loading-border-color
+        --amplify-components-button-primary-disabled-border-color
       );
-      color: var(--amplify-components-button-primary-loading-color);
+      color: var(--amplify-components-button-primary-disabled-color);
     }
 
-    &[disabled] {
+    &[data-loading='true'] {
```
zchenwei, line 141: Need to switch the order here. Otherwise, loading styles will never take effect

packages/ui/src/theme/css/component/button.scss
```diff
@@ -227,6 +227,11 @@
   }
 }
 
+.amplify-button .amplify-button__loader-wrapper {
```
dbanksdesign, line 230: Should this selector just be `.amplify-button__loader-wrapper` to keep it at 1 classname?

packages/ui/src/theme/tokens/components/button.js
```diff
@@ -65,8 +65,8 @@ module.exports = {
     },
     _loading: {
       borderColor: { value: 'transparent' },
-      backgroundColor: { value: '{colors.background.tertiary.value}' },
-      color: { value: '{colors.font.inverse.value}' },
+      backgroundColor: { value: '{colors.background.disabled.value}' },
+      color: { value: '{colors.font.disabled.value}' },
```
dbanksdesign, line 69: Good update!

dbanksdesign (approved):
1 non-blocking comment

packages/ui/src/theme/css/component/button.scss
zchenwei: @dbanksdesign Yea, I do try that at first.

Screen Shot 2022-02-17 at 9 57 18 AM

But the order matters here. The `Flex` styling cannot be overridden if we just use 1 classname

PR #1329, opened 2022-02-15
Increase canary run rate to every 7min

reesscot:
#### Description of changes

* Increase the canary run rate to  run every 7minutes only during working hours
* Change emitted value to 0 for success and 1 for build failure

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1336, opened 2022-02-16
Update Canaries to run 24 hrs / day

reesscot:
Update Canaries to run 24 hrs / day

### DIFF
--- a/.github/workflows/run-builds.yml
+++ b/.github/workflows/run-builds.yml
@@ -2,7 +2,7 @@ name: Run Builds
 
 on:
   schedule:
-    - cron: '*/7 17-23,0-4 * * *' # Run every 7 minutes, from 9am to 8pm Pacific Time
+    - cron: '*/7 * * * *' # Run every 7 minutes
 
 jobs:
   build:


0618 (approved):

0618:
Just to provide more context, this PR is to revert https://github.com/aws-amplify/amplify-ui/pull/1329/files

wlee221 (approved):
LGTM to fill in missing data points in vicinity, but eventually we should look into why data points can be missing in the first place. Looking at the workflow, either of the two `put-metric-data` *must* happen because the two ifs:

https://github.com/aws-amplify/amplify-ui/blob/722e2a93263478aed2a9aee872ab1fcbc86b41ca/.github/workflows/run-builds.yml#L45-L60

and if that's the case, we are safe to ignore missing data.

reesscot:
> LGTM to fill in missing data points in vicinity, but eventually we should look into why data points can be missing in the first place. Looking at the workflow, either of the two `put-metric-data` _must_ happen because the two ifs:
> 
> https://github.com/aws-amplify/amplify-ui/blob/722e2a93263478aed2a9aee872ab1fcbc86b41ca/.github/workflows/run-builds.yml#L45-L60
> 
> and if that's the case, we are safe to ignore missing data.

The main thing we are protecting against is misconfiguration on our part that prevents the canaries from running. Could be misconfigured cron, or even as simple as deleting the workflow files.

Issue #1282, opened 2022-02-09
Allow Content-less TabItem


bestickley:
### On which framework/platform would you like to see this feature implemented?

React

### Which UI component is this feature-request for?

Other

### Please describe your feature-request in detail.

I'm creating a page with tabs that I want to be controlled by the url via React Router. Therefore, I don't want to put the content of my tabs inside `TabItem`, I want it to be below the `Tabs` component utilizing `Routes` and `Route` components. I'm currently forced to add a `<div />` inside the `TabItem` component or else I'll get the error: `react_devtools_backend.js:4061 Amplify UI: <Tabs> component only accepts <TabItem> as children. ` and the tabs won't render.

```ts
import { ReactElement, useEffect, useMemo, useState } from "react";
import { Link, Tabs, TabItem } from "@aws-amplify/ui-react";
import {
  Link as ReactRouterLink,
  Navigate,
  Route,
  Routes,
  useResolvedPath,
  useMatch,
} from "react-router-dom";
import { Box } from "../Box.jsx";
import { UpdateUser } from "./UpdateUser.jsx";
import { CreateUser } from "./CreateUser.jsx";
import { GroupsTable } from "./GroupsTable.jsx";
import { UsersTable } from "./UsersTable.jsx";
import { UsersInGroupTable } from "./UsersInGroupTable.jsx";

interface UserManagementProps {
  groupNames: string[];
}

const usersPath = "users";
const groupsPath = "groups";

export function UserManagement(props: UserManagementProps): ReactElement {
  const { groupNames } = props;
  const groupNameOptions = useMemo(
    () =>
      groupNames.map((g) => ({
        value: g,
        label: g.replace(/([A-Z])/g, " $1").trim(), // add space between capital letters
      })),
    [groupNames]
  );
  const [tab, setTab] = useState(0);
  const { pathname: usersPathname } = useResolvedPath(usersPath);
  const usersMatch = useMatch({ path: usersPathname, end: true });
  const { pathname: groupsPathname } = useResolvedPath(groupsPath);
  const groupsMatch = useMatch({ path: groupsPathname, end: true });
  useEffect(() => {
    if (usersMatch) {
      setTab(0);
    } else if (groupsMatch) {
      setTab(1);
    }
  }, [usersMatch, groupsMatch]);
  return (
    <Box css={{ m: "$2" }}>
      <Tabs currentIndex={tab} onChange={(t) => setTab(t as number)}>
        <TabItem
          title={
[...]

PR #1340, opened 2022-02-16
Allow Content-less TabItem

joebuono:
#### Description of changes

This PR allows the `<TabItem />` primitive to not have `children`, giving customers more flexibility. 

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/1282

Created a sample `<Tabs />` example with no `children` in the `<TabItems />`, then confirmed that there were no `console.warning`'s getting thrown. 

Also caught a tiny bug with newline characters causing the `console.warning`: `'Amplify UI: <Tabs> component only accepts <TabItem> as children.'` This is because React considers newline characters (`\n`) to be children. I added a conditional to ignore whitespace characters as children.  

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/.changeset/famous-tigers-visit.md
+++ b/.changeset/famous-tigers-visit.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+Allow Content-less TabItem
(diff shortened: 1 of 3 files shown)


reesscot (approved):
LGTM

PR #1342, opened 2022-02-16
Add `options` prop to <SelectField>

joebuono:
#### Description of changes

The added `options` prop accepts an array of strings which maps to the value, label and children of each `<option>` in a SelectField. 

If there are children and an `options` prop is set, then the children will win but there will be a warning in the console. 

This requirement will help Amplify Studio's Codegen team create a SelectField easier, plus it has a nice DX if the customer just wants a simple SelectField. 

Added a test and confirmed that a new `options` example in the docs works correctly. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/.changeset/silent-queens-guess.md
+++ b/.changeset/silent-queens-guess.md
@@ -1,0 +1,6 @@
+---
+"docs": patch
+"@aws-amplify/ui-react": patch
+---
+
+Add `options` prop to <SelectField>
--- a/.changeset/silent-queens-guess.md
+++ b/.changeset/silent-queens-guess.md
@@ -1,5 +1,5 @@
 ---
-"@aws-amplify/ui-react": patch
+"@aws-amplify/ui-react": minor
 ---
 
 Add `options` prop to <SelectField>
--- a/docs/src/pages/components/selectfield/examples/SelectFieldOptionsExample.tsx
+++ b/docs/src/pages/components/selectfield/examples/SelectFieldOptionsExample.tsx
@@ -1,0 +1,22 @@
+import { SelectField, Flex } from '@aws-amplify/ui-react';
+
+export const SelectFieldOptionsExample = () => (
+  <Flex direction="column">
+    <SelectField
+      label="Animals"
+      options={['lions', 'tigers', 'bears']}
+    ></SelectField>
+
+    <SelectField label="This is the same as the example above">
+      <option value="lions" label="lions">
+        lions
+      </option>
+      <option value="tigers" label="tigers">
+        tigers
+      </option>
+      <option value="bears" label="bears">
+        bears
+      </option>
+    </SelectField>
+  </Flex>
+);
--- a/docs/src/pages/components/selectfield/react.mdx
+++ b/docs/src/pages/components/selectfield/react.mdx
@@ -58,7 +58,7 @@ To manually control the SelectField state, you can use the `value` and `onChange
 
 ### Options
 
-To create a simple SelectField in which each `<option>` has matching `value`, `label` and `children`, pass a `string[]` to the `options` prop.
+Create a simple `SelectField` by passing a list to the `options` prop.
 
 <Example>
   <SelectFieldOptionsExample />
--- a/docs/src/pages/components/selectfield/react.mdx
+++ b/docs/src/pages/components/selectfield/react.mdx
@@ -13,6 +13,7 @@ import {
   SelectFieldDisabledStateExample,
   SelectFieldValidationErrorExample,
   SelectFieldControlledExample,
+  SelectFieldOptionsExample,
 } from './examples';
 
 ## Demo
@@ -49,6 +50,22 @@ To manually control the SelectField state, you can use the `value` and `onChange
   <ExampleCode>
 
 ```jsx file=./examples/SelectFieldControlledExample.tsx
+
+```
+
+  </ExampleCode>
+</Example>
+
+### Options
+
+To create a simple SelectField in which each `<option>` has matching `value`, `label` and `children`, pass a `string[]` to the `options` prop.
+
+<Example>
+  <SelectFieldOptionsExample />
+  
+  <ExampleCode>
+
+```jsx file=./examples/SelectFieldOptionsExample.tsx
 
 ```
 
--- a/packages/react/src/primitives/SelectField/SelectField.tsx
+++ b/packages/react/src/primitives/SelectField/SelectField.tsx
@@ -10,6 +10,27 @@ import { SelectFieldProps, Primitive } from '../types';
 import { splitPrimitiveProps } from '../shared/styleUtils';
 import { useStableId } from '../shared/utils';
 
+const selectFieldChildren = (
+  children?: React.ReactNode,
+  options?: string[]
+) => {
+  if (options?.length) {
+    if (children) {
+      console.warn(
+        'Amplify UI: <SelectField> component defaults to rendering children over `options`. When using the `options` prop, do not also pass children.'
+      );
+    } else {
+      return options.map((option, index) => (
+        <option label={option} value={option} key={index}>
+          {option}
+        </option>
+      ));
+    }
+  }
+
+  return children;
+};
+
 const SelectFieldPrimitive: Primitive<SelectFieldProps, 'select'> = (
   props,
   ref
@@ -23,6 +44,7 @@ const SelectFieldPrimitive: Primitive<SelectFieldProps, 'select'> = (
     id,
     label,
     labelHidden = false,
+    options,
     size,
     testId,
     ..._rest
@@ -62,7 +84,7 @@ const SelectFieldPrimitive: Primitive<SelectFieldProps, 'select'> = (
         size={size}
         {...rest}
       >
-        {children}
+        {selectFieldChildren(children, options)}
       </Select>
       <FieldErrorMessage hasError={hasError} errorMessage={errorMessage} />
     </Flex>
--- a/packages/react/src/primitives/SelectField/SelectField.tsx
+++ b/packages/react/src/primitives/SelectField/SelectField.tsx
@@ -10,25 +10,29 @@ import { SelectFieldProps, Primitive } from '../types';
 import { splitPrimitiveProps } from '../shared/styleUtils';
 import { useStableId } from '../shared/utils';
 
-const selectFieldChildren = (
-  children?: React.ReactNode,
-  options?: string[]
-) => {
-  if (options?.length) {
-    if (children) {
+interface SelectFieldChildrenProps {
+  children?: React.ReactNode;
+  options?: SelectFieldProps['options'];
+}
+
+const selectFieldChildren = ({
+  children,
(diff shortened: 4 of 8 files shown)


docs/src/pages/components/selectfield/examples/SelectFieldOptionsExample.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const SelectFieldOptionsExample = () => (
+  <Flex direction="column">
+    <SelectField
+      label="Animals"
+      options={['lions', 'tigers', 'bears']}
+    ></SelectField>
+
+    <SelectField label="This is the same as the example above">
+      <option value="lions" label="lions">
```
0618, line 11: How about if we map through the array `[lions, tigers, bears]`?

packages/react/src/primitives/SelectField/SelectField.tsx
```diff
@@ -10,6 +10,27 @@ import { SelectFieldProps, Primitive } from '../types';
...
+  options?: string[]
+) => {
+  if (options?.length) {
+    if (children) {
+      console.warn(
+        'Amplify UI: <SelectField> component defaults to rendering children over `options`. When using the `options` prop, do not also pass children.'
+      );
+    } else {
+      return options.map((option, index) => (
+        <option label={option} value={option} key={index}>
```
0618, line 24: Should we avoid using `index` as `key`?

docs/src/pages/components/selectfield/examples/SelectFieldOptionsExample.tsx
joebuono: We could do that. The reason I did it this way was to explicitly show the output of the `options` prop, which does the mapping.

packages/react/src/primitives/SelectField/SelectField.tsx
joebuono: Yes, if possible. What else could we use as the `key`?

packages/react/src/primitives/SelectField/SelectField.tsx
```diff
@@ -10,6 +10,27 @@ import { SelectFieldProps, Primitive } from '../types';
...
+const selectFieldChildren = (
+  children?: React.ReactNode,
+  options?: string[]
+) => {
+  if (options?.length) {
+    if (children) {
+      console.warn(
+        'Amplify UI: <SelectField> component defaults to rendering children over `options`. When using the `options` prop, do not also pass children.'
+      );
+    } else {
```
reesscot, line 22: What do you think of doing an early return for the `children`, since `children` takes precedence? 

```jsx
  if (children) {
    if (options?.length) {
          console.warn(
             'Amplify UI: <SelectField> component  defaults to rendering children over `options`. When using the `options` prop, omit children.'
          );
    }
    return children;
  }
  
  return options.map((option, index) => (
        <option label={option} value={option} key={index}>
          {option}
        </option>
  ));



```

IMO, else statements are usually a code smell that the logic should be simplified (often using an early return).

packages/react/src/primitives/SelectField/SelectField.tsx
reesscot: How about combining the option value and index: `${option}-${index}`

packages/react/src/primitives/SelectField/SelectField.tsx
```diff
@@ -10,6 +10,27 @@ import { SelectFieldProps, Primitive } from '../types';
 import { splitPrimitiveProps } from '../shared/styleUtils';
 import { useStableId } from '../shared/utils';
 
+const selectFieldChildren = (
+  children?: React.ReactNode,
+  options?: string[]
```
reesscot, line 15: These arguments should be extracted into separate interface, and the options type value should reference `SelectFieldProps['options']` so as to keep the parent component and this function signature in sync.

packages/react/src/primitives/SelectField/__tests__/SelectField.test.tsx
```diff
@@ -269,4 +269,22 @@ describe('SelectField test suite', () => {
       expect(errorText.innerHTML).toContain(errorMessage);
     });
   });
+
+  describe('Options', () => {
```
reesscot, line 273: Please add a test to check for the console warning.

packages/react/src/primitives/types/selectField.ts
```diff
@@ -5,4 +5,9 @@ import { FieldProps } from './field';
 export interface SelectFieldProps
   extends FieldProps,
     FlexContainerStyleProps,
-    SelectProps {}
+    SelectProps {
+  /**
+   * Accepts an array of strings which maps to the value, label and children of each option tag in the SelectField
```
reesscot, line 10: It's probably unnecessary to mention the types in the description, since they are already defined in the type. How about something like: 
```suggestion
   * List of option values for select dropdown
```

docs/src/pages/components/selectfield/react.mdx
```diff
@@ -55,6 +56,22 @@ To manually control the SelectField state, you can use the `value` and `onChange
   </ExampleCode>
 </Example>
 
+### Options
+
+To create a simple SelectField in which each `<option>` has matching `value`, `label` and `children`, pass a `string[]` to the `options` prop.
```
reesscot, line 61: Maybe leave the type information out of the docs, in case it changes and doesn't stay in sync. 
```suggestion
Create a simple `SelectField` by passing a list to the `options` prop.
```

docs/src/pages/components/selectfield/examples/SelectFieldOptionsExample.tsx
reesscot: I like it how it is, since it's here for demonstrating the equivalent example without using the `options` prop.

reesscot (changes requested):
Looks good, just a few comments below

docs/src/pages/components/selectfield/examples/SelectFieldOptionsExample.tsx
0618: Oh, I see. I didn't pay attention to line 4-8

packages/react/src/primitives/SelectField/SelectField.tsx
joebuono: Good point. The reason I wrote it this way is because I wanted to handle the case where there are neither `children` nor `options`.

.changeset/silent-queens-guess.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
+"@aws-amplify/ui-react": patch
+---
+
+Add `options` prop to <SelectField>
```
dbanksdesign, line 6: It would be nice to add an example in here too.

packages/react/src/primitives/SelectField/SelectField.tsx
```diff
@@ -10,6 +10,31 @@ import { SelectFieldProps, Primitive } from '../types';
...
+  children?: React.ReactNode;
+  options?: SelectFieldProps['options'];
+}
+
+const selectFieldChildren = ({
+  children,
+  options,
+}: SelectFieldChildrenProps) => {
+  if (children) {
+    if (options?.length) {
```
dbanksdesign, line 23: https://caniuse.com/mdn-javascript_operators_optional_chaining

Oh we can use javascript optional chaining??

clap

dbanksdesign (approved):
Great job!

great job

packages/react/src/primitives/SelectField/__tests__/SelectField.test.tsx
```diff
@@ -286,5 +286,24 @@ describe('SelectField test suite', () => {
         expect(option.innerHTML).toBe(optionString);
       });
     });
+
+    it('logs a warning to the console if the customer passes both children and options', async () => {
+      const warningMessage =
+        'Amplify UI: <SelectField> component  defaults to rendering children over `options`. When using the `options` prop, omit children.';
+      const spy = jest.spyOn(console, 'warn');
```
reesscot, line 293: These warnings will still show in the console when running the unit tests. Can we temporarily overwrite the console.warn function for your test, and then add the original back?
```
const originalWarn = console.warn;
console.warn = jest.fn();
// do your test stuff
console.warn = originalWarn;
```

packages/react/src/primitives/SelectField/SelectField.tsx
reesscot: Yeah, and I believe TypeScript should handle the conversion to ES5 if not.

.changeset/silent-queens-guess.md
```diff
@@ -0,0 +1,10 @@
+---
+"@aws-amplify/ui-react": patch
```
zchenwei, line 2: Should we mark it as a minor bump?

joebuono: Yes, good idea

packages/react/src/primitives/SelectField/SelectField.tsx
zchenwei, line 23:
```suggestion
  if (children && Array.isArray(options)) {
```
Do you think this would be better? I always try to avoid nested `if...else` blocks as possible as I can

joebuono: If there's a way to handle all four of these cases without nested `if...else` blocks, then yes:
1. There are `children`, but no `options`
2. There are `children` and `options`
3. There are no `children` and no `options`
4. There are no `children`, but there are `options`

packages/react/src/primitives/SelectField/SelectField.tsx
```diff
@@ -10,6 +10,31 @@ import { SelectFieldProps, Primitive } from '../types';
...
+      console.warn(
+        'Amplify UI: <SelectField> component  defaults to rendering children over `options`. When using the `options` prop, omit children.'
+      );
+    }
+    return children;
+  }
+
+  return options?.map((option, index) => (
+    <option label={option} value={option} key={`${option}-${index}`}>
```
zchenwei, line 32: Just think out loud, can we just use the `option` as key? Appending index will somehow affect React's optimization. Like if I have an option `<option label="apple" value="apple" key={'apple-2'}>` at index 2, but after updates, I got the same option `<option label="apple" value="apple" key={'apple-1'}>` move to index 1. However, React will think they are different because their keys are affected by the index and rerender it versus reuse the original one

packages/react/src/primitives/SelectField/SelectField.tsx
zchenwei: Right, my bad. I missed something there and I updated the comment there :)

packages/react/src/primitives/SelectField/SelectField.tsx
zchenwei: Good reading <- https://reactjs.org/docs/reconciliation.html

PR #1343, opened 2022-02-17
Adding deprecation warnings to icons

dbanksdesign:
#### Description of changes

Adding warnings and deprecated flags to built-in iconset to notify users of their removal. 

* Added `@deprecated` flag in JSDoc which will alert users in code
* Added a console warning when an icon is used (not in production to keep applications console clean)
* Added a deprecation warning in the icon docs page

CleanShot 2022-02-16 at 15 21 03

CleanShot 2022-02-16 at 15 22 23

CleanShot 2022-02-16 at 15 22 38@2x

CleanShot 2022-02-16 at 15 24 29@2x

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/primitives/Icon/deprecationWarning.ts
```diff
@@ -0,0 +1,10 @@
+export const deprecationWarning = (iconName) => {
```
reesscot, line 1: Should we make this more generic and add it to `utils`? IT seems like we will be doing this a fair amount as we go along. It would be nice to be able to search for all the `deprecationWarning` calls in the codebase as well.

docs/src/pages/components/icon/react.mdx
```diff
@@ -40,6 +40,13 @@ Import the Icon component and styles.
 
 ### Built-in iconset
 
+<Alert variation="warning" heading="Deprecated">
```
reesscot, line 43: Could be a reusable component to make searching for these across docs easy?

zchenwei (approved):
LGTM! :shipit:

dbanksdesign:
NOTE: do not merge this yet!

reesscot:
I made an update to not add deprecation warnings to icons we use internally. This will prevent us from emitting warnings from our components which use icons.

packages/react/scripts/generateIcons.js
```diff
@@ -6,26 +6,68 @@ const dirPath = `./src/primitives/Icon/icons/`;
...
-      height='1em'
-      className={classNames(ComponentClassNames.Icon, className)}
-      {...rest}
-    >
-      ${source.replace('black', 'currentColor')}
-    </View>
-  );
-}
+const INTERNAL_ICONS = [
+  'Icon',
```
reesscot, line 10: This list comes from the icons that will remain after the 3.0 release:
https://github.com/aws-amplify/amplify-ui/blob/3.0/packages/react/__tests__/exports.ts#L36-L55

packages/react/src/primitives/Icon/icons/IconCheck.tsx
```diff
@@ -5,6 +5,7 @@ import { View } from '../../View';
 
 export const IconCheck = (props) => {
   const { className, ...rest } = props;
+
```
reesscot, line 8: Note that for this interally used icon, we don't add the deprecation warning.

PR #1347, opened 2022-02-17
Store actor transition data in context

wlee221:
#### Description of changes

This PR stores actor return data (`intent`, `user`, `authAttributes`, etc) explicitly in `context` so that we dont rely on `event.data` (fortunately) persisted through series of events.

#### Background

Whenever actor finished and returned back, we've been passing the return data between states by having it persisted in `event.data` through series of events. 

But `event.data` is ephemeral by nature and is risky to persist on its own, so this PR stores that data in `context` immediately after it's available. Doing this will help `actions` to run more independently because this sets the source of truth to `context` instead of depending on series of `events`, which we have much less control over.

_Also, using `event.data` from previous events is really a dependence on side effect that we should avoid._

Did some manual validation over `autoSignIn` and `signIn -> confirmSignUp` flow on Angular example. Relying on e2e for base behaviors. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/ui/src/machines/authenticator/index.ts
```diff
@@ -64,50 +69,92 @@ export function createAuthenticatorMachine() {
...
-          exit: stopActor('signInActor'),
+          initial: 'spawnActor',
+          states: {
+            spawnActor: {
+              always: { actions: 'spawnSignInActor', target: 'runActor' },
+            },
+            runActor: {
+              entry: 'clearActorDoneData',
+              exit: stopActor('signInActor'),
+            },
```
wlee221, line 80: I split the spawn actor step into two parts:
1. `spawnActor` that consumes what's on the context and passes it to the new actor, and
2. `runActor` that removes any residue props that were already passed to the spawned actor, and runs the actor

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -64,50 +69,92 @@ export function createAuthenticatorMachine() {
...
         signIn: {
-          entry: 'spawnSignInActor',
-          exit: stopActor('signInActor'),
+          initial: 'spawnActor',
+          states: {
+            spawnActor: {
+              always: { actions: 'spawnSignInActor', target: 'runActor' },
+            },
+            runActor: {
+              entry: 'clearActorDoneData',
```
wlee221, line 78: `ActorDoneData`, defined below, are any data that are passed between actors when we transition from one to another (e.g. `signIn` -> `signUp`).

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -64,50 +69,92 @@ export function createAuthenticatorMachine() {
...
+              exit: stopActor('signInActor'),
+            },
+          },
           on: {
             SIGN_UP: 'signUp',
             RESET_PASSWORD: 'resetPassword',
             'done.invoke.signInActor': [
               {
                 target: 'signUp',
+                actions: 'setActorDoneData',
```
wlee221, line 88: After actor is done (line 85), immediately set any returned data to `ActorDoneData`

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -128,11 +175,17 @@ export function createAuthenticatorMachine() {
       actions: {
         forwardToActor: forwardTo((context) => context.actorRef),
         setUser: assign({
-          user: (_, event) => event.data.user || event.data,
+          user: (_, event) => event.data as CognitoUserAmplify,
         }),
```
wlee221, line 179: We're making this implementation more precise -- `setUser` will now only be called by `Auth.*` service calls, which has its information on `event.data`. Any `user` data passed between transitions will be handled more explicitly through `setActorDoneData`.

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -187,12 +240,12 @@ export function createAuthenticatorMachine() {
           },
         }),
         spawnSignInActor: assign({
-          actorRef: (context, event) => {
+          actorRef: (context, _) => {
             const { services } = context;
             const actor = signInActor({ services }).withContext({
-              authAttributes: event.data?.authAttributes,
-              user: event.data?.user,
-              intent: event.data?.intent,
```
wlee221, line 195: No more reliance on `event.data` that we have very little control over ✨

packages/ui/src/types/authMachine.ts
```diff
@@ -19,6 +19,7 @@ export interface AuthContext {
   password?: string;
   code?: string;
   mfaType?: AuthChallengeNames.SMS_MFA | AuthChallengeNames.SOFTWARE_TOKEN_MFA;
+  actorDoneData?: Omit<ActorDoneData, 'user'>; // data returned from actors when they finish and reach their final state
```
wlee221, line 22: `user` is set directly on `context.user` instead of in `context.actorDoneData.user`, so omitting that out.

ErikCH (approved):
👍 As long as all the tests passed, looks good

PR #1366, opened 2022-02-18
Pass `formData` on form `SUBMIT` events

wlee221:
#### Description of changes

This PR updates authenticator to submit `formData` on submit, instead of relying `onInput` events to infer submit values. This enables authenticator to catch any `defaultValues` that didn't send `onInput` event (e.g. `defaultPhoneNumber`). This will need updates in the following screens:

- [x] `signIn`
- [x] `confirmSignIn`
- [x] `forceNewPassword`
- [x] `setupTOTP`
- [x] `signUp`
- [x] `confirmSignUp`
- [x] `verifyUser`
- [x] `confirmVerifyUser`
- [x] `resetPassword`
- [x] `confirmResetPassword`

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-ui/issues/1324

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated: #1372

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1324, opened 2022-02-15
PhoneNumberField isn't using defaultCountryCode value on submission


OscarNowell:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator, Other

### How is your app built?

Create React App

### Please describe your bug.

PhoneNumberField doesn't use the provided defaultCountryCode for validation or on form submission.

![Screenshot 2022-02-15 at 12 21 33](https://user-images.githubusercontent.com/43755317/154061230-2fc66e98-bcf9-41f0-aca8-2b42562eb497.png)
 
The set defaultCountryCode will show, but isn't used unless re-selected, as can be seen in this console log of the sign up user attributes. 

![Screenshot 2022-02-15 at 12 19 00](https://user-images.githubusercontent.com/43755317/154060883-ee1cd27b-7a67-4128-88c0-6d0b52cdbfe7.png)
 
It defaults back to +1 instead of the expected +44

### What's the expected behaviour?

I expect the set defaultCountryCode to be used in validation and submission if set to custom value. 

### Help us reproduce the bug!

1. Set up Authenticator to use custom sign up form fields with PhoneNumberField set to have a non default country code.
2. Check form data attributes on submission to see if default country code was used correctly.

### Code Snippet

```javascript

// custom sign up form field component with PhoneNumberField 

export default function CustomSignUpFields () {
  const { updateBlur, validationErrors } = useAuthenticator()

  const handleBlur = (event) => {
    const { name } = event.target
    updateBlur({ name })
  }

  return (
    <>
      <TextField
      autoComplete="username"
      label='Email'
      labelHidden={true}
      name={'email'}
      required
      placeholder={'Email'}
      isRequired
      type={'email'}
    />
      <PasswordField
        autoComplete='new-password'
        data-amplify-password
        hasError={!!validationErrors['confirm-password']}
        isRequired
        name="password"
        label={'Password'}
        labelHidden={true}
        placeholder={'Password'}
        onBlur={handleBlur}
       />
       <PasswordField
        autoComplete='new-password'
[...]

PR #1372, opened 2022-02-19
React test for setting default region code

wlee221:
#### Description of changes

This adds a React test for setting default region code through slots. For now, I used Slots to mock default region code manually. Eventually, we should look to introduce explicit API for setting default region code without having to region code.

I only added test for React as Vue/Angular do not have primitives to mock `PhoneNumberField` through slots. Since we'll be adding an explicit API for setting default form values, let's update the examples after then. This will at least verify the fix for #1324.

#### Issue #, if available

#1324, accompanies #1366.

New e2e test!

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/examples/next/pages/ui/components/authenticator/default-region-code/index.page.tsx
+++ b/examples/next/pages/ui/components/authenticator/default-region-code/index.page.tsx
@@ -1,0 +1,86 @@
+import { Amplify } from 'aws-amplify';
+
+import {
+  PasswordField,
+  PhoneNumberField,
+  TextField,
+  Text,
+  useAuthenticator,
+  Authenticator,
+} from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+
+import awsExports from './aws-exports';
+Amplify.configure(awsExports);
+
+/**
+ * TODO: Introduce and use new API for setting default region code.
+ * For now, we're using Slots to mock default region code manually.
+ */
+const components = {
+  SignUp: {
+    FormFields() {
+      const { updateBlur, validationErrors } = useAuthenticator();
+      return (
+        <>
+          <PhoneNumberField
+            autoComplete="username"
+            countryCodeName="country_code"
+            defaultCountryCode="+44"
+            onCountryCodeChange={(e) => console.log(e.target.value)}
+            isRequired
+            label={'Phone Number'}
+            name="phone_number"
+            labelHidden={true}
+            placeholder={'Phone Number'}
+          />
+          <PasswordField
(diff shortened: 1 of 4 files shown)


examples/next/pages/ui/components/authenticator/default-region-code/index.page.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * TODO: Introduce and use new API for setting default region code.
+ * For now, we're using Slots to mock default region code manually.
+ */
+const components = {
+  SignUp: {
+    FormFields() {
+      const { updateBlur, validationErrors } = useAuthenticator();
+      return (
```
wlee221, line 24: Adapted from @OscarNowell's example on #1324

examples/next/pages/ui/components/authenticator/default-region-code/index.page.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <>
+          <PhoneNumberField
+            autoComplete="username"
+            countryCodeName="country_code"
+            defaultCountryCode="+44"
```
wlee221, line 29: This is where we set the `defaultCountryCode` to `+44`. Eventually, we should provide much easier way to set this without all the slot overrides!

packages/e2e/features/ui/components/authenticator/default-region-code.feature
```diff
@@ -0,0 +1,18 @@
...
+
+  @angular @react @vue  
+  Scenario: Sign up with a new email & password and lowercase the email 
+    Given I intercept '{ "headers": { "X-Amz-Target": "AWSCognitoIdentityProviderService.SignUp" } }' with fixture "sign-up-with-phone"
+    And I type my "phone number" with status "UNCONFIRMED"
+    And I type my password
+    And I confirm my password
+    And I type my "email" with status "UNCONFIRMED"
+    And I click the "Create Account" button
+    Then I verify the body starts with "+44"
```
wlee221, line 18: Since we set `defaultCountryCode` to `+44`, the submitted data should start with `+44` and not the usual `+1`.

ErikCH: This looks good. But this will only detect that it shows `+44` on the page, not that it actually send `+44` to Cognito. Not sure if we've ever made a test like that...

ErikCH: My bad! I see the new `verify` test! Carry on!

PR #1394, opened 2022-02-23
Adding label prop to divider

dbanksdesign:
#### Description of changes

Adding `label` prop to the Divider component. This allows for dividers that say something like "or" like the Authenticator uses for social sign in. This also updates the Authenticator on React, Vue, and Angular to use the new styles so we can get rid of some authenticator-specific styling. 

Tested it in the docs site, and also ran the Angular and Vue examples to verify it looked correct. Note: the Angular example is off a bit due to #1454

React demo:
CleanShot 2022-03-02 at 21 15 46

Angular:
CleanShot 2022-03-02 at 16 46 09@2x

Vue:
CleanShot 2022-03-02 at 16 14 40@2x

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


reesscot (approved):
LGTM

packages/react/src/primitives/Divider/Divider.tsx
```diff
@@ -6,14 +6,19 @@ import { DividerProps, Primitive } from '../types';
...
     aria-orientation={orientation}
     as="hr"
-    className={classNames(ComponentClassNames.Divider, className)}
+    className={classNames(
+      ComponentClassNames.Divider,
+      className,
+      label ? ComponentClassNames.DividerLabel : null
+    )}
     data-size={size}
+    data-label={label}
```
dbanksdesign, line 21: Note: we do need to keep this data attribute around because we use it in CSS with the `attr()` function.

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignIn.tsx
```diff
@@ -1,7 +1,7 @@
 import { FederatedIdentityProviders, translate } from '@aws-amplify/ui';
 
 import { useAuthenticator } from '..';
-import { Divider, Flex, View } from '../../..';
+import { Divider, Flex } from '../../..';
```
zchenwei, line 4: I think we should fix these implicit paths incrementally 🤣

zchenwei (approved):
LGTM 🚢

PR #1399, opened 2022-02-24
fix: add useCreateDataStorePredicate that uses hooks to prevent infinite useEffect loop

### LINKED ISSUE #1396: useDataStoreBinding results in infinite useEffect loop when criteria is set with createDataStorePredicate

React

### Which UI component?

Other

### How is your app built?

Create React App

### Please describe your bug.

With Studio component when a predicate is added to a `useDataStoreBinding` the hook gets stuck in an infinite loop.

Warning from console: `Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`

`createDataStorePredicate` should use `useCallback` with a dependency on inputted object so that a new function is not created each render.

This does result in another issue because in the example code snippet `buttonUserFilterObj` is redefined on each render. There are a few solutions to this.
1. Add custom comparison logic to the useCallback by stringifying the object.
1. define the object with useMemo (this would be difficult for codegen because of the recursive nature of the filter object).
1. define the object with useState.
    1. this could be done in the component or within the createDataStorePredicate function.

```typescript
const [buttonUserFilterObj] = React.useState({
  field: "firstName",
  operand: "Dane",
  operator: "eq",
})

// or 

export const createDataStorePredicateCallback = <Model extends PersistentModel>(
  predicateObject: DataStorePredicateObject
): ProducerModelPredicate<Model> => {
  const [predicateObjectState] = React.useState(predicateObject);
...
```

I think the option 3 is easiest to implement and least error prone.

I tried option with this quick solution and it resolved the issue.
```typescript
const createDataStorePredicate = (predicateObject) => {
  const [predicateObjectState] = React.useState(predicateObject);
  return React.useCallback((...args) => {
    return createDataStorePredicate(predicateObjectState)(...args)
  }, [predicateObjectState]);
}
```

Follow the integration setup on [Amplify Codegen UI tagged-release/q1-release](https://github.com/aws-amplify/amplify-codegen-ui/tree/tagged-release/q1-release).

### Code Snippet

```typescript
const buttonUserFilterObj = {
  field: "firstName",
  operand: "Dane",
  operator: "eq",
};

// creates a new predicate function each time the component renders
const buttonUserFilter = createDataStorePredicate(buttonUserFilterObj);

// this hook will cause infinite loop
const res = useDataStoreCollection({
  type: "collection",
  model: User,
  // the useEffect in useDataStoreCollection has a dependency on criteria
  // however, criteria is a new function on each render so it is always considered not equal and the hook runs
  criteria: buttonUserFilter,
});
```

This error only happens on the tagged-release/q1-release branch of amplify-codegen-react and version 2.6.1 of Amplify UI React. It is unclear how this is currently working on other versions. amplify-codegen-react and Amplify UI did not appear to make changes to code that would affect this feature.

dpilch:
#### Description of changes

Add `useCreateDataStorePredicate` that wraps `createDataStorePredicate` with `useCallback` so that the same function is used on each render.

#### Issue #, if available

Resolves #1396

I changed all occurrences of `createDataStorePredicate` with `useCreateDataStorePredicate`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/primitives/shared/datastore.ts
+++ b/packages/react/src/primitives/shared/datastore.ts
@@ -1,3 +1,4 @@
+import React from 'react';
 import {
   ModelPredicate,
   PersistentModel,
@@ -53,4 +54,22 @@ export const createDataStorePredicate = <Model extends PersistentModel>(
 
     return predicate;
   };
+};
+
+/**
+ * Creates a DataStore compatible predicate function from an object representation using useCallback
+ * @internal
+ */
+export const useCreateDataStorePredicate = <Model extends PersistentModel>(
+  predicateObject: DataStorePredicateObject
+): ProducerModelPredicate<Model> => {
+  const serializedPredicateObject = JSON.stringify(predicateObject);
+
+  return React.useCallback(
+    (predicate: ModelPredicate<Model>) =>
+      createDataStorePredicate<Model>(JSON.parse(serializedPredicateObject))(
+        predicate
+      ),
+    [serializedPredicateObject]
+  );
 };
--- a/packages/react/src/primitives/shared/__tests__/datastore.test.ts
+++ b/packages/react/src/primitives/shared/__tests__/datastore.test.ts
@@ -1,19 +1,22 @@
+import { renderHook } from '@testing-library/react-hooks';
 import { DataStorePredicateObject } from '../../types/datastore';
-import { createDataStorePredicate } from '../datastore';
+import {
+  createDataStorePredicate,
+  useCreateDataStorePredicate,
+} from '../datastore';
 
 type Post = {
   id: string;
   name: string;
   age: string;
 };
+const namePredicateObject = {
+  field: 'name',
+  operator: 'startsWith',
+  operand: 'John',
+};
 
 describe('createDataStorePredicate', () => {
-  const namePredicateObject = {
-    field: 'name',
-    operator: 'startsWith',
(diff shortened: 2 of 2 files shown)


packages/react/src/primitives/shared/datastore.ts
```diff
@@ -54,3 +55,19 @@ export const createDataStorePredicate = <Model extends PersistentModel>(
...
 };
+
+/**
+ * Creates a DataStore compatible predicate function from an object representation using useCallback
+ * @internal
+ */
+export const useCreateDataStorePredicate = <Model extends PersistentModel>(
+  predicateObject: DataStorePredicateObject
+): ProducerModelPredicate<Model> => {
+  const [predicateObjectState] = React.useState(predicateObject);
```
dpilch, line 66: Using `useState` will actually not work here. It will ignore any changes to `predicateObject`.

zchenwei:
Hi @dpilch,  I had a discussion with Scott and he has some context there since he reviewed Hector's PR back then. Having an empty dependencies array is by design. Only running `useEffect` once inside the `useDataStore` hook is an expected behavior because we do not expect the parameters would change dynamically. I have raised a PR to add some comments and disable the eslint rules there. I will close the pull request for now, thanks!

PR #1404, opened 2022-02-24
Add workflow for publishing to geo tag

slaymance:
#### Description of changes

This workflow will publish the Map components to the `@geo` npm tag. It triggers on merges to the `ui-geo/main` branch, which includes e2e tests which must pass prior publishing.

#### Issue #, if available

N/A

Publishing workflow is based upon the `.github/workflows/publish-next.yml` workflow.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes - N/A
- [x] Tests are updated - N/A

### DIFF
--- a/.github/workflows/publish-geo.yml
+++ b/.github/workflows/publish-geo.yml
@@ -47,4 +47,4 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: Publish to geo tag
-        run: yarn changeset publish --tag next
+        run: yarn changeset publish --tag geo
--- a/.github/workflows/publish-geo.yml
+++ b/.github/workflows/publish-geo.yml
@@ -1,0 +1,50 @@
+name: Publish to @geo
+
+on:
+  workflow_run:
+    workflows: ['Tests']
+    types:
+      - completed
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    environment: deployment
+    # We only publish if the previous workflow run (Tests) was successful
+    # and the previous workflow was called from main branch.
+    #
+    # Note that we check against workflow_run.head_branch instead of github.ref
+    # because workflow_run is always run on main, even if it's triggered from
+    # another branch.
+    #
+    # https://github.community/t/workflow-run-not-working-as-expected/139342
+    if: |
+      github.event.workflow_run.conclusion == 'success' &&
+      (
+        github.event.workflow_run.head_branch == 'main' ||
+        github.event.workflow_run.head_branch == 'ui-geo/main'
+      )
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-geo.yml
```diff
@@ -0,0 +1,50 @@
...
+          ref: ${{ github.event.workflow_run.head_branch }}
+      - name: Setup Node.js LTS
+        uses: actions/setup-node@v2
+        with:
+          node-version: lts/*
+          cache: 'yarn'
+      - name: Install and build packages
+        run: yarn --frozen-lockfile && yarn build
+      - name: Bump ui and ui-react packages
+        run: cp .github/changeset-presets/bump-react.md .changeset
```
slaymance, line 40: Are we able to just bump the ui and ui-react packages since the Geo components aren't available for other frameworks?

.github/workflows/publish-geo.yml
```diff
@@ -0,0 +1,50 @@
...
+    #
+    # Note that we check against workflow_run.head_branch instead of github.ref
+    # because workflow_run is always run on main, even if it's triggered from
+    # another branch.
+    #
+    # https://github.community/t/workflow-run-not-working-as-expected/139342
+    if: |
+      github.event.workflow_run.conclusion == 'success' &&
+      (
+        github.event.workflow_run.head_branch == 'main' ||
```
wlee221, line 24: I would remove this line, or any new push to `main` will publish to `geo` as well!

.github/workflows/publish-geo.yml
wlee221: Yep! What you have looks good to me 👍

.github/workflows/publish-geo.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: lts/*
+          cache: 'yarn'
+      - name: Install and build packages
+        run: yarn --frozen-lockfile && yarn build
+      - name: Bump ui and ui-react packages
+        run: cp .github/changeset-presets/bump-react.md .changeset
+      - name: Run changeset version to geo tag
+        run: yarn changeset version --snapshot geo-$(git rev-parse --short=7 HEAD) && yarn angular build
```
wlee221, line 42: What do you think about adding 

`version: geo` and `publish: geo` commands in `package.json` like so: https://github.com/aws-amplify/amplify-ui/blob/a7fcc1457e63456e592608ff3359b797472bcdad/package.json#L22-L23

Doesn't make a lot of difference other than styles, so I'm open to either options.

slaymance: The only reason I didn't opt for a package.json script is that I don't see this workflow necessarily being longterm. I figured that distinction would justify limiting the scope of files changed by addition of this workflow.

wlee221 (approved):

PR #1421, opened 2022-02-25
chore: adding comment to disable eslint rules in useDataStore hook

zchenwei:
#### Description of changes

Adding comments around `useEffect` inside `useDataStore` hook and disabled the eslint rule there

#### Issue #, if available

#1396 

Adding dependencies array makes the useEffect hook inside `useDataStore` running infinitely

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/react/src/hooks/useDataStore.tsx
```diff
@@ -44,6 +44,9 @@ export const useDataStoreCollection = <M extends PersistentModel>({
   };
 
   // Fetch on next render cycle
+  // useEffect should only run once here
+  // Ignore dependencies array here by design
```
reesscot, line 48:
```suggestion
  // Ignore exhaustive dependencies rule here by design
```

packages/react/src/hooks/useDataStore.tsx
```diff
@@ -71,6 +74,9 @@ export const useDataStoreItem = <M extends PersistentModel>({
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   React.useEffect(fetch, []);
```
calebpollman, line 80: Could be off base here, but rather than disable the `eslint` rule, which can lead to difficult to resolve issues later, can we include the hook deps in the array and gate the fetch call with a `useRef` variable? e.g.

```tsx
const hasFetched = useRef(false);

useEffect(() => {
  if (!hasFetched.current) {
    fetch();
    hasFetched.current = true;
  }
}, [fetch]); // not sure if there are other deps
```

reesscot: This won't solve the issue because there are other dependencies that would be required by the exhaustive hooks rule. We can't force the consumers to memoize the`model` or `criteria` props, which I believe is what was causing the infinite loop [here](https://github.com/aws-amplify/amplify-ui/pull/1406). It doesn't make sense for customers to dynamically change which Data Store model they are 'observing', so I think it actually makes sense to ignore the exhaustive hooks rule in this case.

zchenwei: This is a good call out but I think it won't solve the issue there. This `fetch` function would be a new one each time the parameters have changed which Scott mentioned above. It is a gate for calling `fetch` but cannot stop `useEffect` being called.

reesscot (approved):

jacoblogan (approved):
LGTM

PR #1428, opened 2022-02-26
add script to update canary package for release step and fix the clean script

0618:
#### Description of changes

**Issue:** During the release pre-flight process, to run canary, we have to manually change the package version to `next` for each `package.json`

**Fix:** this PR adds a script to do it automatically

**Code Change**:
- add `script/changeCanaryPackageVersion.js`
- fix the `clean` script
- add `clean` script for canary/

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,9 @@
 {
   "private": true,
+  "type": "module",
   "scripts": {
     "build": "turbo run build --include-dependencies --no-deps --scope=*@aws-amplify/*",
-    "clean": "turbo run clean && rm -rf node_modules",
+    "clean": "turbo run clean && rimraf node_modules",
     "setup": "yarn install && yarn build",
     "refresh": "yarn clean && yarn setup",
     "angular": "yarn workspace @aws-amplify/ui-angular",
@@ -24,7 +25,10 @@
     "version:latest": "yarn changeset version && yarn angular build",
     "publish:latest": "yarn changeset publish",
     "canary:build": "yarn --cwd canary/ build",
-    "canary:install": "yarn --cwd canary/ install "
+    "canary:install": "yarn --cwd canary/ install ",
+    "canary:clean": "yarn --cwd canary clean ",
+    "canary:updatePackage": "node script/changeCanaryPackageVersion.js",
+    "release:canary": "yarn install && yarn canary:updatePackage && rm postcss.config.js && yarn canary:install && yarn canary:build"
   },
   "turbo": {
     "baseBranch": "origin/main",
@@ -67,9 +71,11 @@
   "devDependencies": {
     "@changesets/changelog-github": "0.4.1",
     "@changesets/cli": "^2.16.0",
+    "globby": "^13.1.1",
     "husky": ">=6",
     "lint-staged": ">=10",
     "prettier": "2.4.1",
+    "rimraf": "^3.0.2",
     "tsup": "^5.11.9",
     "turbo": "^1.0.23",
     "wait-on": "^6.0.0"
--- a/script/changeCanaryPackageVersion.js
+++ b/script/changeCanaryPackageVersion.js
@@ -1,0 +1,26 @@
+/*
+ * This Script change all package.json fils under canary/ .
+ * It changes `@aws-amplify/ui-{framework}` version to `next`.
+ */
+
+import { globbyStream } from 'globby';
+import fs from 'fs';
+import { pathToFileURL } from 'url';
+
+for await (const path of globbyStream('canary/**/package.json')) {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,8 +1,9 @@
 {
   "private": true,
+  "type": "module",
```
0618, line 3: need to set this for the script to import package

ErikCH: 👍

package.json
```diff
@@ -24,7 +25,10 @@
     "version:latest": "yarn changeset version && yarn angular build",
     "publish:latest": "yarn changeset publish",
     "canary:build": "yarn --cwd canary/ build",
-    "canary:install": "yarn --cwd canary/ install "
+    "canary:install": "yarn --cwd canary/ install ",
+    "canary:clean": "yarn --cwd canary clean ",
+    "canary:updatePackage": "node script/changeCanaryPackageVersion.js",
```
ErikCH, line 30: Could we just change all the canary apps to `next` instead of running this script every time we do a release?

0618: Good point. I haven't looked into out canary yet and I just wrote the script basing on the preflight step in the runbook. 

So I'd like to know how @wlee221 would think

ErikCH: That makes sense! 👍

ErikCH: I was just thinking maybe we want to change the canaries to always run on next. But probably need to talk it over more with @wlee221 and others before we do that.

wlee221: IMO our canaries should still run against `latest`, because that's what the customers are installing in their apps. We should test what customers are install.

`next` makes more sense in release context, in which we want to test what we're pushing onto latest.

script/changeCanaryPackageVersion.js
```diff
@@ -0,0 +1,26 @@
...
+import fs from 'fs';
+import { pathToFileURL } from 'url';
+
+for await (const path of globbyStream('canary/**/package.json')) {
+  const pkgJSON = JSON.parse(fs.readFileSync(path));
+  const { dependencies } = pkgJSON;
+  if (dependencies) {
+    let pkgToUpdate;
+    Object.keys(dependencies)
+      .filter((dependency) => dependency.includes('@aws-amplify'))
```
wlee221, line 16: Very very nit coming from me, what do you think about 
```suggestion
      .filter((dependency) => dependency.includes('@aws-amplify/ui'))
```
so that it filters out any non-ui Amplify packages like `@aws-amplify/auth`? We do not ever use packages directly (they're installed indirectly through `aws-amplify` package), but just for the sake of explicitness.

wlee221 (approved):
Looks good, just one very tiny nit

package.json
wlee221: I believe this is what's causing the test failures, because now it puts all `.js` files in root are now under ES module context. I think we can sandbox this out by declaring `"type": "module"` in a separate `scripts/package.json` instead of in root.

And obviously we should remove that problematic `postcss.config.js`, but I think we can benefit long-term from separating out `script` related stuffs to a separate `package.json`.

package.json
ErikCH: @0618 Can you change the canaries to use `latest` instead of what they are on and include it in this PR? Since you are already updating some of the `package.json` files.

package.json
wlee221: Edited for typo, I cant type on Mondays 🤷‍♂️

0618: To make the minimum change to our previous code, instead of using `type:module`, I used `.mjs`.

Which way do you prefer? Use a separate `package.json` for scripts or remove `postcss.config.js` or the current way (`.mjs`)?

wlee221: IMO removing `postcss.config.js` should happen regardless of the decision here, since I don't see any purpose for it.

I like `.mjs`, we can wait for another script file to see if we need a separate `package.json`.

0618: I'm not sure if we are using `postcss.config.js` anywhere, and it's not related to this PR, so I'll run some tests and open a new PR to remove `postcss.config.js`

0618: I just found we use `postcss` in [`generateCSS.js`](https://github.com/aws-amplify/amplify-ui/blob/d53e5d528215510343e926a48524c2e458ffd20d/packages/ui/scripts/generateCSS.js#L6), so it might not be good to remove it...

wlee221: Hmm, maybe we move `postcss.config.js` to `packages/ui` then? Seems weird this is hoisted at root level. (And this weirdness is on us, pretty sure it's a leftover from our POC in the past).

PR #1429, opened 2022-02-28
Add css classnames

jacoblogan:
#### Description of changes

Update documentation to include target classes of all primitives

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


docs/src/pages/components/alert/react.mdx
```diff
@@ -126,6 +127,10 @@ Use the `onDismiss` prop to pass a function that will run when the Alert is dism
 
 ## CSS Styling
 
+### Target classes
+
+<ComponentClassTable componentName="Alert" />
```
dbanksdesign, line 132: Ah this is so much easier to document!

packages/react/src/primitives/shared/constants.ts
```diff
@@ -1,113 +1,504 @@
...
-  TextAreaField = 'amplify-textareafield',
-  ToggleButton = 'amplify-togglebutton',
-  ToggleButtonGroup = 'amplify-togglebuttongroup',
-  VisuallyHidden = 'amplify-visually-hidden',
-}
+export const ComponentClassObject = {
+  Alert: {
+    className: 'amplify-alert',
+    components: ['Alert'],
+    description: 'Top level element that wraps the Alert primitive',
```
dbanksdesign, line 5: I don't have a better solution, what are people's thoughts about using a JS object here? This would be included in customers' bundles since it is a JS object as opposed to a TS enum. I think this is probably fine for now, its not too much extra JS to ship and it makes it much easier to document. We could think about using JSDoc/TSDoc comments for this depending on the investigation into generating documentation from TS/doc comments. I don't think it is blocking and it is not a 1-way door because customers shouldn't be using this object anyways.

dbanksdesign (commented):
I love the simplification to the docs this adds. Nothing blocking, just one thing we should thing about with the constants.

packages/react/src/primitives/shared/constants.ts
joebuono: Good point. I agree that it's probably fine for now though.

docs/src/components/ComponentClassTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+  Table,
+  ComponentClassObject,
+  View,
+  TableRow,
+  TableCell,
+} from '@aws-amplify/ui-react';
+
+export const ComponentClassTable = ({ componentName }) => {
+  const targetClasses = React.useMemo(() => {
+    return Object.values(ComponentClassObject)
```
joebuono, line 12: I'm getting an error thrown from line 12: 

Screen Shot 2022-03-02 at 3 40 42 PM

docs/src/pages/components/alert/react.mdx
joebuono: Totally! Thanks for doing this, @jacoblogan

joebuono (changes requested):
This looks great, thank you for taking this on!

I'm requesting changes because when I run your branch locally, I keep getting an unhandled runtime error. See comment in `ComponentClassTable.tsx` file

docs/src/components/ComponentClassTable.tsx
dbanksdesign: Did you build the react package before running docs? You'll need to do that in order for this component to work because it is expecting that new object

reesscot (approved):
This is great, thanks for making it easy to document our CSS classes!

PR #1433, opened 2022-02-28
Fix console warnings in unit tests

joebuono:
#### Description of changes

This PR cleans up all the unit tests that were causing something to get logged to the console through `console.log`, `console.warn` or `console.error`. 

I ran `yarn test` and confirmed that nothing extraneous is being logged to the console. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Icon/Icon.tsx
+++ b/packages/react/src/primitives/Icon/Icon.tsx
@@ -35,7 +35,7 @@ const IconPrimitive: Primitive<IconProps, 'svg'> = (
     _children = children;
   }
   if (paths) {
-    _children = paths.map((path) => <path {...path} />);
+    _children = paths.map((path, index) => <path {...path} key={index} />);
   }
   if (pathData) {
     _children = <path d={pathData} fill={fill} />;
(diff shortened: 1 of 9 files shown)


packages/react/src/primitives/Icon/Icon.tsx
```diff
@@ -35,7 +35,7 @@ const IconPrimitive: Primitive<IconProps, 'svg'> = (
     _children = children;
   }
   if (paths) {
-    _children = paths.map((path) => <path {...path} />);
+    _children = paths.map((path, index) => <path {...path} key={index} />);
```
joebuono, line 38: @zchenwei Did we decide on a better approach than using the `index` as the `key`?

zchenwei: We have a util called `useStableId` that could be used to generate id for `key` prop. However, that util function is dependent on Radix `useId`, which we want to remove it in favor our own implementation. I am not sure how much we progress there.

zchenwei (approved):
LGTM 🚢

reesscot (approved):
LGTM

PR #1434, opened 2022-03-01
Convert `padding` and `margin` properties to logical equivalents

joebuono:
#### Description of changes

For example, we convert `paddingTop` to logical `padding-block-start`, and `marginRight` to `margin-inline-end`. 

Regarding both `padding` and `margin` block/inline properties (from MDN):
"The physical property it maps to depends on the values set for [writing-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode), [direction](https://developer.mozilla.org/en-US/docs/Web/CSS/direction), and [text-orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/text-orientation)."

Due to the above, I've also included `writingMode` in the `BaseStyleProps`. Here's an [example](https://stackoverflow.com/questions/59909792/what-is-the-difference-between-margin-block-start-and-margin-top) of how it affects the style prop mappings (note that the same applies for `padding`):

By default (`horizonal-tb`), you will have the following mapping:
```
margin-block-start = margin-top
margin-block-end = margin-bottom
margin-inline-start = margin-left
margin-inline-end = margin-right
```

If we change the `writingDirection` to `vertical-lr`, then the mappings change to:
```
margin-block-start = margin-left
margin-block-end = margin-right
margin-inline-start = margin-top 
margin-inline-end = margin-bottom
```

In the future, we may consider adding `direction` and/or `text-orientation`, both of which appear to be less commonly used. However, `direction` could present a conflict since we are currently mapping `direction` to `flexDirection`. 

After running `yarn react build`, I toyed around with some primitives in the docs and confirmed that adding each property was applied correctly, depending on the `writingMode`. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -240,6 +240,7 @@ const BaseStylePropsMap: Required<{ [key in keyof BaseStyleProps]: true }> = {
   transformOrigin: true,
   width: true,
   whiteSpace: true,
+  writingMode: true,
```
reesscot, line 243: Will customers really need to set this at a component level? Wouldn't they set this at the page level?

packages/react/src/primitives/types/style.ts
```diff
@@ -56,21 +56,21 @@ export interface BaseStyleProps extends FlexItemStyleProps, GridItemStyleProps {
   letterSpacing?: ResponsiveStyle<StyleToken<Property.LetterSpacing>>;
   lineHeight?: ResponsiveStyle<StyleToken<Property.LineHeight>>;
   margin?: ResponsiveStyle<StyleToken<Property.Margin>>;
-  marginBottom?: ResponsiveStyle<StyleToken<Property.MarginBottom>>;
-  marginLeft?: ResponsiveStyle<StyleToken<Property.MarginLeft>>;
-  marginRight?: ResponsiveStyle<StyleToken<Property.MarginRight>>;
-  marginTop?: ResponsiveStyle<StyleToken<Property.MarginTop>>;
+  marginBottom?: ResponsiveStyle<StyleToken<Property.MarginBlockEnd>>;
```
reesscot, line 59: Can you please also add all the logical margin and padding style props?
marginBlockEnd
marginInlineStart
marginInlineEnd
marginBlockStart

paddingBlockEnd
paddingInlineStart
paddingInlineEnd
paddingBlockStart

reesscot (commented):
This is great, just had a few comments.

reesscot (approved):
LGTM

PR #1435, opened 2022-03-01
Update Css selectors to have 010 specificity

jacoblogan:
#### Description of changes
This pull request updates the CSS selectors to use the BEM syntax and keep all selectors at 010 specificity. While this adds new classes and changes the default component styling to use these new selectors, none of the previous data attributes have been removed so any existing css written will continue to apply.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -34,10 +34,16 @@ const AlertPrimitive: Primitive<AlertProps, typeof Flex> = (
...
+  variation = variation || 'default';
+
   return (
     !dismissed && (
       <Flex
-        className={classNames(ComponentClassNames.Alert, className)}
+        className={classNames(
+          ComponentClassNames.Alert,
+          className,
+          `${ComponentClassNames.Alert}--variation-${variation}`
```
dbanksdesign, line 45: We might want to merge your css classes documentation PR before this one and figure out a way to document these classnames. We don't necessarily need to put these classnames in that component classnames object, maybe our solution isn't to add these to that object but dynamically generate them on the docs based on the possible variations and that we follow this specific pattern.

packages/react/src/primitives/Loader/Loader.tsx
```diff
@@ -58,10 +58,17 @@ const LoaderPrimitive: Primitive<LoaderProps, 'svg'> = (
...
+  size = size || 'default';
+  variation = variation || 'default';
+
   return (
     <View
       as="svg"
-      className={classNames(ComponentClassNames.Loader, className)}
+      className={classNames(
+        ComponentClassNames.Loader,
+        `${ComponentClassNames.Loader}--size-${size}--variation-${variation}`,
```
dbanksdesign, line 69: For multiple modifiers, BEM syntax is to usually split each modifier into it's own class: https://css-tricks.com/bem-101/

packages/react/src/primitives/Menu/MenuButton.tsx
```diff
@@ -28,11 +28,18 @@ export const MenuButton = React.forwardRef<
...
+    size = size || 'default';
+    variation = variation || 'default';
 
     return (
       <button
         ref={ref}
-        className={classNames(ComponentClassNames.Button, className)}
+        className={classNames(
+          ComponentClassNames.Button,
+          `${ComponentClassNames.Button}--size-${size}`,
```
dbanksdesign, line 39: How do people feel about removing 'size' and 'variation' from the classnames? So instead this would be

```js
`${ComponentClassNames.Button}--${size}`
```

packages/react/src/primitives/Pagination/PaginationItem.tsx
```diff
@@ -48,7 +48,10 @@ export const PaginationItem: React.FC<PaginationItemProps> = ({
             </Flex>
           ) : (
             <Button
-              className={ComponentClassNames.PaginationItemButton}
+              className={
+                (ComponentClassNames.PaginationItemButton,
+                `${ComponentClassNames.PaginationItemButton}--variation-link`)
```
dbanksdesign, line 53: Wouldn't the Button primitive add this from the variation prop?

packages/ui/src/theme/css/component/alert.scss
```diff
@@ -1,29 +1,29 @@
 .amplify-alert {
   align-items: var(--amplify-components-alert-align-items);
-  justify-content: var(--amplify-components-alert-justify-content);
   background-color: var(--amplify-components-alert-background-color);
+  justify-content: var(--amplify-components-alert-justify-content);
   padding-block: var(--amplify-components-alert-padding-block);
   padding-inline: var(--amplify-components-alert-padding-inline);
+}
+
+.amplify-alert--variation-info {
```
dbanksdesign, line 9: We can use Sass stuff here to make this easier. Nest this under `.amplify-alert` and this could be `&--variation-info`

dbanksdesign (commented):
This is a great start! A few questions open for discussion

0618:
I looked up on the internet but not able to find a definition of "010 specificity". What is that?

jacoblogan:
010 specificity means that we want to keep our selectors at the same specificity level as a single class selector.   https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity has a list for specificity under `Selector Types` and the goal would be that our selectors only contain a single selector of the Class selector level (in the list its the #2 selectors). This is so that other css frameworks and tools (css-in-js solutions) can take precedence over our base css rules.

packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -34,10 +34,16 @@ const AlertPrimitive: Primitive<AlertProps, typeof Flex> = (
     }
   }, [setDismissed, onDismiss, dismissed]);
 
+  variation = variation || 'default';
```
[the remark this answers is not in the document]
jacoblogan: yeah this was the approach I was going with at first but reverted back because I wanted the `data-variation/size` attributes to also be updated.  I was thinking that it would be better for a customer to continue to use the data attribute selectors to update the styling (that way they don't have to worry about import ordering because their rules will all be more specific than ours).

packages/react/src/primitives/Loader/Loader.tsx
jacoblogan: for modifiers where both are needed do you just create a new name for the modifier? for example the sliderfield has a variation and a size modifier that depends on each other.  Vertical sliders change the height with a size modifier and horizontal sliders change the width with a size modifier.

packages/react/src/primitives/Pagination/PaginationItem.tsx
jacoblogan: I like it, it removes a lot of clutter.  We would need modifiers that don't share values though specifically the default state would need a new name.

packages/react/src/primitives/Loader/Loader.tsx
dbanksdesign: Generally, modifiers should be mutually exclusive (size should not modify background color for instance). In the vertical/horizontal cases we can add a local CSS variable like `--slider-size: 1rem` which the size modifier modifies, and the orientation modifier uses in either `width` or `height`.

packages/react/src/primitives/Alert/Alert.tsx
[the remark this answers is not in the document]
dbanksdesign: Customers would still be able to use data attributes, but just 'default' ones would not have a data attribute (which is the case today). We initially didn't include 'default' data attributes to keep the HTML tidy and so customers don't need to add `[data-variation='default']` to target the default variation. Whether or not we have the data attributes is less important, but typically BEM doesn't include 'default' modifiers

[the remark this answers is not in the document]
jacoblogan: currently the overrides to defaults don't affect the variations because the variation selectors have a greater specificity than the default state, with this flattened specificity don't we need a way to target the default state so that we can update it without overriding the other variations?

packages/ui/src/theme/css/component/button.scss
```diff
@@ -225,11 +232,25 @@
...
-}
+  &--disabled {
+    background-color: var(--amplify-internal-button-disabled-background-color);
+    border-color: var(--amplify-internal-button-disabled-border-color);
+    color: var(--amplify-internal-button-disabled-color);
+    text-decoration: var(--amplify-internal-button-disabled-text-decoration);
+    cursor: not-allowed;
+  }
 
-.amplify-button__loader-wrapper {
```
wlee221, line 230: Why was underscore used instead of hyphen?

wlee221 (commented):
Mostly got some questions because I'm unfamiliar with sass and our previous approach 😛

packages/react/src/primitives/Alert/Alert.tsx
```diff
@@ -37,7 +37,11 @@ const AlertPrimitive: Primitive<AlertProps, typeof Flex> = (
   return (
     !dismissed && (
       <Flex
-        className={classNames(ComponentClassNames.Alert, className)}
+        className={classNames(
+          ComponentClassNames.Alert,
+          className,
+          variation ? `${ComponentClassNames.Alert}--${variation}` : null
```
reesscot, line 43: What do you think of having a util function for creating these BEM modifier classes?

```ts
const classNameModifier = ({base, modifier}) => {
  return modifier ? `${base}--${modifier}` : null
}
```

Could use like:

```tsx
className={classNames(
          ComponentClassNames.Alert,
          className,
          classNameModifier({base: ComponentClassNames.Alert, modifier: variation})

```

packages/react/src/primitives/Badge/Badge.tsx
```diff
@@ -8,18 +8,25 @@ import { View } from '../View';
...
-);
+) => {
+  return (
+    <View
+      as="span"
+      className={classNames(
+        ComponentClassNames.Badge,
+        className,
+        variation ? `${ComponentClassNames.Badge}--${variation}` : null,
+        size ? `${ComponentClassNames.Badge}--${size}` : null
```
reesscot, line 19: Could use classNameModifier here to make it a bit cleaner

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -29,6 +29,13 @@ const ButtonPrimitive: Primitive<ButtonProps, 'button'> = (
       className={classNames(
         ComponentClassNames.Button,
         ComponentClassNames.FieldGroupControl,
+        variation ? `${ComponentClassNames.Button}--${variation}` : null,
+        size ? `${ComponentClassNames.Button}--${size}` : null,
+        isDisabled || isLoading
```
reesscot, line 34:
```ts
const classNameDisabledModifier = ({isDisabled, base, modifier}) => {
  return isDisabled ? `${base}--disabled` : null
}
```

packages/ui/src/theme/css/component/table.scss
```diff
@@ -5,75 +5,149 @@
...
 
-  .amplify-table__td {
+  &__td {
     border-color: var(--amplify-components-table-data-border-color);
     border-style: var(--amplify-components-table-data-border-style);
-    border-width: var(--amplify-components-table-data-border-width);
+    border-width: var(--amplify-internal-table-td-border-width);
     color: var(--amplify-components-table-data-color);
     display: var(--amplify-components-table-data-display);
-    font-size: var(--amplify-components-table-data-font-size);
```
reesscot, line 56: Isn't this an breaking change to remove a previously exposed CSS variable and mark it as internal?

packages/ui/src/theme/css/component/sliderField.scss
```diff
@@ -15,36 +15,42 @@
   touch-action: none;
   box-sizing: content-box; // we want the padding to add to the height
   padding-block: var(--amplify-components-sliderfield-padding-block);
+  --amplify-internal-sliderfield__root-height: var(
```
reesscot, line 18: What does it mean that these are `--amplify-internal`?

reesscot (commented):
Thank you for doing this, it will be a great update for customers. Just had some DRY suggestions and questions below.

docs/src/styles/docs/main.scss
```diff
@@ -19,16 +19,17 @@
 
 .docs-content-body {
   padding: var(--amplify-space-xxl);
-  
-  & table {
+
+  & table:not(.amplify-table) {
```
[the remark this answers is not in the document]
jacoblogan: :not is needed here because with this update we are making all of our default styling more generic and here is a spot where our default styling for our table component was more specific than this rule but after the update it becomes less specific.

packages/ui/src/theme/css/component/button.scss
jacoblogan: BEM syntax follows the guide of .{Block}__{Element}--{Modifier} the underscores are saying the `loader-wrapper` is an element of the `amplify-button` block.  Its just the naming convention.

packages/ui/src/theme/css/component/_fieldControl.scss
```diff
@@ -1,4 +1,4 @@
-%amplify-field-control {
```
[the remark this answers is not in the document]
jacoblogan: `%` and `@mixin` fill very similar roles here in that they let us add this block of styling to other rules. `%` is a placeholder selector and just means that this block isn't output in the final css on its own and is only used to extend other sass rules. Because we aren't passing in any variables here we would normally use `%` instead of `@mixin` but the `&--{modifier}` rules weren't being compiled correctly with `%` so that is why this was switched to `@mixin`

packages/react/src/primitives/Divider/__tests__/Divider.test.tsx
```diff
@@ -12,7 +12,7 @@ describe('Divider component', () => {
     const divider = (await screen.findByRole('separator')) as HTMLHRElement;
     expect(divider.nodeName).toBe('HR');
     expect(divider.id).toBe(dividerTestId);
-    expect(divider.dataset['size']).toBeUndefined();
```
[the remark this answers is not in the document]
jacoblogan: no difference here, they are functionally equivalent, this went from `.toBeUndefined()` -> `.toBe('default')` -> `.toBe(undefined)` just because I forgot that these originally started as `.toBeUndefined()` when I was backing out the changes I had for adding in a default state.

packages/ui/src/theme/css/component/table.scss
jacoblogan: That is a really good callout, We should be ok here because this isn't just removed it is instead being set to the new `--amplify-internal-table-td-font-size` which is being set to `--amplify-components-table-data-font-size` on line 8 of this file. But this warrants more investigation to ensure that all of the previously available theming experiences are still available with this new css layout.

packages/ui/src/theme/css/component/sliderField.scss
jacoblogan: these new `--internal` css variables are being used to keep the css structure flat.  Some of our components change different values based on the different states they are in for example the `large` modifier on a divider will change a different value when the divider is horizontal vs vertical.  These new `--internal` variables are used to update these values for the different states they are in.

docs/src/styles/docs/main.scss
zchenwei, line 23: It seems we can get rid of `&` since they are nested. It should be a default behavior

packages/react/src/primitives/Placeholder/__tests__/Placeholder.test.tsx
```diff
@@ -13,7 +13,7 @@ describe('Placeholder: ', () => {
     expect(
       placeholder.classList.contains(ComponentClassNames.Placeholder)
     ).toBe(true);
-    expect(placeholder.dataset['size']).toBeUndefined();
+    expect(placeholder.dataset['size']).toBe(undefined);
```
zchenwei, line 16: I think we should move back to `toBeUndefined()` to be consistent.

packages/react/src/primitives/Divider/__tests__/Divider.test.tsx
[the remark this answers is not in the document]
zchenwei: Plz use `.toBeUndefined()` to be consistent

packages/react/src/primitives/types/card.ts
```diff
@@ -1,6 +1,6 @@
 import { ViewProps } from './view';
 
-export type CardVariations = 'outlined' | 'elevated';
+export type CardVariations = 'outlined' | 'elevated' | 'default';
```
zchenwei, line 3: Why we need a `default`? Shouldn't `undefined` means default like other primitives? I didn't see we use default modifier in CSS

packages/react/src/primitives/Pagination/PaginationItem.tsx
```diff
@@ -84,7 +88,12 @@ export const PaginationItem: React.FC<PaginationItemProps> = ({
       return (
         <View as="li">
           <Button
-            className={ComponentClassNames.PaginationItemButton}
+            className={classNames(
+              ComponentClassNames.PaginationItemButton,
+              isDisabled
```
zchenwei, line 93: Shouldn't we add `${ComponentClassNames.PaginationItemButton}--link` here as well according to the CSS changes you made

packages/ui/src/theme/css/component/checkbox.scss
```diff
@@ -2,8 +2,11 @@
   cursor: var(--amplify-components-checkbox-cursor);
   align-items: var(--amplify-components-checkbox-align-items);
   gap: inherit;
-  &[data-disabled='true'] {
-    cursor: var(--amplify-components-checkbox-disabled-cursor);
+  --amplify-internal-checkbox-disabled-cursor: var(
+    --amplify-components-checkbox-disabled-cursor
+  );
+  &--disabled {
+    cursor: var(--amplify-internal-checkbox-disabled-cursor);
```
zchenwei, line 9: I think we do not need `--amplify-internal-checkbox-disabled-cursor` right? I don't see we need to override this anywhere

packages/ui/src/theme/css/component/checkbox.scss
```diff
@@ -29,42 +38,46 @@
...
-      --amplify-components-checkbox-button-focus-outline-offset
-    );
-
-    border-color: var(--amplify-components-checkbox-button-focus-border-color);
-    box-shadow: var(--amplify-components-checkbox-button-focus-box-shadow);
-  }
-
-  &[data-error='true'] {
+  &--focused {
+    border-color: var(--amplify-components-checkbox-button-before-border-color);
```
zchenwei, line 42: This seems redundant

packages/ui/src/theme/css/component/input.scss
```diff
@@ -1,3 +1,5 @@
+@import 'fieldControl';
```
zchenwei, line 1: Can you make sure we need this import? `@import` makes all variables, mixins, and functions globally accessible. I believe we have done that in `field.scss`. And because of its global nature, `@import` is discouraged in favor of `@use` now. I think we had better looking for a solution using `@use`

packages/ui/src/theme/css/component/_fieldControl.scss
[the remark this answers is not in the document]
zchenwei: That's weird.`%` and `@extend` is supposed to be used with BEM - https://sass-lang.com/documentation/at-rules/extend

packages/ui/src/theme/css/component/textArea.scss
```diff
@@ -1,3 +1,5 @@
+@import 'fieldControl';
```
zchenwei, line 1: I think `@import` will make a file globally accessible. I think `@use` would be the right way to go

packages/ui/src/theme/css/component/select.scss
```diff
@@ -1,3 +1,5 @@
+@import 'fieldControl';
```
zchenwei, line 1: Same comment. I think we just need import this somewhere once

zchenwei (commented):
Great job! Looks like there is quite large amount of work. Anyway, back to the discussion last week. It would be nicer to break up such a large PR because it is kind of risky to this many changes merged in and it also facilitate the reviewing of this PR. I think I probably spent half a day on it  but still cannot 100% ensure I will not miss anything :(

I think we might need to add the new class names to constant. I was under the impression that we use it to generate a table on docs?

jacoblogan:
closing to reopen in smaller chunks with updates

PR #1438, opened 2022-03-01
Updated dependencies for Flutter

abdallahshaban557:
#### Description of changes

Updated dependencies for Flutter

Tested the changes locally using `yarn && yarn build`

#### Checklist

- [X ] PR description included
- [X] `yarn test` passes

### DIFF
--- a/docs/src/pages/getting-started/installation/dependencies.flutter.mdx
+++ b/docs/src/pages/getting-started/installation/dependencies.flutter.mdx
@@ -1,12 +1,6 @@
 import { Link, Alert, Text, Tabs, TabItem } from '@aws-amplify/ui-react';
 
-Add the `amplify_authenticator` package with [pub](https://pub.dev/):
-
-```shell
-flutter pub add amplify_authenticator
-```
-
-This will add a dependency like this to your package's `pubspec.yaml`
+You will need to add dependencies for the `amplify_authenticator` to the `pubspec.yaml` file as shown below.
 
 ```yaml
 dependencies:
--- a/docs/src/pages/getting-started/installation/dependencies.flutter.mdx
+++ b/docs/src/pages/getting-started/installation/dependencies.flutter.mdx
@@ -6,10 +6,12 @@ Add the `amplify_authenticator` package with [pub](https://pub.dev/):
 flutter pub add amplify_authenticator
 ```
 
-This will add a dependency like this to your package's pubspec.yaml
+This will add a dependency like this to your package's `pubspec.yaml`
 
 ```yaml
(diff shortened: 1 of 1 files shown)


ErikCH (approved):
👍

docs/src/pages/getting-started/installation/dependencies.flutter.mdx
```diff
@@ -6,10 +6,12 @@ Add the `amplify_authenticator` package with [pub](https://pub.dev/):
 flutter pub add amplify_authenticator
 ```
 
-This will add a dependency like this to your package's pubspec.yaml
+This will add a dependency like this to your package's `pubspec.yaml`
 
 ```yaml
 dependencies:
+  amplify_flutter: ^0.4.0
```
Jordan-Nelson, line 13: Looks good.

Optionally, we could highlight the line for `amplify_authenticator` since that is the only package that will be added by running the command above.

Changing `yaml` to `yaml{4}` will highlight that line.

Screen Shot 2022-03-01 at 4 35 39 PM

Jordan-Nelson (approved):

docs/src/pages/getting-started/installation/dependencies.flutter.mdx
abdallahshaban557: I feel like that would make it seem that this is the only thing you need to make authenticator work, which is not entirely true. What do you think?

Jordan-Nelson: Yeah, I could see that.

On line 6 it says to run `flutter pub add amplify_authenticator`, but the snippet here shows three dependencies added. The command and the snippet sort of contradict each other. I was thinking the highlight makes it clearer that running `flutter pub add amplify_authenticator` doesn't add all three dependencies.

Maybe we update the lines above to mention the other two packages 

> Add `amplify_flutter`, `amplify_auth_cognito`, and `amplify_authenticator` package with [pub](https://pub.dev/):
> 
> flutter pub add amplify_flutter
flutter pub add amplify_auth_cognito
flutter pub add amplify_authenticator

abdallahshaban557: I am thinking maybe we remove that completely - I feel like the convention for a lot of beginner developers is to immediately add the dependencies in the pubspec.yaml file rather than use `flutter pub`. That is also what we recommend all across the library getting started guide, which I think is the more conventional path. What do you think? @Jordan-Nelson

Jordan-Nelson (approved):
lgtm

PR #1446, opened 2022-03-02
RFC Theme provider

dbanksdesign:
#### Description of changes

Adding a `ThemeProvider`, which the `AmplifyProvider` wraps to be backwards compatible. This will fix some weirdness on the docs site because some examples were using a `AmplifyProvider` in the example and because the docs already had an `AmplifyProvider` and it sets data attrs on the root element, the nested provider was overriding the data attributes.

To solve this, I created a ThemeProvider which has a `root` prop which defaults to false. If `root` is true, it will set the data attrs on the root element which will allow for the CSS selector which sets the theme CSS variables to work at the root level. This allows the Menu primitive which uses a React Portal to work. 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/src/pages/components/menu/react.mdx
```diff
@@ -21,64 +22,92 @@ Menu provides an accessible, interactive menu for selecting actions within an ap
...
 
-<Example>
-  <ControlledExample />
+  </ExampleCode>
 </Example>
 
-## CSS Styling
+## Styling
+
+<Example>
```
reesscot, line 101: Assuming we'll add description text talking about why customers might want to use a ThemeProvider here?

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -1,12 +1,10 @@
 import * as React from 'react';
```
reesscot, line 1: unrelated: does it bother anyone else that this file is an `index.tsx` file?

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -17,31 +15,18 @@ interface AmplifyProviderProps {
...
+        theme,
       }}
     >
       <IdProvider>
-        <div>{children}</div>
-        <style
-          id={`amplify-theme-${name}`}
-          dangerouslySetInnerHTML={{ __html: cssText }}
-        />
+        <ThemeProvider theme={theme} colorMode={colorMode} root={true}>
```
reesscot, line 27: 💯

packages/react/src/components/ThemeProvider/ThemeContext.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { defaultTheme, WebTheme } from '@aws-amplify/ui';
+
+export type ColorMode = 'system' | 'light' | 'dark';
+
+export interface ThemeContextType {
+  theme: WebTheme;
+  colorMode?: ColorMode;
+  setColorMode?: React.Dispatch<React.SetStateAction<ColorMode>>;
+  root?: boolean;
```
reesscot, line 11: Should we use our boolean standard here and prefix with `is`?
```suggestion
  isRootTheme?: boolean;
```

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export { ThemeContext, ColorMode };
+
+export function ThemeProvider({
+  children,
+  colorMode,
+  theme,
+  root,
+}: ThemeProviderProps) {
+  // createTheme will always return a fully formed theme object,
```
reesscot, line 22: Should a modified version of this comment be a JSDoc comment (shown in intellisense) on the ThemeProviderProps.theme property so that customers know that their theme will extend from default theme?

```
interface ThemeProviderProps {
  children: React.ReactNode;
  colorMode?: ColorMode;
  /** 
   * Custom theme extended from the default theme.
  */
  theme?: Theme;
  root?: boolean;
}
```

On that note, shouldn't the other properties have jsdoc comments so customers know what they are for?

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  // extending from the default theme. Doing this allows
+  // users to still use the `useTheme` hook even if they do not
+  // provide a theme
+  const webTheme = createTheme(theme);
+  const { name, cssText } = webTheme;
+
+  React.useEffect(() => {
+    if (root) {
+      if (document && document.documentElement) {
```
reesscot, line 31: Could use optional chaining here?
```suggestion
      if (document?.documentElement) {
```

packages/react/src/primitives/Menu/Menu.tsx
```diff
@@ -25,6 +25,7 @@ const MenuPrimitive: Primitive<MenuProps, 'div'> = (
     trigger,
     triggerClassName,
     onOpenChange,
+    portalled = true,
```
reesscot, line 28:
```suggestion
    isPortalled = true,
```

packages/react/src/primitives/types/menu.ts
```diff
@@ -38,6 +38,12 @@ export interface MenuProps extends FlexProps {
    * Note: only applies if `trigger` prop is null
    */
   triggerClassName?: string;
+
+  /**
+   * Whether or not the menu should use a React Portal
+   * @default false
+   */
+  portalled?: boolean;
```
reesscot, line 46: Should we follow our boolean variable naming convention here?
```suggestion
  isPortalled?: boolean;
```

reesscot: Curious what the use case is for customers not having the menu in a React portal? Mainly for styling?

packages/ui/src/theme/tokens/components/menu.js
```diff
@@ -10,14 +10,6 @@ module.exports = {
   maxWidth: { value: '30rem' },
   minWidth: { value: '14rem' },
 
-  small: {
```
reesscot, line 13: 💯 Thanks for cleaning this up

packages/react/src/primitives/Menu/Menu.tsx
reesscot: Side note, this word sure sounds made up! 😂

reesscot (commented):
Looks great, I think this is a great approach to splitting out the ThemeProvider. Just a few suggestions below

packages/react/src/components/ThemeProvider/index.tsx
reesscot: @dbanksdesign  I withdraw my comment given: https://blog.jim-nielsen.com/2022/a-web-for-all/

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -17,31 +15,18 @@ interface AmplifyProviderProps {
...
-        colorMode
-      );
-    }
-  }, [theme, colorMode]);
   return (
     <AmplifyContext.Provider
       value={{
-        theme: webTheme,
+        theme,
       }}
```
calebpollman, line 24: Do we still need the `AmplifyContext.Provider`?

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { createTheme, Theme } from '@aws-amplify/ui';
+
+import { ThemeContext, ColorMode } from './ThemeContext';
+
+interface ThemeProviderProps {
+  children: React.ReactNode;
+  colorMode?: ColorMode;
+  theme?: Theme;
+  root?: boolean;
```
calebpollman, line 11: I'm slightly concerned with this prop leading to user confusion, but i may not fully grasp the need for it?

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export { ThemeContext, ColorMode };
```
calebpollman, line 14: Any reason not to just export these directly from _ThemeContext.ts_?

calebpollman (commented):
Looking great so far! Left some general feedback and questions

reesscot:
Closing, as we will address again in future PR.

PR #1447, opened 2022-03-02
Fix nested providers

dbanksdesign:
#### Description of changes

Add a cleanup method to the in the useEffect hook of the AmplifyProvider to clean up the data attributes it sets on the root level. Added back in the data attributes on the div the AmplifyProvider wraps its children in. The result of these 2 changes is that nested providers now work mostly as expected. 

Also conditionally adding the inline styles only if a user provides a theme. This will allow customers to more easily use plain CSS for theming where previously they had to mess with selector specificity. 

#### Issue #, if available

The issue was in the docs site we are using nested providers to show some theming examples. Because the AmplifyProvider writes data attrs directly to the root element, a nested provider would override a parent provider's attributes. Also when the nested provider is removed from the tree it doesn't clean itself up. 

Also conditionally rendering the inline stylesheet only if a theme is given. The reason for this is that the default stylesheet users import already has the default CSS variables anyways. This gets rid of the issue where customers wrap their app in a AmplifyProvider and try to theme using CSS variables because the inline styles the provider injects would take precedence.

Looked around on the docs. 

Production (note the data attributes and the weirdness in borders):
https://user-images.githubusercontent.com/321279/156433291-a3124510-33f3-4c22-bfb4-229efec0967c.mp4

Updates:
https://user-images.githubusercontent.com/321279/156433361-f100df59-139d-49e4-b60e-1f21f01f8c09.mp4

https://user-images.githubusercontent.com/321279/156433534-8d580db7-9569-4157-b660-0d9f5f672c9c.mp4

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -23,11 +23,27 @@ export function AmplifyProvider({
...
+          'data-amplify-theme',
+          originalName
+        );
+        document.documentElement.setAttribute(
+          'data-amplify-color-mode',
+          originalColorMode
+        );
+      };
     }
   }, [theme, colorMode]);
```
zchenwei, line 48: Looks like we should put `name` instead of `theme` in the array

zchenwei (commented):
If we add back in the data attributes on the div, what's the benefits of adding these to the root element?

packages/ui/src/theme/baseTheme.ts
```diff
@@ -5,5 +5,5 @@ import { BaseTheme } from './types';
 export const baseTheme: BaseTheme = {
   tokens,
   breakpoints,
-  name: 'base-theme',
+  name: 'default-theme',
```
reesscot, line 8: what does this change do?

reesscot (approved):
LGTM

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -17,31 +16,72 @@ interface AmplifyProviderProps {
...
+          originalName
+        );
+        document.documentElement.setAttribute(
+          'data-amplify-color-mode',
+          originalColorMode
+        );
+      };
     }
-  }, [theme, colorMode]);
+  }, [name, colorMode]);
```
zchenwei, line 55: I am worried that the nested theme will still impact the global one. If we have an example using a nested provider there, wouldn't the global theme be affected as well, though we clean that up right after leaving that page

dbanksdesign: The nested theme will override the root data attributes, but because we are also adding them on a div too, the div node will take precedence. The effect would be if you had nested (or just multiple) providers and you referenced the CSS variables outside the DOM tree of the provider (like in the Menu Portal case). I feel like this is a pretty edge case and it is already the current behavior of the provider

zchenwei: Just think out loud. I was wondering if adding a `isNested` prop on provider makes sense here and use it to gate the DOM attribute calls so that they don't need to be invoked on nested use cases.

zchenwei (approved):

packages/react/src/components/AmplifyProvider/index.tsx
dbanksdesign: Lol yes it does.. and I actually added this in my ThemeProvider PR: https://github.com/aws-amplify/amplify-ui/pull/1446 (need to update the actual prop name). Talking with Scott my thinking is to get this PR merged in to fix our issues we are currently having and then spend some more time on the ThemeProvider PR. The changes this PR has are similar to the ThemeProvider PR too, but with just no API changes.

packages/react/src/index.tsx
```diff
@@ -11,3 +11,4 @@ export * as primitives from './primitives';
  * Re-export public APIs from `@aws-amplify/ui`
  */
 export { defaultTheme, createTheme, translations } from '@aws-amplify/ui';
+export type { Theme } from '@aws-amplify/ui';
```
dbanksdesign, line 14: We were previously exporting this type from within the AmplifyProvider with some weird renaming stuff, this is much cleaner.

PR #1449, opened 2022-03-02
Add href to nav and header links

0618:
#### Description of changes

**Issue**: Our links are missing `href`.
CleanShot 2022-02-02 at 13 48 57@2x

**Code Change**:
- add `href`
- refactor to use `React.forwardRef` as suggested in [Nextjs's doc](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-functional-component)
- add missing `passHref` prop as suggested in [Nextjs's doc](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag)
- fix `NevLink` type error
Screen Shot 2022-03-02 at 13 30 47
- refactor `import` in `Header.tsx` and `LinkButton.tsx` for tree-shaking

#### Issue #, if available

https://app.asana.com/0/1201736086077838/1201759796183693/f

- ran `yarn test`
- local test

image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/components/Layout/Header.tsx
+++ b/docs/src/components/Layout/Header.tsx
@@ -21,6 +21,7 @@ import { useRouter } from 'next/router';
 import { Logo } from '@/components/Logo';
 import { FrameworkChooser } from './FrameworkChooser';
 import { SecondaryNav } from './SecondaryNav';
+import LinkButton from './LinkButton';
 
 const NavLink = ({ href, children, isExternal = false, onClick }) => {
   const { pathname, query } = useRouter();
@@ -35,12 +36,10 @@ const NavLink = ({ href, children, isExternal = false, onClick }) => {
     );
   }
   return (
-    <NextLink href={{ pathname: href, query }}>
-      <div>
-        <a className={className} onClick={onClick}>
-          {children}
-        </a>
-      </div>
+    <NextLink href={{ pathname: href, query }} passHref>
+      <LinkButton href={href} classNames={className} onClick={onClick}>
+        {children}
+      </LinkButton>
     </NextLink>
   );
 };
--- a/docs/src/components/Layout/Header.tsx
+++ b/docs/src/components/Layout/Header.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import { useState } from 'react';
 import NextLink from 'next/link';
 import {
   IconOpenInNew,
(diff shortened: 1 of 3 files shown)


docs/src/components/Layout/LinkButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+    children?: ReactElement;
+  },
+  ref
+) => (
+  <a onClick={onClick} className={classNames} href={href} ref={ref}>
+    {children}
+  </a>
+);
+
+export default forwardRef(LinkButton);
```
dbanksdesign, line 23: Nice to break this out to a separate file!

docs/src/components/Layout/Header.tsx
```diff
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import { useState } from 'react';
```
dbanksdesign, line 1: I want to say there is a reason we use `import * as React from 'react';` as opposed to `import {} from 'react';` but I don't remember why.

dbanksdesign (approved):
LGTM! One general question for the team about importing react, but otherwise lets get this in!

docs/src/components/Layout/Header.tsx
reesscot: We prefer the `import * as React from 'react';` for consistency across the codebase and minimizing import statement code changes. See: 
https://epicreact.dev/importing-react-through-the-ages/

docs/src/components/Layout/Header.tsx
```diff
@@ -21,8 +21,20 @@ import { useRouter } from 'next/router';
 import { Logo } from '@/components/Logo';
 import { FrameworkChooser } from './FrameworkChooser';
 import { SecondaryNav } from './SecondaryNav';
+import LinkButton from './LinkButton';
+import type { MouseEventHandler, ReactElement } from 'react';
```
reesscot, line 25: Nice, love the `import type`

docs/src/components/Layout/Header.tsx
```diff
@@ -21,8 +21,20 @@ import { useRouter } from 'next/router';
...
+const NavLink = ({
+  href,
+  children,
+  isExternal = false,
+  onClick,
+}: {
+  href: string;
+  children: ReactElement;
+  isExternal?: boolean;
+  onClick?: MouseEventHandler<HTMLAnchorElement>;
```
reesscot, line 36: Thanks for fixing the types

reesscot (approved):
LGTM. Would prefer to keep the react namespace import as it was, but not a blocker.

docs/src/components/Layout/Header.tsx
0618: Thanks for sharing! It's fun to know all the histories on how to import react.

Looks like both are fine, but I'll convert my change back for consistency :D

docs/src/components/Layout/Header.tsx
wlee221: I love this too! Nice JS/TS distinguishment.

wlee221 (approved):

PR #1461, opened 2022-03-03
set pagination current page default to 1

jacoblogan:
#### Description of changes

Set the default currentPage value of the Pagination component to 1

Updated the Pagination component to set the currentPage default value to 1.  This fixes an error that was showing NaN if no currentPage was given.

reproducible with `<Pagination totalPages={20} onNext={()=>{}} onPrevious={()=>{}} onChange={()=>{}} />`
CleanShot 2022-02-15 at 16 37 47@2x

Screen Shot 2022-03-03 at 3 57 32 AM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Pagination/PaginationItem.tsx
+++ b/packages/react/src/primitives/Pagination/PaginationItem.tsx
@@ -37,6 +37,7 @@ export const PaginationItem: React.FC<PaginationItemProps> = ({
             <Flex
               as="span"
               className={ComponentClassNames.PaginationItemCurrent}
+              testId="current"
               {...rest}
             >
               {/**
--- a/packages/react/src/primitives/Pagination/PaginationItem.tsx
+++ b/packages/react/src/primitives/Pagination/PaginationItem.tsx
@@ -8,6 +8,9 @@ import { VisuallyHidden } from '../VisuallyHidden';
 import { PaginationItemProps } from '../types/pagination';
 import { ComponentClassNames } from '../shared/constants';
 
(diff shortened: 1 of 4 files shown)


packages/react/src/primitives/Pagination/__tests__/Pagination.test.tsx
```diff
@@ -207,6 +207,21 @@ describe('Pagination component test suite', () => {
       expect(pagination).toHaveTextContent('6');
       expect(pagination).toHaveTextContent('7');
     });
+
+    it('should default the current page to 1 if one is not provided', async () => {
```
0618, line 211: Thanks for adding the test!

0618 (approved):
LGTM, but don't forget to add the change set

packages/react/src/primitives/Pagination/PaginationItem.tsx
```diff
@@ -37,6 +37,7 @@ export const PaginationItem: React.FC<PaginationItemProps> = ({
             <Flex
               as="span"
               className={ComponentClassNames.PaginationItemCurrent}
+              testId="current"
```
zchenwei, line 40: How about defining a const for the `testId` and export it like this <- https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/SliderField/SliderField.tsx#L16-L19

So that it can be used in test file as well

zchenwei (approved):

PR #1464, opened 2022-03-03
Add Concurrency for CI

slaymance:
#### Description of changes
Adds [concurrency](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) condition for our CI workflows. This change will define a group based on the pull request ID and whenever a new CI job is triggered for that group (e.g. a new commit on the PR) any in-progress workflows in that group will be canceled.

This should help alleviate some pressure on the number of concurrent GitHub Action runners we utilize.

#### Issue #, if available
N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated - N/A

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,7 @@
 name: Tests
 
 concurrency:
-  group: ${{ github.event.pull_request.id || github.ref }}
+  group: e2e-${{ github.event.pull_request.id || github.ref }}
   cancel-in-progress: true
 
 on:


.github/workflows/tests.yml
```diff
@@ -3,6 +3,10 @@
 
 name: Tests
 
+concurrency:
+  group: ${{ github.event.pull_request.id }}
```
wlee221, line 7: How does this work for `on: push`?

slaymance: Good callout, I followed the advice for [adding a fallback value](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value).

wlee221: Oh man, I didn't know `||` could be used as fallback operator just like in JS! This is nice.

.github/workflows/tests.yml
```diff
@@ -3,6 +3,10 @@
 
 name: Tests
 
+concurrency:
+  group: ${{ github.event.pull_request.id || github.ref }}
```
wlee221, line 7: Sorry for nits after another :D Let's specify this group name, so it doesn't coincide with other concurrency groups we might add later.

```suggestion
  group: e2e-${{ github.event.pull_request.id || github.ref }}
```

PR #1474, opened 2022-03-04
Adding labelPosition to radio and RadioGroupField

dbanksdesign:
#### Description of changes

Adding `labelPosition` to Radio and RadioGroupField components. If present on the RadioGroupFIeld it will apply to all Radios underneath. A specific Radio can override the group's labelPosition if needed. 

Added and ran tests. Verified behavior on docs. 

CleanShot 2022-03-03 at 21 51 45

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


docs/src/pages/components/radiogroupfield/demo.tsx
```diff
@@ -1,40 +1,65 @@
...
+const propsToCode = ({
+  label,
+  labelPosition,
+  size,
+  direction,
+  name,
+  isDisabled,
+}: RadioGroupFieldProps) => {
+  return (
+    `<RadioGroupField` +
```
0618, line 21: I'm curious if we only use literal template for the entire returned string?

docs/src/pages/components/radiogroupfield/examples/ErrorExample.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export const ErrorExample = () => (
+  <RadioGroupField
+    label="Language"
+    name="example-12"
+    errorMessage="This is a required field."
+    hasError
+    labelHidden
+  >
+    <Radio value="html">html</Radio>
```
0618, line 11: if we re-use `html``css``javascript` at different places, should we use an array of variables?

docs/src/pages/components/radiogroupfield/examples/LabelPositionExample.tsx
```diff
@@ -0,0 +1,35 @@
+import { Radio, RadioGroupField, Flex } from '@aws-amplify/ui-react';
+
+export const LabelPositionExample = () => (
+  <Flex direction="row">
+    <RadioGroupField label="Language" name="language" defaultValue="html">
+      <Radio value="html">html</Radio>
```
0618, line 6: I see we use strings directly in our codebase, I just wonder if we have considered i18n in future? Should we gradually start using variables instead?

docs/src/pages/components/radiogroupfield/examples/ReadOnlyExample.tsx
```diff
@@ -0,0 +1,17 @@
+import { Radio, RadioGroupField } from '@aws-amplify/ui-react';
+
+export const ReadOnlyExample = () => (
+  <>
```
0618, line 4: is `<>` necessary here since it's already wrapped in `<RadioGroupField>`?

docs/src/pages/components/radiogroupfield/examples/DirectionExample.tsx
```diff
@@ -0,0 +1,27 @@
...
+      label="Language"
+      name="example-8"
+      defaultValue="html"
+      direction="row"
+      labelHidden
+    >
+      <Radio value="html">html</Radio>
+      <Radio value="css">css</Radio>
+      <Radio value="javascript">javascript</Radio>
+    </RadioGroupField>
```
0618, line 25: I noticed that we repeatedly using this pattern in this PR, should we consider making a component?

packages/react/src/primitives/Radio/Radio.tsx
```diff
@@ -34,12 +44,26 @@ export const RadioPrimitive: Primitive<RadioProps, 'input'> = (
...
+
+  const labelPosition = radioLabelPosition
+    ? radioLabelPosition
+    : groupLabelPosition;
   return (
     <Flex
       as="label"
       className={classNames(ComponentClassNames.Radio, className)}
       data-disabled={shouldBeDisabled}
+      data-label-position={labelPosition}
```
0618, line 56: How would it work if the language is from right to left?

0618 (commented):
Did you get a chance to test a11y?

docs/src/pages/components/radiogroupfield/examples/DirectionExample.tsx
zchenwei: The code will be displayed on docs side as examples. We probably don't want to encapsulate that

docs/src/pages/components/radiogroupfield/examples/ErrorExample.tsx
zchenwei: Likewise, we want to make the example straightforward for beginners

docs/src/pages/components/radiogroupfield/examples/LabelPositionExample.tsx
zchenwei: This is a good callout. Something we should think about going forward

docs/src/pages/components/radiogroupfield/examples/DirectionExample.tsx
dbanksdesign: Yup anything in an examples folder on the docs site should be stand-alone and be easily readable for users as we display this code in a code block.

docs/src/pages/components/radiogroupfield/examples/LabelPositionExample.tsx
dbanksdesign: There is an asana task to investigate i18n for primitives if you want to take it 😉

docs/src/pages/components/radiogroupfield/demo.tsx
dbanksdesign: I'm open to suggestions to make these functions to generate code from the demo easier. The one difficulty we run into is if a certain prop is undefined, we don't want to show an empty line in the generated code or else we might get:

```jsx
<RadioGroupField
  label="label"


  direction="row"

>
```

packages/react/src/primitives/Radio/__tests__/Radio.test.tsx
```diff
@@ -53,4 +55,69 @@ describe('RadioField test suite', () => {
...
+      </RadioGroupField>
+    );
+    const radioField = await screen.findByTestId('test');
+    expect(radioField.querySelector('.amplify-radio')).toHaveAttribute(
+      'data-label-position',
+      'end'
+    );
+  });
+
+  it('should inherit labelPosition from RadioGroupField', async () => {
```
zchenwei, line 104:
```suggestion
  it('should not inherit labelPosition from RadioGroupField', async () => {
```

packages/ui/src/theme/css/component/radiogroup.scss
```diff
@@ -1,4 +1,5 @@
 .amplify-radiogroup {
   gap: inherit;
   flex-direction: inherit;
+  align-items: flex-start;
```
zchenwei, line 4: nit: since the parent field container has set `align-items: flex-start;`, would it make sense to set it `inherit` to be consistent with other props there?

zchenwei (approved):
No blockers that I see. :shipit:

packages/ui/src/theme/css/component/radiogroup.scss
dbanksdesign: Oh yea, I was playing around with align items at different places and I think you are right.

docs/src/pages/components/radiogroupfield/examples/LabelPositionExample.tsx
reesscot: I would also callout that i18n for primitives versus docs might be different.

reesscot (approved):
LGTM

docs/src/pages/components/radiogroupfield/examples/ErrorExample.tsx
0618: I think it would be cleaner and easier to track and use

docs/src/pages/components/radiogroupfield/examples/DirectionExample.tsx
0618: oh, I see. I didn't notice those are examples

docs/src/pages/components/radiogroupfield/demo.tsx
0618: gotcha, that make sense

packages/react/src/primitives/Radio/Radio.tsx
dbanksdesign: CleanShot 2022-03-07 at 12 59 07

0618: Nice!

PR #1373, opened 2022-02-19
input style props

dbanksdesign:
#### Description of changes

Changing how style props get applied to the elements in field components that have an input 
(TextField, TextAreaField, StepperField, SelectField, PhoneNumberField, PasswordField).
Previously, we tried to map certain style props to the input and others to the Flex component
wrapping the field. Now, all style props go on the Flex component and there is now an
`inputStyles` prop on these fields that accepts an object of style props. These style props
can use the responsive syntax as well. The PhoneNumberField also has a `selectStyles` prop
because it has a select and a text input. 

```jsx
import { TextField, useTheme } from '@aws-amplify/ui-react';
export const TextFieldInputStyleExample = () => {
  const { tokens } = useTheme();
  return (
    <TextField
      label="Name"
      inputStyles={{
        padding: tokens.space.large,
        backgroundColor: tokens.colors.brand.primary[80],
        color: tokens.colors.font.inverse,
      }}
    />
  );
};
```

#### Description of how you validated changes

Added tests and verified expected behavior on docs site locally. 

#### Checklist

- [ ] PR description included
- [x] `yarn test` passes
- [x] Tests are updated
- [x] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1486, opened 2022-03-07
TextField/TextAreafield - width and height style props

reesscot:
#### Description of changes
This PR moves the `width` and `height` style props to the parent container to be consistent with the `SelectField` styling. This fixes an issue where the TextField `height` prop is applied to the input field, whereas the `SelectField` `height` prop is applied to the parent Flex container.
Screen Shot 2022-03-07 at 3 49 09 PM

* Ran unit tests
* Manually tested setting height and width on SelectField and TextField on same line:
Screen Shot 2022-03-07 at 4 01 36 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/primitives/TextField/TextField.tsx
+++ b/packages/react/src/primitives/TextField/TextField.tsx
@@ -30,6 +30,7 @@ const TextFieldPrimitive = <Multiline extends boolean>(
     descriptiveText,
     errorMessage,
     hasError = false,
+    height, // @TODO: remove custom destructuring for 3.0 release
     id,
     label,
     labelHidden = false,
@@ -41,6 +42,7 @@ const TextFieldPrimitive = <Multiline extends boolean>(
     type, // remove from rest to prevent passing as DOM attribute to textarea
     size,
     testId,
+    width, // @TODO: remove custom destructuring for 3.0 release
     ..._rest
   } = props;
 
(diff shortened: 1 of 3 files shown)


joebuono (approved):
LGTM

packages/react/src/primitives/TextField/TextField.tsx
```diff
@@ -30,6 +30,7 @@ const TextFieldPrimitive = <Multiline extends boolean>(
     descriptiveText,
     errorMessage,
     hasError = false,
+    height, // @TODO: remove custom destructuring for 3.0 release
```
wlee221, line 33: For my understanding, wdym by custom destructuring? And how is it a breaking change?

dbanksdesign (approved):

packages/react/src/primitives/TextField/TextField.tsx
dbanksdesign: Removing width and height from this de-structure and instead using this logic for which style props go to which underlying element: https://github.com/aws-amplify/amplify-ui/pull/1373

wlee221: Ah gotcha, thanks!

wlee221 (approved):

packages/react/src/primitives/TextField/TextField.tsx
reesscot: We're specifically pulling the width and height and out of the base style props and applying them to the Flex container in this change. We'll also be doing this for all base style props in the 3.0, which is a much bigger change we will leave in the 3.0 release and won't do here. Thus, I added a comment to remove line when we release 3.0 since it will be done already.

Issue #1451, opened 2022-03-02
Input element loses focus after 1 character is typed (Authenticator re-rendering issue)


nickarocho:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

#### Basic issue

When wrapping my app with the `<Authenticator>` component, the inputs lose focus after typing one character - forcing the user to refocus the input to continue typing.

#### Example

https://user-images.githubusercontent.com/16296496/156466157-1523431e-7f19-447b-bc4c-6f8b1ef01fa3.mov

#### Ideas/Workarounds
After doing a little digging, it sounds like the Authenticator component is forcing the blur on every re-render.

Here's a couple (kinda) similar StackOverflow questions related:
- https://stackoverflow.com/questions/59715158/react-hooks-input-loses-focus-when-1-character-is-typed-in
- https://stackoverflow.com/questions/42573017/in-react-es6-why-does-the-input-field-lose-focus-after-typing-a-character

Works as expected when using the `withAuthenticator` approach (via ["Quick Start" guide](https://ui.docs.amplify.aws/components/authenticator#quick-start)).

### What's the expected behaviour?

The user should be able to type fluidly into the input without having to manually refocus after every keystroke.

### Help us reproduce the bug!

1. Follow the tutorial building a full stack Amplify/React "todo" app at https://docs.amplify.aws/start/getting-started/setup/q/integration/react/
2. After completing the ["Create login UI" section](https://docs.amplify.aws/start/getting-started/auth/q/integration/react/#create-login-ui), try typing into the inputs (this is where the bug should be)
3. Swap out the `Authenticator` with `withAuthenticator` (like [this](https://ui.docs.amplify.aws/components/authenticator#quick-start))
4. Type into the inputs again - should work.

### Code Snippet

- Demo repo with bug: https://github.com/nickarocho/auth-playground/commit/31ce1660d556e66bab1a2a9ee82c74f17fb405bc
[...]

PR #1489, opened 2022-03-08
fix: Authenticator issue where InitMachine useEffect runs every render

reesscot:
#### Description of changes
This PR fixes an issue where the `InitMachine` component is defined on each render. This caused the `useEffect` inside `InitMachine` to also run on each render, which finally caused the `children` of `InitMachine` to be unmounted and remounted. I also found that on initial render, the `useEffect` hook inside `InitMachine` was called many times. This PR fixes all of this by moving `InitMachine` outside of the `Authenticator` function. 

#### Issue #, if available

Fixes: https://github.com/aws-amplify/amplify-ui/issues/1451

Using the app from https://github.com/aws-amplify/amplify-ui/issues/1451 I validated the following:
* useEffect was only called a single time on initial render

* useEffect is not called when typing in the input field, and focus is not lost

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.changeset/nine-camels-cover.md
+++ b/.changeset/nine-camels-cover.md
@@ -1,0 +1,6 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+fix: Authenticator issue where InitMachine useEffect runs every render, causing `children` of 
+`Authenticator` to be unmounted and remounted on every render.
--- a/.changeset/nine-camels-cover.md
+++ b/.changeset/nine-camels-cover.md
@@ -2,5 +2,4 @@
 "@aws-amplify/ui-react": patch
 ---
 
-fix: Authenticator issue where InitMachine useEffect runs every render, causing `children` of 
-`Authenticator` to be unmounted and remounted on every render.
+fix: Authenticator issue where InitMachine useEffect runs every render, causing `children` of `Authenticator` to be unmounted and remounted on every render.
--- a/packages/react/src/components/Authenticator/Authenticator.tsx
+++ b/packages/react/src/components/Authenticator/Authenticator.tsx
@@ -21,18 +21,20 @@ export type AuthenticatorProps = AuthenticatorMachineOptions &
   ComponentsProviderProps;
 
 // Helper component that sends init event to the parent provider
-function InitMachine({ children, ...machineProps }) {
+function InitMachine({ children, ...data }) {
   const { _send, route } = useAuthenticator();
+
+  const hasInitialized = React.useRef(false);
+
   React.useEffect(() => {
-    if (route === 'idle') {
+    if (!hasInitialized.current && route === 'idle') {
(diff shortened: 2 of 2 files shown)


wlee221 (approved):
Looks good to me, thanks for looking into it!

.changeset/nine-camels-cover.md
```diff
@@ -0,0 +1,6 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+fix: Authenticator issue where InitMachine useEffect runs every render, causing `children` of 
+`Authenticator` to be unmounted and remounted on every render.
```
wlee221, line 6:
```suggestion
fix: Authenticator issue where InitMachine useEffect runs every render, causing `children` of `Authenticator` to be unmounted and remounted on every render.
```

NIT: You shouldn't need to manually add newlines, markdowns will handle it in changelogs.

jacoblogan (approved):
LGTM

packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -20,6 +20,22 @@ export type AuthenticatorProps = AuthenticatorMachineOptions &
...
+  const { _send, route } = useAuthenticator();
+  React.useEffect(() => {
+    if (route === 'idle') {
+      _send({
+        type: 'INIT',
+        data: machineProps,
+      });
+    }
+    // Disabling exhaustive hooks rule here because we only want to run a single time
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
calebpollman, line 34: Prefer passing `route` and `machineProps` to the deps array of the `useEffect`, the conditional on ln 27 will prevent the call to `_send` from executing over ignoring the **eslint** rule. 

Assuming there is potential of further calls to `_send` triggering when `route` updates to `'idle'` after the initialization call we can gate with a ref assigned variable (e.g. `hasInitialized`) to ensure that the call to `_send` only happens once. Thoughts?

packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -20,6 +20,22 @@ export type AuthenticatorProps = AuthenticatorMachineOptions &
   RouterProps &
   ComponentsProviderProps;
 
+// Helper component that sends init event to the parent provider
+function InitMachine({ children, ...machineProps }) {
```
calebpollman, line 24: Nit: replace `machineProps` with `data` to allow object shorthand notation on ln 30

calebpollman (commented):
Thanks for taking care of this! Left some feedback and a nit

calebpollman (commented):
LGTM 🚢

calebpollman (approved):

wlee221 (approved):

PR #1492, opened 2022-03-08
Supporting password strength requirements at the sign up experience

ErikCH:
#### Description of changes
When users sign up the should get an error to prompt them what the password qualifications are, instead of only seeing it after they click sign up.

#### Issue #, if available
[Asana](https://app.asana.com/0/1201888820034919/1201880837337067)

image

Created new feature tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-up/sign-up-form-fields/sign-up-form-fields.component.html
```diff
@@ -11,15 +11,25 @@
...
+          autocomplete="new-password"
+        ></amplify-form-field>
+        <div data-amplify-sign-up-errors *ngIf="errors">
+          <div
+            class="amplify-text"
+            data-variation="error"
+            *ngFor="let error of errors"
+          >
+            {{ translate(error) }}
+          </div>
```
ErikCH, line 30: I could have used the Angular error component, but it didn't look right here and didn't match react and Vue. So I used a div instead.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -98,6 +98,14 @@ When('I type an invalid password', () => {
...
 });
 
+When('I type an invalid wrong complexity password', () => {
+  cy.findInputField('Password').type('inv');
+});
+
+When('I type an invalid no lower case password', () => {
+  cy.findInputField('Password').type('INV');
+});
+
```
ErikCH, line 108: New helpers for the new feature test for complexity errors

packages/e2e/features/ui/components/authenticator/sign-up-with-errors.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Sign Up with Errors
```
ErikCH, line 1: New error tests to check the text

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
```diff
@@ -70,15 +76,29 @@ export function FormFields() {
...
+                <PasswordField
+                  autoComplete="new-password"
+                  data-amplify-password
+                  {...propsCreator('password', 'Password', formOverrides, true)}
+                  hasError={!!validationError['confirm_password']}
+                  name="password"
+                  onBlur={handleBlur}
+                />
+                {validationError['password_complexity'] && (
+                  <View data-amplify-sign-up-errors>
```
ErikCH, line 89: I had to add this `View` in here so the spacing would be correct for the errors.

ErikCH: But I didn't want it to show unless there was an error, otherwise it would add this extra blank line after the `password` field.

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +65,69 @@ export const defaultServices = {
...
+  ): Promise<ValidatorResult> {
+    const { password } = formData;
+
+    const { password: touched_password } = touchData;
+
+    /**
+     * If the password is not touched,
+     * or if the password settings are not set, we don't need to validate it.
+     */
+    if (!touched_password || !passwordSettings) return null;
```
ErikCH, line 81: If their zero config doesn't have this in there, it should just skip

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +65,69 @@ export const defaultServices = {
...
+        `Password must contain at least ${policyMinLength} characters`
+      );
+    }
+
+    const passwordPolicyCharacters = passwordSettings?.passwordPolicyCharacters;
+
+    passwordPolicyCharacters?.forEach((errorCheck) => {
+      switch (errorCheck) {
+        case 'REQUIRES_LOWERCASE':
+          if (!/[a-z]/.test(password))
```
ErikCH, line 97: Just using regex here for all these checks. Maybe there is a better way... not sure.

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -78,12 +146,11 @@ export const defaultServices = {
       (password || confirm_password) &&
       password !== confirm_password &&
       ((touched_confirm_password && touched_password) ||
-        (password.length >= 6 && confirm_password.length >= 6))
+        (password?.length >= 6 && confirm_password?.length >= 6))
```
ErikCH, line 149: This was a bug I found and fixed. This would silently fail if the password didn't exist.

packages/ui/src/machines/authenticator/defaultServices.ts
ErikCH: Also if they haven't touched the password yet, it should skip, *same as checking only on blur*

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -210,6 +211,9 @@ export function createAuthenticatorMachine() {
                 s.toLowerCase()
               ) ?? [];
 
+            let cliPasswordSettings =
```
ErikCH, line 214: The other examples converted the values to lower case. But I don't think that's necessary here.

packages/ui/src/theme/css/component/authenticator.scss
```diff
@@ -53,4 +53,10 @@
       --amplify-components-authenticator-state-inactive-background-color
     );
   }
+
+  [data-amplify-sign-up-errors] {
+    @media (max-width: 26rem) {
+      font-size: 0.688rem;
+    }
+  }
```
ErikCH, line 61: This should make it look better on mobile for very small screens.

packages/ui/src/validators/index.ts
```diff
@@ -1,15 +1,18 @@
...
 export const runValidators = async (
   formData: AuthFormData,
   touchData: AuthFormData,
+  passwordSettings: PasswordSettings,
   validators: Validator[]
 ) => {
   const errors = await Promise.all(
-    validators.map((validator) => validator(formData, touchData))
+    validators.map((validator) =>
+      validator(formData, touchData, passwordSettings)
```
ErikCH, line 14: Now I'm adding in the optional passwordSettings to each validator if it needs it.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-up/sign-up-form-fields/sign-up-form-fields.component.html
wlee221: Ah gotcha, I'll add a TODO item to handle it later then.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-up/sign-up-form-fields/sign-up-form-fields.component.html
```diff
@@ -11,15 +11,25 @@
...
+        <amplify-form-field
+          [labelHidden]="grabField('password', 'labelHidden', true)"
+          [placeholder]="grabField('password', 'placeholder', null)"
+          [required]="grabField('password', 'required', true)"
+          [label]="grabField('password', 'label', null)"
+          name="password"
+          autocomplete="new-password"
+        ></amplify-form-field>
+        <div data-amplify-sign-up-errors *ngIf="errors">
+          <div
```
wlee221, line 24:
```suggestion
        <div data-amplify-sign-up-errors *ngIf="errors">
          <!-- TODO: add amplify-alert reusable component -->
          <div
```

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +66,68 @@ export const defaultServices = {
...
+      password_complexity.push(
+        `Password must contain at least ${policyMinLength} characters`
+      );
+    }
+
+    const passwordPolicyCharacters = passwordSettings?.passwordPolicyCharacters;
+
+    passwordPolicyCharacters?.forEach((errorCheck) => {
+      switch (errorCheck) {
+        case 'REQUIRES_LOWERCASE':
```
wlee221, line 97: This is clean!

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +66,68 @@ export const defaultServices = {
...
+        default:
+          break;
+      }
+    });
+
+    /**
+     * Only return an error if there is at least one error.
+     */
+    return password_complexity.length !== 0
+      ? { password_complexity: password_complexity }
```
wlee221, line 128: Should this be `password` instead of `password_complexity`, since these are getting placed under the `password` field? IMO validation should map to each input name so that it matches one-to-one to each form input.

packages/ui/src/types/authenticator/validator.ts
```diff
@@ -6,18 +7,27 @@ import { AuthChallengeNames } from './user';
...
+}
+
 /**
  * Return type of validator. This is `null` if there are no error, and `ValidationError` otherwise.
  */
-export type ValidatorResult = void | null | ValidationError; // null if there are no error, `ValidationError` otherwise
+export type ValidatorResult =
+  | void
+  | null
+  | ValidationError
```
wlee221, line 20: Following up the previous comment, we can consider extending `ValidationError` object to be `Record<string, string | string[]>` instead of introducing a new type. Then `ValidationResult` can collect all validation errors like so:

```
{
  password: [
    `Password must contain a lower case letter`,
    `Password must contain an upper case letter`
  ],
  confirm_password: 'Your passwords must match'
]
```

wlee221 (commented):
Looking good, just got a nit on the interface we're using 🙏

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +66,68 @@ export const defaultServices = {
...
+    /**
+     * If the password is not touched,
+     * or if the password settings are not set, we don't need to validate it.
+     */
+    if (!touched_password || !passwordSettings) return null;
+
+    const password_complexity = [];
+
+    const policyMinLength = +passwordSettings?.passwordPolicyMinLength;
+    if (password.length < policyMinLength) {
```
calebpollman, line 87: What do you think of adding a `MINIMUM_LENGTH` rule? Would allow us to align this test with others in the below switch statement

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-up/sign-up-form-fields/sign-up-form-fields.component.html
ErikCH: 👍

packages/ui/src/machines/authenticator/defaultServices.ts
ErikCH: OK! Yeah, I didn't quite understand how the naming worked, with the helper functions you already created. I'll change this then.

packages/ui/src/types/authenticator/validator.ts
ErikCH: I had some problem with the types. I went back and forth. I'll try this.

calebpollman (commented):
Thanks for doing this! Left some questions/feedback

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +66,68 @@ export const defaultServices = {
...
+      password_complexity.push(
+        `Password must contain at least ${policyMinLength} characters`
```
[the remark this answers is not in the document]
ErikCH: It gets translated in the sign up components instead of here in the xstate machine.

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +66,68 @@ export const defaultServices = {
...
+
+    /**
+     * If the password is not touched,
+     * or if the password settings are not set, we don't need to validate it.
+     */
+    if (!touched_password || !passwordSettings) return null;
+
+    const password_complexity = [];
+
+    const policyMinLength = +passwordSettings?.passwordPolicyMinLength;
```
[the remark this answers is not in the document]
ErikCH: The `+` prefix converts a string to a number. The `passwordPolicyMinLength` is set in the creation of the Cognito User Pool. It's then pulled down into the `aws_exports` file.

packages/ui/src/types/authenticator/validator.ts
```diff
@@ -6,18 +7,27 @@ import { AuthChallengeNames } from './user';
  */
 export type ValidationError = Record<string, string>;
 
+export interface ValidationComplexityError {
+  password_complexity: Array<string>;
```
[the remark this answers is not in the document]
ErikCH: I deleted this interface, and I renamed it to just `password`. I believe in other places in the code we have `confirm_password` but we do that because it matches the `name` of the field.

packages/vue/src/components/confirm-reset-password.vue
```diff
@@ -140,7 +139,7 @@ function onBlur(e: Event) {
               name="password"
               autocomplete="current-password"
               :ariainvalid="
-                  !!(actorContext.validationError as ValidationError)['confirm_password']
+                  !!(actorContext.validationError as Record<string,string>)['confirm_password']
```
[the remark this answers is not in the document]
ErikCH: I went back and forth changing the ValidationError. I'll clean this up with my new changes.

packages/ui/src/types/authenticator/form.ts
```diff
@@ -58,3 +58,13 @@ export interface FormFieldOptions {
...
 }
+
+export type PasswordPolicyCharacters =
+  | 'REQUIRES_LOWERCASE'
+  | 'REQUIRES_NUMBERS'
+  | 'REQUIRES_SYMBOLS'
+  | 'REQUIRES_UPPERCASE';
+export interface PasswordSettings {
+  passwordPolicyMinLength: number;
+  passwordPolicyCharacters: Array<PasswordPolicyCharacters>;
```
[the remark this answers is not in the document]
ErikCH: These keys match the keys names in the `aws_exports` file for the `aws_cognito_password_protection_settings` object. Otherwise I would agree.

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +66,68 @@ export const defaultServices = {
...
+    }
+
+    const passwordPolicyCharacters = passwordSettings?.passwordPolicyCharacters;
+
+    passwordPolicyCharacters?.forEach((errorCheck) => {
+      switch (errorCheck) {
+        case 'REQUIRES_LOWERCASE':
+          if (!/[a-z]/.test(password))
+            password_complexity.push(
+              `Password must contain a lower case letter`
```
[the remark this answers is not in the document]
ErikCH: The sentence is long already, and on smaller screen sizes, I'm worried we'll have to set the font-size to be even smaller. So I think this is fine.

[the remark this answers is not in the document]
ErikCH: +1 on using quotes. Not sure why I used template literals.

packages/ui/src/machines/authenticator/defaultServices.ts
ErikCH: Will update!

packages/ui/src/machines/authenticator/defaultServices.ts
ErikCH: Updated in 11fba8aca5ddf4546428a9d54b9f30d0cffa0ba8

packages/ui/src/types/authenticator/validator.ts
ErikCH: Updated in 11fba8aca5ddf4546428a9d54b9f30d0cffa0ba8

packages/vue/src/components/confirm-reset-password.vue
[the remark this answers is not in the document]
ErikCH: Updated in 11fba8aca5ddf4546428a9d54b9f30d0cffa0ba8

packages/ui/src/machines/authenticator/defaultServices.ts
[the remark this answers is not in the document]
ErikCH: Updated quotes in 11fba8aca5ddf4546428a9d54b9f30d0cffa0ba8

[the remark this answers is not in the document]
calebpollman: Hmmm, concerned that we are implying that users should only include a **single** lowercase/special/uppercase/etc character in their password. Also, with i18n we cannot guarantee the rendering length, correct?

packages/ui/src/types/authenticator/form.ts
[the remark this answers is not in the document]
calebpollman: Just curious, do we need them to match the keys in the `aws_exports` file? What if those keys are updated in the future?

packages/ui/src/types/authenticator/validator.ts
```diff
@@ -1,23 +1,25 @@
+import { PasswordSettings } from '.';
 import { AuthFormData } from './form';
 import { AuthChallengeNames } from './user';
 
 /**
  * Maps each input to its validation error, if any
  */
-export type ValidationError = Record<string, string>;
+export type ValidationError = Record<string, string | string[]>;
```
wlee221, line 8: Sorry on another nit (on what I previously suggested!) -- what do you think about always typing this to `string[]` instead of `string | string[]`?  ie. enforce validators to always return an array of errors.

Then we don't have to do conditional `as string` or `as string[]` on wherever validation errors are, and we would just do a simple for loop like you did for password fields above.

calebpollman: Think that allows for greater extensibility in the future, for example if we have multiple validation rules placed on a username field

ErikCH: Wouldn't this be a breaking change though? We have a `validateCustomSignUp` service that customers can override to add in their own validation rules. To this point we have been assuming the customers sends over an object, with the key being the string. 

Like this

```services={{
        async validateCustomSignUp(formData) {
          if (!formData.acknowledgement) {
            return {
              acknowledgement: 'You must agree to the Terms & Conditions',
            };
          }
        },
      }}
```

If we change the result to be an array, then won't this break?  

It is kind of annoying have to use TypeScript assertion everywhere now. But not sure if this can be done.

packages/ui/src/machines/authenticator/defaultServices.ts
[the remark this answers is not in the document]
ErikCH: What if we changed it to `Password must have lowercase characters`. This would help suggest that it could contain one or more characters. This is also apart of the message Cognito sends back for this error

image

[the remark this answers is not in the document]
calebpollman: Think that works for sure!

packages/ui/src/types/authenticator/validator.ts
wlee221: Oh yeah, you’re right about the breaking change. Let’s leave as-is now, and have it covered in our 3.0 release.

wlee221:
```suggestion
// todo(breaking): enforce error to always be array
export type ValidationError = Record<string, string | string[]>;
```

packages/ui/src/machines/authenticator/defaultServices.ts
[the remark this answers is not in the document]
ErikCH: OK I updated it here 52cf527f4e43993655ff79f7646debd5a120c33f

wlee221 (approved):

packages/ui/src/types/authenticator/form.ts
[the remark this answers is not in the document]
ErikCH: Yeah, it makes sense to match the object coming from the `aws_exports` file, so we can have types when we interface with it. I suppose if it changed , we would need to update this.

calebpollman (approved):
Appreciate you addressing the feedback. Approving with the caveats that:
1) we will discuss naming conventions as a team
2) we will review the UI/UX approach for handling/rendering the validation errors as a team

Thanks for tackling this @ErikCH!

Milan-Shah (approved):
LGTM!

examples/next/pages/ui/components/authenticator/custom-sign-up-fields/index.page.tsx
```diff
@@ -28,7 +28,7 @@ export default function App() {
 
                 {/* Append & require Terms & Conditions field to sign up  */}
                 <CheckboxField
-                  errorMessage={validationErrors.acknowledgement}
+                  errorMessage={validationErrors.acknowledgement as string}
```
0618, line 31: nit: Do we need it? I see it's already recognized as `string` on my local

Screen Shot 2022-03-09 at 15 59 40

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
```diff
@@ -58,7 +58,7 @@ export class FormFieldComponent implements OnInit {
       this.authenticator.authState
     );
     const { validationError } = formContext;
-    return translate(validationError[this.name]);
+    return translate(validationError[this.name] as string);
```
0618, line 61: same here

image

packages/react/src/components/Authenticator/ForceNewPassword/ForceNewPassword.tsx
```diff
@@ -100,7 +100,7 @@ export const ForceNewPassword = (): JSX.Element => {
 
           {!!validationError['confirm_password'] && (
             <Text role="alert" variation="error">
-              {translate(validationError['confirm_password'])}
+              {translate(validationError['confirm_password'] as string)}
```
0618, line 103: same here

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
```diff
@@ -99,7 +119,7 @@ export function FormFields() {
 
                 {validationError['confirm_password'] && (
                   <Text role="alert" variation="error">
-                    {translate(validationError['confirm_password'])}
+                    {translate(validationError['confirm_password'] as string)}
```
reesscot, line 122: Is this required because the formContext isn't properly typed?

reesscot (approved):
Looks, good, just a few type questions.

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
```diff
@@ -70,15 +76,29 @@ export function FormFields() {
...
+              <React.Fragment key={name}>
+                <PasswordField
+                  autoComplete="new-password"
+                  data-amplify-password
+                  {...propsCreator('password', 'Password', formOverrides, true)}
+                  hasError={!!validationError['confirm_password']}
+                  name="password"
+                  onBlur={handleBlur}
+                />
+                {validationError['password'] && (
```
0618, line 88: nit: Do we have a preference in this codebase for bracket notations vs dot notations?

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
```diff
@@ -70,15 +76,29 @@ export function FormFields() {
...
+                  <View data-amplify-sign-up-errors>
+                    {(validationError['password'] as string[])?.map(
```
0618, line 90: also here, not sure why you need to do those `as string` in this PR, I don't see my VS Code complains about them

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
0618, line 122: same here

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -60,6 +65,62 @@ export const defaultServices = {
...
+      );
+    }
+
+    const passwordPolicyCharacters = passwordSettings?.passwordPolicyCharacters;
+
+    passwordPolicyCharacters?.forEach((errorCheck) => {
+      switch (errorCheck) {
+        case 'REQUIRES_LOWERCASE':
+          if (!/[a-z]/.test(password))
+            password_complexity.push('Password must have lower case letters');
```
0618, line 98: just an idea: what do you think if we use a `passwordErrors` (similar as what we do for the `validationError`) so that we can do i18n easier?

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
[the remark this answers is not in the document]
ErikCH: It's because we changed the type of `ValidationError` to include `string | string[]`. So TS gets confused, and you have to use type inference to fix it.

A long term solution will to change `ValidationError` to be `string[]`, but that will be a breaking change. As discussed in an earlier comment .

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
ErikCH: Same issue with `validationError`

examples/next/pages/ui/components/authenticator/custom-sign-up-fields/index.page.tsx
ErikCH: It won't build. So I had to add this.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
ErikCH: Hmm. it was giving me an error in my VS Code, and it won't build.

packages/react/src/components/Authenticator/ForceNewPassword/ForceNewPassword.tsx
ErikCH: See previous comment.

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
ErikCH: I'm ok with either. We don't have a style guide on this.

packages/react/src/components/Authenticator/SignUp/FormFields.tsx
ErikCH: See previous comment.

packages/ui/src/machines/authenticator/defaultServices.ts
ErikCH: Yeah, maybe that's something we can do in a future PR.

PR #1493, opened 2022-03-08
Test GitHub action to release docs

joebuono:
#### Description of changes

The goal of this PR is to test a GitHub action that will release the Amplify UI Docs site after successfully publishing to npm. 

I confirmed that manually merging `main` into `docs-release` and pushing my changes triggers a redeployment of the docs site (or rather, a test version of the docs site). The next step is to automate that process whenever we release our library. 

Instead of creating a separate workflow to test this action, I'm appending it to the end of the `publish-next` workflow for two reasons. 

1. We need access to the `published` output from the [changeset release action](https://github.com/changesets/action#outputs). 
2. This workflow runs more frequently than `publish-latest`, so we'll be able to test it and make sure it works. 

Once we confirm that the GitHub action works as expected, we can append it to the end of the `publish-latest` workflow. I'm still researching if we'll need to change the `fetch-depth` from 0. 

I am open to a better way of testing this GitHub action if you have any suggestions 👍

#### Checklist

- [X] PR description included

### DIFF
--- a/.github/workflows/publish-next.yml
+++ b/.github/workflows/publish-next.yml
@@ -49,3 +49,9 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: Publish to next tag
         run: yarn publish:next
+      - name: Merge main into docs-release branch
+        if: steps.changesets.outputs.published == 'true'
+        run: |
+          git checkout docs-release
+          git merge main
+          git push


wlee221 (approved):
This lgtm, this doesn't affect our workflows while enabling tests!

.github/workflows/publish-next.yml
```diff
@@ -49,3 +49,9 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: Publish to next tag
         run: yarn publish:next
+      - name: Merge main into docs-release branch
+        if: steps.changesets.outputs.published == 'true'
+        run: |
+          git checkout docs-release
```
reesscot, line 55: What kind of protections should we put around this branch since deploying to it can take down our site? We probably shouldn't be able to push to it (wish exception of github action)

reesscot (approved):

PR #1483, opened 2022-03-05
fix(homepage): a11y errors

0618:
#### Description of changes

| before | after |
| --- | --- |
| Screen Shot 2022-03-02 at 16 18 32 |  ~~`image`~~ image |
| 6 X Empty form label | ✅  |
| 1 X Language missing or invalid | ✅  |
| 6 X Empty button | ✅  |
| 5 X Broken ARIA reference |  ✅  |
| image | image |
| image |  image |

**Other code change**

- fix type error
- rename variables
- change format. double quote to single quote.

**Note**

There are still some a11y alerts on the homepage and I didn't fix them because:

- **1 X Missing first level heading**:  I'm not sure what to use as `h1`.
- **Redundant link**: both links on the bottom are "https://docs.amplify.aws/console/"
- **2 X Noscript element**: those are empty tags

**TODO**:
- [x] There are several primitives/components related a11y errors are converted in https://github.com/aws-amplify/amplify-ui/pull/1483/commits/dde5608889d4d22e095a97d70a2101ce9c50a145 and will be done in another PR https://github.com/aws-amplify/amplify-ui/pull/1495

#### Issue #, if available

https://app.asana.com/0/1201736086077838/1201759796183705/f

#### Description of how you validated changes

- ran `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1495, opened 2022-03-08
fix a11y errors for Signin and FieldDescription

0618:
#### Description of changes

This is a follow-up PR for https://github.com/aws-amplify/amplify-ui/pull/1483 . https://github.com/aws-amplify/amplify-ui/pull/1483 only changed docs related code and this PR changes the primitive/component related code.

**Code Change**:
- Fix `aria-reference` ADA error and update unit tests
- Fix FieldSet legend error for react, vue and angular

- `yarn test`
- WAVE checked ADA errors on homepage

image

|   | Before  | After  |
|---|---|---|
| vue  | image  | image  |
| angular   | image  | image  |

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/primitives/Field/FieldDescription.tsx
```diff
@@ -8,12 +9,15 @@ export const FieldDescription: React.FC<FieldDescriptionProps> = ({
...
-  return !labelHidden && descriptiveText ? (
-    <Text className={ComponentClassNames.FieldDescription} {...rest}>
-      {descriptiveText}
-    </Text>
-  ) : null;
-};
+}) => (
+  <Text
+    className={classNames(ComponentClassNames.FieldDescription, {
+      [ComponentClassNames.VisuallyHidden]: labelHidden,
```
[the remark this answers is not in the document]
reesscot: Need to also update the conditional in each field component that uses FieldDescription.
```
const ariaDescribedBy = descriptiveText ? descriptionId : undefined
```

This will ensure we only add the aria-describedby property when customers actually provide descriptive text. Otherwise we are linking to an empty div.

packages/react/src/primitives/Field/FieldDescription.tsx
```diff
@@ -8,12 +9,15 @@ export const FieldDescription: React.FC<FieldDescriptionProps> = ({
   descriptiveText,
   labelHidden,
   ...rest
-}) => {
-  return !labelHidden && descriptiveText ? (
```
reesscot, line 12: We should probably keep the descriptiveText check, because it doesn't make sense to render this div if there is no text to go inside of it.

reesscot (changes requested):

packages/react/src/primitives/Field/FieldDescription.tsx
0618: The a11y errors we have on the homepage have  `descriptiveText` `undefined`  and `labelHidden` `false` 😂

packages/react/src/primitives/Field/FieldDescription.tsx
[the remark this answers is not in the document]
0618: if `ariaDescribedBy` is `undefined`, we would get ADA error?

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
```diff
@@ -61,6 +61,9 @@ export function SignIn() {
             className="amplify-flex"
             disabled={isPending}
           >
+            <VisuallyHidden>
```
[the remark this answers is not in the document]
0618: Done

packages/react/src/primitives/Field/FieldDescription.tsx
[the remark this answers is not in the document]
reesscot: If ariaDescribedBy is undefined, then it shouldn't add the prop to the dom, which should avoid the accessibility error. If there's no descriptiveText provided by the user, then it doesn't make sense to have an ariaDescribedBy at all (thus why I think we should set it to undefined). Does that make sense? Happy to discuss offline if that's easier.

packages/react/src/primitives/Field/FieldDescription.tsx
reesscot: We only need the descriptiveText to render if there's actualy a string available. That's why I suggested the change below

packages/react/src/primitives/Label/Label.tsx
```diff
@@ -13,7 +13,7 @@ const LabelPrimitive: Primitive<LabelProps, 'label'> = (
     <View
       as="label"
       className={classNames(ComponentClassNames.Label, className, {
-        'amplify-visually-hidden': visuallyHidden,
+        [ComponentClassNames.VisuallyHidden]: visuallyHidden,
```
0618, line 16: Double checked this works
image

packages/react/src/primitives/Field/FieldDescription.tsx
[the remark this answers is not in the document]
0618: That makes sense! I updated the PR and also made the same changes for all the `aria-describedby` in the code base 😁

packages/react/src/primitives/Field/FieldDescription.tsx
0618: Good catch! Updated

packages/react/src/primitives/Field/FieldDescription.tsx
```diff
@@ -1,19 +1,27 @@
...
+
 export const FieldDescription: React.FC<FieldDescriptionProps> = ({
   descriptiveText,
   labelHidden,
   ...rest
-}) => {
-  return !labelHidden && descriptiveText ? (
-    <Text className={ComponentClassNames.FieldDescription} {...rest}>
+}) =>
+  !labelHidden && descriptiveText ? (
```
reesscot, line 15:
```suggestion
  descriptiveText ? (
```

packages/react/src/primitives/Field/__tests__/FieldDescription.test.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { FieldDescription, QA_FIELD_DESCRIPTION } from '../FieldDescription';
+import { ComponentClassNames } from '../../shared/constants';
+
+describe('FieldDescription component', () => {
+  it('should not render if it has a "labelHidden" prop is `true`', async () => {
```
reesscot, line 8: Sorry if I wasn't clear. I think it should render if labelHidden is true, but not if descriptiveText is null. Here's the following scenarios we should have tests for:

1. labelHidden = true, descriptiveText = "Provide a valid USPS address." => FieldDescription with className "amplify-visually-hidden"
1. labelHidden = false, descriptiveText = "Provide a valid USPS address." => FieldDescription in DOM with text shown  "Provide a valid USPS address."
1. labelHidden = true, descriptiveText = null => FieldDescription is not present in DOM
1. labelHidden = false, descriptiveText = null => FieldDescription is not present in DOM

packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx
```diff
@@ -37,6 +37,7 @@ const RadioGroupFieldPrimitive: Primitive<RadioGroupFieldProps, typeof Flex> = (
   const fieldId = useStableId(id);
   const labelId = useStableId();
   const descriptionId = useStableId();
+  const ariaDescribedBy = descriptiveText ? descriptionId : undefined;
```
reesscot, line 40: 💯 Yes, this is great!

packages/vue/src/components/sign-in.vue
```diff
@@ -118,12 +118,13 @@ const onForgotPasswordClicked = (): void => {
...
             :disabled="actorState.matches('signIn.submit')"
             class="amplify-flex"
+            p
             style="flex-direction: column"
           >
             <template #fieldSetI="{ slotData }">
               <slot name="signin-fields" :info="slotData"> </slot>
             </template>
-
+            <legend class="amplify-visually-hidden">Sign in</legend>
```
reesscot, line 127: Please post a screenshot of the Sign In Vue, Angular, and React Authenticator Sign In screens. We should also make this accessibility change to all the other screens with forms as well, as they will have the same problem.

packages/vue/src/components/sign-in.vue
```diff
@@ -118,12 +118,13 @@ const onForgotPasswordClicked = (): void => {
           <base-field-set
             :disabled="actorState.matches('signIn.submit')"
             class="amplify-flex"
+            p
```
reesscot, line 121: What does this do?

reesscot (changes requested):

packages/react/src/primitives/Field/__tests__/FieldDescription.test.tsx
0618: Oh, interesting! We want to keep the description even when the label's hidden. Ok...I'll update it. 

Thanks for the detailed clarification!

packages/vue/src/components/sign-in.vue
0618: Oops...good catch!

packages/vue/src/components/sign-in.vue
0618: Screenshots updated in the description. We can cut a ticket for other forms since the [current ticket](https://app.asana.com/0/1201736086077838/1201759796183705/f) is only about fixing a11y for homepage 😁

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
```diff
@@ -61,6 +61,9 @@ export function SignIn() {
...
+              <legend>Sign in</legend>
```
reesscot, line 65: Forgot to mention that this should be translated. See example here:
https://github.com/aws-amplify/amplify-ui/blob/57f1441e4809218a813148d0942de8171d159831/packages/react/src/components/Authenticator/ConfirmSignUp/ConfirmSignUp.tsx#L60

packages/vue/src/components/sign-in.vue
reesscot: Good point, sorry for the scope creep 😭

packages/react/src/primitives/Label/Label.tsx
reesscot: Thank you for checking. So happy this works with classnames now (maybe I was just doing it wrong before, who knows!)

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
0618: Thanks for catching it. Updated

0618:
> @0618 For the "Missing First level heading" error, can you please wrap the main Amplify UI logo in a <Heading level=1"> and add a `<title>Amplify UI</title>` to the SVG file. This should give the SVG an accessible title. https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

Good point. #TIL

reesscot (approved):
LGTM, let's ask @wlee221 or @ErikCH take a look at the Authenticator changes.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-in/sign-in.component.html
```diff
@@ -38,14 +23,8 @@
...
-    <button
-      amplify-button
-      fontWeight="normal"
-      size="small"
-      variation="link"
-      fullWidth="true"
-      (click)="authenticator.toResetPassword()"
-    >
+    <button amplify-button fontWeight="normal" size="small" variation="link" fullWidth="true"
+      (click)="authenticator.toResetPassword()">
```
wlee221, line 27: Style nit -- why is this just two lines now? Shouldn't prettier break it up to multiple lines like it was before?

packages/react/src/components/Authenticator/SignIn/SignIn.tsx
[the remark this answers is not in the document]
wlee221: Thanks @0618 for parity support!

wlee221 (approved):
LGTM, just a prettier nit but actual changes looks good.  And lmk if you need help with resolving conflicts, it's probably from my PR last night 😛

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-in/sign-in.component.html
reesscot: Silly prettier, always changing its mind.

PR #1500, opened 2022-03-09
Fix Dependabot ANSI-HTML version

ErikCH:
#### Description of changes
This will fix a security issue with `ansi-html` found by dependabot. This will make sure we are using the earliest fixed version.

image

#### Issue #, if available

Dependabot

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
   ],
   "resolutions": {
     "ansi-regex": "5.0.1",
+    "ansi-html": "0.0.8",
     "fs-extra": "^10.0.0",
     "jest": "^26.6.3",
     "json-schema": "^0.4.0",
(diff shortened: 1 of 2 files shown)


wlee221 (approved):
LGTM as long as tests pass

package.json
```diff
@@ -56,6 +56,7 @@
   ],
   "resolutions": {
     "ansi-regex": "5.0.1",
+    "ansi-html": "0.0.8",
```
ErikCH, line 59: Force ansi-html to `0.0.8`. This is only used by the angular example dev kit server. So it should NOT affect anything else.

joebuono (approved):

PR #1504, opened 2022-03-10
Resolve conflicts #1492 -> #1484

wlee221:
#### Description of changes

Creating a separate PR because merge conflicts from #1492 to #1484 was non-trivial.

Now this PR puts validation errors under each `FormField` instead of handling them separately.

#### Issue #, if available

Validated manually:

https://user-images.githubusercontent.com/43682783/157737171-c505fef2-6966-4280-a8b3-f59de4771c2b.mp4

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
@@ -34,7 +34,7 @@
     [autocomplete]="formField.autocomplete"
     [type]="formField.type"
   ></amplify-text-field>
-  <div data-amplify-sign-up-errors *ngIf="errors && errors.length > 0">
+  <div data-amplify-sign-up-errors *ngIf="errors?.length > 0">
     <div
       class="amplify-text"
       data-variation="error"
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
@@ -34,8 +34,14 @@
     [autocomplete]="formField.autocomplete"
     [type]="formField.type"
   ></amplify-text-field>
-
-  <amplify-error *ngIf="error">
-    {{ error }}
-  </amplify-error>
+  <div data-amplify-sign-up-errors *ngIf="errors && errors.length > 0">
+    <!-- TODO: add amplify-alert reusable component -->
+    <div
+      class="amplify-text"
+      data-variation="error"
+      *ngFor="let error of errors"
+    >
+      {{ translate(error) }}
+    </div>
+  </div>
 </div>
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
@@ -35,7 +35,6 @@
     [type]="formField.type"
   ></amplify-text-field>
   <div data-amplify-sign-up-errors *ngIf="errors && errors.length > 0">
-    <!-- TODO: add amplify-alert reusable component -->
     <div
       class="amplify-text"
       data-variation="error"
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
@@ -1,10 +1,9 @@
 import { Component, Input } from '@angular/core';
 import {
-  ActorContextWithForms,
-  getActorContext,
   translate,
   countryDialCodes,
   FormFieldOptions,
+  getErrorArray,
 } from '@aws-amplify/ui';
 import { AuthenticatorService } from '../../../../services/authenticator.service';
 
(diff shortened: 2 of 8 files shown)


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.ts
```diff
@@ -42,4 +38,8 @@ export class FormFieldComponent {
   isPhoneField(): boolean {
     return this.formField.type === 'tel';
   }
+
+  translate(pharse: string): string {
```
ErikCH, line 42: Should this be phrase? typo

wlee221: ahhhhh thanks!

wlee221:
```suggestion
  translate(phrase: string): string {
```

packages/vue/src/components/sign-up.vue
```diff
@@ -54,8 +54,8 @@ const submit = (e: Event): void => {
 
     <base-wrapper v-bind="$attrs">
       <base-form
-        @blur="onBlur"
         @input="onInput"
+        @blur.capture="onBlur"
```
ErikCH, line 58: Nice catch!

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
```diff
@@ -34,8 +34,13 @@
...
-    {{ error }}
-  </amplify-error>
+  <div data-amplify-sign-up-errors *ngIf="errors?.length > 0">
+    <div
+      class="amplify-text"
+      data-variation="error"
+      *ngFor="let error of errors"
+    >
+      {{ translate(error) }}
+    </div>
```
ErikCH, line 44: Won't this break the normal errors? So some errors are just red text. But the errors if you send a wrong password should be an `amplify-error` with a red box around it right? Or are those errors somewhere else?

wlee221: These are only for validation error (under each `form-field`). `remoteError`s still will be handled from each route: e.g.

https://github.com/aws-amplify/amplify-ui/blob/aad9a41ecf90a9f081c974f4e7e134a9bb9ff277/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-in/sign-in.component.html#L16-L18

packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
```diff
@@ -0,0 +1,22 @@
...
+  errors: string[];
+}
+export const ValidationErrors = ({ errors }: ValidationErrorsProps) => {
+  if (!errors || errors.length === 0) return null;
+
+  return (
+    <View data-amplify-sign-up-errors>
+      {errors.map((error, idx) => {
+        return (
+          <Text key={idx} role="alert" variation="error">
```
calebpollman, line 15: Not certain it would be really be an issue here, but the value used for `key` should be unique and static to keep the reference the same between renders. Using index values for `keys` is potentially problematic as they can change if the length of the array changes

calebpollman (approved):
LGTM, left some non-blocking feedback/questions

packages/ui/src/helpers/authenticator/form.ts
```diff
@@ -42,3 +42,16 @@ export const isAuthFieldWithDefaults = (
 ): field is AuthFieldsWithDefaults => {
   return authFieldsWithDefaults.includes(field as AuthFieldsWithDefaults);
 };
+
+const isArray = <T>(val: T | T[]): val is T[] => {
+  return Array.isArray(val);
+};
+
+export const getErrorArray = (errors: string | string[]) => {
```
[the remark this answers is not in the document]
wlee221: Done here https://github.com/aws-amplify/amplify-ui/pull/1504/commits/4821531c96b912e0d6608c689a51bd1b94c1aed9

packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
wlee221:
```suggestion
          <Text key={error} role="alert" variation="error">
```

packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
```diff
@@ -0,0 +1,22 @@
+import { translate } from '@aws-amplify/ui';
+import { View } from '../../../primitives/View';
+import { Text } from '../../../primitives/Text';
+
+export interface ValidationErrorsProps {
+  errors: string[];
```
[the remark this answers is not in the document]
wlee221: Hmm, this is confusing with already existing

```
type ValidationError = Record<string, string | string[]>
```

packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
calebpollman: Think that is good for now. We may need to figure something out more long term once if/when we allow users to override the values of `error` since we cannot guarantee uniqueness

packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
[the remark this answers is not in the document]
calebpollman: Agree, we could clean it up as a tech item. Thanks!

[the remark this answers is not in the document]
wlee221: Yeah, I think we can clean it up when we enforce `string | string[]` to be `string[]`. Let's keep it at this for now.

packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
wlee221: Got it.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/form-field/form-field.component.html
ErikCH: 👍

PR #1522, opened 2022-03-15
Fix phone number label in reset-password

wlee221:
#### Description of changes

This fixes a regression I introduced in #1484, in which phone number field would show up as `Enter your phone_number` instead of `Enter your phone number`. This aligns with previous behaviors.

Before fix:
image

After fix:

Screen Shot 2022-03-15 at 11 03 35 AM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/src/helpers/authenticator/formFields/defaults.ts
+++ b/packages/ui/src/helpers/authenticator/formFields/defaults.ts
@@ -105,11 +105,12 @@ const getConfirmSignUpFormFields = (state: AuthMachineState): FormFields => ({
 
 const getResetPasswordFormFields = (state: AuthMachineState): FormFields => {
   const primaryAlias = getPrimaryAlias(state);
+  const { label } = defaultFormFieldOptions[primaryAlias];
   return {
     username: {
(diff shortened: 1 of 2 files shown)


reesscot (approved):
LGTM

packages/ui/src/helpers/authenticator/formFields/defaults.ts
```diff
@@ -105,11 +105,12 @@ const getConfirmSignUpFormFields = (state: AuthMachineState): FormFields => ({
 
 const getResetPasswordFormFields = (state: AuthMachineState): FormFields => {
   const primaryAlias = getPrimaryAlias(state);
+  const { label } = defaultFormFieldOptions[primaryAlias];
```
wlee221, line 108: This gets the label from here:
https://github.com/aws-amplify/amplify-ui/blob/932629520a2f855ae2cadbeb8ec741f5607276ef/packages/ui/src/helpers/authenticator/constants.ts#L89-L91

PR #1527, opened 2022-03-16
fix(docs): Fix docs theme toggle button

### LINKED ISSUE #1526: Theme selector in docs do not work
## Problem
Theme selector toggle buttons in [docs](https://ui.docs.amplify.aws/) are not working. 

https://user-images.githubusercontent.com/43682783/158646618-39a7bcb5-8418-4e1c-9fd4-9df1c2bb5aee.mp4

## Root Cause
This is a regression from #1447, which puts the `[data-amplify-theme]` attribute inside a div in `<AmplifyProvider />`, in addition to the top level `<html />`. But `[data-amplify-theme-override] from the theme toggle buttons is still placed in the top-level `html` like so:

```html
(1) <html data-amplify-theme="amplify-docs" data-amplify-theme-override="classic">
(2)   <body>
(3)     <AmplifyProvider>
(4)       <div data-amplify-theme="amplify-docs">
              ...
```

CSS variables set from `[data-amplify-theme-override]` on (1) gets overridden by css variables set from `[data-amplify-theme]` on (4).

wlee221:
## Description of changes

This fixes the docs theme toggle button by targeting `[data-amplify-theme]` to override the theme css variables.

| Before | After |
|  toggle-fix-before | toggle-fix-after  |

## Details
This is a regression from #1447, which puts the `[data-amplify-theme]` attribute inside a div in `<AmplifyProvider />`, in addition to the top level `<html />`. But `[data-amplify-theme-override] from the theme toggle buttons is still placed in the top-level `html` like so:

```html
(1) <html data-amplify-theme="amplify-docs" data-amplify-theme-override="classic">
(2)   <body>
(3)     <AmplifyProvider>
(4)       <div data-amplify-theme="amplify-docs">
              ...
```

CSS variables set from `[data-amplify-theme-override]` on (1) gets overridden by css variables set from `[data-amplify-theme]` on (4).

## Fix

This PR fixes this issue by changing the override selector 

```diff
- [data-amplify-theme-override="classic"]
+ - [data-amplify-theme-override="classic"] [data-amplify-theme="amplify-docs"]
```

Overrides within nested providers in docs should not be affected because they have different theme name.

## Issue #, if available

Fixes #1526.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/theme.ts
+++ b/docs/src/theme.ts
@@ -69,7 +69,7 @@ export const theme: Theme = {
       },
     },
     {
-      selector: '[data-amplify-theme-override="classic"]',
+      selector: '[data-amplify-theme-override="classic"] [data-amplify-theme]',
       tokens: {
         colors: {
           brand: {
@@ -91,7 +91,7 @@ export const theme: Theme = {
       },
     },
     {
-      selector: '[data-amplify-theme-override="terminal"]',
+      selector: '[data-amplify-theme-override="terminal"] [data-amplify-theme]',
       tokens: {
         colors: {
           green: {


reesscot (approved):
LTGM as a fix. It feels odd that we have to do this though, should we open a ticket to research whether we are doing subthemes in a logical way?

0618 (approved):

reesscot:
> LGTM, but just a small question: do we need to add test to prevent the regression from happening again?

This is a good idea. @wlee221 Can we add an integration test for this in the existing docs integ tests? Can be a follow up PR.

zchenwei (commented):
I might have some context there. Duplicating `[data-amplify-theme]` on both html and div is for menu portal which needs to reference css outside of div. So I feel like adding `[data-amplify-theme-override]` to the div is more in line with what we expected versus increasing the specificity.

zchenwei (approved):

wlee221:
@reesscot 
> It feels odd that we have to do this though, should we open a ticket to research whether we are doing subthemes in a logical way?

Agreed. I think we should pass override theme to `AmplifyProvider`, instead of trying to override it at html level. 

@0618 @reesscot 
> Can we add an integration test for this in the existing docs integ tests? Can be a follow up PR.

Agreed, will follow up right after tis PR!

@zchenwei 
Thanks for the context! Yeah, that `div` is outside of our control as of now. Let's look for a better way to handle overrides in https://app.asana.com/0/1201736086077862/1201980096562763.

PR #1530, opened 2022-03-17
fix: update complex field components to allow absolute positioning

reesscot:
#### Description of changes
Figma to Studio integration will absolutely position components when autolayout is not enabled in Figma. This causes an issue for field components that were passing `position`, `top`, and `left`, and `padding` down to the input element, causing the layout not to match Figma. This is fixed by moving the absolute position props and `padding` up to the container Flex element.

Screen Shot 2022-03-16 at 6 37 34 PM

* Manually published packaged to local Verdaccio and installed in in sample app
* Checked that sample app using code generated from Amplify Studio matches Figma

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.changeset/thick-pumas-guess.md
+++ b/.changeset/thick-pumas-guess.md
@@ -1,0 +1,7 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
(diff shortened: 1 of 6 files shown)


Milan-Shah (approved):
LGTM

wlee221 (approved):

PR #1537, opened 2022-03-17
fix: make shouldWarn required prop on useDeprecationWarning

reesscot:
#### Description of changes
This fixes an issue where when isMultiline is undefined (which is usually true), shouldWarn will default to true and trigger the warning. This causes the TextField component to always complain about the use of isMultiline for fields where it's not set to `true`. Now setting `isMultiline` to any falsy value should not trigger the deprecation warning. I added a unit test to ensure passing `undefined` to `shouldWarn` doesn't trigger a console warning.

* Ran docs site and verified that Multline deprecation message is no longer shown

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/.changeset/unlucky-ducks-melt.md
+++ b/.changeset/unlucky-ducks-melt.md
@@ -2,7 +2,7 @@
 "@aws-amplify/ui-react": patch
 ---
 
-fix: supress erroneous isMultiline deprecation warnings on TextField component
+fix: suppress erroneous isMultiline deprecation warnings on TextField component
 
 Deprecation warning messages are showing for users of TextField for the isMultiline prop even though
 they are not using the prop. This issue is fixed by making the shouldWarn prop required on the internal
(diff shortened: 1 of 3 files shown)


zchenwei (approved):
LGTM!

.changeset/unlucky-ducks-melt.md
```diff
@@ -0,0 +1,9 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+fix: supress erroneous isMultiline deprecation warnings on TextField component
```
joebuono, line 5: nit: suppress

joebuono (approved):

.changeset/unlucky-ducks-melt.md
reesscot: Thanks for the catch!

reesscot, line 5:
```suggestion
fix: suppress erroneous isMultiline deprecation warnings on TextField component
```

PR #1544, opened 2022-03-17
refactor(xstate): Use named actions instead of inline actions

wlee221:
#### Description of changes

Xstate has two options to declare actions/guards/series:

1. Inline

```ts
createMachine({
  entry: () => { console.log('Hello!') }
});
```

2. Named actions
```ts
createMachine(
  {
    entry: 'sayHello'
  },
  {
    actions: {
      sayHello: () => {  console.log('Hello!');  }
    }
  }
);
```

We should use named declaration (2), because:
- Better state machine visualization, because the names appear in the statechart
- Easier-to-understand code (Code split implementation details!)
- Better TS support

XState official document suggests the same thing: [docs](https://xstate.js.org/docs/guides/typescript.html#use-named-actions-guards-services).

We already use and prefer (2), but there are some places we still use inline declarations. This PR makes all those named.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/src/machines/authenticator/actors/signIn.ts
+++ b/packages/ui/src/machines/authenticator/actors/signIn.ts
@@ -48,7 +48,7 @@ export function signInActor({ services }: SignInMachineOptions) {
           exit: ['clearFormValues', 'clearTouched'],
           states: {
             edit: {
-              entry: sendUpdate(),
+              entry: 'sendUpdate',
               on: {
                 SUBMIT: { actions: 'handleSubmit', target: 'submit' },
                 CHANGE: { actions: 'handleInput' },
@@ -57,7 +57,7 @@ export function signInActor({ services }: SignInMachineOptions) {
             },
             federatedSignIn: {
               tags: ['pending'],
-              entry: [sendUpdate(), 'clearError'],
+              entry: ['sendUpdate', 'clearError'],
               invoke: {
                 src: 'federatedSignIn',
                 // getting navigated out anyway, only track errors.
@@ -67,7 +67,7 @@ export function signInActor({ services }: SignInMachineOptions) {
             },
             submit: {
               tags: ['pending'],
-              entry: ['parsePhoneNumber', 'clearError', sendUpdate()],
+              entry: ['parsePhoneNumber', 'clearError', 'sendUpdate'],
               invoke: {
                 src: 'signIn',
                 onDone: [
@@ -118,7 +118,7 @@ export function signInActor({ services }: SignInMachineOptions) {
             },
             verifying: {
               tags: ['pending'],
-              entry: ['clearError', sendUpdate()],
+              entry: ['clearError', 'sendUpdate'],
               invoke: {
                 src: 'checkVerifiedContact',
                 onDone: [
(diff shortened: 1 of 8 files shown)


0618 (approved):

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -313,14 +316,19 @@ export function createAuthenticatorMachine() {
...
-          if (!event.data?.intent) return false;
-          return event.data.intent === 'confirmSignUp';
-        },
-        shouldRedirectToResetPassword: (_, event): boolean => {
-          if (!event.data?.intent) return false;
-          return event.data.intent === 'confirmPasswordReset';
-        },
+        // guards for initial states
+        shouldInitSignUp: (context) => context.config.initialState === 'signUp',
+        shouldInitResetPassword: (context) =>
```
wlee221, line 321: Not satisfied with names here -- maybe `isInitialStateSignUp`? `shouldSignUp`? open for thoughts...

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -271,7 +271,7 @@ export function signInActor({ services }: SignInMachineOptions) {
           exit: ['clearFormValues', 'clearError', 'clearTouched'],
           states: {
             edit: {
-              entry: sendUpdate(),
+              entry: 'sendUpdate',
```
calebpollman, line 274: I'm very new to xstate and do not know the expected patterns so please bear with me here, but would it make sense to use a string `enum` for the `entry` values in this file? Probably out of scope either way 😄

packages/ui/src/machines/authenticator/index.ts
calebpollman: Do think `isInitialStateSignUp` may read cleaner

calebpollman (approved):
LGTM 🚢 . Just left some non-blocking questions

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -42,26 +42,25 @@ export function createAuthenticatorMachine() {
           invoke: [
             {
               // TODO Wait for Auth to be configured
```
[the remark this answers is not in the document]
wlee221: Oh yeah, good question. Yes, ideally we would run this after we know `Amplify.configure` has been completed.  This has been the case because we `Amplifty.configure` are usually placed at app entry point, but we have no semantic guarantee over that 😛

packages/ui/src/machines/authenticator/actors/signIn.ts
wlee221: Nope, unfortunately it's the other way around. We pass non-enum'd string as the key here, and then xstate will use that key to infer the types. Technically, you can still pass strictly typed string here since it `extends string` anyway, but I don't believe that's an established pattern.

But that said, xstate will warn us if there are string typos through its [typegen](https://xstate.js.org/docs/guides/typescript.html#typegen) or through its [vscode extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode).

packages/ui/src/machines/authenticator/index.ts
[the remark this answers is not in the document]
wlee221: edited for wording *

packages/ui/src/machines/authenticator/index.ts
wlee221: Sounds good, adjusting!

PR #1548, opened 2022-03-18
chore(react-authenticator): consolidate Authenticator form event handlers in useFormHandlers hook

calebpollman:
#### Description of changes
- create custom `useFormHandlers` hook to consolidate form handlers used throughout React `Authenticator`
- remove duplicated form handler logic across React `Authenticator` components
- delete some unused utils

#### Issue #, if available
NA

- manually tested
- `yarn e2e dev`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated


packages/react/src/components/Authenticator/hooks/useFormHandlers/useFormHandlers.ts
```diff
@@ -0,0 +1,32 @@
...
+    ({ target: { name } }: React.FocusEvent<HTMLFormElement>) => {
+      updateBlur({ name });
+    },
+    [updateBlur]
+  );
+
+  const handleChange = useCallback(
+    ({ target: { name, value } }: React.ChangeEvent<HTMLFormElement>) => {
+      updateForm({ name, value });
+    },
```
ErikCH, line 19: This looks good!
Just one quick update. We still need to do the check here if it's a checkbox and set the value undefined. Otherwise it will break the `custom-sign-up-fields` example. We should also add another test for this use case as well. 

It occurs when you toggle the checkbox. It works the first time, but not the second time you toggle it. 

https://user-images.githubusercontent.com/65630/159376814-92d8e9e7-d13d-4295-bd04-5b81990b147d.mp4

calebpollman: Ahhh, thanks for the heads up!

wlee221 (approved):

Issue #1556, opened 2022-03-18
Mobile site fixes: Tabs make Authenticator example too wide


joebuono:
### On which framework/platform would you like to see this feature implemented?

React

### Which UI component is this feature-request for?

Other

### Authenticator page

<img src="https://user-images.githubusercontent.com/321279/145105828-6645e6e9-97a6-4691-9299-a5034f38df2a.png" width="400px" align="left" />

- [] Example tabs makes the page too wide

<br clear="both"/>

### Please describe a solution you'd like.

Use an `<Expander />` component instead of `<Tabs />`

### We love contributors! Is this something you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

joebuono:
https://github.com/aws-amplify/amplify-ui/pull/1557

PR #1557, opened 2022-03-18
docs(fix): improve responsiveness for Auth example

joebuono:
#### Description of changes

For the [Labels & Text](https://ui.docs.amplify.aws/components/authenticator#labels--text) example in Authenticator docs, this PR replaces the `Tabs` component with the `Expander` component for better layout on mobile. 

ezgif com-gif-maker (1)

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/1556

Used Chrome DevTools to check that the Expander looks good on mobile screens (see gif above). 

#### Checklist

- [X] PR description included


docs/src/pages/components/authenticator/customization.labels-and-text.web.mdx
```diff
@@ -1,14 +1,14 @@
...
 
 import { LabelsAndTextDemo } from './LabelsAndTextDemo';
 
 Using the same techniques as [Internationalization (I18n)](#internationalization-i18n), you can customize the labels and text of the components:
 
-<Tabs>
-  <TabItem title="Default Values">
+<Expander type="single" isCollapsible>
+  <ExpanderItem title="Default Values" value="default-values">
     Compare the default labels below to those in the customized screens via the tabs above.
```
wlee221, line 12: Since we don't have tabs anymore,

```suggestion
    Compare the default labels here to those in the customized below.
```

reesscot:
Should we only use the Expander on mobile? 
Does it look weird on desktop to have only one example be Expander and all the others be Tabs?
What do you think of having the Default Values open by default?

docs/src/pages/components/authenticator/customization.labels-and-text.web.mdx
wlee221: Also, we should consider separating "Default Values" out of expanders because its purpose is to show reference to default values.

dbanksdesign (commented):
LGTM with @wlee221's comments

joebuono:
> @reesscot  Should we only use the Expander on mobile? Does it look weird on desktop to have only one example be Expander and all the others be Tabs? What do you think of having the Default Values open by default?

1.  Although I think using an `Expander` is an elegant solution for examples with more than three options, I don't think we should _only_ use `Expander` on mobile. When there are three or fewer options (e.g., [Login Mechanisms](https://ui.docs.amplify.aws/components/authenticator#login-mechanisms)), then the `Tabs` component provides a quick way to check the differences between each scenario without needing to scroll up or down to select another option. Of course, I'm about to others' opinions. 
2. IMO, it doesn't look weird on desktop to have only one example be `Expander`, but I'm open to refactoring if it looks weird to anyone. 
3. Yes, Default Values is now open by default.

0618 (approved):
Good idea!

jacoblogan (approved):
LGTM

PR #1563, opened 2022-03-18
chore(docs): add meta description

0618:
#### Description of changes

Add meta-description and title to every page on docs site

image

#### Issue #, if available
https://app.asana.com/0/1201736086077838/1201759796183699/f

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/data/meta.ts
+++ b/docs/src/data/meta.ts
@@ -8,12 +8,12 @@ export const META_INFO: MetaInfo = {
     description:
       'Amplify UI is an open-source design system with cloud-connected components and primitives that simplify building accessible, performant, and beautiful applications in React, Angular, Vue, and Flutter (more coming soon).',
   },
-  '/installation': {
+  '/getting-started/installation': {
     title: 'Installation',
     description:
       'Installation Guide - How to install Amplify UI Packages, Styles, Fonts and Troubleshooting.',
   },
-  '/migration': {
+  '/getting-started/migration': {
     title: 'Migration',
     description:
       'Migration Guide - How to migrate from an older version of Amplify UI to a newer version.',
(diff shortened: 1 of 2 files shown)


zchenwei (approved):
LGTM!

docs/src/pages/_app.page.tsx
```diff
@@ -35,6 +35,10 @@ function MyApp({ Component, pageProps }) {
       <Head>
         <title>Amplify UI</title>
         <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <meta
+          name="description"
+          content="Amplify UI is an open-source design system with cloud-connected components and primitives that simplify building accessible, performant, and beautiful applications in React, Angular, Vue, and Flutter (more coming soon)."
```
jacoblogan, line 40: Are we trying to avoid the "P" word in our documentation now? I don't remember where we ended up on this

reesscot: I think we can continue to use it for now. There isn't a great alternative

reesscot: @0618 Don't you think the meta description should be different for each page though for SEO?

0618: I agree! Good point @reesscot  We should use different meta-descriptions on each page. How do we decide what to use on each?

Should I draft a doc for it?

0618: doc created https://quip-amazon.com/PPRQA0yE9o5J/Docs-Meta-Descriptions

zchenwei:
It seems we are adding a new data now but I was under the impression that the description get by `frontmatter`

docs/src/pages/_app.page.tsx
0618: Instead of creating the meta info separately (as in 5ebc04f36110f3ffa09650b45f0da7a74ff26e24), I created a script `createMetaInfo` to auto update them using the meta info in the mdx file, so that we can keep a single source of truth

0618:
> It seems we are adding a new data file now but I was under the impression that we can get `description` right from each page by [`frontmatter`](https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/utils/getPageFromSlug.tsx#L22). I think we have an example injecting it into layout <- https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/components/Layout/index.tsx#L28

Thanks! Discussed with @reesscot about it and we decided to not dynamically update from the frontmatter so that we have more flexibility to change.

zchenwei:
> > It seems we are adding a new data file now but I was under the impression that we can get `description` right from each page by [`frontmatter`](https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/utils/getPageFromSlug.tsx#L22). I think we have an example injecting it into layout <- https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/components/Layout/index.tsx#L28
> 
> Thanks! Discussed with @reesscot about it and we decided to not dynamically update from the frontmatter so that we have more flexibility to change.

What do you mean by flexibility? Shouldn't we have to maintain both description in `index.page.mdx` and the data now?

0618:
> > > It seems we are adding a new data file now but I was under the impression that we can get `description` right from each page by [`frontmatter`](https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/utils/getPageFromSlug.tsx#L22). I think we have an example injecting it into layout <- https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/components/Layout/index.tsx#L28
> > 
> > 
> > Thanks! Discussed with @reesscot about it and we decided to not dynamically update from the frontmatter so that we have more flexibility to change.
> 
> What do you mean by flexibility? Shouldn't we have to maintain description in both `index.page.mdx` and the data in that case?

yes

docs/src/data/meta.ts
```diff
@@ -0,0 +1,253 @@
...
+    description:
+      'Storage is a simple way to interact with our storage solutions, such as S3 - Amplify UI',
+  },
+  '/components/divider': {
+    title: 'Divider',
+    description:
+      'Divider creates separations in content, helping to organize content and establish visual rhythm - Amplify UI',
+  },
+  '/components/heading': {
+    title: 'Heading',
```
reesscot, line 47: Sorry I didn't notice it before, but can we append Amplify UI at the end of all the titles?
```suggestion
    title: 'Heading - Amplify UI',
```

0618: I've done it in _ampp.page.tsx to save some strolls 😁  https://github.com/aws-amplify/amplify-ui/pull/1563/files#diff-71bc81a2d4c2303176a0670d3591211d2ff148b097962a1d0f84e86f79060649L36

reesscot (approved):
LGTM! Thanks for your hard work making our site more SEO friendly!

PR #1570, opened 2022-03-23
feat(geo): AmplifyMap component implementation

slaymance:
#### Description of changes
This Pull Request adds the implementation for `AmplifyMap` to our feature branch (`feat-geo`) which will be merged to `main` on the date of the scheduled GA release for Geo components.

`AmplifyMap` provides an interactive map to users based upon the [Amplify Geo APIs](https://docs.amplify.aws/lib/geo/getting-started/q/platform/js/). It's built upon [react-map-gl](https://visgl.github.io/react-map-gl/) and provides the same functionality as the library's [default map](https://visgl.github.io/react-map-gl/docs/api-reference/map) while using Amazon Location Services.

Screen Shot 2022-03-23 at 11 36 19

This pull request is part of a series:
- https://github.com/aws-amplify/amplify-ui/pull/1571
- https://github.com/aws-amplify/amplify-ui/pull/1573
- https://github.com/aws-amplify/amplify-ui/pull/1574
- https://github.com/aws-amplify/amplify-ui/pull/1572

This pull request will be part of a series to reduce the amount of changes in each review. Validation includes unit testing and end-to-end testing as part of [PR 1572](https://github.com/aws-amplify/amplify-ui/pull/1572).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated
- [x] Relevant documentation is changed or added (see [PR 1572](https://github.com/aws-amplify/amplify-ui/pull/1572))


packages/react/src/components/Geo/AmplifyMap/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    ...style,
+  } as React.CSSProperties;
+
+  useEffect(() => {
+    const getCredentials = async () => {
+      setCredentials(await Auth.currentUserCredentials());
+    };
+
+    getCredentials();
+  }, []);
```
slaymance, line 46: A note about credentials and amplify config:

I originally had a Geo xstate machine built out to manage map and geocoder state which also handled fetching credentials and Amplify configurations. However, after updating to `react-map-gl` v7, I was able to greatly reduce the amount of state we would need to manage for our components. Thus, I removed the Geo xstate machine implementation since it doesn't currently solve any problems for our Geo components, and fetched credentials and config within the component itself as I need them.

Until the use of xstate provides a benefit, I'm of the opinion that we should not implement it and favor simplified react components.

packages/react/src/components/Geo/AmplifyMap/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+   * mapbox-gl-js, which uses its own copyrighted license. We override the map library with the BSD-licensed
+   * maplibre-gl-js.
+   */
+  return transformRequest ? (
+    <ReactMapGL
+      mapLib={maplibregl}
+      mapStyle={mapId}
+      style={styleProps}
+      transformRequest={transformRequest}
+      {...props}
```
calebpollman, line 72: Just a general callout regarding placing spread props as the last prop in a React component. 


Any time a spread `{...props}` is placed as the last attribute on a React component, we override any props that have been listed above it that have matching keys. So in this case if i used an `AmplifyMap` in my app like so:

```
<AmplifyMap transformRequest={() => null} />
```

it would prevent the `transformRequest` value we are specifying on ln 71 from being passed to the underlying `ReactMapGL`. This may be intentional, as I don't have a huge amount of context here, but figured I should note it just in case :)

packages/react/src/components/Geo/AmplifyMap/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * [react-map-gl default Map](https://visgl.github.io/react-map-gl/docs/api-reference/map), it accepts the same
+ * properties.
+ *
+ * @example
+ * // Basic usage of AmplifyMap:
+ * function App() {
+ *   return <AmplifyMap />
+ * }
+ */
+export const AmplifyMap = ({ style, ...props }: MapProps) => {
```
calebpollman, line 23: This kind of ties in with my comment below on spread props - it may be valuable to export our own `AmplifyMapProps` interface that removes any props that we explicitly use in our `AmplifyMap` component, e.g.
```
export interface AmplifyMapProps extends Omit<MapProps, 'transformRequest' | 'mapLib' | 'mapStyle'>;
```

calebpollman (commented):
Looks like a good start to me 👍  

Reviewed purely from a React perspective, may have missed or misunderstood nuances of the feature itself 🙃

packages/react/src/components/Geo/AmplifyMap/index.tsx
slaymance: This is a good callout and what I had originally written. I'm in agreement and can implement it this way unless there are other thoughts here.

packages/react/src/components/Geo/AmplifyMap/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+   * the request with AWS Sigv4 Auth, provided valid credentials, and is how we integrate react-map-gl with Amplify Geo
+   * and Amazon Location Service. Once the transformRequest is created, we render the map.
+   */
+  useEffect(() => {
+    if (credentials) {
+      const { transformRequest: amplifyTransformRequest } =
+        new AmplifyMapLibreRequest(credentials, region);
+      setTransformRequest(() => amplifyTransformRequest);
+    }
+  }, [credentials]);
```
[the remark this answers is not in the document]
slaymance: Great callout, this simplifies the logic a lot! With the latest change, we fetch the credentials inside the same `useEffect` hook and bypass adding them as state entirely.

packages/react/src/components/Geo/AmplifyMap/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * The mapLib property is used by react-map-gl@v7 to override the underlying map library. The default library is
+   * mapbox-gl-js, which uses its own copyrighted license. We override the map library with the BSD-licensed
+   * maplibre-gl-js.
+   */
+  return transformRequest ? (
+    <ReactMapGL
+      mapLib={maplibregl}
+      mapStyle={mapId}
```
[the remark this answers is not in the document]
slaymance: Yeah the style for Amazon Location Service maps is literally the map ID, but I think this is a clearer change for our purposes.

packages/react/src/components/Geo/AmplifyMap/index.tsx
slaymance: After thinking about this a bit, I think it makes sense for us to omit a user-provided `transformRequest` property since it's the bread and butter of our component and overwriting it defeats the purpose of our library.

We can keep the logic to allow overwriting `mapLib` and `mapStyle` as these are just default values we provide out of the box and a user may want to change them while still taking advantage of Amazon Location Service APIs.

packages/react/src/components/Geo/AmplifyMap/index.tsx
slaymance: See my comment above; we can omit the `transformRequest` property while allowing `mapLib` and `mapStyle` to be overwritten.

packages/react/package.json
```diff
@@ -38,7 +38,7 @@
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "yarn run build:ts && yarn run build:catalog",
-    "build:ts": "tsup --minify --keep-names",
+    "build:ts": "tsup --minify",
```
slaymance, line 41: `--keep-names` was causing a build error for the map component. Can anyone provide a use-case for needing this flag?

packages/react/package.json
```diff
@@ -64,13 +64,17 @@
     "classnames": "2.3.1",
     "deepmerge": "4.2.2",
     "lodash": "4.17.21",
+    "mapbox-gl": "npm:empty-npm-package@1.0.0",
```
slaymance, line 67: We need to alias `mapbox-gl` for users, even though we overwrite it with `maplibre-gl`. `react-map-gl` recommends aliasing the package with an empty npm package: https://visgl.github.io/react-map-gl/docs/get-started/get-started#using-with-a-mapbox-gl-fork

ErikCH (approved):
👍 
Good job!

PR #1575, opened 2022-03-23
Json to es

jacoblogan:
#### Description of changes
Add a new es module export of PrimitiveCatalog to the `/internal/primitive` subpath.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -12,6 +12,10 @@
       "import": "./dist/esm/internal.js",
       "require": "./dist/internal.js"
     },
+    "./internal/primitive": {
+      "import": "./dist/esm/primitive.js",
+      "default": null
+    },
     "./legacy": {
       "import": "./dist/esm/legacy.js",
       "require": "./dist/legacy.js"
--- a/packages/react/scripts/generatePrimitivesCatalog.ts
+++ b/packages/react/scripts/generatePrimitivesCatalog.ts
@@ -169,11 +169,17 @@ for (const [componentName, [node]] of source.getExportedDeclarations()) {
 }
 
 /**
- * Generate the JSON string of the primitives catalog
- * this is being exported under the /primitives.json subpath and can be used by
+ * Generate the JSON string of the PrimitiveCatalog
+ * this is being exported under the /primitives.json subpath and can be used as
  * import PrimitiveCatalog from '@aws-amplify/ui-react/primitives.json'
  */
 const jsonString = JSON.stringify(catalog, null, 2);
+
+/**
+ * Generate the es module of the PrimitiveCatalog
+ * this is being exported under the /internal/primitive subpath and can be used as
+ * import { PrimitiveCatalog } from '@aws-amplify/ui-react/internal/primitive'
+ */
 const exportString = `export const PrimitiveCatalog = ${jsonString};
 export default PrimitiveCatalog;`;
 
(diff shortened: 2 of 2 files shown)


packages/react/package.json
```diff
@@ -12,6 +12,10 @@
       "import": "./dist/esm/internal.js",
       "require": "./dist/internal.js"
     },
+    "./internal/primitive": {
```
reesscot, line 15:
```suggestion
    "./internal/primitives-catalog": {
```

packages/react/package.json
```diff
@@ -12,6 +12,10 @@
...
+      "import": "./dist/esm/primitive.js",
```
reesscot, line 16:
```suggestion
      "import": "./dist/esm/primitives-catalog.js",
```

packages/react/scripts/generatePrimitivesCatalog.ts
```diff
@@ -168,6 +168,28 @@ for (const [componentName, [node]] of source.getExportedDeclarations()) {
...
+ */
+const jsonString = JSON.stringify(catalog, null, 2);
+
+/**
+ * Generate the es module of the PrimitiveCatalog
+ * this is being exported under the /internal/primitive subpath and can be used as
+ * import { PrimitiveCatalog } from '@aws-amplify/ui-react/internal/primitive'
+ */
+const exportString = `export const PrimitiveCatalog = ${jsonString};
+export default PrimitiveCatalog;`;
```
reesscot, line 184: Should we use a named export to stay consistent with the rest of the repo?

```suggestion
export { PrimitiveCatalog }`;
```

reesscot (commented):
This approach LGTM. Just a few non-blocking naming nits.

packages/react/scripts/generatePrimitivesCatalog.ts
```diff
@@ -168,6 +168,27 @@ for (const [componentName, [node]] of source.getExportedDeclarations()) {
...
+
+// Generates dist/primitives.js file
+const JSONoutputPath = path.resolve(__dirname, '..', 'dist', 'primitives.json');
+const internalOutputPath = path.resolve(
+  __dirname,
+  '..',
+  'dist/esm',
+  'primitives-catalog.js'
+);
+fs.writeFileSync(JSONoutputPath, jsonString);
```
jacoblogan, line 193: Continuing to create the JSON version so that this isn't a breaking change for studio, will remove this export once studio has had a chance to migrate over to the new export.

joebuono (approved):
LGTM

packages/react/package.json
```diff
@@ -12,6 +12,10 @@
       "import": "./dist/esm/internal.js",
       "require": "./dist/internal.js"
     },
+    "./internal/primitives-catalog": {
+      "import": "./dist/esm/primitives-catalog.js",
+      "default": null
```
zchenwei, line 17: Do we have a plan to add CJS export in the future? It is a dual(ESM/CJS) package. This seems deviating from the goal

packages/react/package.json
zchenwei: Should we be consistent with `primitive` ? The json version is named `primitive`. I think this will cause confusion but they are the same in different forms.

packages/react/package.json
jacoblogan: I don't know of any plans to add a CJS export as this is an internal package only being used by studio we would do one if they ever needed it, but I'm not sure if there are any plans otherwise. That being said I could add a CJS export for consistency if that is what we agree should happen.

zchenwei (approved):

packages/react/package.json
zchenwei: Ok, that makes sense. CJS is not a must

PR #1578, opened 2022-03-24
fix(examples,e2e): Add examples and tests for #1580

wlee221:
#### Description of changes
This updates the `useAuthenticator` examples and their e2e test to match #1580.

**Details**

Our current `useAuthenticator` examples are really SPA and do not sufficiently test router use cases.

There are many `useAuthenticator` bugs, especially with refresh, so we plan to update examples and tests incrementally as we add more `router` support. For now, this PR adds additional route `useAuthenticator/home`. 

This PR adds:

1. Test starts at `/useAuthenticater` route. This has the authenticator for user to sign in.
1. User signs in with Authenticator.
2. Once user signs in, they'll see a "Navigate to Home" button. (This can be something like "Manage your account" irl)
3. Clicking that button redirects user to /useAuthenticator/home`. This will have basic greetings with a sign out button.
4. Refreshing sustains auth contexts.

**Future Goals**
This example needs to be extended to better resemble customer use cases:

- [ ] Have a separate `/sign-in` route
- [ ] Automatically redirect users without buttons
- [ ] Add current user check in `/home` and make it actually protected.

#### Issue #, if available
Related to #1332, #1497, accompanies #1580

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1497, opened 2022-03-09
Discussion: Use cases for `useAuthenticator` Hook


vymao:
The original issue has since been resolved, however, leaving this open due to good discussion around use cases for the `useAuthenticator` hook

Original ticket below:

--------------------------

### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

### Please describe your bug.

I am trying to use the `useAuthenticator` hook to set a global authentication state. 

However, several problems have arisen: 
1. When I use `const { user, signOut } = useAuthenticator((context) => [context.user]);`, as is mentioned [here](https://ui.docs.amplify.aws/components/authenticator#access-authenticated-user), there is no app-level re-rendering. So several other components that are also using the hook and are dependent on authentication state don't change. 
2. If I refresh the page, React renders everything as signed out. When I go to sign in, it then signs me in automatically without me having to manually sign in. It should ideally render everything as signed in from the start, since I am already signed in. 

### What's the expected behaviour?

When I sign in or out, React should trigger an app-level re-render with the new authentication state. Furthermore, if I refresh the page, the authentication state should not change. 

### Help us reproduce the bug!

Following the instructions, I added `Authenticator.Provider` at the app level: 
```
ReactDOM.render(
  <Authenticator.Provider>
    <HelmetProvider>
      <BrowserRouter>
        <App />
      </BrowserRouter>
    </HelmetProvider>
  </Authenticator.Provider>,
  document.getElementById('root')
);
```
I use `useAuthenticator` to Login: 
```
const { route } = useAuthenticator(context => [context.route]);
    return (
        route === 'authenticated' ? <Navigate to="/dashboard/app" />: (
            <Authenticator
                // Default to Sign Up screen
                initialState="signUp"
                // Customize `Authenticator.SignUp.FormFields`
                signUpAttributes={['preferred_username', 'birthdate']}
                components={components}
                services={{
                    async validateCustomSignUp(formData) {
                        if (!formData.acknowledgement) {
                            return {
[...]

Issue #1332, opened 2022-02-15
Authenticator component returns idle in authenticated state


dmost714:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

vite

### Please describe your bug.

TLDR; If I'm signed-in and reload the browser window, I return to my marketing page instead of the dashboard page.

I'm using the `context.route` from `useAuthenticator` to determine if I should follow the authenticated or unauthenticated code path. If I'm authenticated (signed in via the `Authenticator` component), and reload the webpage, the context.route value returns `idle` instead of the expected `authenticated`. This results in the wrong code path being executed. If the `<Authenticator>` component is used in this state, it skips through sign-in and shows it's children components (my app).

### What's the expected behaviour?

If I'm signed-in and reload the webpage, the route returned by `useAuthenticator` should return `authenticated` instead of `idle`. 

### Help us reproduce the bug!

If a complete working example is needed, let me know.

Here's an explanation of the code snippet below...

I catch the route from useAuthenticator() and use it to return my dashboard routes (authenticated) or the marketing routes (unauthenticated).

The marketing routes show the marketing page and a 'sign-up/sign-in' button that goes to an '/enter' route.
The marketing page shows for "/", any other route resolves to the `<Authenticator />` component.
The marketing page has a button that navigates to 'enter', which leads to the `Authenticator` (per above).
Once authenticated, the dashboard's '/enter' route re-route to '/'; showing the dashboard.
If an unauthenticated user visits a deep link, they sign-in and see the intended page.

Everything works, EXCEPT, if I'm signed in AND on the root page AND reload the browser page i'm taken to the marketing page. This is because useAuthenticator() is returning `idle` after the reload, instead of `authenticated`. Due to the way I've setup routing, any other page will hit the authenticator component, skip though the sign-in and show the intended page.

I believe this is a known issue. I inquired about it in the Discord channel and received this advice:

**From Discord Amplify channel:**
wlee221 — Today at 2:01 PM
[...]

PR #1580, opened 2022-03-24
fix(authenticator): fetch current user on routed apps after refresh

wlee221:
## Problem

Previously, `useAuthenticator` hook/composable/service required an instance of Authenticator in the DOM tree to initiate the machine and look for any current authenticated user in the local storage.

This caused problems when a user refreshed in an authenticated route. For example, imagine the following structure:

- `/` is the entry to the home page and has the `<Authenticator />` instance to enable sign in.
- `/my-account` is an _authenticated_ route that users can go after they signed in. This doesn't have `<Authenticator />` in DOM because they already signed in.

Whenever a user refreshed on `/my-account`, the page wouldn't have an `Authenticator` in the DOM to initiate the authMachine, and so auth contexts from the hooks returned undefined.

## Fix
This PR adjusts `useAuthenticator` hooks so that it first looks for `Auth.currentAuthenticatedUser`. If there are, it'll just transition directly to `authenticated` route and does not require an instance of `<Authenticator />` whatsoever. Otherwise, it would proceed to the `setup` state, in which it will wait for `Authenticator` to send the `INIT` event.

From the framework side, `<Authenticator />` will subscribe to authMachine changes. Once the machine gets to `setup` state and is ready for the `INIT` event, authenticator will send an INIT event with all the machine options.

## Remaining Problems

Note that there are still bugs we have to address for routed app use cases, namely #1581. This PR will at least correct the refresh behavior and remove the need for `Authenticator` in the DOM tree for machines to start.

#### Issue #, if available

Related to https://github.com/aws-amplify/amplify-ui/issues/1332, https://github.com/aws-amplify/amplify-ui/issues/1497

#1578 adds new examples and tests for refreshing. #1578 is to be merged into this PR once approved.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/ui/src/machines/authenticator/index.ts
```diff
@@ -25,47 +25,58 @@ export function createAuthenticatorMachine() {
       context: {
         user: undefined,
         config: {},
-        services: {},
+        services: { ...defaultServices },
         actorRef: undefined,
       },
       states: {
         // See: https://xstate.js.org/docs/guides/communication.html#invoking-promises
         idle: {
```
wlee221, line 33: `idle` state now starts with `Auth.currentAuthenticatedUser` call.

Knowing that, `idle` is a bad name (it's on me!) and should be named more explicitly. I scoped this out of this PR, because that would change our `route` names and I would rather deal with it in 3.0 changes.

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -25,47 +25,58 @@ export function createAuthenticatorMachine() {
...
-          invoke: [
-            {
-              // TODO Wait for Auth to be configured
-              src: 'getCurrentUser',
-              onDone: {
-                actions: 'setUser',
-                target: 'authenticated',
+          initial: 'waitConfig',
+          states: {
+            waitConfig: {
```
wlee221, line 48: At this stage, we confirmed that no user has signed in yet. Now we wait for `Authenticator` to send `INIT` event to pass machine props like `initialState`.

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -25,47 +25,58 @@ export function createAuthenticatorMachine() {
...
+              on: {
+                INIT: {
+                  actions: 'configure',
+                  target: 'applyConfig',
+                },
               },
-              onError: [
+            },
+            applyConfig: {
```
wlee221, line 56: Once we have all the machine props ready, we apply those configs to get `loginMechanisms`, etc.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
```diff
@@ -54,13 +56,24 @@ export class AuthenticatorComponent implements OnInit, AfterContentInit {
...
-      loginMechanisms,
-      services,
-      signUpAttributes,
-      socialProviders,
-      formFields,
+
+    // send INIT event once machine is at 'setup' state
+    this.authenticator.subscribe(() => {
+      const { route } = this.authenticator;
+      if (!this.hasInitialized && route === 'setup') {
```
wlee221, line 63: Previously `authMachine` got stuck in `idle` state until an `Authenticator` in DOM sends an `INIT` event on mount. Now `authMachine` starts independently (see state machine changes below), and will explicitly ask for config from `Authenticator` once it's in `setup` state.

Once `Authenticator` sees that state is in `setup`, it'll send the usual `INIT` event.

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -25,47 +25,58 @@ export function createAuthenticatorMachine() {
       context: {
         user: undefined,
         config: {},
-        services: {},
+        services: { ...defaultServices },
```
calebpollman, line 28: Curious, can we just pass `defaultServices` here?

packages/vue/src/components/authenticator.vue
```diff
@@ -80,19 +80,29 @@ let unsubscribeHub: ReturnType<typeof listenToAuthHub>;
 const { state, send } = useActor(service);
 useAuth(service);
 
+const hasInitialized = ref(false);
+
+service.subscribe((newState) => {
```
calebpollman, line 85: Just a couple questions:
- Should we unsubscribe after we have called `send`? (same question for the listener in Angular)
- What is the type of `newState`? May be missing context but it appears to be `Element` since we are calling `newState.matches` on ln 86, is that correct/expected?

packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -27,7 +27,7 @@ function InitMachine({ children, ...data }) {
   const hasInitialized = React.useRef(false);
 
   React.useEffect(() => {
-    if (!hasInitialized.current && route === 'idle') {
+    if (!hasInitialized.current && route === 'setup') {
```
calebpollman, line 30: Probably out of scope for this PR but what do you think of using `initial` as a route name in place of `setup`?

packages/vue/src/components/authenticator.vue
```diff
@@ -80,19 +80,29 @@ let unsubscribeHub: ReturnType<typeof listenToAuthHub>;
...
+  if (newState.matches('setup')) {
+    if (!hasInitialized.value) {
```
calebpollman, line 87: Nit: Think this can be combined with ln 86 in to one conditional statement

calebpollman (commented):
Granted I am still learning how `xstate` and the `Authenticator` state machine work but LGTM 👍🏽. Left some questions

packages/ui/src/machines/authenticator/index.ts
wlee221: Oh man, I did it again didn't I :D

Yep, you're right.

```suggestion
        services: defaultServices,
```

packages/vue/src/components/authenticator.vue
wlee221: > Should we unsubscribe after we have called send? 
Great point, yep `INIT` is a one time thing. Adjusting!

> What is the type of newState?
It's `XState.State` type, https://xstate.js.org/docs/guides/states.html#api. This is typed like so: 
image

packages/react/src/components/Authenticator/Authenticator.tsx
wlee221: Yep, I don't like the name `setup` either, it's too generic and doesn't say much. Let's scope this out of the PR, this ties to #1581 as well.

packages/vue/src/components/authenticator.vue
wlee221: Yep :p

packages/ui/src/machines/authenticator/index.ts
ErikCH: 👍

ErikCH (approved):
👍 

As long as the tests pass after you merge in your test PR

packages/vue/src/components/authenticator.vue
wlee221: Done in [924e906](https://github.com/aws-amplify/amplify-ui/pull/1580/commits/924e9066bdf25cdce7f9b60ded3bf6376df0a60c)

packages/vue/src/components/authenticator.vue
wlee221: unsubscription done in [c3a7daf](https://github.com/aws-amplify/amplify-ui/pull/1580/commits/c3a7dafcb61eda93c3642f4bd0aaa8c306bf5436) (vue) and [a72d2c3](https://github.com/aws-amplify/amplify-ui/pull/1580/commits/a72d2c314867ead5386d9ab870d62b652d2bcd1d) (angular).

calebpollman (approved):
LGTM 🚢  Thanks for addressing the feedback!

wlee221:
fyi, I added 9 liner commit [b8c36bb](https://github.com/aws-amplify/amplify-ui/pull/1580/commits/b8c36bb60671864136d7b79228c47441fefff696) to fix a autoSignIn test failure. `autoSignIn` flow needed to be adjusted so that it goes to the correct state based on signUp result. This uses the same flow as before, just made more explicit.

I'll merge this tomorrow morning unless there are objections!

slaymance (approved):
Looks good to me!

PR #1587, opened 2022-03-24
feat: type theme component design tokens - part I

reesscot:
#### Description of changes
This PR adds types to our component design tokens. This is the first of several PR's as it's already gotten large with typing just a handful of components.

The following components are typed in this PR: `alert`, `authenticator`, `badge`, `button`, `card`, and `checkbox`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.changeset/beige-insects-return.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
```
reesscot, line 2:
```suggestion
```

packages/ui/src/theme/tokens/components/checkboxField.ts
```diff
@@ -1,4 +1,10 @@
-export const checkboxfield = {
+export interface CheckboxFieldTokens {
+  alignItems: any;
```
reesscot, line 2: I will follow up with more PR's to type all the other component files. For now, typing these as any is the same as what was there before.

packages/ui/src/theme/tokens/components/toggleButton.ts
```diff
@@ -1,4 +1,16 @@
-export const togglebutton = {
+export interface ToggleButtonTokens {
+  borderColor: any;
+  color: any;
+  _hover: any;
```
joebuono, line 4: Just curious, why is an underscore prepended to the ones related to state?

packages/ui/src/theme/tokens/components/index.ts
```diff
@@ -60,7 +105,6 @@ export const components = {
   loader,
   menu,
   pagination,
-  phonenumberfield,
```
joebuono, line 63: Why are we removing `phonenumberfield`?

joebuono (approved):
Just had a few small, non-blocking questions. Thanks for doing all this!

packages/ui/src/theme/tokens/components/index.ts
reesscot: Because it doesn't have a single token in it. :)

packages/ui/src/theme/tokens/components/toggleButton.ts
reesscot: This is a convention we are using to indicate that these are tokens related to [pseudo classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes). We can't use `:` in a property name, or we would be using `:hover`  instead.

0618 (approved):

packages/ui/src/theme/tokens/components/authenticator.ts
```diff
@@ -1,4 +1,66 @@
...
+  color: DesignToken<ColorValue>;
+  orLine: {
+    backgroundColor: DesignToken<BackgroundColorValue>;
+  };
+}
+
+interface AuthenticatorContainerToken {
+  widthMax: DesignToken<SpaceValue>;
+}
+export interface AuthenticatorTokens {
```
calebpollman, line 52: Finding it somewhat difficult to grasp what the structure/contents of a token is meant to be. Looking at the contents of `AuthenticatorTokens` it appears that a token's contents can be:
- a style property (`maxWidth: DesignToken<SpaceValue>`)
- a style object (`form: AuthenticatorFormStyle`)
- another token (`container: AuthenticatorContainerToken`)

Is there some documentation I could look at to better understand this? Is there a semantic difference between a style object (e.g. `AuthenticatorRouterStyle`) versus a design token (e.g. `AuthenticatorContainerToken`)?

packages/ui/src/theme/tokens/components/heading.ts
```diff
@@ -1,4 +1,14 @@
-export const heading = {
+export interface HeadingTokens {
+  color: any;
+  lineHeight: any;
```
calebpollman, line 3: Should `lineHeight` be a property of the number/"size" keys? Does it need to scale with `fontSize` increases?

packages/ui/src/theme/tokens/components/index.ts
```diff
@@ -1,42 +1,87 @@
...
+import { table, TableTokens } from './table';
+import { tabs, TabsTokens } from './tabs';
+import { text, TextTokens } from './text';
+import { togglebutton, ToggleButtonTokens } from './toggleButton';
+import {
+  togglebuttongroup,
+  ToggleButtonGroupTokens,
+} from './toggleButtonGroup';
+
+export interface Components {
```
calebpollman, line 45: With `Components` already being somewhat of an overloaded term, maybe we can name this `ComponentTokens`?

packages/ui/src/theme/tokens/components/index.ts
```diff
@@ -1,42 +1,87 @@
...
+  alert: AlertTokens;
+  authenticator: AuthenticatorTokens;
+  badge: BadgeTokens;
+  button: ButtonTokens;
+  card: CardTokens;
+  checkbox: CheckboxTokens;
+  checkboxfield: CheckboxFieldTokens;
```
calebpollman, line 52: Is there a reason we are using all lowercase key names rather than camelCase in this interface?

calebpollman (commented):
Left mostly questions and a few suggestions. Thanks for doing this @reesscot

packages/ui/src/theme/tokens/components/alert.ts
```diff
@@ -1,4 +1,44 @@
...
+  BackgroundColorValue,
+  ColorValue,
+  DesignToken,
+  FontSizeValue,
+  FontWeightValue,
+  JustifyContentValue,
+  SpaceValue,
+} from '../types/designToken';
+
+interface AlertVariationTokens {
```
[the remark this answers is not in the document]
dbanksdesign: I think that is intentional, this has multiple tokens (color and backgroundColor)

packages/ui/src/theme/tokens/components/authenticator.ts
```diff
@@ -1,4 +1,66 @@
...
+  padding: DesignToken<SpaceValue>;
+}
+
+interface AuthenticatorStateStyle {
+  inactive: {
+    backgroundColor: DesignToken<BackgroundColorValue>;
+  };
+}
+
+interface AuthenticatorOrContainerStyle {
```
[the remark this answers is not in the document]
dbanksdesign: I think these "or line" styles are no longer needed. I removed them and instead used the Divider in the authenticator: https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignIn.tsx#L68

packages/ui/src/theme/tokens/components/authenticator.ts
```diff
@@ -1,4 +1,66 @@
...
+  color: DesignToken<ColorValue>;
+  orLine: {
+    backgroundColor: DesignToken<BackgroundColorValue>;
+  };
+}
+
+interface AuthenticatorContainerToken {
+  widthMax: DesignToken<SpaceValue>;
```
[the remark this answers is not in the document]
dbanksdesign: Yes it should

packages/ui/src/theme/tokens/components/badge.ts
```diff
@@ -1,4 +1,44 @@
...
+
+export interface BadgeVariation {
+  backgroundColor: DesignToken<BackgroundColorValue>;
+  color: DesignToken<ColorValue>;
+}
+export interface BadgeSizes {
+  fontSize: DesignToken<FontSizeValue>;
+  paddingHorizontal: DesignToken<SpaceValue>;
+  paddingVertical: DesignToken<SpaceValue>;
+}
```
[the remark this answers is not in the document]
dbanksdesign: Yea I think this and the alert.ts, we should be consistent. I would say either add `Tokens` at the end or go with a singular name like `BadgeVariation`. I don't have a strong preference because a customer won't use this type explicitly. `BadgeVariationTokens` is a bit more explicit. We could also go with `BadgeVariationTheme`?

packages/ui/src/theme/tokens/components/button.ts
```diff
@@ -1,4 +1,155 @@
...
+  boxShadow: DesignToken<BoxShadowValue>;
+}
+
+interface PrimaryVariationTokens {
+  borderWidth: DesignToken<BorderWidthValue>;
+  borderStyle: DesignToken<BorderStyleValue>;
+  borderColor: DesignToken<BorderColorValue>;
+  backgroundColor: DesignToken<BackgroundColorValue>;
+  color: DesignToken<ColorValue>;
+  _disabled: {
```
dbanksdesign, line 35: This could be `_disabled: StateTokens` right?

packages/ui/src/theme/tokens/components/card.ts
```diff
@@ -1,4 +1,43 @@
...
+  backgroundColor: DesignToken<BackgroundColorValue>;
+  borderRadius: DesignToken<BorderRadiusValue>;
+  borderWidth: DesignToken<BorderWidthValue>;
+  borderStyle: DesignToken<BorderStyleValue>;
+  borderColor: DesignToken<BorderColorValue>;
+  boxShadow: DesignToken<BoxShadowValue>;
+  padding: DesignToken<SpaceValue>;
+  outlined: OutlinedTokens;
+  elevated: ElevatedTokens;
+}
```
dbanksdesign, line 38: This could probably be simplified I think. What these types should really say: the default and all variations (elevated and outlined) have the same tokens. Maybe something like:

```ts
interface CardVariationTokens {
  backgroundColor: DesignToken<BackgroundColorValue>;
  borderRadius: DesignToken<BorderRadiusValue>;
  borderWidth: DesignToken<BorderWidthValue>;
  borderStyle: DesignToken<BorderStyleValue>;
  borderColor: DesignToken<BorderColorValue>;
  boxShadow: DesignToken<BoxShadowValue>;
}

type CardTokens = CardVariationTokens & {
  outlined: CardVariationTokens;
  elevated: CardVariationTokens;
}
```

packages/ui/src/theme/tokens/components/countryCodeSelect.ts
```diff
@@ -1,5 +1,9 @@
-export const countrycodeselect = {
+import { DesignToken, SpaceValue } from '../types/designToken';
+export interface CountryCodeSelectTokens {
+  height: DesignToken<SpaceValue>;
+}
+export const countrycodeselect: CountryCodeSelectTokens = {
```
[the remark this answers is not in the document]
dbanksdesign: We need to also at some point change this to not use "country" and instead use "dialCode" I think?

packages/ui/src/theme/tokens/components/heading.ts
dbanksdesign: lineHeights in the theme are unit-less values so they scale with fontSize by default

packages/ui/src/theme/tokens/components/index.ts
```diff
@@ -1,42 +1,87 @@
...
-import { selectfield } from './selectField';
-import { sliderfield } from './sliderField';
-import { stepperfield } from './stepperField';
-import { switchfield } from './switchField';
-import { table } from './table';
-import { tabs } from './tabs';
-import { text } from './text';
-import { togglebutton } from './toggleButton';
-import { togglebuttongroup } from './toggleButtonGroup';
+import { PaginationInput } from '@aws-amplify/datastore';
```
dbanksdesign, line 1: Do we need this import?

dbanksdesign (commented):
Overall, this looks like a really good start. Just a few minor comments

packages/ui/src/theme/tokens/components/authenticator.ts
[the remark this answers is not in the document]
reesscot: Yes, it should. But I'm typing what's already there, not making structural changes to the types in these PRs.

packages/ui/src/theme/tokens/components/authenticator.ts
```diff
@@ -1,4 +1,66 @@
...
+
+interface AuthenticatorStateStyle {
+  inactive: {
+    backgroundColor: DesignToken<BackgroundColorValue>;
+  };
+}
+
+interface AuthenticatorOrContainerStyle {
+  color: DesignToken<ColorValue>;
+  orLine: {
```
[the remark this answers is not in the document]
reesscot: I'm not sure. This PR types what the existing structure already was.

packages/ui/src/theme/tokens/components/authenticator.ts
```diff
@@ -1,4 +1,66 @@
...
+interface AuthenticatorFooterStyle {
+  paddingBottom: DesignToken<SpaceValue>;
+}
+
+interface AuthenticatorFormStyle {
+  padding: DesignToken<SpaceValue>;
+}
+
+interface AuthenticatorStateStyle {
+  inactive: {
```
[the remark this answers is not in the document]
reesscot: Probably. I don't have much context on authenticator styling.

packages/ui/src/theme/tokens/components/authenticator.ts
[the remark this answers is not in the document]
reesscot: I'll create a ticket to make this change. Since these PR's are focused on types, I don't want to make structural changes to the tokens.

packages/ui/src/theme/tokens/components/countryCodeSelect.ts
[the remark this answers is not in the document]
reesscot: Changing the casing would be a breaking change for customers.

packages/ui/src/theme/tokens/components/index.ts
reesscot: Yes, definitely. Will update

packages/ui/src/theme/tokens/components/index.ts
reesscot: Nope, I think this is a case of a VS Code autoimport error.

packages/ui/src/theme/tokens/components/index.ts
reesscot: I don't remember the original reason, but changing them now would be a breaking change for customers.

packages/ui/src/theme/tokens/components/button.ts
reesscot: Huh, I thought I did that. I got a bit lost in so many components 😅

packages/ui/src/theme/tokens/components/index.ts
calebpollman: Worth adding a todo and ticket for updating them as part of the 3.0 release?

packages/ui/src/theme/tokens/components/card.ts
reesscot: Good call, updated.

packages/ui/src/theme/tokens/components/badge.ts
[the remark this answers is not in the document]
reesscot: Updated to follow the same Tokens suffix on all the types.

packages/ui/src/theme/tokens/components/authenticator.ts
reesscot: The `Style` suffix was a mistake, I went with `Tokens` suffix on all the other tokens.

packages/ui/src/theme/tokens/components/index.ts
reesscot: Yes, I've created an asana task.

PR #1593, opened 2022-03-26
feat: type theme component design tokens - Part II

reesscot:
#### Description of changes
This PR adds types for the following component design tokens:
* checkboxField
* copy
* divider
* expander
* field
* fieldControl
* fieldMessage

#### Testing steps:
Started docs site and smoke test each component (just confirming that a few token changes to meet new types didn't affect styling)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


0618 (dismissed):
Thanks for breaking it to several smaller PRs!

0618 (commented):
Change set 👻

0618 (approved):

packages/ui/src/theme/tokens/components/fieldControl.ts
```diff
@@ -1,28 +1,101 @@
...
+  outlineStyle: DesignToken<OutlineStyleValue>;
+  outlineWidth: DesignToken<OutlineWidthValue>;
+  outlineOffset: DesignToken<OutlineOffsetValue>;
+
+  small: FieldControlSizeTokens;
+  large: FieldControlSizeTokens;
+  quiet: FieldControlQuietTokens;
+  _focus: FieldControlFocusTokens;
+  _disabled: FieldControlDisabledTokens;
+  _error: FieldControlErrorTokens;
```
dbanksdesign, line 98: Thinking out loud here... would it be more readable to inline these interfaces that aren't shared? `FieldControlErrorTokens` is only used here as far as I can tell. Something we might want to do across the board (if we think it is a good idea). I'm not super opinionated one way or the other, just wanted to bring it up

packages/ui/src/theme/tokens/components/expander.ts
```diff
@@ -1,17 +1,96 @@
...
+interface ExpanderItemFocusTokens {
+  boxShadow: DesignToken<BoxShadowValue>;
+}
+
+interface ExpanderItemTokens {
+  marginTop: DesignToken<SpaceValue>;
+  boxShadow: DesignToken<BoxShadowValue>;
+  borderBottomLeftRadius: DesignToken<RadiusValue>;
+  borderBottomRightRadius: DesignToken<RadiusValue>;
+  borderTopLeftRadius: DesignToken<RadiusValue>;
```
dbanksdesign, line 26: Why do we have both positional (top/left) and semantic (start/end) versions of border radius? Should we stick with one?

dbanksdesign (approved):
2 non-blocking minor comments. Other than that it looks good!

packages/ui/src/theme/tokens/components/expander.ts
reesscot: That's very strange indeed, I'll create a ticket for this.

reesscot: Oh now I remember, it's because Safari didn't support semantic border radius, so we had to have both specified in the CSS. But I don't think there's a reason to keep both tokens, we can just use the same positional tokens.

packages/ui/src/theme/tokens/components/fieldControl.ts
reesscot: Honestly went back and forth on this a few times. 😅 I decided to create an interface for every one because it was too hard to decide _which_ ones to split out and which ones not to split out. By creating an interface for all of them, it's more consistent.

Issue #1599, opened 2022-03-29
Extra Field When Customizing Issuer for Setup TOTP Screen


bestickley:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Vite

### Please describe your bug.

When I set the `formFields` property on Authenticator to: `{ setupTOTP: { QR: { totpIssuer: "My Web App" } } };` there is an extra field added to the setupTOTP form. See screenshot below. When I use Google Authenticator to add the TOTP it correctly changes the issuer name so that's good!

<img width="563" alt="image" src="https://user-images.githubusercontent.com/35735118/160668090-4f3093a9-8785-428b-aad4-9cd4173193ca.png">

HTML:
![image](https://user-images.githubusercontent.com/35735118/160668201-4b6a73da-79d0-413e-a016-c8d2001c1e4c.png)

### What's the expected behaviour?

Extra field is not added.

### Help us reproduce the bug!

See screenshot above.

ErikCH:
Hi @bestickley !

I was able to reproduce this, and it's a quick fix. We should have it up soon!

bestickley:
Thanks, @ErikCH!

ErikCH:
This has been merged closing. Should be out in the next release

PR #1600, opened 2022-03-29
fix(Authenticator): Setup TOTP form fields

ErikCH:
#### Description of changes
Form fields in the Setup TOTP page is showing a duplicate form field. This fix updates the form fields so it does not display.

#### Issue #, if available

#1599 

Created Cypress Test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


examples/next/pages/ui/components/authenticator/sign-in-totp-mfa/index.page.tsx
```diff
@@ -7,9 +7,13 @@ import awsExports from './aws-exports';
 Amplify.configure(awsExports);
 
 export default function SignInTotpMfa() {
+  const formFields = { setupTOTP: { QR: { totpIssuer: 'My Web App' } } };
```
ErikCH, line 10: Added this extra form fields for the new test.

packages/ui/src/helpers/authenticator/formFields/formFields.ts
```diff
@@ -31,6 +31,7 @@ export const getFormFields = (
   const customFormFields =
     getActorState(state).context?.formFields?.[route] || {};
   const formFields = applyDefaults(defaultFormFields, customFormFields);
+  delete formFields['QR'];
```
ErikCH, line 34: QR isn't a field, so it doesn't need to be in this list. It's only used in the Setup TOTP page, to check for the `totpIssuer` and `totpUsername`. Which this should not affect.

calebpollman: In that case seems like we should prevent the `QR` key from being added to `formFields` altogether?

jacoblogan (approved):
LGTM maybe we can add a follow up item to add a step where we remove all formfields that aren't fields (even if for now it is just QR marking the delete with a note so that if we ever introduce a second one we can break it out then)

packages/ui/src/helpers/authenticator/formFields/formFields.ts
ErikCH: That's a good point. 

The formFields is passed in from the customer. To change it, it would break. https://ui.docs.amplify.aws/components/authenticator#update-setup-totp-qr-issuer-and-username

But maybe something we can add in 3.0?

We could try to delete the QR code earlier on, in the `getSortedFormField`, but I hesitated doing it there, because it would require me to do a deep clone, and that's already done in the `applyDefaults`. Otherwise when you delete it, it deletes it everywhere.

calebpollman: Updating for the 3.0 release makes sense 👍🏽 Do you think it make sense to remove the `QR` key/values from `formFields` as part of the breaking change? I might be misunderstanding the docs, but based on the example it looks like the `QR` field is not actually a form field.

ErikCH: Yeah, I'm thinking maybe we can do that. That would require us to create yet another prop that the Authenticator would accept...

PR #1601, opened 2022-03-29
chore: pin docs and examples react dep to 17

reesscot:
#### Description of changes
This PR updates our react dependency in docs and next-example to 17

#### Checklist

- [x] PR description included


joebuono (approved):
LGTM

0618 (approved):

jacoblogan (approved):

docs/package.json
```diff
@@ -38,9 +38,9 @@
     "next-sitemap": "^1.6.203",
     "prism-react-renderer": "1.2.1",
     "raw-loader": "^4.0.2",
-    "react": "latest",
+    "react": "17",
```
0618, line 41: should we use `^17.0.0`?

reesscot: Isn't that pretty much equivalent to `17`?

0618: I need to look it up 😂

reesscot: Probably wouldn't actually ever want `17.0.0` anyway. With `17` we should always get the latest version of 17, which is what we want for docs.

ErikCH (approved):
👍

docs/package.json
0618: Cool! Verified here https://semver.npmjs.com/ 😁

ErikCH, line 41: Maybe ^17 instead?

PR #1611, opened 2022-03-31
fix(Geocoder): prevent conditional hook calls in Geocoder

calebpollman:
#### Description of changes
Add `StandaloneGeocoder` and `ControlledGeocoder` components to prevent conditional React hook calls

#### Issue #, if available
NA

- ran integ tests
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Geo/Geocoder/index.tsx
+++ b/packages/react/src/components/Geo/Geocoder/index.tsx
@@ -1,6 +1,6 @@
 import maplibregl from 'maplibre-gl';
 import { createAmplifyGeocoder } from 'maplibre-gl-js-amplify';
-import React, { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 import { useControl, useMap } from 'react-map-gl';
 import type { IControl } from 'react-map-gl';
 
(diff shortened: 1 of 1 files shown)


ErikCH (approved):
👍

packages/react/src/components/Geo/Geocoder/index.tsx
```diff
@@ -45,29 +45,37 @@ export const Geocoder = ({
...
+
+  return null;
+};
+
+const StandaloneGeocoder = (props: GeocoderProps) => {
+  const hasMounted = useRef(false);
+
+  useEffect(() => {
+    if (!hasMounted.current) {
+      (createAmplifyGeocoder(props) as unknown as GeocoderControl).addTo(
```
ErikCH, line 70: I've never seen `as unknown as GeocoderControl` . That's interesting way of doing it.

joebuono (approved):
LGTM

packages/react/src/components/Geo/Geocoder/index.tsx
calebpollman: I'm not quite sure why the casting was done this way and due to my lack of familiarity with the feature I chose to do the least impactful change for this PR. In my experience casting to unknown is only required when casting from type A to type B where type A has nothing in common with type B

PR #1612, opened 2022-04-01
Mitigate risk of using `dangerouslySetInnerHTML`

joebuono:
### Potential Security Risk

Within the React package of the Amplify UI codebase, we are using React’s `dangerouslySetInnerHTML` to insert CSS into a `<style>` tag and then inject it into the DOM:
```jsx
<style
  id={`amplify-theme-${name}`}
  dangerouslySetInnerHTML={{ __html: cssText }}
/>
```
We are doing this because we need to dynamically insert CSS into the page. 

The reason this is potential security risk is because a developer could dynamically set a theme based on user input. This could allow an attacker to execute malicious code on the webpage since React’s `dangerouslySetInnerHTML` does not sanitize the HTML. 

However, as mentioned in the internal T Corp ticket: “the risk of an unknown attacker being able to exploit this condition would be highly unlikely because the React DOM escapes everything in JSX before rendering, making XSS almost impossible. The CSS that's being set is in control of the Amplify UI team.”

Additionally, the internal [`createTheme` function](https://github.com/aws-amplify/amplify-ui/blob/9a095ae46ee6639f7c8fe9a3a7b2b871449867b8/packages/ui/src/theme/createTheme.ts) escapes all input to strings and does not execute any of the passed-in code.

### Solution

To mitigate this risk, we are using the [DOMPurify](https://github.com/cure53/DOMPurify) to sanitize the HTML and prevent XSS attacks. DOMPurify is the library that [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#:~:text=You%20must%20regularly%20patch%20DOMPurify%20or%20other%20HTML%20Sanitization%20libraries%20that%20you%20use.%20Browsers%20change%20functionality%20and%20bypasses%20are%20being%20discovered%20regularly.) recommends. Specifically, we are using [isomorphic-dompurify  ](https://www.npmjs.com/package/isomorphic-dompurify) in order to get the e2e and docs tests to pass (by rendering on both the client and server side). 

```jsx
import DOMPurify from 'isomorphic-dompurify';

<style
  id={`amplify-theme-${name}`}
  dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(cssText) }}
/>
```

#### Tradeoffs to using DOMPurify

* Adding DOMPurify as a dependency means that we will have to regularly patch it since browsers change functionality and bypasses are being discovered regularly (see [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#:~:text=You%20must%20regularly%20patch%20DOMPurify%20or%20other%20HTML%20Sanitization%20libraries%20that%20you%20use.%20Browsers%20change%20functionality%20and%20bypasses%20are%20being%20discovered%20regularly)) 
* It will also slightly increase our [bundle size](https://bundlephobia.com/package/dompurify@2.3.6) by up to 19.6kB

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/packages/react/jest.setup.ts
+++ b/packages/react/jest.setup.ts
@@ -1,1 +1,5 @@
 import '@testing-library/jest-dom';
+import { TextDecoder, TextEncoder } from 'util';
+
+global.TextEncoder = TextEncoder;
+global.TextDecoder = TextDecoder;
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -68,7 +68,7 @@
     "@xstate/react": "1.6.3",
     "classnames": "2.3.1",
     "deepmerge": "4.2.2",
-    "dompurify": "2.3.6",
+    "isomorphic-dompurify": "0.18.0",
     "lodash": "4.17.21",
     "qrcode": "1.5.0",
     "react-generate-context": "1.0.1"
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -64,9 +64,11 @@
     "@radix-ui/react-dropdown-menu": "0.1.6",
     "@radix-ui/react-slider": "0.1.4",
     "@radix-ui/react-tabs": "0.1.5",
+    "@types/dompurify": "2.3.3",
     "@xstate/react": "1.6.3",
     "classnames": "2.3.1",
     "deepmerge": "4.2.2",
+    "dompurify": "2.3.6",
     "lodash": "4.17.21",
     "qrcode": "1.5.0",
     "react-generate-context": "1.0.1"
--- a/.changeset/strong-cats-search.md
+++ b/.changeset/strong-cats-search.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+Mitigate risk of `dangerouslySetInnerHTML` by using DOMPurify.sanitize
--- a/packages/react/src/components/AmplifyProvider/__tests__/AmplifyProvider.test.tsx
+++ b/packages/react/src/components/AmplifyProvider/__tests__/AmplifyProvider.test.tsx
@@ -2,6 +2,7 @@ import { render, screen } from '@testing-library/react';
 
 import { AmplifyProvider } from '../index';
 import { Heading } from '../../../primitives';
+import { Theme } from '@aws-amplify/ui-react';
 
 const App = () => {
   return <Heading>Howdy</Heading>;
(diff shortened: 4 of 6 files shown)


packages/react/jest.setup.ts
```diff
@@ -1 +1,5 @@
 import '@testing-library/jest-dom';
+import { TextDecoder, TextEncoder } from 'util';
+
+global.TextEncoder = TextEncoder;
+global.TextDecoder = TextDecoder;
```
joebuono, line 5: This is necessary in order to use isomorphic dompurify: https://github.com/kkomelin/isomorphic-dompurify/issues/91#issuecomment-1012645198

packages/react/package.json
```diff
@@ -64,9 +64,11 @@
     "@radix-ui/react-dropdown-menu": "0.1.6",
     "@radix-ui/react-slider": "0.1.4",
     "@radix-ui/react-tabs": "0.1.5",
+    "@types/dompurify": "2.3.3",
```
joebuono, line 67: This import fixed the following error: https://github.com/rollup/rollup-plugin-commonjs/issues/361

packages/react/package.json
```diff
@@ -64,9 +64,11 @@
...
     "@xstate/react": "1.6.3",
     "classnames": "2.3.1",
     "deepmerge": "4.2.2",
+    "isomorphic-dompurify": "0.18.0",
```
joebuono, line 71: Importing the isomorphic version of dompurify in order to get the e2e and docs tests to pass. Isomorphic dompurify is basically a wrapper around dompurify that allows it to run on the server side as well as the client.

packages/react/package.json
reesscot: This should be in `devDependencies` since it's not a runtime dependency, just a build-time dependency.

packages/react/jest.setup.ts
reesscot: please put this comment in the actual jest.setup file

packages/react/package.json
reesscot: I'm nervous about depending on this package which hasn't been battle tested as much as the original. I'd prefer that we create our own universal isomorphic wrapper around DOMPurify.

joebuono:
Closing this PR because we've found a better solution: https://github.com/aws-amplify/amplify-ui/pull/1622

calebpollman:
Out for JD

calebpollman:
Taking an extended weekend ?

POC for JS Issues
Ashish Nanda : ***@***.******@***.***>

POC for UI Issues
Milan Shah : ***@***.******@***.***>

PR #1616, opened 2022-04-04
update react authenticator styling, remove inline style tags, add aut…

jacoblogan:
This pull request has two goals the first is to remove all of the inline style tags from the react Authenticator, the second is to add and document new css classes and design tokens that will specifically target pieces of the Authenticator.  The goal is to provide a more consistent customization experience with Authenticator as with our primitive components.

authClasses

The styling documentation under the Authenticator page is updated to contain just the classes and css variables that target the Authenticator. The list of all css variables is being removed and might want to be moved to another section (maybe theming or a seperate style guide) to preserve the entire list. 

This allows our customers to customize the Authenticator in a more direct and straightforward manner.
old style rules using data attributes moves to classes `[data-amplify-authenticator]` --> `.amplify-authenticator`

customizing the authenticator by using rules that target primitives within the authenticator move to classes
`[data-amplify-authenticator] .amplify-button[data-variation="primary"]` --> `.amplify-authenticator-sign-in-button`

overriding specific css variables for authenticator at a root level becomes possible
```
:root {
  --amplify-authenticator-sign-in-button-background: red;
}
```

this also opens up the authenticator to being more fully themed through our built in theming system by overriding the newly created authenticator design tokens in the amplify provider
```
const theme = {
  name: 'my-theme',
  tokens: {
    components: {
      authenticator: {
        signIn: { 
          button: {
             background: {
                 value: '{colors.brand.secondary.10.value}'
             }
          }
        },
      },
    },
  },
};
```

This is not a breaking change as the currently used methods for styling the authenticator are going to remain and customers can continue to use them.  Their current rules will only be affected if they are relying on the current specificity of the authenticator rules to override other rules they have specified with lower specificity because the new class rules will be a flat css structure with 010 specificity.  

** This pull request is not ready to be merged as additional classes and design tokens need to be added, this is just to demonstrate the concept and continue the discussion around the topic **

### DIFF
--- a/docs/src/components/ComponentStyleDisplay.tsx
+++ b/docs/src/components/ComponentStyleDisplay.tsx
@@ -6,6 +6,7 @@ import {
   TabItem,
   Flex,
   View,
+  ScrollView,
   Heading,
   useTheme,
 } from '@aws-amplify/ui-react';
@@ -14,7 +15,7 @@ export const ComponentStyleDisplay = ({ componentName }) => {
   const { tokens } = useTheme();
 
   return (
-    <View className="docs-component-styles">
+    <ScrollView className="docs-component-styles" maxHeight="70vh">
       <Flex
         direction={{
           base: 'column',
@@ -41,6 +42,6 @@ export const ComponentStyleDisplay = ({ componentName }) => {
           </Tabs>
         </Flex>
       </Flex>
-    </View>
+    </ScrollView>
   );
 };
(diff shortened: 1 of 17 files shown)


calebpollman (commented):
Thinking the approach looks good! Left some minor feedback 😅

Additionally, think we should move the `ComponentClassNames` values for the `Authenticator` in to their own _constants.ts_ file located somewhere near the `Authenticator` code. Think ideally we want to keep a separation of concerns between UI primitive constants and connected component constants that directly reference the `Authenticator`. What do you think?

PR #1619, opened 2022-04-05
perf(react): optimize `useAuthenticator`s used internally

### LINKED ISSUE #1602: "Error: Maximum update depth exceeded" when attempting to render Authenticator component with React 18's createRoot API

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

### Please describe your bug.

Trying to render the `Authenticator` or `withAuthenticator` components with React 18's `createRoot` API results in errors being thrown.

This newer render API might also be the cause of the Authenticator not rendering properly on newer Next.js (>= v[12.1.1](https://github.com/vercel/next.js/releases/tag/v12.1.1)) apps built using `create-next-app`.

Authenticator components should render without errors

1. `npx create-react-app my-app`
2. `cd my-app`
3. `npm i aws-amplify @aws-amplify/ui-react`
4. copy paste provided code snippets into respective files
5. `npm run start`
6. Authenticator may render for a few seconds, try interacting with it to trigger the errors

### Code Snippet

```js
// index.js

import React from "react";
import * as ReactDOMClient from "react-dom/client";
import App from "./App";
import { Amplify } from "aws-amplify";
import awsmobile from "./aws-exports";

import "./index.css";

Amplify.configure(awsmobile);

const container = document.getElementById("root");

const root = ReactDOMClient.createRoot(container);

root.render(<App />);
```

```js
// App.js

import { withAuthenticator } from "@aws-amplify/ui-react";

const App = () => {
  return <h1>Welcome</h1>;
};

export default withAuthenticator(App);
```

Screen Shot 2022-03-29 at 5 38 22 PM

wlee221:
#### Description of changes

Optimizes `useAuthenticator`s used internally in `Authenticator`. This was causing way too many re-renders, such that it was pushing nested re-renders to >50 which would cause the error described in #1602.

By using `selector` option in `useAuthenticator`, we can guarantee components do not re-render unless context value of interest changes.

#### Improvements

I counted the re-renders in `Router` when I signed in with email + password. It reduced the number of re-renders from 174 to 12.

**Before**

https://user-images.githubusercontent.com/43682783/161739119-34fe2c73-f34f-4394-a392-6c61672a8d64.mp4

**After**

https://user-images.githubusercontent.com/43682783/161739517-079f185a-552d-4024-a864-18f916a9f992.mp4

#### Issue #, if available

Resolves #1602.

Manually tested against React 18 example. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Authenticator/Authenticator.tsx
+++ b/packages/react/src/components/Authenticator/Authenticator.tsx
@@ -22,7 +22,8 @@ export type AuthenticatorProps = AuthenticatorMachineOptions &
 
 // Helper component that sends init event to the parent provider
 function InitMachine({ children, ...data }) {
-  const { _send, route } = useAuthenticator();
+  // TODO: `INIT` event should be removed so that `_send` doesn't need to be extracted
+  const { _send, route } = useAuthenticator((context) => [context.route]);
 
   const hasInitialized = React.useRef(false);
 
--- a/packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
+++ b/packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
@@ -60,6 +60,7 @@ export const VerifyUser = (): JSX.Element => {
     },
   } = useCustomComponents();
 
+  // TODO: expose unverifiedAttributes from `useAuthenticator`
   const { _state, isPending } = useAuthenticator();
   const { handleChange, handleSubmit } = useFormHandlers();
   const context = getActorContext(_state) as SignInContext;
--- a/packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
+++ b/packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
@@ -61,7 +61,9 @@ export const VerifyUser = (): JSX.Element => {
   } = useCustomComponents();
 
   // TODO: expose unverifiedAttributes from `useAuthenticator`
-  const { _state, isPending } = useAuthenticator();
+  const { _state, isPending } = useAuthenticator((context) => [
+    context.isPending,
+  ]);
   const { handleChange, handleSubmit } = useFormHandlers();
   const context = getActorContext(_state) as SignInContext;
 
--- a/packages/react/src/helpers/utils.ts
+++ b/packages/react/src/helpers/utils.ts
@@ -1,8 +1,15 @@
+import isEmpty from 'lodash/isEmpty';
+
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
 
 export const areArraysEqual = (arr1: Array<any>, arr2: Array<any>) => {
   if (arr1.length !== arr2.length) return false;
-  return arr1.every((value, index) => value === arr2[index]);
+  return arr1.every((value, index) => {
+    // edge case: both values are empty object/array. This can often happen with empty spreads.
+    if (isEmpty(value) && isEmpty(arr2[index])) return true;
+
+    return value === arr2[index];
+  });
 };
 
 // Error message handling source:
--- a/packages/react/src/helpers/utils.ts
+++ b/packages/react/src/helpers/utils.ts
@@ -1,14 +1,29 @@
 import isEmpty from 'lodash/isEmpty';
+import isArray from 'lodash/isArray';
+import isObject from 'lodash/isObject';
 
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
 
+const isEmptyObj = (val: any) => isObject(val) && isEmpty(val);
+
+const isEmptyArr = (val: any) => isArray(val) && isEmpty(val);
+
 export const areArraysEqual = (arr1: Array<any>, arr2: Array<any>) => {
   if (arr1.length !== arr2.length) return false;
-  return arr1.every((value, index) => {
-    // edge case: both values are empty object/array. This can often happen with empty spreads.
-    if (isEmpty(value) && isEmpty(arr2[index])) return true;
+  return arr1.every((elem1, index) => {
+    const elem2 = arr2[index];
+    /**
+     * edge cases: if both values are empty object/array, we consider them equal.
+     * These can catch empty default values (`[]`, `{}`) that unintentionally point
+     * to different refernces.
+     *
+     * We can consider doing a deep comparison, but left it here for efficiency
+     * + practicality for authenticator state comparison purposes.
+     */
+    if (isEmptyArr(elem1) && isEmptyArr(elem2)) return true;
+    if (isEmptyObj(elem1) && isEmptyObj(elem2)) return true;
 
-    return value === arr2[index];
+    return elem1 === elem2;
   });
 };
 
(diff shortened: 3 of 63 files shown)


packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -22,7 +22,8 @@ export type AuthenticatorProps = AuthenticatorMachineOptions &
 
 // Helper component that sends init event to the parent provider
 function InitMachine({ children, ...data }) {
-  const { _send, route } = useAuthenticator();
+  // TODO: `INIT` event should be removed so that `_send` doesn't need to be extracted
+  const { _send, route } = useAuthenticator((context) => [context.route]);
```
ayush987goyal, line 26: Just a minor callout that xstate docs recommends to define the selector functions outside out the component definition to optimize further. [Ref](https://xstate.js.org/docs/packages/xstate-react/#api) 

> // tip: optimize selectors by defining them externally when possible

wlee221: Thanks for the callout! I think it's referring to this `selector` we pass to `useSelector`:

https://github.com/aws-amplify/amplify-ui/blob/e3f1186df5bb936db14c0897b98e12cc3983aea1/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx#L124

I had it inside the component because it needed a references to a `React.useRef`, but will keep this optimization in mind for our 3.0 changes.

packages/react/src/helpers/utils.ts
```diff
@@ -1,8 +1,30 @@
...
+  return arr1.every((elem1, index) => {
+    const elem2 = arr2[index];
+    /**
+     * edge cases: if both values are empty object/array, we consider them equal.
+     * These can catch empty default values (`[]`, `{}`) that unintentionally point
+     * to different refernces.
+     *
+     * We can consider doing a deep comparison, but left it here for efficiency
+     * + practicality for authenticator state comparison purposes.
+     */
```
wlee221, line 22: This optimization is particularly targeted against comparing default empty values, like https://github.com/aws-amplify/amplify-ui/blob/e20720a894ccb2cfcc7ede7160299e082ec76fd2/packages/ui/src/helpers/authenticator/facade.ts#L59

which will always create a **new** empty object even if `validationError` doesn't exist, and thus would fail shallow equality check. This would cause a re-render every single state change! Adding these empty arr/obj checks will optimize these cases.

wlee221: Alternatively, we could do a deep value comparison with something like `lodash/isEqual`. But this is expensive for large objects like `user`, so I left it at shallow comparison + treating empty array/object as equal.

packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
```diff
@@ -60,6 +60,7 @@ export const VerifyUser = (): JSX.Element => {
     },
   } = useCustomComponents();
 
+  // TODO: expose unverifiedAttributes from `useAuthenticator`
   const { _state, isPending } = useAuthenticator();
```
ErikCH, line 64: So in situations where you are pulling out the `_state` there is no context to be providedd? I noticed the same in the `FormFields.tsx` file too.

packages/react/src/helpers/utils.ts
ErikCH: 👍

ErikCH (approved):
Just a couple of questions, but looks good.

packages/react/src/helpers/utils.ts
```diff
@@ -1,8 +1,30 @@
...
+import isArray from 'lodash/isArray';
+import isObject from 'lodash/isObject';
+
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
 
+const isEmptyObj = (val: any) => isObject(val) && isEmpty(val);
+
+const isEmptyArr = (val: any) => isArray(val) && isEmpty(val);
+
 export const areArraysEqual = (arr1: Array<any>, arr2: Array<any>) => {
```
calebpollman, line 11: We **may** want to rename this util since two arrays that are empty are not equal due to identity

packages/react/src/helpers/utils.ts
```diff
@@ -1,8 +1,30 @@
+import isEmpty from 'lodash/isEmpty';
+import isArray from 'lodash/isArray';
+import isObject from 'lodash/isObject';
+
 export const isDevelopment = () => process.env.NODE_ENV !== 'production';
 
+const isEmptyObj = (val: any) => isObject(val) && isEmpty(val);
```
calebpollman, line 7: Can we narrow the argument of `val` here (same with `isEmptyArr`)?

packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
calebpollman: Should we optimize for `isPending` here or would that prevent us from picking up changes in `_state`?

calebpollman (approved):
LGTM, just left some minor questions and a non-blocking piece of feedback. Thanks for working through this

packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
wlee221: @ErikCH Yep, `unverifiedAttributes` is not currently provided from our facade. We should to make codes simpler, but I'll do that in a separate PR. (And with conversations about refactoring facade values, we could use a discussion before that).

wlee221: @calebpollman you're right! I forgot to optimize this part.

I was initially torn here because `useAuthenticator((context) => [context.isPending])` wouldn't satisfy exhausive deps. But considering it's internal and that we're in so much re-rendering hell, we should make this optimization.

packages/react/src/helpers/utils.ts
wlee221: I intentially had it as `any` here, because we don't know the type the context `selector` has chosen. But `isObject`, etc provides TypeScript type guards, so this is type-safe.

packages/react/src/helpers/utils.ts
wlee221: Changing it to `AreArrayValuesEqual` as per offline discussion. Technically we don't do deep equality, but this conveys the point that we're interested in the values!

wlee221: Done here! [ac5153d](https://github.com/aws-amplify/amplify-ui/pull/1619/commits/ac5153db3fb572ffa761f07f70280a0ca056e896)

packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
wlee221: Done here cabac4e3d16f57610fdc5d9431c871fc07d904de

wlee221: Oops wrong commit -- it should be 8fa81743bfbd41942fbc2b290b519ed51f5c2ce5

PR #1620, opened 2022-04-05
fix(react): Keep a local copy of previous facade value 

wlee221:
#### Description of changes

`useAuthenticator` did not optimize re-renders as expected when it selects an actor context. For example,

```tsx
const { error } = useAuthenticator((context) => [context.error])
```

would always re-render, even if `error` hasn't changed. This was because `prevState` and `nextState` we use in comparators holds a *reference* to actor contexts, which could easily change between each compare calls. 

This PR resolves it by explicitly holding onto actor snapshot values that wouldn't change between compares. More details on code comments.

#### Issue #, if available

Related to #1602

E2e test, + some manual validation on `examples/next`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -97,9 +97,16 @@ export const useAuthenticator = (selector?: Selector) => {
     [service]
   );
 
-  const getFacade = (state: AuthMachineState) => {
-    return { ...sendAliases, ...getServiceContextFacade(state) };
-  };
+  const getFacade = (state: AuthMachineState) => ({
+    ...sendAliases,
+    ...getServiceContextFacade(state),
+    /** @deprecated For internal use only */
+    _state: state,
+    /** @deprecated For internal use only */
+    _send: send,
+  });
+
+  const prevFacadeRef = React.useRef<ReturnType<typeof getFacade>>();
 
   /**
    * For `useSelector`'s selector argument, we just return back the `state`.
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -125,9 +137,17 @@ export const useAuthenticator = (selector?: Selector) => {
      * We only trigger re-render if any of values in specified selected
      * values change. First compute the facade for prev and next state.
      */
-    const prevFacade = getFacade(prevState);
+    const prevFacade = prevFacadeRef.current;
```
ErikCH, line 140: So the previous state is just a ref to the current state? Since the next state keeps changing?

ErikCH (approved):
Looks good, just a quick question earlier.

calebpollman (approved):
LGTM 🚢

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: On each state transition, this comparator function gets called with `prevState` and `nextState`. `prevState` is the previous authenticator machine xstate state values. It'll look something like this:

```tsx
{
  value: 'signIn.runActor',
  context: {
    user: {...},
    actorRef: {...},
  },
   ...
}
```

While static / non-ref context values like `prevState.value` will be compared fine, static references like `actorRef` would always be considered equal even if its value in the ref has changed. That led to this code change.

reesscot (approved):
LGTM, just one non-blocking comment

PR #1622, opened 2022-04-05
Server side XSS prevention

joebuono:
#### Description of changes

This PR addresses our use of React's `dangerouslySetInnerHTML`. Comments are included in the code itself so that our thought process is preserved for future reference. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


wlee221 (approved):
I appreciate the comments you've put to this. Thanks for deep diving!

packages/react/src/components/AmplifyProvider/__tests__/AmplifyProvider.test.tsx
```diff
@@ -37,4 +38,48 @@ describe('AmplifyProvider', () => {
...
+      </AmplifyProvider>
+    );
+
+    const styleTag = container.querySelector(`#amplify-theme-${name}`);
+    expect(styleTag).toBe(null);
+  });
+
+  it('filters out XSS attacks which attempt to escape the CSS context using whitespace characters', async () => {
+    const name = 'maliciousTheme';
+    const xss = `</style\r ><script>alert('XSS')</script>`;
```
reesscot, line 66: Can we add another test like this?( backticks preserve carriage returns)
```suggestion
    const xss = `</style
    ><script>alert('XSS')</script>`;
```

ErikCH (approved):
👍

zchenwei (commented):
I was wondering if we would like to do the filtering on token level by just filtering out the malicious value containing closing style tag etc. versus removing the entire css text?

zchenwei (approved):

reesscot (approved):
LGTM

robrich:
This is a naïve solution. See https://stackoverflow.com/a/482088/702931 If you can assume the server won't allow any CSS execution like `with()` or `exec()` or similar, maybe, but now I need only HTML-encode it.

reesscot:
> This is a naïve solution. See https://stackoverflow.com/a/482088/702931 If you can assume the server won't allow any CSS execution like `with()` or `exec()` or similar, maybe, but now I need only HTML-encode it.

@robrich Can you provide an example? The [stackoverflow](https://stackoverflow.com/a/482088/702931) that you linked to references IE10 which is retired, and XBL with Firefox. XBL has been [removed from Firefox](https://groups.google.com/g/firefox-dev/c/3UBCgl3hjQE/m/HHZy28pKCAAJ) since 2019. Are there current browsers with these vulnerabilities?

robrich:
I googled it fast and found https://codepen.io/3n3a/pen/RwGKwrX and https://www.invicti.com/blog/web-security/private-data-stolen-exploiting-css-injection/  I could also trick the solution by including `&lt;/st&#121;le&gt;` in the string.

dbanksdesign:
I tried that trick and it didn't work: https://stackblitz.com/edit/nextjs-pcwxmh?file=pages%2F_app.js

PR #1625, opened 2022-04-05
fix(ui): fix references in base theme

dbanksdesign:
#### Description of changes

* Updated some theme references that were failing and added a test to make sure it won't happen in the future
* Updated style dictionary version
* Changing `defaultTheme` to be of type `BaseTheme` which makes more sense and is what people would use. 

tests pass and builds pass

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/ui/src/theme/index.ts
```diff
@@ -1,7 +1,7 @@
 import { createTheme } from './createTheme';
 import { baseTheme } from './baseTheme';
 
-export { createTheme };
+export { createTheme, baseTheme };
```
zchenwei, line 4: `baseTheme` has a confusing name `default-theme` 🤣

packages/ui/src/theme/tokens/components/copy.ts
```diff
@@ -1,6 +1,6 @@
 export const copy = {
-  fontSize: { value: '{font.sizes.xs.value}' },
-  gap: { value: '{space.relative.medium.value}' },
+  fontSize: { value: '{fontSizes.xs}' },
```
zchenwei, line 2: Why `.value` could be skipped here? It seems we always need that when referencing other tokens value

dbanksdesign: Correct, `.value` can be skipped now

jacoblogan (approved):
LGTM

packages/ui/src/theme/tokens/components/copy.ts
reesscot: +1 to this question, thought I was just missing something

reesscot: Do we need to do this on all the other tokens too, or is it an optional change?

packages/ui/src/theme/__tests__/baseTheme.test.ts
```diff
@@ -0,0 +1,21 @@
+import { baseTheme } from '../baseTheme';
+import StyleDictionary from 'style-dictionary';
+
+describe('@aws-amplify/ui', () => {
+  describe('baseTheme', () => {
+    it('all references should resolve', () => {
+      expect(() => {
+        StyleDictionary.extend({
+          // This is to make typescript happy, the types don't jive well
+          tokens: Object.assign({ ...baseTheme.tokens }),
```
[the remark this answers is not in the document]
zchenwei: Agree

packages/ui/src/theme/tokens/components/copy.ts
dbanksdesign: It is optional, we could do a PR to remove it for all the other tokens because it is cleaner

reesscot (approved):

PR #1630, opened 2022-04-06
chore: updating dependencies to ESM equivalents

zchenwei:
#### Description of changes

1. Updating `@aws-amplify/ui-react` to use `lodash-es`, which is the ESM equivalent of `lodash` for better tree shaking
2.  `@aws-amplify/ui` will stick with `lodash`. `lodash-es` has issues with `ts-node` since it is defined as ESM package explicitly with `type: module` in its package.json. Also, had a discussion with @calebpollman. It is consumed by `ui-react-native` and React Native does not support tree shaking. I don't see there is much benefit to make the update. As long as we import each method from individual module instead of the whole lodash library, bundle size is still minimized.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201951124298517


zchenwei:
`lodash-es` package has declared `type: module` which causes issues with jest and ts-node. By default, all `.js` files will be treated as commonjs module, a commonjs module cannot use `require` statement to load a ESM file.

PR #1638, opened 2022-04-06
chore: updating import paths to reduce circular dependencies

zchenwei:
#### Description of changes

There is quite amount of circular dependencies due to implicit import paths. They will affect tree shaking negatively. A good example is the newly added maplibre cannot be tree shaken correctly for Authenticator users.

***Example code with @aws-amplify/ui-react 2.15.0***

```jsx
import { Authenticator } from '@aws-amplify/ui-react';

function App() {
  return (
    <Authenticator>
      {({ signOut, user }) => (
        <main>
          <h1>Hello {user.username}</h1>
          <button onClick={signOut}>Sign out</button>
        </main>
      )}
    </Authenticator>
  );
}

export default App;
```

| Before     | After |
| ----------- | ----------- |
| Screen Shot 2022-04-07 at 12 36 17 PM | Screen Shot 2022-04-07 at 1 28 06 PM |

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201951124298521


packages/react/src/components/Authenticator/hooks/useFormHandlers/useFormHandlers.ts
```diff
@@ -3,7 +3,7 @@ import { useCallback } from 'react';
 import { getFormDataFromEvent } from '../../../../helpers/utils';
 import { useAuthenticator } from '../useAuthenticator';
 
-export default function useFormHandlers() {
+export function useFormHandlers() {
```
calebpollman, line 6: Curious why we need to update this from a `default` export?

packages/react/src/components/Authenticator/shared/RemoteErrorMessage.tsx
```diff
@@ -1,5 +1,5 @@
-import { useAuthenticator } from '..';
-import { Alert } from '../../..';
+import { useAuthenticator } from '../hooks/useAuthenticator';
+import { Alert } from '../../../primitives/Alert';
```
calebpollman, line 2: Is there a reason to prefer the path of `'../../../primitives/Alert'` over `'../../../primitives'` here? Asking since we are using `'../../../primitives'` in the files above, but I'm guessing it is because we are only using the `Alert` component in this file

calebpollman (commented):
Thanks for doing this @zchenwei Left some questions

packages/react/src/components/Authenticator/shared/RemoteErrorMessage.tsx
zchenwei: Yes, that's true. There should be no difference since we have defined `sideEffects` saying each primitive module is side effect free. With that, doing `'../../../primitives/xxx'` is always encouraged because importing from individual module can help skip unused modules when a bundler is doing code analysis even without providing hints via `sideEffect`

packages/react/src/components/Authenticator/hooks/useFormHandlers/useFormHandlers.ts
zchenwei: Just for consistency, the other hooks are using named export. We prefer named export because it allows us to shared the exact same name.

packages/react/src/components/Authenticator/shared/RemoteErrorMessage.tsx
calebpollman: Should we break up the grouped import statements and align the rest of the imports paths in these files for consistency? e.g.
```
import { View } from '../../../primitives/View';
import { Flex } from '../../../primitives/Flex';
```

zchenwei: For sure, I will make the updates

ErikCH (approved):
👍

dbanksdesign (approved):
While the code isn't a nice, the customer benefit is much better! LGTM!

two thumbs. up

0618 (approved):
LGTM, Thanks! 
just 2 questions:
1. Do we also want to make this change for other code outside authenticator?
2. do we need to document this somewhere, like the code standard, or you think we can just follow the code itself?

zchenwei:
> LGTM, Thanks! just 2 questions:
> 
> 1. Do we also want to make this change for other code outside authenticator?
> 2. do we need to document this somewhere, like the code standard, or you think we can just follow the code itself?

I only found the circular dependencies under Authenticator folder lol. So feel free to change it whenever we find some.

We do have a code standard doc but i don't think people reference it before coding :(

PR #1654, opened 2022-04-08
Add `nonce` prop to AmplifyProvider

joebuono:
#### Description of changes

This PR exposes the capability to add a [nonce](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) to the`<style>` tag rendered by the `AmplifyProvider`:

Example:
`<AmplifyProvider nonce="randomString"/> => <style nonce="randomString">`

This will allow customers to use strict CSP rules while also using a custom theme. See the [example below](https://content-security-policy.com/examples/allow-inline-style/):

Allow Inline Styles using a Nonce

I spent quite a bit of time testing this out on the UI Docs site and NextJS example app, as well as in different browsers. I can confirm that:
- If there is a nonce in headers and it matches the nonce passed to the `<style>` tag, then those styles will get applied to the page. 
- If there is a nonce in headers but it does not match the nonce passed to the `style` tag (or if there is no nonce on the `style` tag), then the browser will log a console error and the styles will not get applied. 
- If there is not a nonce in the headers, then all styles will get applied. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/.changeset/wild-comics-carry.md
+++ b/.changeset/wild-comics-carry.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+Add `nonce` prop to AmplifyProvider to allow strict CSP rules
--- a/packages/react/src/components/AmplifyProvider/index.tsx
+++ b/packages/react/src/components/AmplifyProvider/index.tsx
@@ -10,12 +10,14 @@ interface AmplifyProviderProps {
   children: React.ReactNode;
   colorMode?: ColorMode;
   theme?: Theme;
+  nonce?: string;
 }
 
 export function AmplifyProvider({
   children,
   colorMode,
   theme,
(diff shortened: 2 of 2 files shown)


zchenwei (approved):
LGTM!

0618 (commented):
Do we also need to update our docs and examples so that users know to use it?

joebuono:
@0618 Yes, there is a separate Asana task to update the docs once the other security tasks are also completed.

0618 (approved):

vladpoltorin1:
hey @joebuono,
does adding static value make any sense in terms of security? The value will be the same for every request within one build..
Or is it just to make CSP work with amplify-ui and get rid of `unsafe-inline`?

PR #1655, opened 2022-04-08
fix: accessibly hide label text for CheckboxField when labelHidden is…

hbuchel:
#### Description of changes

This updates the Checkbox label to use the visuallyHidden class (ComponentClassNames.VisuallyHidden) to hide the label Text when the labelHidden prop = true. Currently, when true, it removes the label text from the DOM completely making it inaccessible.

Using the docs example, we can compare the DOM before and after.
```
<CheckboxField label="Subscribe" name="subscribe" value="yes" labelHidden={true} />
```

**Before**

Screen Shot 2022-04-08 at 2 13 06 PM

**After**
Screen Shot 2022-04-08 at 2 16 00 PM

[VoiceOver video](https://user-images.githubusercontent.com/376920/162500869-9a2013bb-5292-4295-a8cf-ee4110a7d620.mp4)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/.changeset/lucky-cooks-shave.md
+++ b/.changeset/lucky-cooks-shave.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+fix: accessibly hide label text for CheckboxField when labelHidden is…
--- a/packages/react/src/primitives/Checkbox/Checkbox.tsx
+++ b/packages/react/src/primitives/Checkbox/Checkbox.tsx
@@ -74,10 +74,12 @@ const CheckboxPrimitive: Primitive<CheckboxProps, 'input'> = (
           {...rest}
         />
       </VisuallyHidden>
-      {label && !labelHidden && (
+      {label && (
         <Text
           as="span"
-          className={ComponentClassNames.CheckboxLabel}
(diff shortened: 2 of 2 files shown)


0618 (approved):

dbanksdesign (approved):
LGTM! thanks for fixing this!

thank you

zchenwei (approved):
Great! Regarding the way we use `classNames` to conditionally add a class. I think we should decide on a format to be consistent across the code base. @jacoblogan is using it in a different way, for example <- https://github.com/aws-amplify/amplify-ui/pull/1435/files#diff-e0aa256921de3c08bb1625deaf417b89742da7e030663a394a2bc7065bc00d4eR91

hbuchel:
> Great! Regarding the way we use `classNames` to conditionally add a class, I think we should decide on a format to be consistent across the code base. @jacoblogan is using it in a different way, for example <- https://github.com/aws-amplify/amplify-ui/pull/1435/files#diff-e0aa256921de3c08bb1625deaf417b89742da7e030663a394a2bc7065bc00d4eR91

Good point, I took this format from another component I saw (Label), but funnily in the past I've typically wrote them the way you linked there.

Issue #1651, opened 2022-04-08
Error "Cannot read properties of undefined (reading 'send')" when calling useAuthenticator


techandmedia:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Next.js

### Please describe your bug.

```code
TypeError: Cannot read properties of undefined (reading 'send')

This error happened while generating the page. Any console logs will be displayed in the terminal window.
```

### What's the expected behaviour?

No error when calling useAuthenticator

### Help us reproduce the bug!

Error occured when using React v17 and v18

https://github.com/techandmedia/next-amplify-react18/tree/react18-useAuthenticator

### Code Snippet

```typescript
// Put your code below this line.
import { useEffect } from "react";
import { Amplify } from "aws-amplify";
import { Authenticator, useAuthenticator } from "@aws-amplify/ui-react";
import "@aws-amplify/ui-react/styles.css";
import awsExports from "../../aws-exports";

Amplify.configure({ ...awsExports, ssr: true });

export default function MyApp({ Component, pageProps }) {
  const { user, signOut } = useAuthenticator((context) => [context.user]);

  useEffect(() => {
    console.log("App", user);
    console.log("App", signOut);
  }, [user, signOut]);

  return (
    <main>
      <h1>Hello {user.username}</h1>
      <button onClick={signOut}>Sign out</button>
      <Component {...pageProps} />
    </main>
    // <Authenticator>
    //   {({ signOut, user }) => (
    //     <main>
    //       <h1>Hello {user.username}</h1>
    //       <button onClick={signOut}>Sign out</button>
    //       <Component {...pageProps} />
    //     </main>
    //   )}
    // </Authenticator>
  );
}

```

### Additional information and screenshots

Error occured when using React v17 and v18

```json
{
  "private": true,
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start"
  },
  "dependencies": {
    "@aws-amplify/ui-react": "^2.14.1",
    "aws-amplify": "^4.3.19",
[...]

PR #1657, opened 2022-04-08
chore: Add error message for no provider found

wlee221:
#### Description of changes

Whenever customer are trying to use `useAuthenticator` without wrapping their App with `Authenticator.Provider`, they'll get the cryptic error message 

```
TypeError: Cannot read properties of undefined (reading 'send')
```

This PR adds a check for that and adds a more dev-friendly message.

#### Issue #, if available

#1651: dev confusion with errors when no provider is added/

No runtime behavior changes unless the app is already on faulty configuration. I might have missed edge cases like `Authenticator.Provider` not being ready by the time `useAuthenticator` first runs, but e2e test should catch that.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -89,6 +89,15 @@ export type Selector = (context: AuthenticatorContext) => Array<any>;
 
 export const useAuthenticator = (selector?: Selector) => {
   const { service } = React.useContext(AuthenticatorContext);
+
+  if (!service) {
+    console.error(
+      'No `Authenticator.Provider` was found above where this hook is being used. ' +
+        'Please ensure you wrap your App with `Authenticator.Provider` like so: ' +
+        'https://ui.docs.amplify.aws/components/authenticator#useauthenticator-hook'
+    );
+  }
+
   const send = service.send;
 
   // send aliases are static and thus can be memoized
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -89,6 +89,16 @@ export type Selector = (context: AuthenticatorContext) => Array<any>;
...
 export const useAuthenticator = (selector?: Selector) => {
   const { service } = React.useContext(AuthenticatorContext);
+
+  if (!service) {
+    console.error(
+      'No `Authenticator.Provider` was found above where `useAuthenticator` hook is. ' +
+        'Please ensure you wrap your App with `Authenticator.Provider` like so: ' +
+        'https://ui.docs.amplify.aws/components/authenticator#useauthenticator-hook'
+    );
+    return null;
```
wlee221, line 99: Wasn't sure whether we should `throw` or just return `null` here. 

I went with `null` to avoid breaking customer apps unexpectedly. But on the other hand, it also makes sense to `throw` because not having `Provider` will break their app anyway. Up to discussion!

0618 (approved):

joebuono (approved):

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
ErikCH: 👍

PR #1662, opened 2022-04-11
Allow React 17 and 18 as peer deps

jonasws:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -77,8 +77,8 @@
   },
   "peerDependencies": {
     "aws-amplify": "3.x.x || 4.x.x",
-    "react": ">= 16.8.0",
-    "react-dom": ">= 16.8.0"
+    "react": ">= 16.8.0 || ^17 || ^18",
+    "react-dom": ">= 16.8.0 || ^17 || ^18"
   },
   "peerDependenciesMeta": {
     "aws-amplify": {


packages/react/package.json
```diff
@@ -77,8 +77,8 @@
   },
   "peerDependencies": {
     "aws-amplify": "3.x.x || 4.x.x",
-    "react": ">= 16.8.0",
-    "react-dom": ">= 16.8.0"
+    "react": ">= 16.8.0 || ^17 || ^18",
```
zchenwei, line 80: Shouldn't `>=` allow 17 and 18?

reesscot: +1. @jonasws Was there a particular error message you were getting that prompted this PR?

jonasws: No, [in the newer versions of npm the peer dependency resolution changed, and requires us to define these explicitly](https://docs.npmjs.com/cli/v8/commands/npm-install).  Since React 18 is now considered stable, it makes sense to allow this version when using Amplify.

reesscot: @jonasws As far as I know `>= 16.8.0` will allow both versions `17` and `18` already since they are both greater than 16, so there shouldn't be a conflict even in newer versions of npm. 

We haven't heard of anyone having trouble installing `@aws-amplify/ui-react` alongside React 17 or 18 using newer versions of npm, but if you are experiencing errors, we'd be interested in learning more.

jonasws: It may be that my npm installation has simply been misconfigured and wrongfully reporting Amplify to be a "culprit". In that case, my apologies! I've tried to ship in as a happy consumer (as I've done for some other libraries that was not using the `>=` operator as Amplify claerly is.

Keep up the good work!

reesscot: @jonasws +1 to NPM error messages being confusing. It looks like Radix is in the process of updating their peerDependencies to allow React 18: https://github.com/radix-ui/primitives/pull/1329

If this is blocking you, you may need to try the `--legacy-peer-deps` flag in the meantime.

reesscot:
Closing for now, as this was an issue with Radix UI, which is being worked on here:
https://github.com/radix-ui/primitives/pull/1329

benoitgrelard:
Hey @reesscot, it's fantastic to see Amazon using Radix for Amplify UI!
We'd love it if you could participate in our case studies here: https://www.radix-ui.com/case-studies
Let me know if you'd be interested 🙏

PR #1666, opened 2022-04-11
chore(ui-react): updates to allow eslint react-hooks/rules-of-hooks set to error

calebpollman:
#### Issue #, if available

- update eslint `'rules-of-hooks/exhaustive-deps'` from `'warning'` to `'error'`
- turn linting on for entire `ui-react` package
- add missing or update dep arrays as needed to satisfy `'rules-of-hooks/exhaustive-deps'`
- and `DefaultFooter` component in `SignIn`
- add `DefaultHeader` component in `ConfirmSignUp`
- add `DefaultComponents` interface and remove `PartialDeep` and related files
- add `useAuthenticatorService` and error throw in `useAuthenticator` to prevent conditional hook calls
- ~~refactor `useDataStoreCollection`, `useDataStoreBinding`, and `useDataStoreItem` hooks to call `useDataStoreResult` and add some additional unit tests~~ (moving to own PR)

- ran unit and e2e tests
- smoke tested error throwing for un-wrapped `useAuthenticator`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -87,24 +93,27 @@ export type InternalAuthenticatorContext = {
...
 
-export const useAuthenticator = (selector?: Selector) => {
+const useAuthenticatorService = () => {
   const { service } = React.useContext(AuthenticatorContext);
 
   if (!service) {
-    console.error(
-      'No `Authenticator.Provider` was found above where `useAuthenticator` hook is. ' +
-        'Please ensure you wrap your App with `Authenticator.Provider` like so: ' +
-        'https://ui.docs.amplify.aws/components/authenticator#useauthenticator-hook'
```
calebpollman, line 97: Removed as url is fragile, and the `Provider`/`Context` pattern in React is well documented, but open to putting it back

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -52,9 +52,15 @@ export const Provider = ({ children }) => {
...
+  const isListening = React.useRef(false);
   React.useEffect(() => {
+    if (isListening.current) {
+      return;
+    }
+
+    isListening.current = true;
     return listenToAuthHub(send);
-  }, []);
+  }, [send]);
```
ErikCH, line 63: Is this because of the exhaustive deps eslint rule? So it only runs once?

calebpollman: That's correct.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
ErikCH: This is from this PR. https://github.com/aws-amplify/amplify-ui/pull/1657

From this issue https://github.com/aws-amplify/amplify-ui/issues/1651.

It might be worth keeping to help customers from cryptic error messages.

calebpollman: I only removed the url since it can easily become out of sync with the docs site or is that your concern?

wlee221: I'm open to either ways. I added a link to docs because it was a major confusion point.

Since `#useauthenticator-hook` part is fragile, I'm wondering if we could still word it like

> Please ensure you wrap your App with `Authenticator.Provider`. See `useAuthenticator` section on https://ui.docs.amplify.aws/components/authenticator

The url without hash will be stable.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -87,24 +93,27 @@ export type InternalAuthenticatorContext = {
...
-    return null;
   }
 
+  return service;
+};
+
+export const useAuthenticator = (selector?: Selector) => {
+  const service = useAuthenticatorService();
+
   const send = service.send;
```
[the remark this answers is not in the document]
wlee221: FYI, I was considering even `throw`ing, but decided not to: https://github.com/aws-amplify/amplify-ui/pull/1657#discussion_r846453427. Open to thoughts!

wlee221 (commented):
Looking good! Adding comments for Authenticator changes. I'll err on @reesscot for changes in non-authenticator hooks.

packages/react/src/hooks/useDataStore.tsx
```diff
@@ -11,46 +11,103 @@ import {
...
-      return () => subscription.unsubscribe();
+      // Unsubscribe from query updates on unmount
+      if (subscription) {
+        return () => {
+          subscription.unsubscribe();
+        };
+      }
     }
-  };
+  }, [isRecord, props]);
```
zchenwei, line 87: I like the solution of using `useRef` to prevent additional datastore calls. I think the infinite `useEffect` loop will arise again. People will normally pass in a new object `useDataStoreBinding({})` when they using the hook.

calebpollman: Maybe I'm missing context here but not sure what the concern fully is, a `useEffect` with an empty dep array still gets called on each render, it just doesn't execute its body (unless I am mistaken)

zchenwei: That's true from js perspective. I don't know how React emit this errors internally but this will arise when dependencies are changing constantly. They use a term `prevent re-running` on `useEffect` to explain the effect passing an empty array. It is close to the familiar `componentDidMount` and `componentWillUnmount` mental model.

calebpollman: A potential solution would be to assign props to a ref which would allow us to remove it from the dep array, which would then prevent the useEffect from running unless isRecord updates. Thoughts?

zchenwei: I like it. That's a way recommended by React actually. <- https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
calebpollman: @wlee221 [Updated](https://github.com/aws-amplify/amplify-ui/pull/1666/commits/474f7e540dd398abbab34beea33db478ba512519)

ErikCH (approved):
👍

wlee221 (approved):

.changeset/purple-flies-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+chore(ui-react): updates to allow eslint react-hooks/rules-of-hooks set to error
```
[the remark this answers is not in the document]
calebpollman: Done 👍

packages/react/src/hooks/useDataStore.tsx
calebpollman: Per offline discussion moving the refactoring of `useDataStoreCollection` and `useDataStoreItem` to their own PR

zchenwei (approved):
LGTM :shipit:

PR #1667, opened 2022-04-11
fix: update focus styling for switchField 

hbuchel:
#### Description of changes

This updates the focus style for the switchField to be more visually distinct and to match that of the checkbox field. It also makes some adjustments to the state names for the tokens to use an _ prefix, and updates related tests.

~~FYI, consulted with @dbanksdesign and we don't currently have a concrete pattern for defining our tokens in regards to variations and state, or how they should be nested, but suggested:~~

```
component[variant][state][child]
```
~~This would allow us to group state styles together in a file, instead of potentially having them scattered throughout a file. This way, you could quickly identify all of the default focus styles, or those associated with specific variants.~~

**NOTE: I removed the changes to token nesting order. After some offline discussion, we were concerned this could create potential breaking changes (i.e. css custom properties would have new names) if consumers were relying on tokens or css custom properties to re-theme their app.** Let's save these refactors for our next major version.

Additionally, the preferred naming convention for states like hover or focus should be
```
_state: {}
```
I created an additional task in our Docs project to define/document these conventions on the Theming tab.

**Before (focused)**

Screen Shot 2022-04-11 at 5 42 13 PM

Screen Shot 2022-04-11 at 6 15 45 PM

**After (focused)**

Screen Shot 2022-04-11 at 5 42 02 PM

Screen Shot 2022-04-11 at 6 17 51 PM

#### Issue #, if available

N/A

Checked changes in local docs instance and in local next-example instance

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


0618 (approved):

slaymance (dismissed):
Looks good! For my understanding, what's the rationale for prepending `_` to the state properties?

hbuchel:
> Looks good! For my understanding, what's the rationale for prepending `_` to the state properties?

This is to more readily differentiate a state (focus, disabled, loading) from a variation name (small, medium, large). We have quite a few components following this convention already, though I've found a couple here and there that aren't.

packages/ui/src/theme/tokens/components/switchField.ts
```diff
@@ -36,14 +17,45 @@ export const switchfield = {
...
+          offsetX: '0px',
+          offsetY: '0px',
+          blurRadius: '0px',
+          spreadRadius: '2px',
+          color: '{colors.border.focus.value}',
+        },
+      },
+    },
+  },
+  _checked: {
```
hbuchel, line 45: FYI, pushed a commit that updates 'checked' to '_checked' to follow our _state guidance, and so I can use this as an example in docs for token definitions :)

packages/ui/src/theme/__tests__/defaultTheme.test.ts
```diff
@@ -528,24 +528,24 @@ describe('@aws-amplify/ui', () => {
...
         --amplify-components-switchfield-label-padding: var(--amplify-space-xs);
         --amplify-components-switchfield-thumb-background-color: var(--amplify-colors-background-primary);
         --amplify-components-switchfield-thumb-border-color: var(--amplify-colors-border-tertiary);
         --amplify-components-switchfield-thumb-border-radius: var(--amplify-radii-xxxl);
-        --amplify-components-switchfield-thumb-checked-transform: var(--amplify-transforms-slide-x-medium);
         --amplify-components-switchfield-thumb-transition-duration: var(--amplify-time-medium);
         --amplify-components-switchfield-thumb-width: var(--amplify-space-relative-medium);
         --amplify-components-switchfield-track-background-color: var(--amplify-colors-background-tertiary);
         --amplify-components-switchfield-track-border-radius: var(--amplify-radii-xxxl);
-        --amplify-components-switchfield-track-checked-background-color: var(--amplify-colors-brand-primary-60);
```
reesscot, line 544: We need to be careful of breaking changes like this. If there are customers using this variable SwitchField styling will break. Can we link this one to the new variable for backwards compatibility?

reesscot: Same comment for these as well:
```
--amplify-components-switchfield-thumb-checked-transform
--amplify-components-switchfield-focused-shadow:
```

packages/ui/src/theme/tokens/components/switchField.ts
```diff
@@ -109,14 +106,45 @@ export const switchfield: SwitchFieldTokens = {
   track: {
     backgroundColor: { value: '{colors.background.tertiary.value}' },
     borderRadius: { value: '{radii.xxxl.value}' },
-    checked: {
-      backgroundColor: { value: '{colors.brand.primary.60.value}' },
```
reesscot, line 113: Don't we need to leave this for backwards compatibility?

reesscot (commented):
This is great, thank you so much for fixing these. I just think we need to link the deprecated CSS variables back to the new ones to prevent breaking customers.

dbanksdesign (dismissed):
LGTM! :shipit:

packages/ui/src/theme/tokens/components/switchField.ts
hbuchel: Updated to revert these breaking token name changes.

packages/ui/src/theme/__tests__/defaultTheme.test.ts
hbuchel: Updated to revert these breaking token name changes.

packages/ui/src/theme/__tests__/overrides.test.ts
```diff
@@ -563,7 +563,7 @@ describe('@aws-amplify/ui', () => {
         --amplify-components-stepperfield-flex-direction: column;
         --amplify-components-stepperfield-input-text-align: center;
         --amplify-components-switchfield-disabled-opacity: var(--amplify-opacities-60);
-        --amplify-components-switchfield-focused-shadow: var(--amplify-shadows-small);
+        --amplify-components-switchfield-focused-shadow: 0px 0px 0px 2px var(--amplify-colors-border-focus);
```
0618, line 566: Do we also need to update the CSS variables here? https://ui.docs.amplify.aws/components/authenticator?platform=react#amplify-css-variables

hbuchel: HMMM yes, I just updated this PR to include that change. Since it currently exists, we should keep it up to date. Outside of this PR, I'm wondering if:
- should we show just those css vars specific to the authenticator on this page? I wonder if it makes more sense to just show --amplify-components-authenticator-* variables and maybe a link to a full list elsewhere.
- we should have a less manual way of updating that page?

zchenwei (approved):
LGTM :shipit:

packages/ui/src/theme/tokens/components/switchField.ts
hbuchel: Reverted this change to avoid breaking change to property names.

PR #1687, opened 2022-04-13
feat: adding equivalent props with value default to false

zchenwei:
#### Description of changes

To simplify code generation process, we would like to have all boolean props default to `false`.

1. Alert `isIconHidden` is added with default of `false`
2. FieldGroupIcon `isHidden` is added with default of `false`
3. SearchField `isLabelShown` is added with default of `false`
4. Updating `CONTRIBUTING.md` for `@aws-amplify/ui-react`

Added unit tests to cover new changes

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Alert/__tests__/Alert.test.tsx
+++ b/packages/react/src/primitives/Alert/__tests__/Alert.test.tsx
@@ -86,6 +86,20 @@ describe('Alert: ', () => {
     ).toBe(false);
   });
 
+  it('should have isIconHidden default to false', async () => {
+    render(<Alert variation="info" testId="testIcon" />);
+    const alert = await screen.findByTestId('testIcon');
+    expect(alert.childElementCount).toBe(2);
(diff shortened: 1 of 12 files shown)


packages/react/src/primitives/Alert/__tests__/Alert.test.tsx
```diff
@@ -135,7 +149,7 @@ describe('Alert: ', () => {
     expect(alert.dataset['demo']).toBe('true');
   });
 
-  describe.only('Forward ref: ', () => {
```
reesscot, line 138: good catch!

reesscot (changes requested):
Temporarily blocking, as codegen team determines if they can solve their issue without this.

zchenwei:
Codegen team has the issue fixed on their end without the need of doubling up the props. Close it now

PR #1689, opened 2022-04-14
chore: update PR template to include side effects check

zchenwei:
#### Description of changes

Updating contributing guide and PR template checklist to include `sideEffects`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -23,6 +23,7 @@ the requirements below.
 - [ ] PR description included
 - [ ] `yarn test` passes
 - [ ] Tests are updated
+- [ ] `sideEffects` field in `package.json` is updated
```
reesscot, line 26: I'm not sure how I feel about this being here. I feel like we could address the side effects in a different way. Do we think this will catch side effects being introduced? My hunch is that it will just get ignored by people who don't understand what it means.

zchenwei: I wouldn't say this will catch new introduced side effects. It is more like a reminder

zchenwei: I cannot think of a different way but people will find their files with side effects are not included in the build during dev 🤣

reesscot: What if we change it to `"No side effects or `sideEffects` field updated`, with a link to docs on avoiding sideeffects in contributing? That way we are raising awareness with a more detailed explanation

reesscot (approved):

0618 (approved):

packages/react/CONTRIBUTING.md
```diff
@@ -12,6 +12,42 @@ This will start building `@aws-amplify/ui-react` in watch mode. To test your cha
...
+### Imports
+
+Separate all imports organized, alphabetical blocks of `third_party → internal → local` for easier reading.
+
+```js
+import { isEmpty } from 'lodash/isEmpty';
+import * as React from 'react';
+import { Auth } from 'aws-amplify';
+import { Button } from 'primitives/Button';
+import { THIS_ENUM } from 'utils/types';
```
wlee221, line 26: Should this be

```suggestion
import { Button } from './primitives/Button';
import { THIS_ENUM } from './utils/types';
```
?

packages/react/CONTRIBUTING.md
```diff
@@ -12,6 +12,54 @@ This will start building `@aws-amplify/ui-react` in watch mode. To test your cha
...
+```
+
+### Variable Naming
+
+1. Component names should be capitalized.
+2. Boolean props should be prefixed with `is` and default to `false`.
+
+### Others
+
+Do **NOT** introduce any [side effects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) unless you have to with rational reasons. A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports. An example of this are polyfills, which affect the global scope and usually do not provide an export. For example,
```
calebpollman, line 51:
```suggestion
Do **NOT** introduce any [side effects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) unless you have to with rational reasons. A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports. An example of this are polyfills, which affect the global scope and usually do not provide an export. For example:
```

packages/react/CONTRIBUTING.md
```diff
@@ -12,6 +12,54 @@ This will start building `@aws-amplify/ui-react` in watch mode. To test your cha
...
+
+```js
+import './polyfill';
+```
+
+Any imported file is subject to tree shaking, that means CSS file needs to be added to the side effect list so it will not be unintentionally dropped,
```
calebpollman, line 57:
```suggestion
Any imported file is subject to tree shaking, that means CSS file needs to be added to the side effect list so it will not be unintentionally dropped:
```

calebpollman (approved):

PR #1690, opened 2022-04-14
chore: use tsup for dev build

reesscot:
#### Description of changes
This PR adds tsup back as the dev build method because rollup is only really designed for production builds. This increases dev build speed, especially in watch mode.

1. `yarn ui build && yarn react build`
2. Then I ran both `yarn docs dev` and `yarn react-example dev` builds and successfully saw changed files reflected in the docs and sample apps (without crashing them)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


calebpollman (approved):
LGTM 🚢

0618 (approved):
Do we also need to update README?

reesscot:
> Do we also need to update README?

I double-checked the contributing and it already references using `yarn react dev` and `yarn ui dev`

PR #1692, opened 2022-04-14
chore(ui-react): refactor DataStore query hooks

calebpollman:
#### Description of changes

Follow up to: https://github.com/aws-amplify/amplify-ui/pull/1666

Combine the functionality of both the `useDataStoreItem` and `useDataStoreCollection` hooks in the `useDataStoreBinding` hook. Allows removal of `eslint` ignore comments, and fixes `rules-of-hooks` errors.

#### Issue #, if available
NA

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/hooks/useDataStore.tsx
```diff
@@ -11,99 +11,107 @@ import {
...
-    items: [],
-    isLoading: false,
-    error: undefined,
-  });
+export function useDataStoreBinding<Model extends PersistentModel>(
+  props: DataStoreBindingProps<Model, 'record'>
+): DataStoreItemResult<Model>;
+export function useDataStoreBinding<Model extends PersistentModel>(
+  props: DataStoreBindingProps<Model, 'collection'>
+): DataStoreCollectionResult<Model>;
```
alharris-at, line 24: I'm not familiar w/ this approach exporting just the method signature for a union input of props/results. Does this help the type system narrow the output type based on the input for an otherwise generic method?

packages/react/src/hooks/useDataStore.tsx
```diff
@@ -11,99 +11,107 @@ import {
...
+export function useDataStoreBinding<Model extends PersistentModel>(
+  props:
+    | DataStoreBindingProps<Model, 'record'>
+    | DataStoreBindingProps<Model, 'collection'>
+): DataStoreItemResult<Model> | DataStoreCollectionResult<Model> {
```
alharris-at, line 29: I'm sad that we did this, but it looks like codegen-ui is dependent on this behavior rather than just the underlying hooks.

alharris-at, line 29: I'm sad that we did this, but it looks like codegen-ui is dependent on this behavior rather than just the underlying hooks.

alharris-at: Actually, I just confirmed we no longer actually use `record`, only `collection` (then we manually unpack items[0] similar to what you're doing in the response for `record`) https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap

alharris-at: Actually, I just confirmed we no longer actually use `record`, only `collection` (then we manually unpack items[0] similar to what you're doing in the response for `record`) https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap

alharris-at (approved):

packages/react/src/hooks/useDataStore.tsx
calebpollman: Valid question. Since I didn't write this the overload initially (just moved the function declaration above the `useDataStoreCollection` and `useDataStoreItem` hooks) I'm just taking my best guess here but I believe In order for the function overload to work correctly all the signatures need to be exported

packages/react/src/hooks/useDataStore.tsx
calebpollman: Just spitballing here, and probably not pertinent to this PR, but in the future maybe we can consider migrating to only using `useDataStoreCollection`?

alharris-at: Agreed, probably something you'll want to chat w/ the Studio UI team about updating on their end, then building a deprecation path here.

alharris-at (dismissed):

PR #1702, opened 2022-04-15
chore: update menu demo to new format

reesscot:
#### Description of changes

Updates the Menu docs to use the new demo format:

#### Issue #, if available
https://github.com/aws-amplify/amplify-ui/issues/938

`yarn docs dev` then smoke tested the Menu demo 

#### Checklist

- [x] PR description included


docs/src/pages/components/menu/demo.tsx
```diff
@@ -1,37 +1,61 @@
...
-import { useMenuProps } from '@/components/useMenuProps';
-import { Example } from '@/components/Example';
+import { Demo } from '@/components/Demo';
+import { MenuPropControls, MenuPropControlsProps } from './MenuPropControls';
+import { useMenuProps } from './useMenuProps';
+
+const propsToCode = (props: MenuPropControlsProps) => {
+  const size = props.size ? `\n  size=${props.size}` : '';
+  return `<Menu ${size}
+  menuAlign={${props.menuAlign}}
```
joebuono, line 11: Can we demonstrate passing this prop as a string instead of as a variable?

`menuAlign="${props.menuAlign}"`

docs/src/pages/components/menu/demo.tsx
```diff
@@ -1,37 +1,61 @@
...
+      </MenuItem>
+      <Divider />
+      <MenuItem isDisabled onClick={() => alert('Delete')}>
+        Delete
+      </MenuItem>
+      <MenuItem onClick={() => alert('Attend a workshop')}>
+        Attend a workshop
+      </MenuItem>
+  </Menu>`;
+};
```
joebuono, line 30: nit: Can we adjust the indentation in the example above to match what it looks like after running prettier?

docs/src/pages/components/menu/demo.tsx
reesscot: good catch, will update

docs/src/pages/components/menu/demo.tsx
reesscot: Not sure I'm following the ask, should it look different?

joebuono: It looks like the closing `</Menu>` tag isn't vertically aligned with the opening `<Menu>` tag, and the rest appears to be indented with tabs rather than two spaces.

joebuono (approved):

docs/src/pages/components/menu/demo.tsx
reesscot: gotcha, thanks, will update

hbuchel (approved):
LGTM!

PR #1709, opened 2022-04-15
docs: updating Link docs

zchenwei:
#### Description of changes

Updating `Link` docs

Screen Shot 2022-04-15 at 4 37 46 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


docs/src/pages/components/link/demo.tsx
```diff
@@ -14,21 +27,23 @@ export const LinkDemo = () => {
...
         <Link
           color={linkProps.color}
           isExternal={linkProps.isExternal}
           textDecoration={linkProps.textDecoration}
         >
-          {linkProps.children}
+          My Demo Link
         </Link>
-        <Text fontSize={'.75em'}>
+        <Text fontSize="0.75em">
```
jacoblogan, line 42: can we use a design token here?
`const { tokens } = useTheme();`
`<Text fontSize={tokens.space.relative.small}>`

docs/src/pages/components/link/react.mdx
```diff
@@ -80,74 +106,68 @@ The Link component primarily uses four css pseudo classes. These are `:active`,
...
-  all: unset;
-  /* Add your styling here */
-}
-```
+<Example>
+  <LinkGlobalStylingExample />
+  <ExampleCode>
+    ```css
+    /* styles.css */
+    :root {
```
jacoblogan, line 114: I don't think `:root` works for overriding css variables because of how we do out theme, global styling rule should be `[data-amplify-theme]`

docs/src/pages/components/link/examples/LinkGlobalStylingExample.tsx
```diff
@@ -0,0 +1,6 @@
+import { Link, useTheme } from '@aws-amplify/ui-react';
+
+export const LinkGlobalStylingExample = () => {
+  const { tokens } = useTheme();
+  return <Link color={tokens.colors.red[80]}>Link Global Styling</Link>;
```
jacoblogan, line 5: The example and the code don't match here, the example is global styling but here you are using local styling.  We don't want to override the global styling across the entire page but using a class name here and including the updated style in the corresponding css file would be a more accurate representation of what we are trying to demonstrate

jacoblogan (approved):
LGTM just a couple of minor questions

docs/src/pages/components/link/demo.tsx
```diff
@@ -14,21 +27,23 @@ export const LinkDemo = () => {
...
           <sup>*</sup>Rel Attribute=
```
joebuono, line 43: Maybe "Rel Attribute=" should only be displayed when `isExternal` is `true`?

zchenwei: I like this idea

joebuono (approved):

PR #1710, opened 2022-04-15
chore: update view demo to new format

reesscot:
#### Description of changes

Update View documentation demo to new format.

Tested local docs

#### Checklist

- [x] PR description included


docs/src/pages/components/view/ViewPropControls.tsx
```diff
@@ -0,0 +1,110 @@
+import * as React from 'react';
+import { Flex, ViewProps, SelectField, TextField } from '@aws-amplify/ui-react';
+import _ from 'lodash';
```
joebuono, line 3: Why are we importing lodash?

docs/src/pages/components/view/ViewPropControls.tsx
```diff
@@ -0,0 +1,110 @@
...
+    <Flex direction="column">
+      <SelectField
+        value={asElementType as string}
+        label="As element type"
+        onChange={(event) =>
+          setAsElementType(event.target.value as ViewProps['as'])
+        }
+      >
+        <option value="div">div</option>
+        <option value="button">button</option>
```
joebuono, line 58: Just for my own clarity: when the view is set `as="button"`, should we expect the cursor to change to a hand `onmouseover`?

docs/src/pages/components/view/ViewPropControls.tsx
```diff
@@ -0,0 +1,110 @@
...
+        value={boxShadow as string}
+        onChange={(event) => setBoxShadow(event.target.value)}
+        label="Box Shadow"
+      />
+      <TextField
+        value={color as string}
+        onChange={(event) => setColor(event.target.value)}
+        label="Color"
+      />
+      <TextField
```
joebuono, line 92: I saw that @0618 was using a `SliderField` for numeric props in her TextField PR (see [FontWeight](https://github.com/aws-amplify/amplify-ui/pull/1711) in demo). Is this something we'd like to explore for props like height, borderRadius, padding, width, etc?

docs/src/pages/components/view/demo.tsx
```diff
@@ -1,286 +1,82 @@
...
+    ariaLabel: 'View example',
+    width: '20rem',
+    height: '4rem',
+    color: 'var(--amplify-colors-blue-60)',
+    backgroundColor: 'var(--amplify-colors-white)',
+    boxShadow: '3px 3px 5px 6px var(--amplify-colors-neutral-60)',
+    padding: '1rem',
+    border: '1px solid var(--amplify-colors-black)',
+    borderRadius: '6px',
+    opacity: '100%',
```
joebuono, line 42: I don't see the opacity prop as an option in the demo?

docs/src/pages/components/view/demo.tsx
```diff
@@ -1,286 +1,82 @@
...
-            }}
-          />
-        </FieldSet>
+export const ViewDemo = () => {
+  const props = useViewProps({
+    as: 'div',
+    ariaLabel: 'View example',
+    width: '20rem',
+    height: '4rem',
+    color: 'var(--amplify-colors-blue-60)',
```
joebuono, line 36: It's an interesting idea to use CSS variables in the example. Do you think this will be intuitive to customers though?

joebuono (commented):
Thanks for refactoring this old demo! Just had a few comments

docs/src/pages/components/view/ViewPropControls.tsx
reesscot: great question, I'll remove it

docs/src/pages/components/view/ViewPropControls.tsx
reesscot: No, because a button wouldn't have a hand cursor. I think you're thinking of an `<a>` tag?

docs/src/pages/components/view/ViewPropControls.tsx
reesscot: That makes sense for `FontWeight` because it doesn't have a unit. I think we would just want a TextField here because a user may want to experiment with `px`, `em`, `rem`, etc. We also probably don't want to limit the range of numbers customers can put in, which would have to be decided when using a `SliderField`.

docs/src/pages/components/view/demo.tsx
reesscot: Good point, I forgot to remove it here. Will update.

docs/src/pages/components/view/demo.tsx
reesscot: I'm also on the fence, what does everyone else think? I liked it because it exposed the idea to customers that they could just use the CSS variables. But maybe there could be a better user experience for exposing the CSS variables, like a searchable dropdown instead?

docs/src/pages/components/view/demo.tsx
```diff
@@ -1,286 +1,78 @@
...
-            type="text"
-            placeholder="Set opacity"
-            value={opacity}
-            onChange={(event: any) => {
-              setOpacity(event.target.value);
-            }}
-          />
-        </FieldSet>
+    `<View` +
+    (props.ariaLabel ? `\n  ariaLabel="${props.ariaLabel}"` : '') +
```
jacoblogan, line 10: can we add a simple shared function to do this that we can use across all our docs pages.  might help clean up the docs pages a little bit instead of having these ternaries everywhere.

joebuono (approved):

jacoblogan (approved):
LGTM

docs/src/pages/components/view/demo.tsx
dbanksdesign: :shipit:

docs/src/pages/components/view/demo.tsx
reesscot: Good idea. I updated PR to include `getPropString` util.

PR #1734, opened 2022-04-21
fix: issue where number unions ignored in primitivesCatalog

reesscot:
#### Description of changes
Fixes an issue where in our `generatePrimitivesCatalog` script  where union types of numbers were being ignored. This made it so that the `Header` primitive's `level` prop was not included in the `primitives.json` or `primitivesCatalog` ESM export. In Studio UI, it get's filtered out for not being a valid prop from the definition.

I also fixed a few other warnings in the script because `Radio` doesn't have a `label` prop, and `TextField` `isMultiline` prop has been deprecated and is no longer used in Studio.

Diff of the JSON files before and after these changes:

```diff

                      "variation": {
                        "type": "string"
                      },
+                   "level": {
+                     "type": "number",
+                     "priority": true
+                   },
                     "isTruncated": {
                         "type": "boolean"
```

</body></html>
```

#### Checklist

<!-- Remove items that do not apply. For completed items, change [ ] to [x]. -->

- [x] PR description included

### DIFF
--- a/packages/react/scripts/generatePrimitivesCatalog.ts
+++ b/packages/react/scripts/generatePrimitivesCatalog.ts
@@ -34,7 +34,7 @@ const priorityProps = {
   Pagination: ['totalPages', 'currentPage', 'siblingCount'],
   PasswordField: [...fieldProps, 'hideShowPassword', 'labelHidden'],
   PhoneNumberField: [...fieldProps, 'labelHidden'],
-  Radio: ['label', 'value', 'checked', 'isDisabled', 'size', 'labelPosition'],
+  Radio: ['value', 'checked', 'isDisabled', 'size', 'labelPosition'],
   Rating: ['value', 'size', 'maxValue', 'fillColor', 'emptyColor'],
   SearchField: [...fieldProps, 'variation', 'labelHidden'],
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by zchenwei — APPROVED** [review]
LGTM!

**Review by cwoolum — APPROVED** [review]
LGTM

File: packages/react/scripts/generatePrimitivesCatalog.ts [hunk 1]
```diff
@@ -115,10 +115,20 @@ const getCatalogComponentProperty = (
   } else if (propType.isNumber() || propType.isNumberLiteral()) {
     return { type: PrimitiveCatalogComponentPropertyType.Number };
   } else if (propType.isUnion()) {
+    const hasNumber = propType
+      .getUnionTypes()
+      .every((prop) => prop.isNumber() || prop.isNumberLiteral());
```
slaymance, line 120: This should be `.every` and not `.some` like in the `hasString` check below?

reesscot: With 'some' it was falsely changing lots of types to "number" when the union included number. We only want to catch union types of all number types here.

PR #1740, opened 2022-04-21
use flattened css in vue

jacoblogan:
#### Description of changes
Update Vue code to use the flattened css selectors

### DIFF
--- a/packages/vue/src/components/confirm-sign-up.vue
+++ b/packages/vue/src/components/confirm-sign-up.vue
@@ -105,7 +105,7 @@ const onLostCodeClicked = (): void => {
             <amplify-button
               class="amplify-field-group__control"
               :fullwidth="false"
-              :laoding="false"
+              :loading="false"
               :variation="'primary'"
               type="submit"
               style="font-weight: normal"
--- a/packages/vue/src/components/confirm-sign-up.vue
+++ b/packages/vue/src/components/confirm-sign-up.vue
@@ -104,9 +104,9 @@ const onLostCodeClicked = (): void => {
             </base-alert>
             <amplify-button
               class="amplify-field-group__control"
-              data-fullwidth="false"
-              data-loading="false"
-              data-variation="primary"
+              :fullwidth="false"
+              :laoding="false"
+              :variation="'primary'"
               type="submit"
               style="font-weight: normal"
               :disabled="isPending"
@@ -115,8 +115,8 @@ const onLostCodeClicked = (): void => {
             </amplify-button>
             <amplify-button
               class="amplify-field-group__control"
-              data-fullwidth="false"
-              data-variation="default"
+              :fullwidth="false"
+              :variation="'default'"
               style="font-weight: normal"
               type="button"
               @click.prevent="onLostCodeClicked"
--- a/packages/vue/src/components/primitives/amplify-button.vue
+++ b/packages/vue/src/components/primitives/amplify-button.vue
@@ -32,6 +32,7 @@ const { type, fullWidth, size, variation, fontWeight } = withDefaults(
       :data-variation="variation"
       :data-loading="loading"
       :data-disabled="disabled"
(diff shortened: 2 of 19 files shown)


packages/vue/src/components/primitives/amplify-button.vue
```diff
@@ -5,6 +5,8 @@ interface PropsInterface {
   size?: 'small' | 'medium' | 'large';
   variation?: 'primary' | 'default' | 'link';
   fontWeight?: 'normal' | 'bold' | 'lighter';
+  loading?: boolean | string;
+  disabled?: boolean | string;
```
ErikCH, line 9: Since you added `disabled` as a prop here, it's no longer adding the `disabled` attribute via `$attrs` to the button. You'll need to explicitly set `:disabled=disabled` on the button to correct this.

packages/vue/src/components/primitives/base-heading.vue
```diff
@@ -12,16 +12,21 @@ export default defineComponent({
   setup(props, { slots, attrs }): Record<string, unknown> | (() => unknown) {
     const defaultSlot = slots.default ? slots.default() : [];
     const headingI = slots.headingI ? slots.headingI() : [];
+    const headerClass = `${attrs.class || ''} amplify-heading--${attrs.level}`;
```
ErikCH, line 15: This is a little confusing on Vue's part but the `level` is passed in as a prop, not attrs. `attrs` only includes non prop attributes passed into the component.
```suggestion
    const headerClass = `${attrs.class || ''} amplify-heading--${props.level}`;
```

packages/vue/src/components/primitives/base-heading.vue
```diff
@@ -12,16 +12,21 @@ export default defineComponent({
...
       ];
     } else {
       return () =>
-        h(`h${props.level}`, { 'data-amplify-heading': '', ...attrs }, [
-          //@ts-ignore
-          headingI[0] ? h(headingI[0].children[0]) : h(defaultSlot[0]),
-        ]);
+        h(
+          `h${props.level}`,
+          { 'data-amplify-heading': '', class: headerClass, ...attrs },
```
ErikCH, line 24: The `...attrs` has a class being passed into it. So it's overriding the `class: headerClass`. You could destructuring the class out of `attrs`, since you are already adding it into `headerClass`

```const {class, ...rest} = attrs;```

Then use `rest` instead of `attrs`

**Review by ErikCH — CHANGES_REQUESTED** [review]
Just a few changes! Looks good

File: packages/vue/src/components/confirm-sign-up.vue [hunk 4]
```diff
@@ -104,9 +104,9 @@ const onLostCodeClicked = (): void => {
             </base-alert>
             <amplify-button
               class="amplify-field-group__control"
-              data-fullwidth="false"
-              data-loading="false"
-              data-variation="primary"
+              :fullwidth="false"
+              :laoding="false"
```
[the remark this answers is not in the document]
jacoblogan: yes thanks, updated

packages/vue/src/components/primitives/amplify-button.vue
jacoblogan: thanks good catch, `:disabled="disabled"` has been added onto the button

packages/vue/src/components/primitives/base-heading.vue
jacoblogan: good catch, thanks that has been updated

packages/vue/src/components/primitives/base-heading.vue
jacoblogan: good catch, yes `...attrs` was overwriting the class property.  I've reversed the order and broken out the class property previous to that and added in the desired header class.

ErikCH (approved):
👍

reesscot (approved):
LGTM

PR #1741, opened 2022-04-21
update react code to use the flattened css selectors

jacoblogan:
#### Description of changes

Update react code to use the flattened css selectors

### DIFF
--- a/packages/react/src/primitives/Checkbox/Checkbox.tsx
+++ b/packages/react/src/primitives/Checkbox/Checkbox.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import classNames from 'classnames';
 
-import { classNameModifier } from '../shared/utils';
+import { classNameModifierByFlag } from '../shared/utils';
 import { CheckboxProps } from '../types/checkbox';
 import { ComponentClassNames } from '../shared/constants';
 import { Flex } from '../Flex';
@@ -50,13 +50,48 @@ const CheckboxPrimitive: Primitive<CheckboxProps, 'input'> = (
   const buttonTestId = useTestId(testId, ComponentClassNames.CheckboxButton);
   const iconTestId = useTestId(testId, ComponentClassNames.CheckboxIcon);
   const labelTestId = useTestId(testId, ComponentClassNames.CheckboxLabel);
+  const flexClasses = classNames(
+    ComponentClassNames.CheckboxButton,
+    classNameModifierByFlag(
+      ComponentClassNames.CheckboxButton,
+      'disabled',
+      isDisabled
+    ),
+    classNameModifierByFlag(
+      ComponentClassNames.CheckboxButton,
+      'error',
+      hasError
+    ),
+    classNameModifierByFlag(
+      ComponentClassNames.CheckboxButton,
+      'focused',
+      dataFocus
+    )
+  );
+  const iconClasses = classNames(
+    ComponentClassNames.CheckboxIcon,
+    classNameModifierByFlag(
+      ComponentClassNames.CheckboxButton,
+      'checked',
+      dataChecked
+    ),
+    classNameModifierByFlag(
+      ComponentClassNames.CheckboxButton,
+      'disabled',
+      isDisabled
+    )
+  );
 
   return (
     <Flex
       as="label"
       className={classNames(
         ComponentClassNames.Checkbox,
-        classNameModifier(ComponentClassNames.Checkbox, 'disabled', isDisabled),
+        classNameModifierByFlag(
+          ComponentClassNames.Checkbox,
+          'disabled',
+          isDisabled
+        ),
         className
       )}
       data-disabled={isDisabled}
@@ -94,24 +129,7 @@ const CheckboxPrimitive: Primitive<CheckboxProps, 'input'> = (
       <Flex
         aria-hidden="true"
         as="span"
-        className={classNames(
-          ComponentClassNames.CheckboxButton,
-          classNameModifier(
-            ComponentClassNames.CheckboxButton,
-            'disabled',
-            isDisabled
-          ),
-          classNameModifier(
-            ComponentClassNames.CheckboxButton,
-            'error',
-            hasError
-          ),
-          classNameModifier(
-            ComponentClassNames.CheckboxButton,
-            'focused',
-            dataFocus
-          )
-        )}
+        className={flexClasses}
         data-checked={dataChecked}
         data-disabled={isDisabled}
         data-focus={dataFocus}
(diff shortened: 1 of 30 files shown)


packages/react/src/primitives/shared/utils.ts
```diff
@@ -159,3 +159,15 @@ export const mergeVariantsAndOverrides = (
     ...merged,
   };
 };
+
+export type Modifiers = string | number | null;
+export const classNameModifier = (
+  base: string,
```
calebpollman, line 165: Can this be typed as `ComponentClassNames`?

calebpollman (commented):
Thanks for doing this! Left some feedback but overall looks like a great start

packages/react/src/primitives/shared/utils.ts
```diff
@@ -159,3 +159,15 @@ export const mergeVariantsAndOverrides = (
     ...merged,
   };
 };
+
+export type Modifiers = string | number | null;
+export const classNameModifier = (
```
hbuchel, line 164: Can we have comments (or is this the best place to doc this...?) that explain how/why this is used or needed? Just looking through the uses of it for each component and they generally follow the same pattern of:

```
className={classNames(
  ComponentClassNames.ComponentName,
  classNameModifier(ComponentClassNames.ComponentName, size),
  className
)}
```

So I'm wondering if we should have a small blurb to say if these need to be in a specific order or if there is any additional nuance there. If I were to create a brand new component that is something I would be curious about.

jacoblogan: good call, I've added comments to clarify what this function is and how it is used, also to simplify things I've separated the two use cases into two functions.

packages/react/src/primitives/Loader/Loader.tsx
```diff
@@ -34,6 +35,13 @@ const LoaderPrimitive: Primitive<LoaderProps, 'svg'> = (
   percentage = Math.max(percentage, 0);
 
   const percent = `${percentage}%`;
+  const componentClasses = classNames(
+    ComponentClassNames.Loader,
+    classNameModifier(ComponentClassNames.Loader, size),
+    classNameModifier(ComponentClassNames.Loader, variation),
+    isDeterminate ? ComponentClassNames.LoaderDeterminate : null,
```
zchenwei, line 42: Use `classNameModifierByFlag`?

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -31,6 +32,16 @@ const SelectPrimitive: Primitive<SelectProps, 'select'> = (
   // value === undefined is to make sure that component is used in uncontrolled way so that setting defaultValue is valid
   const shouldSetDefaultPlaceholderValue =
     value === undefined && defaultValue === undefined && placeholder;
+  const componentClasses = classNames(
+    ComponentClassNames.Select,
+    ComponentClassNames.FieldGroupControl,
+    classNameModifier(ComponentClassNames.Select, size),
+    classNameModifier(ComponentClassNames.Select, variation),
+    classNameModifierByFlag(ComponentClassNames.Select, 'error', hasError),
+    classNameModifier(ComponentClassNames.Select, size),
```
zchenwei, line 41: Duplicate one

packages/react/src/primitives/SliderField/SliderField.tsx
```diff
@@ -112,7 +118,12 @@ const SliderFieldPrimitive: Primitive<SliderFieldProps, typeof Root> = (
         outerEndComponent={outerEndComponent}
       >
         <Root
-          className={classNames(ComponentClassNames.SliderFieldRoot, className)}
+          className={classNames(
+            ComponentClassNames.SliderFieldRoot,
+            `${ComponentClassNames.SliderFieldRoot}--${orientation}`,
+            size ? `${ComponentClassNames.SliderFieldRoot}--${size}` : null,
```
zchenwei, line 124: Use util class?

packages/react/src/primitives/shared/utils.ts
```diff
@@ -159,3 +159,15 @@ export const mergeVariantsAndOverrides = (
...
+  base: string,
+  modifier: Modifiers,
+  override?: boolean
+): string => {
+  if (override === undefined) {
+    return modifier ? `${base}--${modifier}` : null;
+  }
+  return override ? `${base}--${modifier}` : null;
```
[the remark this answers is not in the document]
jacoblogan: Yes that is correct, I've updated the PR to separate this logic into 2 functions to make things more clear

packages/react/src/primitives/shared/__tests__/utils.test.tsx
```diff
@@ -336,3 +337,27 @@ describe('mergeVariantsAndOverrides', () => {
...
+
+  it('should return the modified className', () => {
+    expect(classNameModifier(myClass, modifier)).toEqual(modifiedClassName);
+  });
+
+  it('should return null', () => {
+    expect(classNameModifier(myClass, undefined)).toEqual(null);
+  });
+
+  it('should return the modified className', () => {
```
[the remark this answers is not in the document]
jacoblogan: thanks updating those test names

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -89,15 +94,44 @@ const CheckboxPrimitive: Primitive<CheckboxProps, 'input'> = (
...
+            ComponentClassNames.CheckboxButton,
+            'error',
+            hasError
+          ),
+          classNameModifier(
+            ComponentClassNames.CheckboxButton,
+            'focused',
+            dataFocus
+          )
+        )}
```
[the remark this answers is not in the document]
jacoblogan: great suggestion, I've updated the PR to move the long className functions into the body of the component to make the props easier to read

packages/react/src/primitives/Loader/Loader.tsx
jacoblogan: The classNameModifier function is used to add modifiers onto the classnames, here we don't want to modify the classname instead just add the classname in if it `isDeterminate`

packages/react/src/primitives/shared/utils.ts
```diff
@@ -159,3 +159,15 @@ export const mergeVariantsAndOverrides = (
     ...merged,
   };
 };
+
+export type Modifiers = string | number | null;
```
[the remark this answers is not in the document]
calebpollman: 👍🏽 Still curious, can we narrow the allowed values of `Modifiers`? Also, think it can be the singular form `Modifier` since it is not an array or an object

packages/react/src/primitives/SliderField/SliderField.tsx
jacoblogan: yes

zchenwei (approved):
LGTM

hbuchel (approved):
LGTM, thank you for adding the additional comments!

PR #1742, opened 2022-04-21
flatten css to 010 specificity

jacoblogan:
#### Description of changes
Update css selectors to be less specific

### DIFF
--- a/packages/ui/src/theme/css/component/badge.scss
+++ b/packages/ui/src/theme/css/component/badge.scss
@@ -1,42 +1,41 @@
 .amplify-badge {
-  color: var(--amplify-components-badge-color);
   background-color: var(--amplify-components-badge-background-color);
   border-radius: var(--amplify-components-badge-border-radius);
+  color: var(--amplify-components-badge-color);
   display: inline-flex;
   font-size: var(--amplify-components-badge-font-size);
   font-weight: var(--amplify-components-badge-font-weight);
+  line-height: var(--amplify-components-badge-line-height);
   padding: var(--amplify-components-badge-padding-vertical)
     var(--amplify-components-badge-padding-horizontal);
   text-align: var(--amplify-components-badge-text-align);
-  line-height: var(--amplify-components-badge-line-height);
-
-  &[data-variation='info'] {
+  &--info {
     color: var(--amplify-components-badge-info-color);
     background-color: var(--amplify-components-badge-info-background-color);
   }
 
-  &[data-variation='error'] {
+  &--error {
     color: var(--amplify-components-badge-error-color);
     background-color: var(--amplify-components-badge-error-background-color);
   }
 
-  &[data-variation='warning'] {
+  &--warning {
     color: var(--amplify-components-badge-warning-color);
     background-color: var(--amplify-components-badge-warning-background-color);
   }
 
-  &[data-variation='success'] {
+  &--success {
     color: var(--amplify-components-badge-success-color);
     background-color: var(--amplify-components-badge-success-background-color);
   }
 
-  &[data-size='small'] {
+  &--small {
     font-size: var(--amplify-components-badge-small-font-size);
     padding: var(--amplify-components-badge-small-padding-vertical)
       var(--amplify-components-badge-small-padding-horizontal);
   }
 
-  &[data-size='large'] {
+  &--large {
     font-size: var(--amplify-components-badge-large-font-size);
     padding: var(--amplify-components-badge-large-padding-vertical)
       var(--amplify-components-badge-large-padding-horizontal);
--- a/packages/ui/src/theme/css/component/button.scss
+++ b/packages/ui/src/theme/css/component/button.scss
@@ -8,22 +8,43 @@
   align-items: center;
   border-radius: var(--amplify-components-button-border-radius);
   box-sizing: border-box;
+  font-size: var(--amplify-components-button-font-size);
   font-weight: var(--amplify-components-button-font-weight);
   transition: all var(--amplify-components-button-transition-duration);
-  border-width: var(--amplify-components-button-border-width);
   border-style: var(--amplify-components-button-border-style);
   border-color: var(--amplify-components-button-border-color);
   border-radius: var(--amplify-components-button-border-radius);
+  border-width: var(--amplify-components-button-border-width);
   color: var(--amplify-components-button-color);
-  font-size: var(--amplify-components-button-font-size);
   line-height: var(--amplify-components-button-line-height);
+  user-select: none;
   padding-block-start: var(--amplify-components-button-padding-block-start);
   padding-block-end: var(--amplify-components-button-padding-block-end);
   padding-inline-start: var(--amplify-components-button-padding-inline-start);
   padding-inline-end: var(--amplify-components-button-padding-inline-end);
-  user-select: none;
   // all:unset removes this, so we have to add it back.
   cursor: pointer;
+
+  --amplify-internal-button-disabled-color: var(
+    --amplify-components-button-disabled-color
+  );
+  --amplify-internal-button-disabled-background-color: var(
+    --amplify-components-button-disabled-background-color
+  );
+  --amplify-internal-button-disabled-border-color: var(
+    --amplify-components-button-disabled-border-color
+  );
+  --amplify-internal-button-disabled-text-decoration: initial;
+  --amplify-internal-button-loading-background-color: var(
+    --amplify-components-button-loading-background-color
+  );
+  --amplify-internal-button-loading-border-color: var(
+    --amplify-components-button-loading-border-color
+  );
+  --amplify-internal-button-loading-color: var(
+    --amplify-components-button-loading-color
+  );
+  --amplify-internal-button-loading-text-decoration: initial;
 
   &:hover {
     background-color: var(--amplify-components-button-hover-background-color);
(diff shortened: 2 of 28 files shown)


packages/ui/src/theme/css/component/badge.scss
```diff
@@ -1,42 +1,41 @@
 .amplify-badge {
-  color: var(--amplify-components-badge-color);
   background-color: var(--amplify-components-badge-background-color);
   border-radius: var(--amplify-components-badge-border-radius);
+  color: var(--amplify-components-badge-color);
```
dbanksdesign, line 4: Ohh making these alphabetic order. Nice!

packages/ui/src/theme/css/component/button.scss
```diff
@@ -8,23 +8,44 @@
...
+  user-select: none;
   padding-block-start: var(--amplify-components-button-padding-block-start);
   padding-block-end: var(--amplify-components-button-padding-block-end);
   padding-inline-start: var(--amplify-components-button-padding-inline-start);
   padding-inline-end: var(--amplify-components-button-padding-inline-end);
-  user-select: none;
   // all:unset removes this, so we have to add it back.
   cursor: pointer;
 
+  --amplify-internal-button-disabled-color: var(
```
dbanksdesign, line 28: Why are we adding these internal variables? Maybe let's add a comment here for that.

packages/ui/src/theme/css/component/loader.scss
```diff
@@ -5,6 +5,24 @@
   fill: none;
   stroke: var(--amplify-components-loader-stroke-empty);
 
+  // these are css variables used to keep the structure flat but allow multiple states to effect the value
```
dbanksdesign, line 8: Ahhh I see now.

packages/ui/src/theme/css/component/loader.scss
```diff
@@ -47,23 +53,46 @@
...
+    );
+    --amplify-internal-loader-linear-font-size: var(
+      --amplify-components-loader-linear-small-font-size
+    );
+  }
+
+  &--large {
+    width: var(--amplify-components-loader-large-width);
+    height: var(--amplify-components-loader-large-height);
+    font-size: var(--amplify-internal-loader-large-font-size);
```
dbanksdesign, line 75: Could these also use internal variables too?

packages/ui/src/theme/css/component/sliderField.scss
```diff
@@ -15,36 +15,42 @@
...
-  }
-  
-  [data-size='small'] & {
-    &[data-orientation='horizontal'] {
-      height: var(--amplify-components-sliderfield-small-thumb-height);
-    }
-    &[data-orientation='vertical'] {
-      width: var(--amplify-components-sliderfield-small-thumb-height);
-    }
+    width: var(--amplify-internal-sliderfield__root-width);
```
dbanksdesign, line 35: nit: are we using BEM naming for other css vars too?

dbanksdesign (commented):
Just a few minor nits and questions, overall it looks good!

packages/ui/src/theme/css/styles.scss
```diff
@@ -3,6 +3,7 @@
  */
 
 @import 'dist/theme';
+@import './component/fieldControl';
```
zchenwei, line 6: Moving this to line 54? That way we can have all component styles group together

packages/ui/src/theme/css/component/checkbox.scss
```diff
@@ -2,7 +2,7 @@
   cursor: var(--amplify-components-checkbox-cursor);
   align-items: var(--amplify-components-checkbox-align-items);
   gap: inherit;
-  &[data-disabled='true'] {
+  &--disabled {
```
zchenwei, line 5: Moving this down and combine two selector?

packages/ui/src/theme/css/component/pagination.scss
```diff
@@ -14,64 +14,64 @@
...
-  padding-inline-end: var(
-    --amplify-components-pagination-button-padding-inline-end
-  );
-  transition-property: var(
-    --amplify-components-pagination-button-transition-property
-  );
-  transition-duration: var(
-    --amplify-components-pagination-button-transition-duration
-  );
+  &__item-button--link {
```
zchenwei, line 17: I feel like we can get rid of `link` without appending a new `xxx--link` class. What do you think?

packages/ui/src/theme/css/component/sliderField.scss
```diff
@@ -91,18 +109,17 @@
   background-color: var(
     --amplify-components-sliderfield-range-background-color
   );
-  
+
   &[data-disabled] {
```
zchenwei, line 113: Why not flatten `disabled`?

packages/ui/src/theme/css/component/sliderField.scss
```diff
@@ -130,19 +147,22 @@
   &:focus {
     box-shadow: var(--amplify-components-sliderfield-thumb-focus-box-shadow);
   }
-  
+
   &[data-disabled] {
```
zchenwei, line 151: Why not flatten `disabled`?

reesscot: I think it's because we can't access the underlying Radix component to add our own flattened classes.

packages/ui/src/theme/css/component/sliderField.scss
reesscot: I think it's because we can't access the underlying Radix component to add our own flattened classes.

packages/ui/CONTRIBUTING.md
```diff
@@ -5,3 +5,24 @@
 1. Navigate to the _root_ of your local clone of [aws-amplify/amplify-ui](https://github.com/aws-amplify/amplify-ui)
 1. Run `yarn setup`
 1. Run `yarn ui dev`
+
+## Style Guide
+
+The @aws-amplify/ui package uses the [BEM](http://getbem.com/) styling syntax to standardize its css code and maintain flat css selectors.  The goal is to keep all css selectors at a 010 specificity so that customers can easily restyle their application using any of the available styling tools that exist. This is achieved by using specific classes that follow the syntax of `{Block}__{Element}--{Modifier}`.  Blocks should be a standalone entity that has meaning on its own. An Element is a part of a block and only has meaning within the context of a block.  A Modifier is state on a block or element that changes its behavior or appearance.
```
reesscot, line 11:
```suggestion
The @aws-amplify/ui package uses the [BEM](http://getbem.com/) styling syntax to standardize its CSS code and maintain flat CSS selectors.  The goal is to keep all CSS selectors at a 010 specificity so that customers can easily restyle their application using any of the available styling tools that exist. This is achieved by using specific classes that follow the syntax of `{Block}__{Element}--{Modifier}`.  Blocks should be a standalone entity that has meaning on its own. An Element is a part of a block and only has meaning within the context of a block.  A Modifier is state on a block or element that changes its behavior or appearance.
```

packages/ui/CONTRIBUTING.md
```diff
@@ -5,3 +5,24 @@
...
+- A class name following this syntax can only have a single Block, Element, and Modifier, but there are some cases where multiple states influence a css properties value.  In these instances we are using an additional css variable to update the targeted value when multiple states are applied.
```
reesscot, line 12:
```suggestion
- A class name following this syntax can only have a single Block, Element, and Modifier, but there are some cases where multiple states influence a css properties value.  In these instances we are using an additional css variable prefixed with `amplify-internal` to update the targeted value when multiple states are applied.
```

packages/ui/CONTRIBUTING.md
```diff
@@ -5,3 +5,24 @@
...
+  --amplify-internal-divider-size: 0.75rem;
+}
+.amplify-divider--horizontal {
+  border-bottom-width: var(--amplify-internal-divider-size);
+}
+.amplify-divider--vertical {
+  border-left-width: var(--amplify-internal-divider-size);
+}
+```
+In this example the `amplify-divider` block can have 2 modifiers that will influence each other.  The size modifier `--small` is responsible for updating the border width, but cannot update it directly because it targets different css properties based on the orientation modifier.  Here we are using a css variable to update the desired value.  The orientation modifier applies the css variable to the desired css property while the size modifier updates the value of that css variable so it is the correct value being applied to the correct property.
```
reesscot, line 28: This is a really clear description! 💯

spelling nit: `css` => `CSS`
```suggestion
In this example the `amplify-divider` block can have 2 modifiers that will influence each other.  The size modifier `--small` is responsible for updating the border width, but cannot update it directly because it targets different CSS properties based on the orientation modifier.  Here we are using a CSS variable to update the desired value.  The orientation modifier applies the CSS variable to the desired CSS property while the size modifier updates the value of that CSS variable so it is the correct value being applied to the correct property.
```

packages/ui/src/theme/css/component/checkbox.scss
reesscot: +1

packages/ui/src/theme/css/component/loader.scss
```diff
@@ -5,6 +5,18 @@
   fill: none;
   stroke: var(--amplify-components-loader-stroke-empty);
 
+  // these are css variables used to keep the structure flat but allow multiple states to effect the value
+  // because these ultimately refer back to our design tokens, these do not break the contract with our customers
```
reesscot, line 9: Probably don't need this comment now that you documented this in contributing
```suggestion
```

reesscot (approved):
LGTM, just a few spelling nits in contributing.

zchenwei (approved):
LGTM

packages/ui/src/theme/css/styles.scss
jacoblogan: yes that has been moved

packages/ui/src/theme/css/component/checkbox.scss
jacoblogan: I've moved the flex property that was below this up but I'm not sure what else can be combined.  There are the Element blocks below this that could be combined under the Block.

packages/ui/src/theme/css/component/pagination.scss
jacoblogan: yes good call, link has been removed

packages/ui/src/theme/css/component/checkbox.scss
zchenwei: I mean we have 2 `&--disabled` on the same level? This could be moving down there?

PR #1749, opened 2022-04-21
fix(geo): Add position option to GeocoderControl

slaymance:
#### Description of changes
The `position` property for the `Geocoder` component needs to be passed as the second argument to the `useControl` hook when `Geocoder` is used as a map control.

#### Issue #, if available
N/A

Passing a `position` property to the `Geocoder` component in the `map-with-geocoder` example is reflected in the map.
```jsx
import { MapView, Geocoder } from '@aws-amplify/ui-react';
import { Amplify } from 'aws-amplify';

import awsExports from './aws-exports';

export default function MapWithGeocoder() {
  return (
    <MapView>
      <Geocoder position="top-left" />
    </MapView>
  );
}
```
Screen Shot 2022-04-21 at 09 14 47

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated - N/A
- [x] Relevant documentation is changed or added (and PR referenced) - docs are up to date

### DIFF
--- a/packages/react/src/components/Geo/Geocoder/index.tsx
+++ b/packages/react/src/components/Geo/Geocoder/index.tsx
@@ -44,13 +44,13 @@ export const Geocoder = (props: GeocoderProps) => {
    * This logic determines whether the Geocoder exists as part of a Map component or if it is a standalone component.
    * The `useControl` hook inside `GeocoderControl` from `react-map-gl` makes it easy to add a control to a map,
    * but throws an error if that map doesn't exist. If the map doesn't exist, the Geocoder is mounted to a container
-   * upon rendering inside the `StandaloneGeocoder`.
+   * upon rendering inside the `GeocoderStandalone`.
    */
   if (map) {
     return <GeocoderControl {...GEOCODER_OPTIONS} {...props} />;
   }
 
-  return <StandaloneGeocoder {...GEOCODER_OPTIONS} {...props} />;
+  return <GeocoderStandalone {...GEOCODER_OPTIONS} {...props} />;
 };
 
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Geo/Geocoder/index.tsx
```diff
@@ -37,27 +37,26 @@ type GeocoderControl = IControl & {
...
-  position = 'top-right',
-  ...props
-}: GeocoderProps) => {
+export const Geocoder = (props: GeocoderProps) => {
   const { current: map } = useMap();
 
   /**
    * This logic determines whether the Geocoder exists as part of a Map component or if it is a standalone component.
-   * The `useControl` hook inside `ControlledGeocoder` from `react-map-gl` makes it easy to add a control to a map,
+   * The `useControl` hook inside `GeocoderControl` from `react-map-gl` makes it easy to add a control to a map,
```
slaymance, line 45: I changed the name since calling a React component "controlled" is distinct from a map control.

calebpollman: Makes sense to me. Should we update the name of `StandaloneGeocoder` as well so we have a consistent naming pattern?

slaymance: I'm down with that in this PR. Any good ideas? Should it just be `GeocoderStandalone`?

calebpollman: SGTM!

TreTuna (approved):

PR #1752, opened 2022-04-21
security: remove dangerouslySetInnerHTML from AmplifyProvider

joebuono:
#### Description of changes

Our previous solution for preventing server-side XSS involved adding a regex to filter for closing `</style>` tags, which could be used to escape from the CSS context and then run scripts. 

However, we'd like to make some improvements
- use the `textContent` property instead 
- avoid using a blacklist regex, since we don't know all the possible payloads which could be used to trigger an XSS vulnerability.

Thus, we’ve come up with a better, simpler solution which avoids using `dangerouslySetInnerHTML` altogether:

```jsx
        <style id={`amplify-theme-${name}`} nonce={nonce} suppressHydrationWarning>
          {cssText}
        </style>
```

Surprisingly, setting the CSS string directly inside the `<style>` tag works. We originally thought that JSX would escape meaningful CSS characters (such as `'`, `"`', `<` and `>`) and break the CSS. However, CSS text gets escaped on the server side (encoded as `[data-amplify-theme=&quot;amplify-docs&quot;]`), but then hydrated correctly on the front end.

The only issue we saw was a "text content did not match" warning from NextJS. This warning is easily suppressed by passing the `suppressHydrationWarning` prop, and it does not cause any additional re-renders. 

I ran the docs and a sample app while making changes to the AmplifyProvider. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/packages/react/src/components/AmplifyProvider/index.tsx
+++ b/packages/react/src/components/AmplifyProvider/index.tsx
@@ -76,52 +76,14 @@ export function AmplifyProvider({
           without having to worry about specificity issues because this stylesheet
           will likely come after a user's defined CSS.
         */}
-      {/*
-          Q: Why are we using dangerouslySetInnerHTML?
-          A: We need to directly inject the theme's CSS string into the <style> tag without typical HTML escaping. 
-             For example, JSX would escape characters meaningful in CSS such as ', ", < and >, thus breaking the CSS. 
-
-          Q: Why not use a sanitization library such as DOMPurify?
-          A: For our use case, we specifically want to purify CSS text, *not* HTML. 
-             DOMPurify, as well as any other HTML sanitization library, would escape/encode meaningful CSS characters 
-             and break our CSS in the same way that JSX would. 
-
-          Q: Are there any security risks in this particular use case?
-          A: Anything set inside of a <style> tag is always interpreted as CSS text, *not* HTML.
-             Reference: “Restrictions on the content of raw text elements” https://html.spec.whatwg.org/dev/syntax.html#cdata-rcdata-restrictions
-             And in our case, we are using dangerouslySetInnerHTML to set CSS text inside of a <style> tag. 
-            
-             Thus, it really comes down to the question: Could a malicious user escape from the context of the <style> tag? 
-             For example, when inserting HTML into the DOM, could someone prematurely close the </style> tag and add a <script> tag?
-               e.g., </style><script>alert('hello')</script>
-             The answer depends on whether the code is rendered on the client or server side. 
-
-             Client side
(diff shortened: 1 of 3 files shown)


packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -76,52 +76,14 @@ export function AmplifyProvider({
...
-             as this the only way to escape from the context of the browser interpreting the text as CSS. 
-             We also need to catch cases where there is any kind of whitespace character </style[HERE]>, such as tabs, carriage returns, etc:
-             </style
-             
-             >
-             Therefore, by only rendering CSS text which does not include a closing '</style>' tag, 
-             we ensure that the browser will correctly interpret all the text as CSS. 
-        */}
-      {typeof theme === 'undefined' || /<\/style/i.test(cssText) ? null : (
+      {typeof theme === 'undefined' ? null : (
```
0618, line 79: Thanks! It looks a lot cleaner! ❤️ 

Would it be even more cleaner if we shorten this line as well?
```suggestion
      {!theme &&
```

0618 (approved):

packages/react/src/components/AmplifyProvider/index.tsx
reesscot: @0618 I think the existing check is safer since we need to explictly return `null` if the user has no theme. We might want to add the case of the `theme` being null as well though:
```jsx
{theme == null ? null : (
```

dbanksdesign (approved):
I try to avoid `&&` for conditional rendering in react: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx so I like the ternary operator. I just have a few questions to make sure this is buttoned up but mostly just to confirm it is working.

packages/react/src/components/AmplifyProvider/index.tsx
```diff
@@ -76,52 +76,14 @@ export function AmplifyProvider({
...
-             we ensure that the browser will correctly interpret all the text as CSS. 
-        */}
-      {typeof theme === 'undefined' || /<\/style/i.test(cssText) ? null : (
+      {theme == null ? null : (
         <style
           id={`amplify-theme-${name}`}
-          dangerouslySetInnerHTML={{ __html: cssText }}
           nonce={nonce}
-        />
+          suppressHydrationWarning
```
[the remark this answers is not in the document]
joebuono: Yes, just tested it in CRA and it works. I'm not sure how to test if there's a delay in styles getting applied

reesscot (changes requested):
Blocking PR as it causes a "flash of un-themed content"

Issue #1738, opened 2022-04-21
SelectField text not visible on docs in dark mode (Chrome on Windows)


cwoolum:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Other

### Which UI component?

Other

### How is your app built?

Docs

### Please describe your bug.

When opening a select list in dark mode on the docs, the select options are not themed. Not sure if this is only a docs issue.

https://ui.docs.amplify.aws/components/selectfield

![20279BCF-5311-4E7F-A391-3DE1CEA72BA3](https://user-images.githubusercontent.com/942415/164387194-ce75ba85-d652-453e-92e8-a35de6d29cc2.GIF)
d

### What's the expected behaviour?

The options be themed

### Help us reproduce the bug!

Open the docs, set the site to dark mode with any theme, open a select field anywhere on the page.

### Code Snippet

```typescript
// Put your code below this line.

```

zchenwei:
@cwoolum, which browser are you using? I don't see this issue happen in Chrome and FireFox

cwoolum:
Chrome 100 on Windows. It seems to work fine in FireFox

PR #1755, opened 2022-04-21
fix: Resets select option text color to fix windows darkmode bug

hbuchel:
#### Description of changes

This is essentially setting the text color of our `<option />` elements to the browser default to fix the following issue:  

The SelectField options are not legible in Chrome on Windows in dark mode. While the text color from dark mode cascades down (a very light gray), we do not update the background color.

Alternatively, we could consider theming of the option dropdown for browsers that allow it; Windows on Chrome allows you to style the background color of the options dropdown. That will require some further consideration for creating tokens for the font color as well. The fix in this PR works with our current theming options so seems the better fit for now.

**Before**

before

**After**

after

#### Issue #, if available

Fixes: https://github.com/aws-amplify/amplify-ui/issues/1738

Checked in a number of other browsers with the change in place in local docs instance:

dropdowns

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/ui/src/theme/css/component/select.scss
+++ b/packages/ui/src/theme/css/component/select.scss
@@ -30,6 +30,11 @@
     }
   }
 
+  // For Chrome on Windows, fixes darkmode text color
+  option {
+    color: initial;
+  }
+
   &[data-size='small'] {
     min-width: var(--amplify-components-select-small-min-width);
   }
--- a/.changeset/chatty-scissors-change.md
+++ b/.changeset/chatty-scissors-change.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui": patch
+---
+
+fix: Resets select option text color to fix windows darkmode bug


zchenwei (approved):
LGTM

reesscot (approved):
🛳

packages/ui/src/theme/css/component/select.scss
```diff
@@ -30,6 +30,11 @@
     }
   }
 
+  // For Chrome on Windows, fixes darkmode text color
+  option {
```
reesscot, line 34: My only question is whether we want to namespace this inside of our own class? This could change the default for other library code the app developer may be using?

hbuchel: Hm do you mean making this a flat `.amplify-select__option {}`? or something else? Currently it's `.amplify-select option {}`. 

I thought about that first option but our SelectField allows passing native `<option>` elements as [children](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/SelectField/SelectField.tsx#L28), so I think we'd have to do some fancy iterating on `children` to place this class name on them (and making sure we're not removing any classes/attributes a customer might customize on their `<option>`s)

PR #1758, opened 2022-04-21
chore(ui-react): add additional linting rules and ignorePatterns for eslint

calebpollman:
#### Description of changes

Introduce new `eslint` rules for _@aws-amplify/ui-react_:
- add `eslint` rules and related devDeps
- update existing linting devDep versions
- update _ui-react_ `build` command to include `lint` to catch missed development linting errors
- fix linting errors in _src/helpers_
- add `eslint` vscode extension to recommended vscode extension list
- add _tsconfig.dev.json_ for type checking entire package
- add linting on precommit for _ui-react_
- add temporary `eslint.ignorePatterns` values for:
```
  'src/components', //  277 problems (276 errors, 1 warning)
  'src/hooks', // 147 problems (147 errors, 0 warnings)
  'src/primitives', // 772 problems (772 errors, 0 warnings)
  'src/primitives/Icon/icons', // 9669 problems (9669 errors, 0 warnings)
```

Incremental follow up pull requests will be made removing the temporary values from `eslint.ignorePatterns` as the code in their respective directories is updated to pass the new rules (with the exception of _src/primitives/Icon/icons_ as the _icons_ directory will be removed in a future release).

#### Issue #, if available
NA

`yarn react lint`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated


packages/react/tsconfig.json
```diff
@@ -17,6 +17,6 @@
     "module": "esnext",
     "target": "es2015"
   },
-  "exclude": ["node_modules", "src/**/*.test.ts", "src/**/*.test.tsx"],
+  "exclude": ["node_modules"],
```
ErikCH, line 20: I'm not sure why we excluded tests from our tsconfig. But I guess it's fine not to exclude them.

ErikCH, line 20: I'm not sure why we excluded tests from our tsconfig. But I guess it's fine not to exclude them.

zchenwei: Last time I was trying to remove them but then I found `.d.ts` will also be generated for test files and shipped along with other code. Can you verify that?

calebpollman: That was a miss on my end, going to put them back. 

That being said, it did allow me to catch an issue where one of the `AmplifyProvider` unit tests had an import from `@aws-amplify/ui-react`, which makes me think that there is probably value in adding a `tsconfig.dev.json` to run our type checking against in dev, but that should be a separate task

zchenwei: Does ESslint always reference the settings in `tsconfig.json`? Is it able to override the exclude in its own config file?

calebpollman: Type checking and eslint are tangental to each other, but for TS related rules we need to point eslint towards a `tsconfig` file. The `excludes` property is specific for type compilation in TS

zchenwei (approved):

packages/react/.eslintrc.js
```diff
@@ -0,0 +1,122 @@
...
+  root: true,
+  ignorePatterns: [
+    '.eslintrc.js',
+    'jest.config.js',
+    'tsup.config.ts',
+    'scripts',
+    'dist',
+    'node_modules',
+
+    // TODO Delete this file once the above directories are no longer ignored
```
reesscot, line 12: What does this comment mean?

reesscot (approved):
Shipit

packages/react/.eslintrc.js
calebpollman: Kind of just took action on this but since the `plugin:react/recommended` rules include importing React in related files, and that we **sometimes** do it already, that we should delete _react-shim.js_ once the entire package is passing linting 😉

reesscot: I think we may not need the `react-shim.js` file at all.  My understanding is that importing `React` is no longer required unless it's actually used in the component:
https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

Looks like we would need to add `plugin:react/jsx-runtime` according to docs: https://www.npmjs.com/package/eslint-plugin-react

Issue #678, opened 2021-11-11
Switching themes wipes out state


dbanksdesign:
https://user-images.githubusercontent.com/321279/141253210-78b99eaf-6b41-4c35-bc1e-36f3d1246406.mp4

Need to investigate whether this happens with Class components as well.

PR #1761, opened 2022-04-22
Docs bug fix (draft) - Switching themes wipes out state

joebuono:
#### Problem
- When a customer toggles light/dark mode on the Amplify UI Docs site, it wipes out state in the component demos: https://github.com/aws-amplify/amplify-ui/issues/678
- This docs bug has been outstanding for 5 months, so it’d be nice to get it solved. 

#### What we’ve tried
- @dbanksdesign told me that they’d narrowed it down to how the docs site does dynamic imports, which give us a lot of performance benefits. 
- After spending a few days diving into the weeds of NextJS, I wasn’t able to find a way to prevent NextJS from re-rendering the page and wiping out state when the user toggles light/dark mode. 
- I also tried memoizing various components in the render path, but that didn’t work either. 

#### Proposed solution
- Basically, we just want to lift state up from the demos so that it doesn’t get wiped out after each re-render. 
- To do this, I’ve declared a `Map` object above `MyApp` at the top of the component tree, which functions as a very simple datastore:

```js
export const DemoState = new Map();

function MyApp({ Component, pageProps }) {
   // code...
}
```

Then, in each demo, we can use the `Map` to get and set demo props:

```js
import { DemoState } from 'src/pages/_app.page';

export const AlertDemo = () => {
  const alertProps = useAlertProps(
    DemoState.get('alert') || {
      isDismissible: false,
      hasIcon: true,
      heading: 'Alert heading',
      body: 'This is the alert message',
    }
  );

  React.useEffect(() => {
    DemoState.set('alert', alertProps);
  });
// code...
```

In the example above, we make a call to get the props for the `'alert'` demo from the `Map` object. If they exist, then `alertProps` are set to what they were before the alert demo got re-rendered, thus persisting the demo’s state. If `get` returns `undefined`, then the default props get applied. 

Every time the `alertProps` change, this triggers the `useEffect` hook to set the demo's state on the `Map` object. 
- Ideally, we’d only have to set state when the customer toggles light/dark mode, but we have no way of knowing when they’ll toggle and re-render the page

#### Tradeoffs
- I’ll have to add a few lines of code to each component demo in order to get and set component state
- Worst case, the `Map` could store state for every component if a customer went through each demo without doing a browser refresh. But even so, that wouldn’t take up much memory at all. 

I am totally open to comments/suggestions. Please let me know if you can think of a better way to solve this bug!

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/678

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/pages/_app.page.tsx
+++ b/docs/src/pages/_app.page.tsx
@@ -15,6 +15,8 @@ import '../styles/index.scss';
 // @ts-ignore Cannot assign to 'useLayoutEffect' because it is a read-only property.ts(2540)
 if (typeof window === 'undefined') React.useLayoutEffect = React.useEffect;
 
+export const DemoState = new Map();
+
 function MyApp({ Component, pageProps }) {
   const router = useRouter();
   const { platform = 'react' } = router.query;
--- a/docs/src/pages/components/alert/demo.tsx
+++ b/docs/src/pages/components/alert/demo.tsx
@@ -1,8 +1,9 @@
+import * as React from 'react';
 import { Alert } from '@aws-amplify/ui-react';
-
 import { Demo } from '@/components/Demo';
 import { AlertPropControls } from './AlertPropControls';
 import { useAlertProps } from './useAlertProps';
+import { DemoState } from 'src/pages/_app.page';
 
 const propsToCode = (props) => {
   return (
@@ -21,11 +22,17 @@ const propsToCode = (props) => {
 };
 
 export const AlertDemo = () => {
-  const alertProps = useAlertProps({
-    isDismissible: false,
-    hasIcon: true,
-    heading: 'Alert heading',
-    body: 'This is the alert message',
+  const alertProps = useAlertProps(
+    DemoState.get('alert') || {
+      isDismissible: false,
+      hasIcon: true,
+      heading: 'Alert heading',
+      body: 'This is the alert message',
+    }
+  );
+
+  React.useEffect(() => {
+    DemoState.set('alert', alertProps);
   });
 
   return (


dbanksdesign:
I like this direction. This could also help keep the demo state in the URL as well so users could share a URL that has the demo props set. Is there a performance hit for storing the demo state outside of React in a simple Map? Like would it cause unnecessary re-renders? Is there a benefit to using Map versus creating a React context? These aren't leading questions, using a plain Map might be the best performance-wise, I just want to make sure.

joebuono:
> 

@dbanksdesign Regarding the questions in your comment:

_Is there a benefit to using Map versus creating a React context?_
- Since the props for each component's demo do not need to be shared across the entire docs site, I wouldn't really consider these props to be "global" state in the same way that a current authenticated user, theme, or preferred language would be ([see React Context docs](https://reactjs.org/docs/context.html#:~:text=Context%20is%20designed%20to%20share%20data%20that%20can%20be%20considered%20%E2%80%9Cglobal%E2%80%9D%20for%20a%20tree%20of%20React%20components%2C%20such%20as%20the%20current%20authenticated%20user%2C%20theme%2C%20or%20preferred%20language.)). 
- Since React Context is primarily used when some data needs to be accessible by _many_ components at different nesting levels ([see React Context docs](https://reactjs.org/docs/context.html#:~:text=Context%20is%20primarily%20used%20when%20some%20data%20needs%20to%20be%20accessible%20by%20many%20components%20at%20different%20nesting%20levels.)), I don’t think our particular docs bug fits that use case. 
- The main benefit of using `Map` is that it provides a straightforward way to persist data without the complexity of state management. 
- Beyond the simple `get` and `set` methods exposed by the `Map` object, I can't think of any additional patterns of updating demo state that we'd need to explicitly codify through context/hooks or Redux. 
- For our use case, `Map` is a lightweight solution to persist component demo state across NextJS re-renders: no more, no less. 

_Is there a performance hit for storing the demo state outside of React in a simple Map? Like would it cause unnecessary re-renders?_ 
- Using a `Map` does not cause unnecessary re-renders, while using React Context did (at least in my implementation, which could certainly be incorrect/suboptimal). We could also implement memoization for each demo, but that seems excessive compared to the simple `Map` solution.

dbanksdesign:
Thanks for the explanation, I'm sold on this approach. I think my only change would be to move the DemoState out of 'src/pages/_app.page' because it doesn't seem like it belongs there. Maybe put it in 'utils' or 'data'?

Issue #1790, opened 2022-04-28
MapView component is missing the possibility to forward ref and doesn't seem to work with useMap-hook


Defrosted:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Create React App

### Please describe your bug.

The MapView wrapper component doesn't take a ref parameter at all so that it would be possible to access the Map API of the underlying Map-component from the react-map-gl package (as described in the [documentation](https://visgl.github.io/react-map-gl/docs/api-reference/map) for the Map component). The useMap hook provided by react-map-gl doesn't seem to work with the MapView component either, forcing me to use the pure JS maplibre-gl-js-amplify package instead of React components. I didn't delve a lot into the useMap-hook option, since it would require me to do extensive refactoring.

### What's the expected behaviour?

The expected behaviour would be to be able to pass refs and use the useMap-hook as documented by [react-map-gl](https://visgl.github.io/react-map-gl/docs/api-reference/map) documentation the MapView is claiming to be fully compatible with.

Example for original Map component from the react-map-gl documentation
```typescript
import * as React from 'react';
import Map from 'react-map-gl';

function App() {
  const mapRef = React.useRef();

  const onMapLoad = React.useCallback(() => {
    mapRef.current.on('move', () => {
      // do something
    });
  }, []);

  return <Map ref={mapRef} onLoad={onMapLoad} />;
}
```

### Help us reproduce the bug!

Trying to pass a ref to the MapView component isn't supported and is caught by Typescript.

```typescript
import { MapView } from '@aws-amplify/ui-react'
import { useRef } from 'react';
import { MapRef } from 'react-map-gl';
const Component: React.FC = () => {
  const mapRef = useRef<MapRef | undefined>(undefined);
  return (
    <MapView ref={mapRef} />
  );
}
```

### Additional information and screenshots

@aws-amplify/ui-react version 2.15.6
maplibre-gl version 1.15.3
react-map-gl version 7.0.11

Defrosted:
[...]

PR #1791, opened 2022-04-28
feat(geo): Forward ref to react-map-gl component

slaymance:
#### Description of changes
`react-map-gl` supports passing a `ref` to the `Map` component to grant access to the underlying map object to add event callbacks and such ([see the docs](https://visgl.github.io/react-map-gl/docs/api-reference/map#methods)). This PR provides this functionality for users of `MapView` by forwarding the `ref` prop to the wrapped component and updating the TypeScript type.

#### Issue #, if available
#1790 

Ensuring e2e tests function is sufficient validation.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


.changeset/nine-pillows-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/ui-react': minor
+---
+
+Add forward ref functionality to MapView
```
slaymance, line 5: Minor makes sense here? The `ref` property was not valid prior to this change so that was my reasoning.

zchenwei: Right, we publish `ref` on other primitives as minor

zchenwei (commented):
Overall, I am good with it. Can we add a unit test to make sure the `ref` prop gets passed through? Like how we do in primitive tests

packages/react/src/components/Geo/MapView/index.tsx
```diff
@@ -21,66 +21,64 @@ export type MapViewProps = Omit<MapProps, 'transformRequest'>;
...
-      amplifyConfig.geo?.AmazonLocationService ??
-      {},
-    [amplifyConfig]
-  );
-  const [transformRequest, setTransformRequest] = useState<
-    TransformRequestFunction | undefined
-  >();
+export const MapView = forwardRef<MapRef, MapViewProps>(
+  ({ mapLib, mapStyle, style, ...props }, ref) => {
+    const amplifyConfig = Amplify.configure() as any;
```
reesscot, line 26: non-blocking: can we do anything to improve the typing here?

reesscot (approved):
LGTM

slaymance:
I don't think unit testing this will be straightforward since the map needs a functioning Amplify config to render.

calebpollman (approved):
LGTM 🚢

zchenwei (approved):

PR #1792, opened 2022-04-28
remove all inline styling from angular authenticator

jacoblogan:
#### Description of changes
This pull request removes all of the inline styling from the authenticator component in angular

Visual inspection through examples and cypress e2e tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-in/sign-in.component.html
```diff
@@ -18,8 +21,14 @@
 
 <amplify-slot name="sign-in-footer" [context]="context">
   <div data-amplify-footer>
-    <button amplify-button fontWeight="normal" size="small" variation="link" fullWidth="true"
-      (click)="authenticator.toResetPassword()">
+    <button
+      amplify-button
+      fontWeight="normal"
```
zchenwei, line 26: Maybe I miss something here. Shouldn't we avoid using styling props as well? This will block customers using strict CSP rules as well.

zchenwei (approved):
LGTM! Just have a question

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-in/sign-in.component.html
jacoblogan: Yes you are correct, and this will still throw errors for the strictest of CSP rules, but it does work with angular's recommended csp rule https://angular.io/guide/security#content-security-policy

wlee221 (approved):
lgtm, thanks!

reesscot (approved):
LGTM

Issue #1784, opened 2022-04-27
Reset password submit button has incorrect label 


mikejpeters:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Angular

### Which UI component?

Authenticator

### How is your app built?

Webpack

### Please describe your bug.

When using the `<amplify-authenticator>` Angular component, with no customizations, the reset password button says "Send Code" even though the code was sent in the  previous step. 

![image](https://user-images.githubusercontent.com/3334654/165633586-a3a4d249-d495-4e8b-bdd9-318b81c072fc.png)

### What's the expected behaviour?

The button should be labelled something more clear like "Submit" or "Change Password"

### Help us reproduce the bug!

I think this happens in any situation the Angular authenticator is used, during the reset password flow.

Looking at the Angular component, it seems like the wrong label is being set here:

https://github.com/aws-amplify/amplify-ui/blob/8b72277ab8bd7ad64fa298a9d509572318ac8db2/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-reset-password/amplify-confirm-reset-password.component.ts#L18

https://github.com/aws-amplify/amplify-ui/blob/8b72277ab8bd7ad64fa298a9d509572318ac8db2/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-reset-password/amplify-confirm-reset-password.component.html#L16-L18

Compared to the Vue component, which seems more correct:

https://github.com/aws-amplify/amplify-ui/blob/8b72277ab8bd7ad64fa298a9d509572318ac8db2/packages/vue/src/components/confirm-reset-password.vue#L38

https://github.com/aws-amplify/amplify-ui/blob/8b72277ab8bd7ad64fa298a9d509572318ac8db2/packages/vue/src/components/confirm-reset-password.vue#L100-L108

### Code Snippet

```html
<div class="ion-padding">
  <amplify-authenticator>
    <ng-template amplifySlot="authenticated">
      <ion-app>
        <ion-router-outlet></ion-router-outlet>
      </ion-app>
    </ng-template>
[...]

PR #1793, opened 2022-04-28
fix(authenticator): Update ConfirmResetPassword submit text for Angular

slaymance:
#### Description of changes
The text for the submit button on the `ConfirmResetPassword` screen incorrectly showed `Send Code` even when the code was already sent during the `ResetPassword` state. This PR updates the text to `Submit` to match the experience in the React and Vue implementations.

Before:
Screen Shot 2022-04-28 at 14 00 58

After:
Screen Shot 2022-04-28 at 13 57 44

#### Issue #, if available
#1784 

Manual visual validation.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated - N/A

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-reset-password/amplify-confirm-reset-password.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-reset-password/amplify-confirm-reset-password.component.ts
@@ -3,6 +3,7 @@ import { AuthenticatorService } from '../../../../services/authenticator.service
 import {
   FormFieldsArray,
   getFormDataFromEvent,
+  hasTranslation,
   translate,
 } from '@aws-amplify/ui';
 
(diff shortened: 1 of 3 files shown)


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-reset-password/amplify-confirm-reset-password.component.ts
```diff
@@ -15,7 +15,7 @@ export class ConfirmResetPasswordComponent {
   @Input() public headerText = translate('Reset your password');
 
   // translated strings
-  public sendCodeText = translate('Send Code');
+  public submitText = translate('Submit');
```
wlee221, line 18: Open to your opinion, do you think this is a breaking change?

Technically yes for customers who were translating the "submit" text, but at the same time this is a technically bug.

wlee221: If yes, we do have a pattern for adding backwards compatible translations:

https://github.com/aws-amplify/amplify-ui/blob/331df831fd7ddeb8c6f28b4cb385d23d1aa524be/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/sign-in/sign-in.component.ts#L24-L28

jacoblogan: +1 for supporting backwards compatible translations, how do we feel about moving this into the translate function, so that the translate function can also accept a list of strings and if so returns the first one with a match?

jacoblogan (approved):
LGTM, lets just decide how we want to handle the translation issue.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/confirm-reset-password/amplify-confirm-reset-password.component.ts
slaymance: I agree we should try to make this as backwards compatible as possible. However, since `Send Code` is a string that could appear elsewhere in the Authenticator, we would be defaulting to whatever a user translates for that string instead of `'Submit'`.

In other words, if a user translates `Send Code`, they can never translate the `Submit` text for confirm-reset-passwrd.

wlee221 (approved):
🛳

PR #1798, opened 2022-04-29
fix(react): Compatibility with React 18 `useSyncExternalStore` hook

### LINKED ISSUE #1780: NextJS/Authenticator error: "Missing getServerSnapshot, which is required for server-rendered content."

React

### Which UI component?

Authenticator

### How is your app built?

Create Next App

### Please describe your bug.

When scaffolding a new Next.js app via `npx create-next-app` and trying to use the `Authenticator` component as documented, the app immediately breaks with a `500` internal server error:
```
Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
```

Screen Shot 2022-04-26 at 3 43 00 PM

The Authenticator component should render without any error and the app should build properly with `yarn dev`.

When the `Authenticator` is removed from `index.js` the app renders as expected.

Also, I tried both with/without wrapping `_app.js` with `Authenticator.Provider` and the error occurs in both approaches.

I feel like I'm maybe doing something wrong with my setup but can't put my finger on it. The [Amplify "Getting Started" docs for Next.js](https://docs.amplify.aws/start/getting-started/data-model/q/integration/next/#api-with-server-side-rendering-ssr) still uses `import { AmplifyAuthenticator } from '@aws-amplify/ui-react-v1'`, but I wanted to work on some use cases for the new Authenticator with Next and haven't found any relevant documentation on either the JS or UI docs site. I will update docs to use the new Authenticator component, but first want to make sure I am setting things up properly myself first.

Thanks for the help!

1. `npx create-next-app@latest next-example`
2. `cd next-example && code .`
3. `yarn add aws-amplify @aws-amplify/ui-react`
4. `amplify add auth`
5. `amplify push`
6. `amplify init` (default options)
7. Import the Authenticator component into `index.js` (snippet attached)
8. `yarn dev`
9. The error occurs immediately

### Code Snippet

```js
// pages/index.js

import { Authenticator } from "@aws-amplify/ui-react";
import "@aws-amplify/ui-react/styles.css";

export default function Home() {
  return (
    <Authenticator>
      {({ signOut, user }) => (
        <main>
          <h1>Hello {user.username}</h1>
          <button onClick={signOut}>Sign out</button>
        </main>
      )}
    </Authenticator>
  );
}
```

```js
// pages/_app.js

import { Amplify } from "aws-amplify";
import { Authenticator } from "@aws-amplify/ui-react";

import "../styles/globals.css";
import "@aws-amplify/ui-react/styles.css"; // default theme
import awsExports from "../src/aws-exports";

Amplify.configure({ ...awsExports, ssr: true });

export default function App({ Component, pageProps }) {
  return (
    <Authenticator.Provider>
      <Component {...pageProps} />
    </Authenticator.Provider>
  );
}
```

```
node -v
v16.14.2
```

```
{
  "name": "webpack-5-testing",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start",
    "lint": "next lint"
  },
  "dependencies": {
    "@aws-amplify/ui-react": "^2.15.5",
    "aws-amplify": "^4.3.20",
    "next": "12.1.5",
    "react": "18.0.0",
    "react-dom": "18.0.0"
  },
  "devDependencies": {
    "eslint": "8.14.0",
    "eslint-config-next": "12.1.5"
  }
}
```

wlee221:
## tl;dr

This PR bumps `@xstate/react` to `3.0.0` and also resolves compatibility problems with `useSyncExternalStore` hook shim that `@xstate/react` now depends on.

## Background

`Next.js` with React 18 and `@aws-amplify/ui-react` was breaking (#1780) due to compatibility problems with React 18  [`useSyncExternalStore` hook](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore). Note that xstate uses an official [shim](https://www.npmjs.com/package/use-sync-external-store) of the hook for backwards compatibility.

`@xstate/react@3.0.0` resolved this by rewriting its implementations to use `useSyncExternalStore` shim internally ([xstate#2939](https://github.com/statelyai/xstate/pull/2939)). While it fixed the build issue (#1780), we found that `useSyncExternalStore` does not work well with the `useSelector` pattern we used for Authenticator and caused it to be stuck. 

This PR resolves the problems above by rewriting our `useSelector` usage to be more idiomatic and compatible with `useSyncExternalStore`. More details in the comments.

#### Issue #, if available

Resolves #1780.

`npm pack`ed `packages/react` and verified CRA builds with React 18. Also checked with current React 17 examples.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -150,16 +150,7 @@ export const useAuthenticator = (selector?: Selector) => {
     return areArrayValuesEqual(prevDepsArray, nextDepsArray);
   };
 
-  const facade = useSelector(
-    service,
-    xstateSelector,
-    comparator,
-    /**
-     * Below parameter was needed to avoid "The result of getSnapshot should be
-     * cached to avoid an infinite loop" error on React 18
-     */
-    (interpreter) => interpreter.getSnapshot()
-  );
+  const facade = useSelector(service, xstateSelector, comparator);
 
   return {
     ...facade,
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -41,7 +41,7 @@ export const Provider = ({ children }) => {
    *
    * Leaving this as is for now in the interest of suggested code guideline.
    */
-  const service = useInterpret(createAuthenticatorMachine);
+  const service = useInterpret(createAuthenticatorMachine, { devTools: true });
   const currentProviderVal = { service };
 
   const value = isEmpty(parentProviderVal)
@@ -59,11 +59,11 @@ export const Provider = ({ children }) => {
     }
 
     isListening.current = true;
-    return listenToAuthHub(send);
+    return listenToAuthHub(send as any);
   }, [send]);
 
   return (
-    <AuthenticatorContext.Provider value={value}>
+    <AuthenticatorContext.Provider value={value as any}>
       {children}
     </AuthenticatorContext.Provider>
   );
@@ -105,7 +105,7 @@ const useAuthenticatorService = () => {
   return service;
 };
 
-export const useAuthenticator = (selector?: Selector) => {
+export const useAuthenticator = (selector?: Selector, debug = false) => {
   const service = useAuthenticatorService();
 
   const { send } = service;
@@ -174,12 +174,25 @@ export const useAuthenticator = (selector?: Selector) => {
     const prevDepsArray = selector(prevFacade);
     const nextDepsArray = selector(nextFacade);
 
+    if (debug) console.log(_prevState, nextState);
+
     // Shallow compare the array values
     // TODO: is there a reason to compare deep at the cost of expensive comparisons?
-    return areArrayValuesEqual(prevDepsArray, nextDepsArray);
+    const equal = areArrayValuesEqual(prevDepsArray, nextDepsArray);
+    console.log(equal);
+    return equal;
   };
 
-  const state = useSelector(service, xstateSelector, comparator);
+  const state = useSelector(
+    service,
+    xstateSelector,
+    comparator,
+    (interpreter) => {
+      console.log(interpreter.getSnapshot());
+      return interpreter.getSnapshot();
+    }
+  );
+  if (debug) console.log('state', state);
 
   return {
     ...getFacade(state),
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -133,12 +133,6 @@ export const useAuthenticator = (selector?: Selector) => {
    * re-render. Does a deep equality check.
    */
   const comparator = (
-    /**
-     * We do not use `_prevState`, because it holds a *reference* to actor
-     * object, of which value could easily mutate between compare calls.
-     *
-     * Instead, we'll use prevFacadeRef for comparison.
-     */
     prevFacade: ReturnType<typeof getFacade>,
     nextFacade: ReturnType<typeof getFacade>
   ) => {
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -128,13 +128,7 @@ export const useAuthenticator = (selector?: Selector, debug = false) => {
    * To provide a consistent set of facade, we let the `selector` trivially return
    * itself and let comparator decide when to re-render.
    */
-  const xstateSelector = (state: AuthMachineState) => state;
-
-  /**
-   * Holds a snapshot copy of last previous facade values. Will be used
-   * on state changes to see if any of facade values have changed.
-   */
-  const prevFacadeRef = React.useRef<ReturnType<typeof getFacade>>();
+  const xstateSelector = (state: AuthMachineState) => getFacade(state);
 
   /**
    * comparator decides whether or not the new authState should trigger a
@@ -147,25 +141,10 @@ export const useAuthenticator = (selector?: Selector, debug = false) => {
      *
      * Instead, we'll use prevFacadeRef for comparison.
      */
-    _prevState: AuthMachineState,
-    nextState: AuthMachineState
+    prevFacade: ReturnType<typeof getFacade>,
+    nextFacade: ReturnType<typeof getFacade>
   ) => {
     if (!selector) return false;
-
-    /**
-     * We only trigger re-render if any of values in specified selected
-     * values change. First compute the facade for prev and next state.
-     */
-    const prevFacade = prevFacadeRef.current;
-    const nextFacade = getFacade(nextState);
-
-    /**
-     * prevFacadeRef can now be updated with new facade values
-     */
-    prevFacadeRef.current = nextFacade;
-
-    // If this is the first time comparator is called, return false
-    if (!prevFacade) return false;
 
     /**
      * Apply the passed in `selector` to get the value of their desired
@@ -174,30 +153,26 @@ export const useAuthenticator = (selector?: Selector, debug = false) => {
     const prevDepsArray = selector(prevFacade);
     const nextDepsArray = selector(nextFacade);
 
-    if (debug) console.log(_prevState, nextState);
+    if (debug) console.log(prevFacade.route, nextFacade.route);
+    if (debug) console.log(prevDepsArray, nextDepsArray);
 
     // Shallow compare the array values
     // TODO: is there a reason to compare deep at the cost of expensive comparisons?
-    const equal = areArrayValuesEqual(prevDepsArray, nextDepsArray);
-    console.log(equal);
-    return equal;
+    return areArrayValuesEqual(prevDepsArray, nextDepsArray);
   };
 
-  const state = useSelector(
+  const facade = useSelector(
     service,
     xstateSelector,
     comparator,
-    (interpreter) => {
-      console.log(interpreter.getSnapshot());
-      return interpreter.getSnapshot();
-    }
+    (interpreter) => interpreter.getSnapshot()
   );
-  if (debug) console.log('state', state);
+  if (debug) console.log('state', service.getSnapshot());
 
   return {
-    ...getFacade(state),
+    ...facade,
     /** @deprecated For internal use only */
-    _state: state,
+    _state: service.getSnapshot(),
     /** @deprecated For internal use only */
     _send: send,
   };
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -105,7 +105,7 @@ const useAuthenticatorService = () => {
   return service;
 };
 
-export const useAuthenticator = (selector?: Selector, debug = false) => {
+export const useAuthenticator = (selector?: Selector) => {
   const service = useAuthenticatorService();
 
   const { send } = service;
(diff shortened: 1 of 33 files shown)


packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -121,52 +121,23 @@ export const useAuthenticator = (selector?: Selector) => {
...
+   * This is to hide the internal xstate implementation details to customers.
    */
-  const xstateSelector = (state: AuthMachineState) => state;
-
-  /**
-   * Holds a snapshot copy of last previous facade values. Will be used
-   * on state changes to see if any of facade values have changed.
-   */
-  const prevFacadeRef = React.useRef<ReturnType<typeof getFacade>>();
+  const xstateSelector = (state: AuthMachineState) => getFacade(state);
```
wlee221, line 129: Instead of using the hacky `selector` that returns itself, we make the `selector` more explicit by having it return ("select") facade values inferred from current state.

This aligns with the intended purpose of `useSelector`, and works well with `useSyncExternalStore` hook under the hood as well.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -121,52 +121,23 @@ export const useAuthenticator = (selector?: Selector) => {
...
-   * itself and let comparator decide when to re-render.
+   * This is to hide the internal xstate implementation details to customers.
    */
-  const xstateSelector = (state: AuthMachineState) => state;
-
-  /**
-   * Holds a snapshot copy of last previous facade values. Will be used
-   * on state changes to see if any of facade values have changed.
-   */
-  const prevFacadeRef = React.useRef<ReturnType<typeof getFacade>>();
```
wlee221, line 137: This is no longer needed -- see comment below on `_prevState`.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -121,52 +121,23 @@ export const useAuthenticator = (selector?: Selector) => {
...
    * comparator decides whether or not the new authState should trigger a
    * re-render. Does a deep equality check.
    */
   const comparator = (
-    /**
-     * We do not use `_prevState`, because it holds a *reference* to actor
-     * object, of which value could easily mutate between compare calls.
-     *
-     * Instead, we'll use prevFacadeRef for comparison.
-     */
```
wlee221, line 149: This is no longer needed for two reasons:

1. `ref` comparison is done internally from React 18's `useSyncExternalStore` hook:

https://github.com/facebook/react/blob/ce13860281f833de8a3296b7a3dad9caced102e9/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js#L131-L140

2. With `selector` now returning a flat facade values instead of xstate `state` references, we're free of the reference comparison problem mentioned in this comment.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -121,52 +121,23 @@ export const useAuthenticator = (selector?: Selector) => {
...
-    const nextFacade = getFacade(nextState);
-
-    /**
-     * prevFacadeRef can now be updated with new facade values
-     */
-    prevFacadeRef.current = nextFacade;
-
-    // If this is the first time comparator is called, return false
-    if (!prevFacade) return false;
-
```
wlee221, line 169: With us no longer relying on `React.useRef`s, these can be removed ✨

packages/ui/src/types/authenticator/stateMachine/authMachine.ts
```diff
@@ -9,7 +9,7 @@ import { AuthEvent } from './event';
 /**
  * Intefrace for `authMachine` machine interpreter
  */
-export type AuthInterpreter = Interpreter<AuthContext, any, AuthEvent>;
+export type AuthInterpreter = Interpreter<AuthContext, any, AuthEvent, any>;
```
wlee221, line 12: The new fourth argument is for `xstate`'s `TypeState` TS [feature](https://xstate.js.org/docs/guides/typescript.html#typestates), which we don't use yet. TypeScript was getting confused without us explicitly passing this, so adding `any`. Shouldn't affect any of our typing as we don't use this.

packages/ui/src/types/authenticator/stateMachine/authMachine.ts
```diff
@@ -9,7 +9,13 @@ import { AuthEvent } from './event';
...
  * Intefrace for `authMachine` machine interpreter
  */
-export type AuthInterpreter = Interpreter<AuthContext, any, AuthEvent>;
+export type AuthInterpreter = Interpreter<
+  AuthContext,
+  any,
+  AuthEvent,
+  any,
+  any
+>;
```
wlee221, line 18: The last two generics argument is for `xstate`'s [`TypeState`](https://xstate.js.org/docs/guides/typescript.html#typestates) and [`Typegen`](https://xstate.js.org/docs/guides/typescript.html#typegen) features. We don't use these features, but TS was getting confused when these generics weren't provided. Adding `any`s to resolve it. This should not affect actual typings we depend on.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: This is where the compatibility problem occurred -- `useSyncExternalStore` was thinking `state` was not changing because the previous `selector` was just returning the same state object ref.

packages/ui/src/types/authenticator/stateMachine/authMachine.ts
calebpollman: Can we please add a backlog task (and corresponding code comment) to address this? Even if we are not using these features currently, tightening up on `any` usage will make it much easier to transition the package over to TS strict mode when we are ready to move forward with that

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -179,12 +150,21 @@ export const useAuthenticator = (selector?: Selector) => {
...
-  const state = useSelector(service, xstateSelector, comparator);
+  const facade = useSelector(
+    service,
+    xstateSelector,
+    comparator,
+    /**
+     * Below parameter was needed to avoid "The result of getSnapshot should be
+     * cached to avoid an infinite loop" error on React 18
+     */
+    (interpreter) => interpreter.getSnapshot()
```
calebpollman, line 161: I'm not entirely certain what is happening here but what is preventing us from caching the result of `getSnapshot`? Prefer to avoid adding more workarounds (maybe this is not a workaround, but that's what i inferred from the comment)

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -179,12 +150,21 @@ export const useAuthenticator = (selector?: Selector) => {
...
+  );
 
   return {
-    ...getFacade(state),
+    ...facade,
     /** @deprecated For internal use only */
-    _state: state,
+    _state: service.getSnapshot(),
```
calebpollman, line 167: Assuming `service.getSnapshot` returns a new object on each call, does computing it for each call to `useAuthenticator` have any impact on re-renders? Follow up question, from what i understand, since our usage of `_state` is not for handling dynamic values, maybe we should consider assigning its value to a ref that only computes on the first call to `useAuthenticator`?

calebpollman (commented):
Thanks for addressing this @wlee221. Left some non-blocking questions/feedback

ErikCH (approved):
Looks good to me. I am interested to hear more about `getSnapShot` but other then that this makes sense.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
slaymance: Very nice! Good find, totally makes sense why we weren't getting errors then.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
slaymance: I'm not seeing the same error if I delete this `getSnapshot` parameter from the `useSelector` call. The [xstate docs](https://xstate.js.org/docs/packages/xstate-react/#useselector-actor-selector-compare-getsnapshot) mention the parameter is optional, but don't really give an example of when you would want to specify it. Are you able to walk me through replicating that error you saw?

packages/ui/src/types/authenticator/stateMachine/authMachine.ts
wlee221: I think it'd tie with existing task here: https://app.asana.com/0/1201736086077862/1201802820587074

I'll add additional code comment.

wlee221: 2f60d1a4bb0e0dd839ae79803986

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: Oops, I think I worded that incorrectly. This is not a workaround, and is an intended use of the fourth argument in [`useSelector`](https://xstate.js.org/docs/packages/xstate-react/#useselector-actor-selector-compare-getsnapshot) here. 

If we don't pass in this fourth argument, xstate guesses the usage and passes in a generic `getSnapshot` defined here https://github.com/statelyai/xstate/blob/c0b787d2a2334665af0ecb4ebc1ce5653fa3725c/packages/xstate-react/src/useSelector.ts#L12-L17. This led to the error mentioned in the comment, so I tightened it up by which `getSnapshot` our interpreter uses.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: Great catch! Here was where I was not feeling 100% too.

(1) This would not trigger more re-renders, but yes this is additional computation that we didn't have. This was now needed that we're using `selector` to get `facade` directly. This does not change the runtime complexity though, because `getSnapshot` was already getting called through `useSelector`.

 A long term solution would be to not depend on `_state` at all, or (2):

(2) I like the idea! I do want to scope it out of this PR though, because I want to audit and make sure what you said about `_state` is true.  e.g.

https://github.com/aws-amplify/amplify-ui/blob/3729ac84f3dd6b81e28e055edab41f9761384f9a/packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignIn.tsx#L11

depends on dynamic value. It might be worth more effort to investigate expanding `useAuthenticator` facade values so that `_state` doesn't get used like that there.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: @slaymance, sorry I commented before I saw yours!

I saw this error on our `examples/next` without the fourth option. Let me try to repro again in a clean environment.

wlee221: Oh wait, you might be totally correct @slaymance, I don't see that error anymore. Thanks for run testing my code!

slaymance (approved):
Great work!

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: Took a quick look through the implementation, `getSnapshot` just returns `_state` underneath the interpreter and does not return a new object. 

https://github.com/statelyai/xstate/blob/c0b787d2a2334665af0ecb4ebc1ce5653fa3725c/packages/core/src/interpreter.ts#L1394

So this is not an expensive operation. This explains why we were having `ref` comparison problems too!

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
wlee221: Removed [7af3c68](https://github.com/aws-amplify/amplify-ui/pull/1798/commits/7af3c68d948a4483c23192d0b69f16adf5ed251f) here, will resolve this.

calebpollman (approved):
LGTM 🚢

PR #1808, opened 2022-05-02
Fix: Geo test failing

ErikCH:
#### Description of changes
Fix to our CI tests, since GEO test is failing.

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/e2e/cypress/integration/ui/components/geo/map-with-geocoder/map-with-geocoder.steps.ts
+++ b/packages/e2e/cypress/integration/ui/components/geo/map-with-geocoder/map-with-geocoder.steps.ts
@@ -13,7 +13,7 @@ When('I press the enter key', () => {
 });
 
 When('I click on a map marker', () => {
-  cy.get('.maplibregl-marker').first().click();
+  cy.get('.maplibregl-marker').first().click({ force: true });
 });
 
 Then('I see markers equal to my default search results', () => {


packages/e2e/cypress/integration/ui/components/geo/map-with-geocoder/map-with-geocoder.steps.ts
```diff
@@ -13,7 +13,7 @@ When('I press the enter key', () => {
 });
 
 When('I click on a map marker', () => {
-  cy.get('.maplibregl-marker').first().click();
+  cy.get('.maplibregl-marker').first().click({ force: true });
```
calebpollman, line 16: The error i saw in my [test PR](https://github.com/aws-amplify/amplify-ui/pull/1806) was:
```
  1) Map with Geocoder
       Marker information popup:
     CypressError: Timed out retrying after 15050ms: `cy.click()` failed because this element:

`<div aria-label="Map marker" class="maplibregl-marker mapboxgl-marker maplibregl-marker-anchor-center mapboxgl-marker-anchor-center" tabindex="0" style="transform: translate(-50%, -50%) translate(674px, 272px) rotateX(0deg) rotateZ(0deg);">...</div>`

is being covered by another element:

`<path d="M27,13.5 C27,19.074644 20.250001,27.000002 14.75,34.500002 C14.016665,35.500004 12.983335,35.500004 12.25,34.500002 C6.7499993,27.000002 0,19.222562 0,13.5 C0,6.0441559 6.0441559,0 13.5,0 C20.955844,0 27,6.0441559 27,13.5 Z"></path>`

Fix this problem, or use {force: true} to disable error checking.
```

Somewhat concerned with force clicking since it implies that there is a covering element, do you think that this could lead to false positives in the future?

calebpollman (commented):
Thanks for picking this up so quickly @ErikCH! Do we know what the root cause is? It is a little scary to me that our [last merged commit](https://github.com/aws-amplify/amplify-ui/commit/04132afc16167b3fd2015d097a840c555a0cb3da) passed the integ tests but somehow they are failing now with no code changes

packages/e2e/cypress/integration/ui/components/geo/map-with-geocoder/map-with-geocoder.steps.ts
ErikCH: Since I can't reproduce it outside of the e2e tests, by manually clicking, this doesn't seem to be a real issue. But we could check with @slaymance if he has any guidance.

slaymance (approved):
I'm ok with this workaround to get CI passing.

packages/e2e/cypress/integration/ui/components/geo/map-with-geocoder/map-with-geocoder.steps.ts
ErikCH: It seems that the documentation on the cypress site states that force, forces the click, instead of waiting for `actionabality`

image


So it is abled to be clicked, it just gets confused because of the SVG icon. I don't think this would cause any `false` positives in the future, and we only use this method in one place so far. 

I can't predict how this `When` will be used in the future.

calebpollman (approved):
Still concerned about the root cause but that can be determined outside of this PR. Thanks again @ErikCH

Issue #1773, opened 2022-04-25
Browser crashes when refresh token is expired


ptaranto:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

NX/webpack

### Please describe your bug.

After successful login with tokens saved in local storage, launching the app when the refresh token is expired the browser tab crashes.

I was able to breakpoint it and check that tokens in local storage are cleared by CognitoUser.js, dispatchAuthEvent('tokenRefresh_failure', error, "Failed to retrieve new token"); is called by AuthClass, then it gets lost when running a generator function in Auth.js

### What's the expected behaviour?

user is redirected to the Authenticator login form

### Help us reproduce the bug!

- log in to the app to have the tokens saved to local storage
- wait for the refresh token to expire (or edit the tokens in local storage)
- refresh the page and notice the app/browser crash without any console error or other information

### Code Snippet

https://codesandbox.io/s/zealous-star-xhj15y

```typescript
import { Amplify } from "aws-amplify";
import { Authenticator, Flex } from "@aws-amplify/ui-react";
import "./styles.css";
import "@aws-amplify/ui-react/styles.css";

Amplify.configure({
  Auth: {
    region: "eu-west-1",
    userPoolId: "eu-west-1_POOLID",
    userPoolWebClientId: "POOL_ID",
    authenticationFlowType: "USER_PASSWORD_AUTH"
  }
});

export default function App() {
  return (
    <Flex justifyContent="center" alignItems="center" height="100vh">
      <Authenticator.Provider>
        <Authenticator>
          {(props) => {
            console.log(props.user);
            return (
              <>
                <p>{props.user.username}</p>
                <button onClick={props.signOut}>sign-out</button>
              </>
            );
          }}
        </Authenticator>
      </Authenticator.Provider>
    </Flex>
  );
}


```

### Additional information and screenshots

libs in package.json:
"@aws-amplify/ui-react": "2.13.0"
"amazon-cognito-identity-js": "^5.2.8"
"aws-amplify": "^4.3.19"

[...]

PR #1809, opened 2022-05-02
feat(authenticator): listen to tokenRefresh events

### LINKED ISSUE #1683: On token refresh useAuthenticator returns an unauthenticated user object

React

### Which UI component?

Authenticator

### How is your app built?

CRA with webpack 5

### Please describe your bug.

This issue might be related to #1333 and/or #1497 

For testing purposes I reduced the token refresh time to 7 minutes, but I assume it will not be different with longer refresh times.

When the token is expired and it should be refreshed and you than navigate to a page which uses useAuthenticator I get the error:
image

That the token is refreshed by the library and useAuthenticator would still return a valid user object

- Log in to the application
- Wait until the token needs to be refreshed (I my case 7 minutes)
- Navigate to a page which uses useAuthenticator and uses user.getUserData() and see the error

### Code Snippet

```typescript
// Index.js
ReactDOM.render(
  <Provider store={store}>
    <React.StrictMode>
      <Authenticator.Provider>
        <App />
      </Authenticator.Provider>
    </React.StrictMode>
  </Provider>,
  document.getElementById('root')
);
```

```typescript
// App.js
<Authenticator hideSignUp loginMechanisms={['email']}>
        {() => (
          <main>
            ... Components ...
          </main>
        )}
      </Authenticator>
```

```typescript
// MyComponent.js
const { user } = useAuthenticator((context) => [context.user]);
// Here user cannot be used because its not authenticated
user.getUserData()
```

_No response_

wlee221:
#### Description of changes

This updates Authenticator to listen to `tokenRefresh` and `tokenRefresh_failure`.

#### e2e tests

There is no good way to mock token refresh directly (the lowest ttl you can set is 5 minutes), so I instead mocked the hub events.

#### Angular

Angular default change detection is based on async network events and html events. And because hub events are just synchronous (and doesn't necessarily accompany network requests), it needed some manual change detection triggers. 

#### Issue #, if available

Resolves #1683, #1773

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] e2e tests are added to match feature


packages/ui/src/machines/authenticator/index.ts
```diff
@@ -178,6 +178,24 @@ export function createAuthenticatorMachine() {
...
+            },
+            refreshUser: {
+              invoke: {
+                src: 'getCurrentUser',
+                onDone: {
+                  actions: 'setUser',
+                  target: 'idle',
+                },
+                onError: { target: '#authenticator.signOut' },
+              },
```
wlee221, line 196: on token refresh, call `getCurrentUser` again to get up to date user.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
```diff
@@ -61,12 +66,47 @@ export class AuthenticatorComponent
...
+      /**
+       * Hub events that we handle can lead to multiple state changes:
+       * e.g. `authenticated` -> `signOut` -> initialState.
+       *
+       * We want to ensure change detection runs all the way, until
+       * we reach back to the initial state. Setting the below flag
+       * to true to until we reach initial state.
+       */
+      this.isHandlingHubEvent = true;
+      return state;
```
wlee221, line 89: In other words -- on hub events we handle, we turn this flag on that will trigger manual change detection until state machine is back at the initial state (as result of a sign out).

This is based on current implementation of `listenToHub`, which only has `signOut` behavior or `refreshUser` behavior. 

This might change if we add more hub event handling. But this at least shouldn't be breaking when that happens; there just will be wasted change detections that we can look to optimize at then.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -8,6 +8,28 @@ import { get, escapeRegExp } from 'lodash';
 let language = 'en-US';
 let window = null;
 
+/**
+ * Given dot delimited paths to a method (e.g. Amplify.Auth.signIn) on window,
+ * returns the object that holds the method (Amplify.Auth) and the method (signIn).
+ *
+ * Used for mocking and spying Amplify methods.
+ */
+const getMethodFromWindow = (path: string) => {
```
wlee221, line 17: This was getting reused in some places, so extracted it out.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -290,3 +302,25 @@ When('I see {string} as the {string} input', (custom, order) => {
...
+  if (!window) {
+    throw new Error(`window has not been set in the Cypress tests`);
+  }
+
+  const Hub = window['Hub'];
+  if (!Hub) {
+    throw new Error('Hub is not available on the window.');
+  }
+
+  Hub.dispatch('auth', { event: eventName });
```
wlee221, line 316: Sends a fake hub event.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
wlee221: This could happen with `tokenRefresh_failure`, because token expires silently instead of it firing off some html button press for example.

.changeset/two-turkeys-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aws-amplify/ui": patch
```
wlee221, line 2:
```suggestion
"@aws-amplify/ui": minor
```

ErikCH (approved):
LGTM!

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -8,6 +8,28 @@ import { get, escapeRegExp } from 'lodash';
...
+ *
+ * Used for mocking and spying Amplify methods.
+ */
+const getMethodFromWindow = (path: string) => {
+  let paths = path.split('.');
+  const method = paths.pop();
+  const obj = get(window, paths);
+
+  if (!window) {
+    throw new Error(`window has not been set in the Cypress tests`);
```
calebpollman, line 23: Nit: prefer single quotes

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -290,3 +302,25 @@ When('I see {string} as the {string} input', (custom, order) => {
 
   // cy.findByLabelText(custom).type(Cypress.env('VALID_PASSWORD'));
 });
+
+When('I mock {string} event', (eventName: string) => {
+  if (!window) {
+    throw new Error(`window has not been set in the Cypress tests`);
```
calebpollman, line 308: Nit: prefer single quotes

calebpollman (approved):
Left some a question and some nits but looks fairly straightforward

packages/e2e/cypress/integration/common/shared.ts
wlee221:
```suggestion
    throw new Error('window has not been set in the Cypress tests');
```

packages/e2e/cypress/integration/common/shared.ts
wlee221:
```suggestion
    throw new Error('window has not been set in the Cypress tests');
```

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -8,6 +8,28 @@ import { get, escapeRegExp } from 'lodash';
...
 let window = null;
 
+/**
+ * Given dot delimited paths to a method (e.g. Amplify.Auth.signIn) on window,
+ * returns the object that holds the method (Amplify.Auth) and the method (signIn).
+ *
+ * Used for mocking and spying Amplify methods.
+ */
+const getMethodFromWindow = (path: string) => {
+  let paths = path.split('.');
```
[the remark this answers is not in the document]
wlee221:
```suggestion
  const paths = path.split('.');
```

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -178,6 +178,24 @@ export function createAuthenticatorMachine() {
...
         },
         authenticated: {
+          initial: 'idle',
+          states: {
+            idle: {
+              on: {
+                TOKEN_REFRESH: 'refreshUser',
+              },
+            },
+            refreshUser: {
```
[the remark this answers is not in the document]
wlee221: It will not, they will be caught under https://github.com/aws-amplify/amplify-ui/blob/4c6ca0abd73a6b7df51193c27223ac2befb43c9a/packages/ui/src/helpers/authenticator/facade.ts#L74-L75

PR #1811, opened 2022-05-02
chore(ui-react): refactor Authenticator Router, add RouteContainer

calebpollman:
#### Description of changes

Refactor `Router` and add `RouteContainer`:
- consolidate HTML `View` components from `Authenticator` and `Router` components in `RouteContainer`
- wrap "route" components (e.g. `ConfirmSignUp`) with `RouteContainer`
- fix circular dependency issues in `Authenticator` by removing `defaultComponents` from `CustomComponents` initialization
- add/consolidate/update some types

#### Issue #, if available
NA

- ran integ tests
- manual functionality testing
- created snapshots tests in sample app
- checked imported types in sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/Authenticator/ConfirmSignIn/ConfirmSignIn.tsx
```diff
@@ -29,25 +33,27 @@ export const ConfirmSignIn = (): JSX.Element => {
...
-    <form
-      data-amplify-form=""
-      data-amplify-authenticator-confirmsignin=""
-      method="post"
-      onChange={handleChange}
-      onSubmit={handleSubmit}
-    >
-      <Flex as="fieldset" direction="column" isDisabled={isPending}>
-        <Header />
+    <RouteContainer className={className} variation={variation}>
```
calebpollman, line 36: For the `route` component files, this PR just wraps the components in a `RouteContainer` component that consists of the HTML related elements that were previously inside the `Router`

packages/react/src/components/Authenticator/Router/Router.tsx
```diff
@@ -0,0 +1,91 @@
...
+  variation,
+}: RouterProps) {
+  const { route } = useAuthenticator(({ route }) => [route]);
+  const RouterChildren = useRouterChildren(children);
+
+  return (
+    <RouterChildren
+      className={className}
+      hideSignUp={isSignInOrSignUpRoute(route) ? hideSignUp : undefined}
+      variation={variation}
```
ErikCH, line 88: This looks good, but I had a question. So with this new design you have to prop drill `className` and `variation` From the `Authenticator` -> `Router` -> `<component i.e. signUp, signIn>` -> `RouteContainer` -> finally to the `View` component so it can be added to the `data-variation` and `className`. 

I suppose that's fine, it's just a pattern we'll have to make sure we remember to do on all future components we add.

ErikCH (approved):
Just a few questions, looks good

packages/react/src/components/Authenticator/Router/Router.tsx
calebpollman: That is a possibility. The idea of having the route options in the components context is to prevent the need to pass UI specific properties to the state machine and keep a separation of concerns between UI needs and state machine logic

wlee221 (approved):
Nothing stands out to me, thanks @calebpollman!

PR #1814, opened 2022-05-03
Remove inline styling vue

jacoblogan:
#### Description of changes
This pull requests removes all of the inline styling from the vue authenticator.  Some of the new classes being used were previously added to the ui package in the pull request to remove the react inline styling.  The other classes that are used in only the vue authenticator have been added to the local vue css file.

#### Checklist

- [x] PR description included


packages/vue/src/components/confirm-sign-up.vue
```diff
@@ -74,50 +74,43 @@ const onLostCodeClicked = (): void => {
...
-            style="flex-direction: column"
+            class="amplify-flex amplify-authenticator__column"
             :disabled="isPending"
           >
             <base-form-fields route="confirmSignUp"></base-form-fields>
           </base-field-set>
 
-          <base-footer
-            class="amplify-flex"
-            style="flex-direction: column; align-items: unset"
```
jacoblogan, line 100: the flex-direction is now being set by `amplify-authenticator__column` but the align-items rule doesn't seem to be having an effect on the layout so it was just removed.

ErikCH (approved):
👍

joebuono (approved):

PR #1815, opened 2022-05-03
docs fix: switching themes wipes out state

joebuono:
#### Description of changes

This PR is based on the solution proposed in this [draft PR](https://github.com/aws-amplify/amplify-ui/pull/1761). 

There has been a [longstanding docs bug](https://github.com/aws-amplify/amplify-ui/issues/678) in which switching themes (e.g., toggling from light to dark mode) wipes out the component's demo state. This has to do with NextJS dynamic imports, which give us performance benefits. 

As a solution, we are using a simple `Map` to persist component demo state across NextJS re-renders without the complexity of state management (e.g., Redux). 

We are not using React Context for two reasons (based on the [React Context docs](https://reactjs.org/docs/context.html)):
- Since the props for each component's demo do not need to be shared across the entire docs site, 
  these props aren't "global" state in the same way that a current authenticated user, theme, or preferred language would be
- React Context is primarily used when some data needs to be accessible by _many_ components at different nesting levels, which does not match our use case

The Map's key-value pairs are the component's display name and its demo's props. Its `get` and `set` methods are used accordingly in each component's demo, which do not cause unnecessary re-renders. 

These changes have thus far only been applied to updated docs pages, but will need to be applied when we update the remaining pages. 

#### Issue #, if available

Fixes https://github.com/aws-amplify/amplify-ui/issues/678

Ran the docs locally. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


docs/src/utils/demoState.ts
```diff
@@ -0,0 +1,13 @@
...
+
+We are not using React Context for two reasons (based on the React Context docs: https://reactjs.org/docs/context.html):
+- Since the props for each component's demo do not need to be shared across the entire docs site, 
+  these props aren't "global" state in the same way that a current authenticated user, theme, or preferred language would be
+- React Context is primarily used when some data needs to be accessible by many components at different nesting levels, which does not match our use case
+
+The Map's key-value pairs are the component's name and its demo's props. 
+*/
+
+export const demoState = new Map<string, any>();
```
reesscot, line 13: Would prefer not to use `any`. Could we narrow down this type?

```suggestion
type DemoStateValue = string | number | boolean | null;

export const demoState = new Map<string, DemoStateValue>();
```

docs/src/utils/demoState.ts
```diff
@@ -0,0 +1,13 @@
+/*
+
+We are using a Map as a lightweight solution to persist component demo state across NextJS re-renders without the complexity of state management (e.g., Redux).
```
reesscot, line 3: Add a comment on why we need to persist demo state with a link to the original github issue.

docs/src/pages/components/view/demo.tsx
```diff
@@ -26,18 +27,26 @@ const propsToCode = (props: ViewProps) => {
...
-    padding: '1rem',
-    border: '1px solid var(--amplify-colors-black)',
-    borderRadius: '6px',
+  const props = useViewProps(
+    demoState.get(View.displayName) || defaultViewProps
+  );
+
+  React.useEffect(() => {
+    demoState.set(View.displayName, props);
   });
```
reesscot, line 50: Do you have the EsLint plugin installed? You should be getting a rules of hooks error here. I don't think we want to set the demoState on every render, only if props change:
```suggestion
  }, [props]);
```

docs/src/pages/components/view/demo.tsx
```diff
@@ -26,18 +27,26 @@ const propsToCode = (props: ViewProps) => {
...
-    ariaLabel: 'View example',
-    width: '20rem',
-    height: '4rem',
-    color: 'var(--amplify-colors-blue-60)',
-    backgroundColor: 'var(--amplify-colors-white)',
-    boxShadow: '3px 3px 5px 6px var(--amplify-colors-neutral-60)',
-    padding: '1rem',
-    border: '1px solid var(--amplify-colors-black)',
-    borderRadius: '6px',
+  const props = useViewProps(
```
reesscot, line 44: I think all of our `use[Component]Props` hooks need to be updated to internally use `React.useMemo`, otherwise they will always return a new object literal, which will always run the `useEffect` below even if `props` is added to the dependency array.

**Simplified example**
```jsx
export const useViewProps: UseViewProps = (initialValues) => {
  const [ariaLabel, setAriaLabel] = React.useState<ViewProps['ariaLabel']>(
    initialValues.ariaLabel
  );

  return React.useMemo(() => ({
    ariaLabel,
    setAriaLabel,
  }), [ariaLabel, setAriaLabel);
};
```

reesscot (commented):
Looks good. Just had some performance suggestions below.

docs/src/pages/components/view/demo.tsx
calebpollman: +1

docs/src/utils/demoState.ts
joebuono: Added

reesscot (approved):
LGTM

docs/src/utils/demoState.ts
```diff
@@ -0,0 +1,20 @@
...
+
+Problem: Component demo state gets wiped out by NextJS when a customer switches themes (e.g., toggles between light/dark mode). 
+- See: https://github.com/aws-amplify/amplify-ui/issues/678
+- This has to do with how we are doing dynamic imports, which give us performance benefits
+
+Solution: We are using a simple Map to persist component demo state across NextJS re-renders without the complexity of state management (e.g., Redux)
+
+We are not using React Context for two reasons (based on the React Context docs: https://reactjs.org/docs/context.html):
+- Since the props for each component's demo do not need to be shared across the entire docs site, 
+  these props aren't "global" state in the same way that a current authenticated user, theme, or preferred language would be
```
0618, line 11: I think the reason might not matter in our case because we NextJS only render the page user is looking at... 😁

PR #1824, opened 2022-05-04
refactor(angular): Use explicit change detection trigger

wlee221:
#### Description of changes
Previously, we've been relying on default Angular change detection to know when Authenticator should re-render. In particular, Angular runs [change detection](https://blog.angular-university.io/how-does-angular-2-change-detection-really-work/) whenever async promise resolves (e.g. `Auth.getCurrentAuthenticatedUser`) or html event (e.g. button click). This worked well until now, because these two were what would cause change in our state machine.

But with us trying to add support to Hub events (#1809), this default behavior falls short; hub events are synchronous and thus Angular doesn't run change detection.

#### Solution

This PR resolves this by using `OnPush` change detection strategy to:

1. Turn off default change detection and
2. Explicitly invoke change detection whenever state machine gets an update.

This aligns with the previous behavior, but just more explicit so Angular doesn't miss edge cases like hub events. This should not cause more re-renders, since Authenticator was already being updated on every state change. 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
@@ -1,5 +1,7 @@
 import {
   AfterContentInit,
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   ContentChildren,
   Input,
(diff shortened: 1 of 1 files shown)


wlee221:
Closing, the effort is not trivial

Issue #1818, opened 2022-05-03
Reset Password SMS is not being delivered due to missing dialCode (country_code) in the phone number.


budevelop:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

### Please describe your bug.

Problem - Reset Password code is not delivered to the user via SMS.
Existing Environment - Authenticator is used with phone_number as the only login mechanism.
Root Cause - dialCode is not included in the phone number (which is username in resetPassword formFields) when trying to send the SMS. SMS is sent to a number without any country code.
More Info - Override handleForgotPassword and check the formData. It only includes the phone number without the country code.
Possible Workaround - If the app is getting users only from a single country, override handleForgotPassword function and add country_code at the beginning of the formData (which is the phone number).

### What's the expected behaviour?

DialCode needs to be prepended to the phone number in the Reset Password; so that SMSs including the reset code would be delivered to the users. handleForgotPassword formData should include the dialCode as well.

### Help us reproduce the bug!

Create a Cognito User Pool with only phone_number registration. Use the phone_number as the only login mechanism in the new Authenticator UI. Try to reset password. SMS would not be delivered.

### Code Snippet

```typescript
const services = {
    async handleForgotPassword(formData) {
        console.log(formData);
        return Auth.forgotPassword(formData);
    },
};


<AmplifyProvider>
   <Authenticator loginMechanisms={['phone_number']} formFields={formFields} components={components} services={services}>
                        {({ signOut, user }) => (
                            <main>
                                     ...
                            </main>
                        )}
   </Authenticator>
</AmplifyProvider>
```

budevelop:
Same problem also applies to handleForgotPasswordSubmit where users submit the received code with the new password.

ErikCH:
@budevelop I pushed up and merged a fix for this. It should be available later this afternoon after our release.

PR #1833, opened 2022-05-04
fix: Fixed error with country code on reset password page

ErikCH:
#### Description of changes

When users try to reset their password, and they are using phone, it doesn't include the country_code

#### Issue #, if available
#1818 

Additional automated test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -290,3 +290,30 @@ When('I see {string} as the {string} input', (custom, order) => {
...
   // cy.findByLabelText(custom).type(Cypress.env('VALID_PASSWORD'));
 });
+
+When('I type a valid code', () => {
+  /**
+   * Confirmation code differs on React/Vue vs Angular. Testing for both for
+   * now, but we can look to make them consistent on next major release.
+   */
+  const regex = new RegExp(`^(confirmation )?code( *)?`, 'i');
+  cy.findByRole('spinbutton', { name: regex }).type('1234');
```
ErikCH, line 300: Moved these to common so they could be accessed by more then the reset password feature

packages/e2e/features/ui/components/authenticator/sign-in-with-phone.feature
```diff
@@ -10,6 +10,24 @@ Feature: Sign In with Phone Number
...
     Given I'm running the example "ui/components/authenticator/sign-in-with-phone"
 
+  @angular @react @vue
+  Scenario: Reset Password with valid phone with country code
+    When I click the "Forgot your password?" button
+    When I select my country code with status "CONFIRMED"
+    And I type my "phone number" with status "CONFIRMED"
+    And I intercept '{ "headers": { "X-Amz-Target": "AWSCognitoIdentityProviderService.ForgotPassword" } }' with fixture "reset-password"
+    And I click the "Send code" button
+    And I verify the body has "+19995554444" included
```
ErikCH, line 20: I check the username twice in the body to verify to make sure that it's being sent correctly over and not missing the country_code.

packages/ui/src/machines/authenticator/actors/resetPassword.ts
```diff
@@ -183,7 +183,7 @@ export function resetPasswordActor({ services }: ResetPasswordMachineOptions) {
       },
       services: {
         async resetPassword(context) {
-          const username = context.formValues?.username ?? context.username;
+          const { username } = context;
```
ErikCH, line 186: I couldn't find anywhere why we couldn't use `context.username` here. In fact, the `confirmResetPassword`, the next step, uses the `context.username` so it has to be there otherwise everything would fail.

wlee221: Hmm, technically `context.formValues?.username` is where username values lives in initially.

It is then assigned to `context.username` as part of `setUsername` entry action in `resetPassword.submit`, so that's why `context.username` is always there for `resetPassword` and `confirmResetPassword` flow.

`setUsername` entry action:

https://github.com/aws-amplify/amplify-ui/blob/194ba9fff4ea20fd87d6b40d4c98b19f77bad727/packages/ui/src/machines/authenticator/actors/resetPassword.ts#L47-L49

wlee221: So that's the technical background why I think that check existed in the first place... but anyway as you wrote, it works just as well if we simply use 

```ts
const { username } = context;
```

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -290,3 +290,30 @@ When('I see {string} as the {string} input', (custom, order) => {
 
   // cy.findByLabelText(custom).type(Cypress.env('VALID_PASSWORD'));
 });
+
+When('I type a valid code', () => {
+  /**
+   * Confirmation code differs on React/Vue vs Angular. Testing for both for
+   * now, but we can look to make them consistent on next major release.
+   */
```
wlee221, line 298: Yep! This is on me, let's make them consistent on 3.0 🙏

wlee221 (approved):
LGTM!

packages/ui/src/machines/authenticator/actors/resetPassword.ts
ErikCH: Ok that makes sense

reesscot (approved):

PR #1846, opened 2022-05-06
docs(update): SliderField

joebuono:
#### Description of changes

This PR updates the SliderField docs page. The best way to review this PR is to pull it and run the docs locally. 

ezgif com-gif-maker (1)

Ran the docs locally. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


docs/src/pages/components/sliderfield/demo.tsx
```diff
@@ -1,40 +1,69 @@
...
+import { useSliderFieldProps } from './useSliderFieldProps';
+import { getPropString } from '../utils/getPropString';
+import { demoState } from '@/utils/demoState';
 
-import { Flex, SliderField } from '@aws-amplify/ui-react';
+const propsToCode = (props: SliderFieldProps) => {
+  return (
+    `<SliderField` +
+    `\n  label="${props.label}"` +
+    (props.min ? `\n  min={${props.min}}` : '') +
```
jacoblogan, line 13: can you use the getPropString helper function for more of these properties? looks like maybe min and step could use it and maybe we create a second util function that does the same thing for boolean flags isDisabled, isValueHidden, labelHidden.

joebuono: Good idea. `min`, `max` and `step` are all numbers though, so we can't use getPropString on them. I was thinking about making a general `propsToCode` util function, which is maybe a better longterm solution

jacoblogan: yeah we should probably just have one of these util functions for each of the common types, number, string, and boolean.

jacoblogan (approved):
LGTM

docs/src/pages/components/sliderfield/demo.tsx
```diff
@@ -1,40 +1,69 @@
...
+    (props.step !== 1 ? `\n  step={${props.step}}` : '') +
+    getPropString(props.size, 'size') +
+    getPropString(props.trackSize, 'trackSize') +
+    getPropString(props.emptyTrackColor, 'emptyTrackColor') +
+    getPropString(props.filledTrackColor, 'filledTrackColor') +
+    getPropString(props.thumbColor, 'thumbColor') +
+    (props.orientation === 'vertical'
+      ? `\n  orientation="${props.orientation}"`
+      : '') +
+    (props.isDisabled ? `\n  isDisabled` : '') +
```
jacoblogan, line 24: previously we've tried to use the full `isDisabled={true}` for our examples instead of the shorthand `isDisabled` should we do the same here?

docs/src/pages/components/sliderfield/demo.tsx
reesscot: `getPropString` takes a number or a string and returns a string, so you should be able to use it. I'm open to better names for the function though if it's confusing.

reesscot (approved):

docs/src/pages/components/sliderfield/demo.tsx
joebuono: If we used `getPropString` for a prop which is type `number`, then it would return something like:
`min="7"`

However, we'd want to show that prop set to a number, for example:
`min={7}`

I can adjust the `getPropString` function to handle both the number and string cases.

PR #1862, opened 2022-05-10
Docs/a11y roundup

hbuchel:
#### Description of changes

Small roundup of "low hanging fruit" accessibility fixes that are spread around the docs site (ignoring any that are component specific). Added some comments to Files changed to describe the changes.

**Before: Component page (Button) scan results.**

Screen Shot 2022-05-10 at 1 35 26 PM

**After: Component page (Button) scan results**

Screen Shot 2022-05-10 at 1 36 07 PM

**What about the other 10 issues?**
The other issues are specific to the Button docs examples (buttons with the same name, links with the same name, etc).

Additionally, axe picks up that the code examples are not keyboard navigable since they cannot be scrolled; adding tabindex="0" would likely fix that but might require some further investigation for implementing since we're generating those `<pre>` elements within an mdx file. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/docs/src/components/Layout/SecondaryNav.tsx
+++ b/docs/src/components/Layout/SecondaryNav.tsx
@@ -185,12 +185,12 @@ export const SecondaryNav = (props) => {
 
 export const Sidebar = () => {
   return (
-    <aside className="docs-sidebar">
+    <nav aria-label="Section navigation" className="docs-sidebar">
       <div className="docs-sidebar-inner">
-        <nav className="docs-sidebar-nav">
+        <div className="docs-sidebar-nav">
           <SecondaryNav />
-        </nav>
+        </div>
       </div>
-    </aside>
+    </nav>
   );
 };
--- a/docs/src/components/Layout/index.tsx
+++ b/docs/src/components/Layout/index.tsx
@@ -123,6 +123,7 @@ export default function Page({
                 >
                   <Icon
                     ariaLabel=""
+                    aria-hidden="true"
                     as={SiReact}
                     marginInlineEnd={tokens.space.xs}
                   />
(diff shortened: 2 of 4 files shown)


docs/src/components/Layout/SecondaryNav.tsx
```diff
@@ -185,12 +185,12 @@ export const SecondaryNav = (props) => {
 
 export const Sidebar = () => {
   return (
-    <aside className="docs-sidebar">
```
hbuchel, line 188: We also have two asides; the other is the table of contents for the page itself. Since this sidebar element only houses the navigation, I changed this element to a <nav> (with aria label) with nested divs inside of it since the nav is the only semantic element needed.

docs/src/components/Layout/index.tsx
```diff
@@ -95,7 +95,7 @@ export default function Page({
                   isExternal
                 >
                   <Icon
-                    ariaLabel=""
+                    ariaLabel="W3C"
```
hbuchel, line 98: This link will now read as "W3C Aria Pattern".

docs/src/components/Layout/index.tsx
```diff
@@ -123,6 +123,7 @@ export default function Page({
                 >
                   <Icon
                     ariaLabel=""
+                    aria-hidden="true"
```
hbuchel, line 126: This icon doesn't provide any context since the label already says "React". Axe complains because the SVG has a role of img and an empty title tag.

joebuono (approved):
Thank you for doing this!

jacoblogan (approved):

reesscot (approved):
Thanks for fixing these!

zchenwei (commented):
I am curious what tool you are using for scanning. 🤣

PR #1863, opened 2022-05-10
fix(authenticator): fix infinite loop on refresh after refresh token expires

wlee221:
#### Description of changes

While #1809 resolves happy cases, refresh token could still cause an infinite loop when it expires (default 30 days). 

This is because we have

https://github.com/aws-amplify/amplify-ui/blob/04c49c44999d62287a6e932f7d0188c84a90ede6/packages/ui/src/machines/authenticator/index.ts#L208

which forwards any `SIGN_OUT` event to the actor. 

But when refresh token is expired and the app is refreshed, it calls `SIGN_OUT` event before state machine has any actor. This confuses the `forwardTo` action. Specifically, `forwardTo` action uses `send` under the hood, and it sends the event to itself if [the target is undefined](https://xstate.js.org/docs/guides/actions.html#send-action). So when `SIGN_OUT` is called when actor doesn't exist yet, it sends that event over and over to itself.

This PR fixes it by having `hasActor` condition to `forwardTo`.

I verified the fix on a mock environment, but I'm working on verifying the fix on actual app. Will update once done.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated
- [ ] 

### DIFF
--- a/packages/ui/src/machines/authenticator/index.ts
+++ b/packages/ui/src/machines/authenticator/index.ts
@@ -206,7 +206,6 @@ export function createAuthenticatorMachine() {
         SUBMIT: { actions: 'forwardToActor' },
         FEDERATED_SIGN_IN: { actions: 'forwardToActor' },
         RESEND: { actions: 'forwardToActor' },
-        SIGN_OUT: { actions: 'forwardToActor' },
         SIGN_IN: { actions: 'forwardToActor' },
         SKIP: { actions: 'forwardToActor' },
       },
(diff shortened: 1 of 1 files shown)


ErikCH (approved):
👍

packages/ui/src/machines/authenticator/index.ts
```diff
@@ -205,14 +206,18 @@ export function createAuthenticatorMachine() {
         SUBMIT: { actions: 'forwardToActor' },
         FEDERATED_SIGN_IN: { actions: 'forwardToActor' },
         RESEND: { actions: 'forwardToActor' },
-        SIGN_OUT: { actions: 'forwardToActor' },
```
wlee221, line 208: `SIGN_OUT` here is unneeded, we have no reason to pass this to any child.

hbuchel (approved):

zchenwei (approved):
LGTM!

PR #1868, opened 2022-05-10
fix(authenticator): Only pass hub event when authenticator is in correct state

wlee221:
#### Description of changes

Instead of blindly accepting hub events and passing them to state machine, this PR adds a state check to ensure that the authenticator is in the correct state. 

This will prevent sending hub events (e.g. `tokenRefresh`) over and over when Authenticator is already handling it.

#### Refactor

`listenToAuthHub` now needs access to current state, so I refactored the signature to accept `authService`, which is a static context that provides access to `state` and `send`. I also abstracted out the default hub handler, and enabled other libraries to pass custom hub handler. This helps Angular because it has been using extra handling for change detection (see #1809).

#### Issue #, if available

Related to #1683.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
+++ b/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
@@ -12,6 +12,7 @@ import {
 } from '@angular/core';
 import {
   AuthenticatorMachineOptions,
+  defaultAuthHubHandler,
   listenToAuthHub,
   SocialProvider,
   translate,
@@ -66,7 +67,10 @@ export class AuthenticatorComponent
       formFields,
     } = this;
 
-    this.unsubscribeHub = listenToAuthHub((event) => {
+    const { authService } = this.authenticator;
+
+    this.unsubscribeHub = listenToAuthHub(authService, (data, service) => {
+      defaultAuthHubHandler(data, service);
       /**
        * Hub events aren't properly caught by Angular, because they are
        * synchronous events. Angular tracks async network events and
@@ -74,7 +78,6 @@ export class AuthenticatorComponent
        *
        * On any notable hub events, we run change detection manually.
        */
-      const state = this.authenticator.authService.send(event);
       this.changeDetector.detectChanges();
 
       /**
@@ -86,7 +89,6 @@ export class AuthenticatorComponent
        * to true to until we reach initial state.
        */
       this.isHandlingHubEvent = true;
-      return state;
     });
 
     /**
--- a/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
+++ b/packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
@@ -41,26 +41,22 @@ export const Provider = ({ children }: { children: React.ReactNode }) => {
    *
    * Leaving this as is for now in the interest of suggested code guideline.
    */
-  const service = useInterpret(createAuthenticatorMachine);
+  const service = useInterpret(createAuthenticatorMachine, { devTools: true });
   const currentProviderVal = { service };
 
   const value = isEmpty(parentProviderVal)
     ? currentProviderVal
     : parentProviderVal;
 
-  const {
-    service: { send },
-  } = value;
+  const { service: activeService } = value;
 
   const isListening = React.useRef(false);
   React.useEffect(() => {
-    if (isListening.current) {
-      return;
-    }
+    if (isListening.current) return;
 
     isListening.current = true;
-    return listenToAuthHub(send);
-  }, [send]);
+    return listenToAuthHub(activeService);
+  }, [activeService]);
 
   return (
     <AuthenticatorContext.Provider value={value}>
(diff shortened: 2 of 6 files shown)


packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
```diff
@@ -66,15 +67,17 @@ export class AuthenticatorComponent
       formFields,
     } = this;
 
-    this.unsubscribeHub = listenToAuthHub((event) => {
+    const { authService } = this.authenticator;
+
+    this.unsubscribeHub = listenToAuthHub(authService, (data, service) => {
+      defaultAuthHubHandler(data, service);
```
wlee221, line 73: So this is more of a decorator pattern now -- use the default auth hub handler, then angular-specific custom handling is appended.

Note that this is Angular only.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -41,26 +41,22 @@ export const Provider = ({ children }: { children: React.ReactNode }) => {
    *
    * Leaving this as is for now in the interest of suggested code guideline.
    */
-  const service = useInterpret(createAuthenticatorMachine);
+  const service = useInterpret(createAuthenticatorMachine, { devTools: true });
```
ErikCH, line 44:
```suggestion
  const service = useInterpret(createAuthenticatorMachine);
```

wlee221: Good catch!

ErikCH (approved):
👍 

This looks ok! Just make sure it doesn't break the refresh fix you added earlier. I think it's fine

packages/ui/src/helpers/authenticator/utils.ts
```diff
@@ -32,6 +33,27 @@ export const censorPhoneNumber = (val: string): string => {
...
+      }
+      break;
+    case 'signOut':
+    case 'tokenRefresh_failure':
+      if (state.matches('authenticated.idle')) {
+        send('SIGN_OUT');
+      }
+      break;
+  }
+};
```
wlee221, line 55: Here is the actual code fix; the rest is refactors.

packages/ui/src/helpers/authenticator/utils.ts
```diff
@@ -32,6 +33,27 @@ export const censorPhoneNumber = (val: string): string => {
   return split.join('');
 };
 
+export const defaultAuthHubHandler: HubHandler = (data, service) => {
```
calebpollman, line 36: What do you think about adding unit testing for this utility? Just thinking that it could be useful since we don't cover this with integ testing (afaik)

calebpollman (commented):
LGTM, just left some thoughts

packages/ui/src/helpers/authenticator/utils.ts
```diff
@@ -32,6 +33,27 @@ export const censorPhoneNumber = (val: string): string => {
...
+        send('TOKEN_REFRESH');
+      }
+      break;
+    case 'signOut':
+    case 'tokenRefresh_failure':
+      if (state.matches('authenticated.idle')) {
+        send('SIGN_OUT');
+      }
+      break;
+  }
```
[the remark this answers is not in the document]
wlee221:
```suggestion
    case default:
      break;
  }
```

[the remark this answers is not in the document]
wlee221: ^ ignore that, that had a bad syntax. 

Done here https://github.com/aws-amplify/amplify-ui/pull/1868/files/8e87825aa412c3301e38279d14f4a36605cd9a5e..c5cd8982003b330c8e8985d8076d290b6158503c

packages/ui/src/helpers/authenticator/utils.ts
wlee221: I like it. There is however an integration test for these though at `/hub-events` ([.feature file](https://github.com/aws-amplify/amplify-ui/blob/main/packages/e2e/features/ui/components/authenticator/hub-events.feature)), which mocks and tests hub events and see if authenticator reacts accordingly.

That said though, it only tests for happy cases (as is with other e2e tests), so I do see value in adding unit tests. But currently `ui` packages do not have unit tests set up, so I'm thinking we want to scope this out to a bigger PR that adds low-level testing on UI repo. What do you think? For now, integration tests at least covers all the known auth hub events.

calebpollman: That works for me 😄  

If we don't already, can we add a backlog ticket for setting up unit tests where needed?

calebpollman (approved):
LGTM 🚢

packages/ui/src/helpers/authenticator/utils.ts
wlee221: Done! https://app.asana.com/0/1201736086077862/1202277146130103/f

PR #1772, opened 2022-04-25
Merge flattened css into main branch

jacoblogan:
#### Description of changes
This pull request our current styles across react, angular, and vue to use flattened css selectors following the BEM styling syntax.  This is to produce a better experience for customers who are using css frameworks and tools.  By lowering our selector specificity it allows our customers to in turn use lower specificity selectors to overwrite our default styles.

- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1872, opened 2022-05-10
Remove is disabled

ErikCH:
#### Description of changes
Removing non-existant prop `isDisabled` on DOM element on View primitive.

Was causing this warning

image

Ran manually

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/react/src/primitives/View/View.tsx
+++ b/packages/react/src/primitives/View/View.tsx
@@ -27,7 +27,6 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
       'aria-label': ariaLabel,
       'data-testid': testId,
       disabled: isDisabled,
-      isDisabled: isDisabled,
       ref,
       style: propStyles,
       ...nonStyleProps,
--- a/.changeset/rude-wasps-hammer.md
+++ b/.changeset/rude-wasps-hammer.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ui-react': patch
+---
+
+Removed isDisabled prop, since it doesn't exist as an HTML element


packages/react/src/primitives/View/View.tsx
```diff
@@ -27,7 +27,6 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
       'aria-label': ariaLabel,
       'data-testid': testId,
       disabled: isDisabled,
-      isDisabled: isDisabled,
```
zchenwei, line 30: @jacoblogan Can you double check? - https://github.com/aws-amplify/amplify-ui/pull/1772/files#r865784417

zchenwei (commented):
LGTM! But would like to have @jacoblogan to double check this.

reesscot (approved):
LGTM

packages/react/src/primitives/View/View.tsx
reesscot: > this was added in because of the FieldGroupIconButton which is using the as property to render our primitive Button but the isDisabled flag was not being propagated down to the Button primitive.

I don't understand why this problem would show up after refactoring the CSS. Do you think this was a preexisting bug?

jacoblogan:
Yes this was a bug introduced with the css refactor, this was needed because of the `FieldGroupIconButton` which uses `as={Button}` to render the button but the view component was striping out the `isDisabled` prop so the final button being rendered was missing the disabled classes.

jacoblogan (approved):
Thanks for catching this, after discussing with @reesscot I'll create a follow up PR to move this check onto the Button component so that it doesn't get passed through to the underlying dom element.

Issue #1795, opened 2022-04-28
Bug: PhoneNumber field dial code dropdown doesn't honor readonly prop


reesscot:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

N/A

### Please describe your bug.

When PhoneNumberField has prop `isReadOnly={true}`, dial code dropdown can still be edited:
![2022-04-28 16 31 54](https://user-images.githubusercontent.com/6165315/165863935-dbe42ee6-5f94-4d67-bda3-d1cc9a9678f7.gif)

### What's the expected behaviour?

When PhoneNumberField has prop `isReadOnly={true}`, dial code dropdown should NOT be changeable.

### Help us reproduce the bug!

See:
https://ui.docs.amplify.aws/components/phonenumberfield#states

### Code Snippet

```typescript
// Put your code below this line.

```

reesscot:
Further research on this issue shows that this is due to the [HTML select](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) element not supporting the `readonly` property. Need to do further research on whether this is something we want to try to address outside of the HTML spec.

PR #1878, opened 2022-05-11
bug(fix): PhoneNumberField does not honor readonly prop

joebuono:
#### Description of changes

This PR address [this issue](https://github.com/aws-amplify/amplify-ui/issues/1795): "PhoneNumberField dial code dropdown doesn't honor readonly prop"

As a result of these changes, if the `isReadOnly` prop is passed to the `PhoneNumberField`:
- The `CountryCodeSelect` primitive is marked with the `aria-disabled=“true”` attribute. The reason we’re not marking it with `aria-readonly=“true”` is because that does not cause screen readers to announce anything to customers. @hbuchel 
- All the options in the `CountryCodeSelect` are marked with the `disabled` attribute because the select field should still be readable and focusable, but disallow changing the readonly selection. 
- Form data is still submitted if the `isReadOnly` prop is passed. 

readonlygif

#### Issue #, if available

Fixes https://github.com/aws-amplify/amplify-ui/issues/1795

Ran the docs locally and wrote some unit tests. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
+++ b/packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
@@ -12,6 +12,7 @@ const CountryCodeSelectPrimitive: Primitive<CountryCodeSelectProps, 'select'> =
     const countryCodeOptions = React.useMemo(
       () =>
         dialList.map((dialCode) => (
+          // Regarding the `disabled` attribute, see comment in SelectField below
           <option key={dialCode} value={dialCode} disabled={isReadOnly}>
             {dialCode}
           </option>
@@ -21,6 +22,10 @@ const CountryCodeSelectPrimitive: Primitive<CountryCodeSelectProps, 'select'> =
 
     return (
       <SelectField
+        /*
+          Since <select> elements do not support the `readonly` html attribute, it is suggested to use the `disabled` html attribute 
+          so that a screen reader will announce something to the user about the interactivity of the options list ( https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly)
+        */
         aria-disabled={isReadOnly}
         autoComplete="tel-country-code"
         className={classNames(ComponentClassNames.CountryCodeSelect, className)}
--- a/packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
+++ b/packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
@@ -7,20 +7,21 @@ import { CountryCodeSelectProps, Primitive } from '../types';
 import { SelectField } from '../SelectField';
 
 const CountryCodeSelectPrimitive: Primitive<CountryCodeSelectProps, 'select'> =
-  ({ className, dialCodeList, ...props }, ref) => {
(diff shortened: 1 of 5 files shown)


0618 (approved):

jacoblogan (approved):
LGTM

packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
```diff
@@ -7,20 +7,21 @@ import { CountryCodeSelectProps, Primitive } from '../types';
...
 
 const CountryCodeSelectPrimitive: Primitive<CountryCodeSelectProps, 'select'> =
-  ({ className, dialCodeList, ...props }, ref) => {
+  ({ className, dialCodeList, isReadOnly, ...props }, ref) => {
     const dialList = dialCodeList ?? countryDialCodes;
     const countryCodeOptions = React.useMemo(
       () =>
         dialList.map((dialCode) => (
-          <option key={dialCode} value={dialCode}>
+          <option key={dialCode} value={dialCode} disabled={isReadOnly}>
```
reesscot, line 15: Should we add a comment of why we are mapping disabled to isReadOnly? We will wonder later on.

packages/react/src/primitives/PhoneNumberField/CountryCodeSelect.tsx
```diff
@@ -7,20 +7,21 @@ import { CountryCodeSelectProps, Primitive } from '../types';
...
             {dialCode}
           </option>
         )),
-      [dialList]
+      [dialList, isReadOnly]
     );
 
     return (
       <SelectField
+        aria-disabled={isReadOnly}
```
reesscot, line 24: Same here, I think there should be a comment on why this is more accessible.

packages/react/src/primitives/PhoneNumberField/__tests__/PhoneNumberField.test.tsx
```diff
@@ -29,6 +31,34 @@ describe('PhoneNumberField primitive', () => {
     };
   };
 
+  const originalLog = console.log;
```
reesscot, line 34: nit: put this above the describe, and reassign `originalLog` in an `afterEach`.

packages/react/src/primitives/PhoneNumberField/__tests__/PhoneNumberField.test.tsx
```diff
@@ -135,4 +165,24 @@ describe('PhoneNumberField primitive', () => {
 
     expect(onCountryCodeChange).toHaveBeenCalled();
   });
+
+  it('should set aria-disabled="true" when the isReadOnly prop is passed, and disable all the select options', async () => {
```
reesscot, line 169: Add comment on why this behavior exists

reesscot (commented):
This is looking great, just had some comments. Also, the demo is showing a button that looks too tall. Can we fix that?

zchenwei (commented):
Humm, SR wouldn't announce anything from elements marked by `aria-readonly=true`? It feels weird.

PR #1892, opened 2022-05-12
docs(fix): Reorder homepage framework images

joebuono:
#### Description of changes

This PR simply reorders the framework images on the homepage to match the order of the FrameworkChooser (React, Angular, Vue, Flutter). 

Homepage-frameworks

#### Checklist

- [X] PR description included


docs/src/pages/index.page.tsx
```diff
@@ -203,17 +203,17 @@ const HomePage = ({ colorMode, setThemeOverride, themeOverride }) => {
             <Image
               alt=""
               className="docs-home-vue"
-              src="/svg/integrations/vue.svg"
+              src="/svg/integrations/react.svg"
```
0618, line 206: the `className` doesn't have to change?

joebuono: Good catch!

0618:
I noticed that we have a lot of patterns like this in this repo. If we use a `framework` constant for them, we wouldn't have this kind of mis-order

reesscot (approved):

docs/src/pages/index.page.tsx
```diff
@@ -201,22 +201,22 @@ const HomePage = ({ colorMode, setThemeOverride, themeOverride }) => {
             display={{ base: 'none', large: 'initial' }}
           >
             <Image
-              alt=""
-              className="docs-home-vue"
-              src="/svg/integrations/vue.svg"
-            />
-            <Image
-              alt=""
+              alt="React framework image"
```
reesscot, line 204: Alt text shouldn't have the word image in it, since the screen reader already knows it's an image. Alt text is meant to replace the image if it doesn't show up, so maybe `React logo` would make more sense here.

0618 (approved):

zchenwei (approved):

PR #1887, opened 2022-05-12
fix(geo): Point to react-map-gl es5 bundle

slaymance:
#### Description of changes
Some users have reported issues with ES 2020 features that are used by our `react-map-gl` dependency and certain build tools that they're using (e.g. Create React App v4). This PR updates our use of `react-map-gl` to point to the ES5-compatible bundle provided by the library which will mitigate issues related to ES 2020 features.

#### Issue #, if available
#1656 

#### Description of how you validated changes
Manually validated this change by sym-linking it to a CRA4 sample application using the `<MapView>` component and verified that there were no build issues and the [workaround suggested in #1656](https://github.com/aws-amplify/amplify-ui/issues/1656#issuecomment-1095389881) is not needed.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated - N/A
- [x] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated - N/A
- [x] Relevant documentation is changed or added (and PR referenced) - N/A

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1917, opened 2022-05-17
Revert "fix(geo): Point to react-map-gl es5 bundle"

slaymance:
Reverts aws-amplify/amplify-ui#1887

Pointing `react-map-gl` to the es5 bundle causes an error when an application uses the package with >=es6 module resolution. This represents a breaking change to customer applications.


thaddmt (approved):

joebuono (approved):

reesscot (approved):
LGTM

PR #1910, opened 2022-05-17
fix(authenticator): Move `autoSignIn` logic out of `signUp` actor

wlee221:
#### Description of changes

Previously, `autoSignIn` flow was part of the `signUp` actor. But `autoSignIn` is quite incomplete compared to what `signIn` actor does; `autoSignIn` only tries `Auth.signIn` and does not check for any auth challenges or edge cases like sms mfa, totp, etc.

Instead of having incomplete `autoSignIn` in `signUp` actor, this moves `autoSignIn` to `signIn` actor. This makes semantic sense, because `autoSignIn` is part of sign in flow anyway and `signUp` actor should only be dealing with sign up logic. 

#### Issue #, if available

Fixes #1660.

#### Description of how you validated changes

I confirmed manually that end-users are be automatically redirected to `confirmSignIn` after confirm sign up.

Also added new e2e test that goes through this process.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated
- [x] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated
- [x] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1937, opened 2022-05-19
fix(angular,vue): remove floating border between transitions

wlee221:
#### Description of changes

Angular and Vue Authenticator displays a floating border between transitions (e.g. `autoSignIn`) that shouldn't be there. This PR removes it by only rendering router containers only when current `route` has a valid component associated with it.

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/pull/1910#discussion_r877251711

Checked manually:

<details>
 <summary>Angular</summary>
  <h4>Before</h4>

https://user-images.githubusercontent.com/43682783/169414772-ab2e3fda-22a0-4c32-8e01-bfae5a0db3e4.mp4

  <h4>After</h4>

https://user-images.githubusercontent.com/43682783/169414808-671a5db9-83ee-409f-b20d-fc63802ff5b7.mp4

</details>

<details>
 <summary>Vue</summary>
  <h4>Before</h4>

https://user-images.githubusercontent.com/43682783/169414859-f911b06f-8e6e-4071-9e8f-130cfd8833f1.mp4

  <h4>After</h4>

https://user-images.githubusercontent.com/43682783/169414867-b4ec980a-1698-4b7c-a5f4-e35eb596c6c3.mp4

</details>

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/vue/src/components/authenticator.vue
```diff
@@ -239,14 +239,24 @@ const hasTabs = computed(() => {
...
 });
+
+const hasRouteComponent = computed(() => {
+  return !(
+    state.value.matches('authenticated') ||
+    state.value.matches('idle') ||
+    state.value.matches('setup') ||
+    state.value.matches('signOut') ||
+    actorState.value?.matches('autoSignIn')
+  );
```
wlee221, line 250: On our next capacity, we should look into using `useAuthenticator` in this file so we can just use `route` instead of `state.matches`.

This will also let us to share `hasRouteComponent` between the three frameworks too. I left it as-is because the three frameworks handle routes quite differently, but this will be worth looking into after we refactor this file to use `useAuthenticator`.

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
```diff
@@ -171,10 +171,25 @@ export class AuthenticatorComponent
   }
 
   public hasTabs() {
-    const route = this.authenticator.route;
+    const { route } = this.authenticator;
     return route === 'signIn' || route === 'signUp';
   }
 
+  public hasRouteComponent() {
```
0618, line 178: since we have a "pure function" check, I guess I need to double check here, do we want to make it a pure function?

0618 (approved):

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
wlee221: Talked offline; This is a pure function but a not a side effect, so this is good to go.

Since Angular is OOP based, this function cannot be made pure because output is affected by `this.*`. 

I took a quick googling, and [some people see](https://softwareengineering.stackexchange.com/questions/96429/getters-and-setters-in-functional-languages#comment174337_96429) `this` as an implicit input to the function, which will qualify the function as pure. Couldn't find more source though 🤷

wlee221: But nevertheless, tldr; this is idiomatic and good to go 🙏

packages/vue/src/components/authenticator.vue
ErikCH: Good point. Vue never got refactored to use every `useAuthenticator` composable internally. I wasn't sure how that would affect re-rerendering everywhere. However, eventually we should do that, and test to make sure that it doesn't cause any issues.

ErikCH (approved):
👍

PR #1940, opened 2022-05-19
feat(e2e): Add custom Cypress command to wait for idle map

slaymance:
#### Description of changes
In our Geo end-to-end tests, we often would come up against a problem where a test step would need to wait for map animations to complete before executing additional Cypress commands which interact with the map or make assertions about the map. This PR adds a custom Cypress command to wait for the map to enter an idle state before continuing with additional Cypress test steps.

It achieves this by separately rendering `geo/` sample apps in the Next.js examples with the `MapProvider` component, allowing us to access the `map` object to place a handler on the `on.idle` callback.

#### Issue #, if available
N/A

Notice in the example below that the location search step executes before the map even loads.
no-wait
```javascript
When('I search for {string}', (searchTerm: string) => {
  cy.intercept(/.*places.*/).as('searchResults');
  cy.findByRole('textbox', {
    name: /search/i,
  }).type(searchTerm);
  cy.wait('@searchResults');
});
```

By adding the new custom command `cy.waitForIdleMap` in the test, we can indicate to the test runner to wait for the map to be loaded and enter an idle state before executing the location search step:
wait-for-idle
```javascript
When('I search for {string}', (searchTerm: string) => {
  ///
  cy.waitForIdleMap();
  ///

  cy.intercept(/.*places.*/).as('searchResults');
  cy.findByRole('textbox', {
    name: /search/i,
  }).type(searchTerm);
  cy.wait('@searchResults');
});
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated - N/A
- [x] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated - N/A


examples/next/pages/_app.page.tsx
```diff
@@ -4,13 +4,51 @@
...
 
 if (typeof window !== 'undefined') {
   window['Amplify'] = Amplify;
   window['Hub'] = Hub;
 }
 
+const SignalIdleMap = () => {
+  const { default: map } = useMap();
+
+  map?.once('load', () => {
```
thaddmt, line 17: Oh have not seen this before but im guessing this helps resolve multiple map renders?

thaddmt (approved):

examples/next/pages/_app.page.tsx
slaymance: Eh, in the case of loading it could be either `map.on` or `map.once` since the load callback is only called once. I just felt that this made it clearer since we only want to set the map object on the window once.

zchenwei (approved):

examples/next/pages/_app.page.tsx
```diff
@@ -4,13 +4,47 @@
...
 if (typeof window !== 'undefined') {
   window['Amplify'] = Amplify;
   window['Hub'] = Hub;
 }
 
+const SetCypressProperties = () => {
+  const { default: map } = useMap();
+
+  map?.once('load', () => {
+    if (window['Cypress']) {
```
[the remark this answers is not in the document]
slaymance: Good callout, I did this and actually added the property setters to a `useEffect` callback since they are technically side effects.

Issue #1737, opened 2022-04-21
Accessibility: Authenticator error messages should use `aria-errormessage`


reesscot:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React, Angular, Vue

### Which UI component?

Authenticator

### How is your app built?

N/A

### Please describe your bug.

When Authenticator fields are invalid and have `aria-invalid` equal to true, they should also have the `aria-errormessage` attribute which points to the error message text. This allows screenreaders to announce an error message on invalid field entry.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage

![Uploading Screen Shot 2022-04-20 at 8.03.31 PM.png…]()

### What's the expected behaviour?

`aria-errormessage` on invalid fields should point to DOM element which contain the corresponding error message. 

### Help us reproduce the bug!

1. Go to https://ui.docs.amplify.aws/components/authenticator#sign-up-attributes,  Sign Up Attributes section
2. Enter two passwords which don't match to trigger error
3. Inspect input element and see that no `aria-errormessage` is there.

### Code Snippet

```typescript
// Put your code below this line.

```

hbuchel:
I think we should consider using aria-describedby to associate the error message to the input. I was googling around and found [this post](https://www.davidmacd.com/blog/test-aria-describedby-errormessage-aria-live.html) and [this page on a11ysupport.io](https://a11ysupport.io/tech/aria/aria-errormessage_attribute) which documents only JAWS supporting aria-errormessage. I tested with [this pen](https://codepen.io/hbuchel/pen/GRQEKRK/3799dd675f2c1e2498730b847a80bb52) and that seems to still be the case that NVDA and VoiceOver do not read out the associated error message :(

wlee221:
@hbuchel, I could just confirm that with VoiceOver as well 😢 . I opened #1945 to use `aria-describedBy`.

PR #1945, opened 2022-05-20
fix(authenticator): Add aria-describedBy

wlee221:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Authenticator/shared/FormField.tsx
+++ b/packages/react/src/components/Authenticator/shared/FormField.tsx
@@ -6,6 +6,7 @@ import { PhoneNumberField } from '../../../primitives/PhoneNumberField';
 import { TextField } from '../../../primitives/TextField';
 import { useAuthenticator } from '../hooks/useAuthenticator';
 import { ValidationErrors } from './ValidationErrors';
+import { useStableId } from '../../../primitives/utils/useStableId';
 
 export interface FormFieldProps extends Omit<FormFieldOptions, 'label'> {
   // label is a required prop for the UI field components used in FormField
@@ -29,6 +30,8 @@ export function FormField({
     [name, validationErrors]
   );
   const hasError = errors?.length > 0;
+  const errorId = useStableId();
+  const ariaDescribedBy = hasError ? errorId : undefined;
 
   if (type === 'tel') {
     return (
@@ -40,8 +43,9 @@ export function FormField({
           countryCodeName="country_code"
           autoComplete={autoComplete}
           hasError={hasError}
+          aria-describedby={ariaDescribedBy}
         />
-        <ValidationErrors errors={errors} />
+        <ValidationErrors errors={errors} id={errorId} />
       </>
     );
   } else if (type === 'password') {
@@ -52,8 +56,9 @@ export function FormField({
           name={name}
           autoComplete={autoComplete}
           hasError={hasError}
+          aria-describedby={ariaDescribedBy}
         />
-        <ValidationErrors errors={errors} />
+        <ValidationErrors errors={errors} id={errorId} />
       </>
     );
   } else {
@@ -65,8 +70,9 @@ export function FormField({
           autoComplete={autoComplete}
           hasError={hasError}
           type={type}
+          aria-describedby={ariaDescribedBy}
         />
-        <ValidationErrors errors={errors} />
+        <ValidationErrors errors={errors} id={errorId} />
       </>
     );
   }
--- a/packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
+++ b/packages/react/src/components/Authenticator/shared/ValidationErrors.tsx
@@ -5,12 +5,13 @@ import { Text } from '../../../primitives/Text';
 
 export interface ValidationErrorsProps {
   errors: string[];
+  id?: string;
 }
-export const ValidationErrors = ({ errors }: ValidationErrorsProps) => {
+export const ValidationErrors = ({ errors, id }: ValidationErrorsProps) => {
   if (!(errors?.length > 0)) return null;
 
   return (
-    <View data-amplify-sign-up-errors>
+    <View data-amplify-sign-up-errors id={id}>
       {errors.map((error, idx) => {
         return (
           <Text key={error} role="alert" variation="error">


packages/react/src/components/Authenticator/shared/FormField.tsx
```diff
@@ -40,8 +43,9 @@ export function FormField({
           countryCodeName="country_code"
           autoComplete={autoComplete}
           hasError={hasError}
+          aria-describedby={ariaDescribedBy}
```
reesscot, line 46: I think you meant to use aria-errormessage?
https://github.com/aws-amplify/amplify-ui/issues/1737

wlee221: Update after offline discussion -- `aria-errormessage` is not well supported: [descriptiveText](https://github.com/aws-amplify/amplify-ui/issues/1737#issuecomment-1132205724) and we should use `aria-describedBy`. I could confirm this with MacOS voiceover as well.

One concern is that `aria-describedBy` is a property of `TextField` used for descriptive texts ([docs](https://ui.docs.amplify.aws/react/components/textfield)). But Authenticator doesn't use any descriptive texts, so IMO we can proceed with this change. Lmk what you think 🙏

ErikCH (approved):
Looks good, but could you add another task to make sure we get these changes in Angular and Vue too? 

Perhaps it will be a longer task since we don't have `stableId` in Vue and React.

calebpollman (commented):
LGTM but can we add unit testing since this is not covered by an integ test?

wlee221:
> Looks good, but could you add another task to make sure we get these changes in Angular and Vue too?

Yep! I wanted to React first to set the pattern, then I'll open another PR to add it in Vue/Angular.

wlee221:
> LGTM but can we add unit testing since this is not covered by an integ test?

Would it be okay to add this to e2e test? This is something that will be shared across all frameworks, so IMO this makes sense to be an e2e test. And we already have utility for checking attributes (and it is easier to type something on cypress).

That said though, I'm definitely with you on adding more unit tests. I think `packages/ui` ([asana](https://app.asana.com/0/1201736086077862/1202277146130103/f)) is a good place to start, because it has a lot of shared utility functions. Then we can start expanding them to `packages/*`to complement e2e / ui unit tests. Does that sound reasonable to you?

calebpollman:
> > LGTM but can we add unit testing since this is not covered by an integ test?
> 
> Would it be okay to add this to e2e test? This is something that will be shared across all frameworks, so IMO this makes sense to be an e2e test. And we already have utility for checking attributes (and it is easier to type something on cypress).
> 
> That said though, I'm definitely with you on adding more unit tests. I think `packages/ui` ([asana](https://app.asana.com/0/1201736086077862/1202277146130103/f)) is a good place to start, because it has a lot of shared utility functions. Then we can start expanding them to `packages/*`to complement e2e / ui unit tests. Does that sound reasonable to you?

Have a couple thoughts: 
- afaik our integ tests currently only test for happy path scenarios, and the changes in this PR are related to errors if i'm not mistaken?
- Think it's great that we have a task for adding unit tests to the _ui_ package, and fully agree unit testing it would give us some quick coverage on shared utils, but that shouldn't prevent us from adding unit tests to new code/bug fixes/refactors to decrease the accumulation of tech debt in other packages

wlee221:
> afaik our integ tests currently only test for happy path scenarios, and the changes in this PR are related to errors if i'm not mistaken?

We already test for validation errors in e2e: https://github.com/aws-amplify/amplify-ui/blob/b7887bde3d6cf4e2c9beb3c184010327a56fb0f7/packages/e2e/features/ui/components/authenticator/sign-up-with-errors.feature#L18-L25

There we test for invalid inputs and confirm that error messages are there. I was thinking about adding a line "And I see validation errors are accessible". 

Note that there are other unhappy test cases in e2e (e.g. [confirm-sign-up](https://github.com/aws-amplify/amplify-ui/blob/b7887bde3d6cf4e2c9beb3c184010327a56fb0f7/packages/e2e/features/ui/components/authenticator/confirm-sign-up.feature#L22-L23), [totp](https://github.com/aws-amplify/amplify-ui/blob/b7887bde3d6cf4e2c9beb3c184010327a56fb0f7/packages/e2e/features/ui/components/authenticator/sign-in-totp-mfa.feature#L25))

> but that shouldn't prevent us from adding unit tests to new code/bug fixes/refactors to decrease the accumulation of tech debt in other packages

+1, and I'm not at all against adding more unit tests in other packages. For Authenticator though, we've found e2e test more effective in doing cross-framework testing. So for cases like this we're testing for the same behavior, I think I prefer e2e because it reduces test duplication between `packages/*`. 

For more nuanced unhappy cases that can't be captured through e2e, or for framework specific API like `Provider`s, I agree with you that we should let unit tests handle them. This is something our library has been lacking for sure.

packages/react/src/components/Authenticator/shared/FormField.tsx
```diff
@@ -29,6 +30,8 @@ export function FormField({
     [name, validationErrors]
   );
   const hasError = errors?.length > 0;
+  const errorId = useStableId();
+  const ariaDescribedBy = hasError ? errorId : undefined;
```
hbuchel, line 34: Do we currently allow customization of these fields (via the Authenticator) that would allow someone to pass through descriptiveText for their own initial aria-describedBy attribute? As we do for the [primitive PasswordField](https://ui.docs.amplify.aws/react/components/passwordfield#descriptive-text)? If not, this is fine. If yes, we would have to do some managing of the id reference list that aria-describedBy supports (you can have more than one ID for the aria-describedBy attribute on an element)

zchenwei (approved):

wlee221:
Had an offline discussion on testing strategy. We'll go with snapshot test + e2e here.

Snapshot and e2e tests are complementary; the former guarantees visual consistency, and e2e tests simulate real human-like interaction and and verifies real use cases. So here, I'll add unit test targeting `FormField.tsx`, then extend pre-existing `sign-up-with-email` e2e test (this will help vue/angular too). ETA EOD today.

packages/react/src/components/Authenticator/shared/FormField.tsx
wlee221: We do not -- https://ui.docs.amplify.aws/react/components/authenticator#form-field-customization.

I'll go ahead with this approach then!

wlee221:
Upon actually writing code, imo unit tests I wrote are sufficient as we already have e2e testing invalid inputs 🤷 

Unit tests added here a0c8ba9194515fcf3a38158f78522386e50188c2 /cc @calebpollman

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+const usernameFormFieldProps: FormFieldProps = {
+  ...defaultFormFieldOptions.username,
+  name: 'username',
+  label: 'Username',
+};
+
+const useAuthenticatorSpy = jest.spyOn(
+  UseAuthenticatorModule,
+  'useAuthenticator'
+);
```
wlee221, line 18: I found `jest.spyOn` easier to work with than `jest.mock` when we want its return value to change between tests.

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
```diff
@@ -0,0 +1,76 @@
+import { render, screen } from '@testing-library/react';
+import * as React from 'react';
+
+import { defaultFormFieldOptions } from '@aws-amplify/ui';
+import { FormField, FormFieldProps } from '../FormField';
```
calebpollman, line 5: Prefer to name test files based on the file they are testing. As we are only testing the `FormField` component and not the `FormFields` think we can name this file _FormField.text.tsx_, what do you think?

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  name: 'username',
+  label: 'Username',
+};
+
+const useAuthenticatorSpy = jest.spyOn(
+  UseAuthenticatorModule,
+  'useAuthenticator'
+);
+
+const errorMessage = 'Username cannot be empty.';
```
calebpollman, line 20: Nit: prefer SCREAMING_SNAKE_CASE for string constants

calebpollman (commented):
LGTM, pending approval on some updates to the test file

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
wlee221: Oops, that's a typo ;)

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const mockInvalidFormField = () => {
+  useAuthenticatorSpy.mockReturnValue({
+    validationErrors: { username: errorMessage },
+  } as any);
+};
+
+const useStableIdSpy = jest.spyOn(UseStableIdModule, 'useStableId');
+
+describe('FormField Snapshots: ', () => {
```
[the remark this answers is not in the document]
wlee221: 84fd4566e2a391ab4603b93f9f41e81a199b647a

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  beforeAll(() => {
+    // mocking ids so snapshots are consistent
+    useStableIdSpy.mockReturnValue('mock-id-0');
```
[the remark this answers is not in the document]
wlee221: 84fd456, I used `mockReturnValue` followed with `.mockRestore`, because this hooks are called more than once on re-renders and we aren't really concerned with the exact number it gets called.

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
wlee221: 84fd456

packages/react/src/components/Authenticator/shared/__tests__/FormFields.test.tsx
wlee221: 84fd456

calebpollman (approved):
LGTM 🚢

PR #1955, opened 2022-05-23
Use reusable e2e workflow

wlee221:
#### Description of changes

This uses [Reusable Workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to create shared `reusable-e2e.yml` workflow. This removes a lot of duplicated code we had!

Note that this is a refactor and not an additive change. I added walkthrough comments for finer details. 

I created a fork [wlee221/amplify-ui](https://github.com/wlee221/amplify-ui) and ran the changes there. Even then, I had to comment out chunks of `e2e` code that required repo secrets because these secrets are not (and shouldn't be) available in my fork. 

That said, I confirmed the existing workflows to be triggered correctly:
- Test and deploy main to `@next`: https://github.com/wlee221/amplify-ui/actions/runs/2374092675
- Test PRs: https://github.com/wlee221/amplify-ui/pull/2
- Fork PRs run when `run-tests` label is added: https://github.com/wlee221/amplify-ui/pull/3

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


.github/workflows/reusable-e2e.yml
```diff
@@ -0,0 +1,328 @@
+on:
+  workflow_call:
```
wlee221, line 2: "workflow_call" declares this file to be a reusable workflow. Other workflows can **call** this file to run e2e files as part of their steps.

wlee221, line 2: "workflow_call" declares this file to be a reusable workflow. Other workflows can **call** this file to run e2e files as part of their steps.

.github/workflows/reusable-e2e.yml
```diff
@@ -0,0 +1,328 @@
...
+    inputs:
+      commit:
+        # commit to run tests against
+        required: true
+        type: string
+      repository:
+        # repository to run tests against (possibly a fork of amplify-ui)
+        required: true
+        type: string
```
wlee221, line 11: Here's where we get the `ref` to run e2e tests against. This follows the recommended security practice: [link](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-action-instead-of-an-inline-script-recommended).

.github/workflows/reusable-e2e.yml
```diff
@@ -0,0 +1,328 @@
...
+      DOMAIN:
+        required: true
+      PHONE_NUMBER:
+        required: true
+      USERNAME:
+        required: true
+      NEW_PASSWORD:
+        required: true
+      VALID_PASSWORD:
+        required: true
```
wlee221, line 26: Here are the required secrets to run `e2e` tests. Note that we're not introducing any new secrets here than what we already had. This PR is a refactor and is not an additive change.

wlee221, line 26: Here are the required secrets to run `e2e` tests. Note that we're not introducing any new secrets here than what we already had. This PR is a refactor and is not an additive change.

.github/workflows/reusable-e2e.yml
```diff
@@ -0,0 +1,320 @@
...
+
+jobs:
```
wlee221, line 28: Below is copy + paste of e2e test codes we already had. 

The only difference is that now we're using `inputs.commit` and `inputs.repository` that we defined above.

.github/workflows/test-deploy-main.yml
```diff
@@ -11,300 +11,22 @@ on:
...
+      commit: ${{ github.event.pull_request.head.sha }}
+      repository: ${{ github.event.pull_request.head.repo.full_name }}
+    secrets:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      DOMAIN: ${{ secrets.DOMAIN }}
+      PHONE_NUMBER: ${{ secrets.PHONE_NUMBER }}
+      USERNAME: ${{ secrets.USERNAME }}
+      NEW_PASSWORD: ${{ secrets.NEW_PASSWORD }}
+      VALID_PASSWORD: ${{ secrets.VALID_PASSWORD }}
```
wlee221, line 26: Here's what the caller workflow syntax looks like! 
- `uses: aws-amplify/amplify-ui/.github/workflows/reusable-e2e.yml@main` calls the workflow we created above. 
- `with: ` passes the `input` fields.
- `secrets` pass the secrets reusable workflow needs.

.github/workflows/test-deploy-main.yml
```diff
@@ -11,300 +11,22 @@ on:
...
-          persist-credentials: false
-      - name: Setup Node.js LTS
-        uses: actions/setup-node@v2
-        with:
-          node-version: lts/*
-          cache: 'yarn'
-      - name: Install packages
-        run: yarn --no-lockfile
-      - name: Build ui package
-        run: yarn ui build
```
wlee221, line 37: Just like that, repeated codes are gone!

.github/workflows/test-prs.yml
```diff
@@ -24,11 +24,7 @@ jobs:
     if: |
       (github.event.pull_request.head.repo.full_name == github.repository) ||
       (github.event.action == 'labeled' && github.event.label.name == 'run-tests')
-    outputs:
-      ref: ${{ steps.get-ref.outputs.ref }}
-      commit: ${{ steps.get-ref.outputs.commit }}
```
wlee221, line 29: These were previously needed for e2e testing, but now is no longer needed now that it's being done by the reusable workflow

.github/workflows/test-deploy-main.yml
```diff
@@ -11,300 +11,22 @@ on:
...
+      commit: ${{ github.sha }}
+      repository: ${{ github.repository }}
+    secrets:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      DOMAIN: ${{ secrets.DOMAIN }}
+      PHONE_NUMBER: ${{ secrets.PHONE_NUMBER }}
+      USERNAME: ${{ secrets.USERNAME }}
+      NEW_PASSWORD: ${{ secrets.NEW_PASSWORD }}
+      VALID_PASSWORD: ${{ secrets.VALID_PASSWORD }}
```
wlee221, line 26: Here's what the caller workflow syntax looks like! 
- `uses:` calls the workflow we created above. 
- `with:` passes the `input` fields.
- `secrets:` passes the secrets reusable workflow needs. 

While it seems redundant to pass repo secrets here when the reusable workflow already has access to them, this is a weird syntax limitation by GitHub actions [discussions](https://github.community/t/reusable-workflows-secrets-and-environments/203695/46).

zchenwei (approved):
This looks really neat and good! Just a question why not call it `reusable-tests` since the file also includes unit tests and docs tests

.github/workflows/reusable-e2e.yml
```diff
@@ -0,0 +1,320 @@
...
+      - name: Get CLI versions
+        id: cli-version
+        run: echo "::set-output name=version::$(amplify --version)"
+
+      - name: Create or restore environments cache
+        id: environments-cache
+        uses: actions/cache@v2
+        with:
+          path: environments/**/aws-exports.js
+          key: ${{ runner.os }}-environments-${{ steps.cli-version.outputs.version }}-${{ hashFiles('environments/**/amplify/**') }}
```
0618, line 240: What do you think if we use an environment, like `${{ secrets.CACHE_VERSION }}`, so that we can force clean the cache?

ref: https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions

0618 (approved):
with a question

.github/workflows/reusable-e2e.yml
wlee221: Ah, this cache key is dynamic already! This key will be changed based on what the latest cli version is and the sha sum of environments files. So whenever either those change, the previous cache key will be no longer be used and be cleaned.

wlee221:
> Just a question why not call it reusable-tests [...]

Good point! I called it reusable-e2e to distinguish from other test workflows like our canaries (and we'll have more to come). Open to suggestions 👍

PR #1956, opened 2022-05-23
re export amplify provider as theme provider and update all documenta…

jacoblogan:
#### Description of changes
This PR re exports the AmplifyProvider as ThemeProvider and updates all documentation to use the newly exported ThemeProvider.  This PR does not remove AmplifyProvider and is not a breaking change.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


zchenwei (approved):
This is looking good! You may want to update the exports snapshot :)

docs/src/pages/[platform]/components/card/examples/CardThemeExample.tsx
```diff
@@ -21,7 +21,7 @@ const cardTheme = {
 
 export const CardThemeExample = () => {
   return (
-    <AmplifyProvider theme={cardTheme}>
+    <ThemeProvider theme={cardTheme}>
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/card/examples/CardThemeExample.tsx:24]: We should add a note in an Info alert above our Theme examples that the AmplifyProvider was renamed to ThemeProvider, and the ThemeProvider export is available since version 2.x.x (specific version added after we release?)

docs/src/pages/[platform]/getting-started/installation/dependencies.react.mdx
```diff
@@ -36,19 +36,19 @@ export default function App() {
 
 ### Custom Theme
 
-To add a [custom theme](/theming), wrap your app in the `AmplifyProvider` component and include the default styling.
+To add a [custom theme](/theming), wrap your app in the `ThemeProvider` component and include the default styling.
```
Comment (hbuchel) [inline docs/src/pages/[platform]/getting-started/installation/dependencies.react.mdx:39]: Just a note, when you merge in latest from main,  this section was removed since it doesn't provide much information about how to actually add a custom theme. We can probably keep it removed?

Comment (jacoblogan) [reply docs/src/pages/[platform]/getting-started/installation/dependencies.react.mdx]: thanks this has been removed

docs/src/pages/[platform]/components/card/examples/CardThemeExample.tsx
Comment (jacoblogan) [reply docs/src/pages/[platform]/components/card/examples/CardThemeExample.tsx:24]: Thanks that is a good call out, I've added an alert to all of the component examples and the theming getting started page.  For the version number I went with 2.18.3 because it looks like that will be our next release. But we can update that if we need to.

joebuono (commented):
Thanks for doing this. Can we also:
- rename the `packages/react/src/components/AmplifyProvider` directory
- `export * from './ThemeProvider'` in `packages/react/src/components/index.tsx`
- use `ThemeProvider` in `AmplifyProvider.test.tsx`

joebuono (approved):
Thanks for making the changes!

PR #1965, opened 2022-05-24
Replacing sandpack with react-live

dbanksdesign:
#### Description of changes

Removing Sandpack (codesandbox) with react-live. This should really speed up the homepage. 

https://react-live.d2q023dzpor7ad.amplifyapp.com/

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


hbuchel (commented):
Our StepperField is getting squashed somehow (so the current value is hidden)? Does it do this for you?

Screen Shot 2022-05-26 at 1 10 52 PM

joebuono (approved):
Nice! Excited to use this

docs/src/pages/_document.page.tsx
```diff
@@ -25,7 +25,7 @@ const getCSPContent = (context: Readonly<HtmlProps>) => {
       style-src 'self' 'unsafe-inline';
       font-src 'self' data:;
       frame-src 'self' *.codesandbox.io;
-      img-src 'self' cm.everesttech.net amazonwebservices.d2.sc.omtrdc.net dpm.demdex.net https://images.unsplash.com;
+      img-src 'self' cm.everesttech.net amazonwebservices.d2.sc.omtrdc.net dpm.demdex.net https://images.unsplash.com data:;
```
0618, line 28: Not sure if we want to accept all `data:`. @reesscot  and I talked about it and trying to only allow the specific hash, but I didn't figure out how

docs/src/pages/_document.page.tsx
```diff
@@ -36,9 +36,9 @@ const getCSPContent = (context: Readonly<HtmlProps>) => {
     style-src 'self' 'unsafe-inline';
     font-src 'self';
     frame-src 'self' *.codesandbox.io aws.demdex.net;
-    img-src 'self' cm.everesttech.net amazonwebservices.d2.sc.omtrdc.net dpm.demdex.net https://images.unsplash.com;
+    img-src 'self' cm.everesttech.net amazonwebservices.d2.sc.omtrdc.net dpm.demdex.net https://images.unsplash.com data:;
     connect-src 'self' *.shortbread.aws.dev amazonwebservices.d2.sc.omtrdc.net dpm.demdex.net https://*.algolia.net https://*.algolianet.com;
-    script-src 'self' '${cspInlineScriptHash}' a0.awsstatic.com;
+    script-src 'unsafe-eval' 'self' '${cspInlineScriptHash}' a0.awsstatic.com;
```
0618, line 41: Why do we need to add this to prod? 

Is it safe? Is there an alternative way for `eval`?

0618:
I see there are already 2 approvals. To prevent it from accidentally merged, I'm adding a blocker

0618 (dismissed):
I see there are already 2 approvals. To prevent it from accidentally merged, I'm adding a blocker

docs/src/pages/_document.page.tsx
0618: also, the warning is already in `main`, so it might be more clear to resolve it in a separated PR. What do you think?

docs/src/pages/_document.page.tsx
dbanksdesign: This was reviewed in an appsec meeting and it is ok. We need it because that is how react-live works (evaluating the code in the browser). The risk is low because there is no customer data on the site and it all runs in the browser so the only issue would be if the user themselves wrote malicious code that affected themself. We didn't need this previously because codesandbox was running in an iframe.

docs/src/pages/_document.page.tsx
dbanksdesign: We can remove it, I'll just need to rewrite the inline background image in the CSS :/

reesscot (approved):
LGTM!

0618 (approved):

PR #1967, opened 2022-05-24
fix(ui): Remove default css margins for <body>

slaymance:
#### Description of changes
Browsers will set default margins for elements (known as User Agent Style sheets). Since `<body>` elements in Amplify UI applications are set to have a `min-height: 100vh;`, whenever a browser sets a default `margin` for `<body>` a scroll bar will always appear on the page even if there aren't enough elements on the page to warrant scrolling.

This PR resets the `margin` for `<body>` (and other elements) to ensure setting `min-height: 100vh;` will only take up the available screen real estate. You can review [https://piccalil.li/blog/a-modern-css-reset/](https://piccalil.li/blog/a-modern-css-reset/) which we use as a reference for our CSS reset.

#### Issue #, if available
N/A

Below is an example of how an Amplify UI application currently looks without a margin reset. Note that I've made the `<html>` element have a background of `lightpink` and the `<body>` element have a background of `lightblue` for visual clarity. Notice that a scroll bar appears even though this app only contains an `<h1>` element:
with-margins

With this PR, the `<body>` margin is reset and the height takes up the entire screen with no scroll bar:
without-margins

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated - N/A
- [x] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated - N/A

### DIFF
--- a/packages/ui/src/theme/css/styles.scss
+++ b/packages/ui/src/theme/css/styles.scss
@@ -20,6 +20,23 @@ html:focus-within {
   scroll-behavior: smooth;
 }
 
+// Remove default margin
+// https://piccalil.li/blog/a-modern-css-reset/
+body,
+h1,
+h2,
+h3,
+h4,
+p,
+li,
+figure,
+figcaption,
+blockquote,
+dl,
+dd {
+  margin: 0;
+}
+
 // https://piccalil.li/blog/a-modern-css-reset/
 body {
   min-height: 100vh;
--- a/packages/ui/src/theme/css/styles.scss
+++ b/packages/ui/src/theme/css/styles.scss
@@ -20,25 +20,9 @@ html:focus-within {
   scroll-behavior: smooth;
 }
 
-// Remove default margin
-// https://piccalil.li/blog/a-modern-css-reset/
-body,
-h1,
-h2,
-h3,
-h4,
-p,
-li,
-figure,
-figcaption,
-blockquote,
-dl,
-dd {
-  margin: 0;
-}
-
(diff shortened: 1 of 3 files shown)


jacoblogan (approved):

packages/ui/src/theme/css/styles.scss
```diff
@@ -20,6 +20,23 @@ html:focus-within {
   scroll-behavior: smooth;
 }
 
+// Remove default margin
+// https://piccalil.li/blog/a-modern-css-reset/
+body,
```
reesscot, line 25: Can we scope this to body only since that's the root of the issue? I don't see a reason to reset the other elements at this time and it may mess up the styling of our other primitives.

reesscot (changes requested):

packages/ui/src/theme/css/styles.scss
slaymance: I made this change and decided to remove the `<h1>` tag from the next examples since it isn't necessary and the other examples don't have a similar tag.

slaymance:
Upon reviewing how other UI libraries treat global styles as this PR is proposing to do, we think it's best to minimize how much global styling the `@aws-amplify/ui-react` library does for users. One of the only libraries that does employ global styling (including a browser margin reset) is Ant Design, and that library has an outstanding popular issue to [remove global styles](https://github.com/ant-design/ant-design/issues/9363).

PR #1980, opened 2022-05-25
feat: adding missing label props to make text in i18n file customizable

zchenwei:
#### Description of changes

There are a few texts in `i18n` file for primitives's labels and since we do not want to bring in any i18n lib and leave the flexibility to our users, we want make them customizable through props. This PR does the following:

- Adding missing related label props to make texts in `i18n` file customizable by users themselves
- Move texts from `i18n` file to `constants` file since the file name is misleading right now
- Update the Accessibility section for `Pagination` to make the description there more clear
- Updating unit tests to cover the newly added props use cases

Screen Shot 2022-05-24 at 6 43 52 PM

Add new unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


0618:
No Changeset found 👮

docs/src/pages/[platform]/components/pagination/examples/PaginationAccessibilityExample.tsx
```diff
@@ -0,0 +1,14 @@
+import { usePagination, Pagination } from '@aws-amplify/ui-react';
+
+export const PaginationAccessibilityExample = () => {
+  const paginationProps = usePagination({ totalPages: 5 });
+  return (
+    <Pagination
+      pageLabel="Jump to page"
+      currentPageLabel="You are on page"
+      leftArrowLabel="Back to previous page"
+      rightArrowLabel="Forward to next page"
```
Comment (dbanksdesign) [inline docs/src/pages/[platform]/components/pagination/examples/PaginationAccessibilityExample.tsx:10]: Can we change these to `nextLabel` and `previousLabel` so they make sense for RTL?

packages/react/src/primitives/Field/FieldClearButton.tsx
```diff
@@ -3,16 +3,20 @@ import * as React from 'react';
...
 import { FieldGroupIconButton } from '../FieldGroupIcon';
 import { IconClose } from '../Icon';
-import { SharedText } from '../shared/i18n';
+import { SharedText } from '../shared/constants';
 
 const ariaLabelText = SharedText.Fields.ariaLabel.clearField;
 
 const FieldClearButtonPrimitive: Primitive<FieldClearButtonProps, 'button'> = (
-  props,
+  { ariaLabel, size, ...rest },
```
dbanksdesign, line 11: Nit: should we do `ariaLabel = SharedText.Fields.ariaLabel.clearField` here like you do above in the Collection component?

packages/react/src/primitives/PasswordField/__tests__/PasswordField.test.tsx
```diff
@@ -97,7 +97,7 @@ describe('PasswordField component', () => {
       />
     );
 
-    const button = await screen.queryByRole('button');
```
dbanksdesign, line 100: Just curious why this change is needed

packages/react/src/primitives/StepperField/StepperField.tsx
```diff
@@ -35,6 +34,8 @@ const StepperFieldPrimitive: Primitive<StepperFieldProps, 'input'> = (
     isDisabled,
     isReadOnly,
     isRequired,
+    increaseButtonLabel = SharedText.StepperField.ariaLabel.IncreaseTo,
+    decreaseButtonLabel = SharedText.StepperField.ariaLabel.DecreaseTo,
```
dbanksdesign, line 38: Same here, I prefer `increaseButtonLabel`

dbanksdesign (commented):
Overall this is great! Just some comments about naming

docs/src/pages/[platform]/components/pagination/examples/PaginationAccessibilityExample.tsx
Comment (zchenwei) [reply docs/src/pages/[platform]/components/pagination/examples/PaginationAccessibilityExample.tsx:10]: Thanks for the suggestion! I am kinda on the fence when naming this two. I was actually thinking either `prev` or `next` might be confusing because they can be sibling buttons  of current page semantically versus `<` and `>` buttons. But I am good with `next` and `previous`

docs/src/pages/[platform]/components/pagination/react.mdx
```diff
@@ -29,7 +30,7 @@ Import the Pagination component. To use Pagination as an uncontrolled component,
   <ExampleCode>
 ```tsx file=./examples/DefaultPaginationExample.tsx
 
-```
+````
```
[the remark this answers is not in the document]
Comment (zchenwei) [reply docs/src/pages/[platform]/components/pagination/react.mdx:33]: ah, I will fix it

packages/react/src/primitives/Pagination/usePaginationItems.tsx
```diff
@@ -19,6 +19,11 @@ export const usePaginationItems = (
   totalPages: number,
   hasMorePages: boolean,
   siblingCount: number,
+  currentPageLabel: string = SharedText.PaginationItem.invisibleLabel
+    .currentPage,
+  pageLabel: string = SharedText.PaginationItem.ariaLabel.goToPage,
+  leftArrowLabel: string = SharedText.PaginationItem.ariaLabel.goToPrevPage,
+  rightArrowLabel: string = SharedText.PaginationItem.ariaLabel.goToNextPage,
```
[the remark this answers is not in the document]
zchenwei: Good call

packages/react/src/primitives/Pagination/usePaginationItems.tsx
```diff
@@ -66,12 +71,13 @@ export const usePaginationItems = (
             type="page"
             page={item as number}
             currentPage={currentPage}
+            currentPageLabel={currentPageLabel}
             onClick={onChange}
-            ariaLabel={`Go to page ${item}`}
+            ariaLabel={`${pageLabel} ${item}`}
```
hbuchel, line 76: I see that we're really just porting these changes over so as to not introduce a breaking change, so this is also out of scope, but want to point this out so we might reference it for future discussions: When we do a review of our i18n capabilities, we should consider how we would support interpolation in our string translations. Example, as is this works for "Go to page 31" or "translatedText {s}" as the supplied string, but for Arabic or Japanese or some other languages the supplied string might look like: 
"{s} translatedText" or even "translatedText {s} translatedText". No suggested changes now, just putting this down somewhere so we can point to it.

packages/react/src/primitives/StepperField/StepperField.tsx
zchenwei: I can make the changes for pagination label. But I think for the other ones, it would be a breaking change I think? Actually that's the reason why I didn't refactor the `ShareText` structure to make it more consistent because I found that we export it in our bundle, which we shouldn't IMO 🤣

packages/react/src/primitives/shared/constants.ts
```diff
@@ -515,6 +515,47 @@ export const ComponentClassObject = {
...
+    },
+  },
+  StepperField: {
+    ariaLabel: {
+      IncreaseTo: 'Increase to',
+      DecreaseTo: 'Decrease to',
+    },
+  },
+};
+
```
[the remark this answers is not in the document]
zchenwei: Exactly, I want to do something here but I dare not to because it could break things. Same reason above. I can update the newly added one from Pagination

packages/react/src/primitives/Pagination/usePaginationItems.tsx
zchenwei: Oh yes, definitely a good call out here. We have some accessibility debt here. Thanks for the insights! Maybe I should add a TODO there to remind us as well

zchenwei:
> No Changeset found 👮

You save me 🤣

packages/react/src/primitives/shared/constants.ts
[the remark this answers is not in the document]
zchenwei: Discuss offline, we will mark `ShareText` as deprecated and remove it in the future and use `ComponentText` internally moving forward.

dbanksdesign (approved):
One potentially out-of-scope comment about composed primitives passing props. Other than that, agree with @hbuchel's comments, but I think this is good to go for now.

slaymance (approved):
Looks great, nice improvement for users!

packages/react/src/primitives/Collection/Collection.tsx
```diff
@@ -98,7 +98,7 @@ export const Collection = <Item,>({
       {isSearchable ? (
         <Flex className={ComponentClassNames.CollectionSearch}>
           <SearchField
-            label={SharedText.Collection.SearchFieldLabel}
+            label={searchLabel}
             placeholder={searchPlaceholder}
```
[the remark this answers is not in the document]
zchenwei: Yea, I like this ideas. That way we do not have to define quite a lot duplicate props on parents just for sub-primitives customization

PR #1686, opened 2022-04-13
chore: align typescript version at root devDependencies

reesscot:
#### Description of changes

This PR aligns the typescript package version used by our monorepo to the latest version `4.6.3`. This prevents us from using different versions in different packages accidentally. I've left angular out of this update since I think it needs a very particular version. I also left the canary apps out as well, since they are meant to be used directly without the root node_modules folder.

#### Description of how you validated changes
```
yarn clean && yarn build
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1983, opened 2022-05-26
fix(security): ping eventsource version to fix security#42

0618:
#### Description of changes

- Update for the dependency security issue https://github.com/aws-amplify/amplify-ui/security/dependabot/42
- Move @angular-devkit/build-angular to root to prevent miss-updating versions in different folders in the future

```sh
~~yarn clean && yarn build~
```

Update: `yarn clean && yarn build` doesn't work because after `yarn clean`, the packages never got installed.

```
yarn angular-example build
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated


package.json
```diff
@@ -53,6 +53,7 @@
     "ws": "^7.4.6"
   },
   "devDependencies": {
+    "@angular-devkit/build-angular": "^13.3.7",
```
calebpollman, line 56: Is there a reason that this needs to move up to the root devDeps? Would prefer we keep platform specific deps scoped to their platforms unless we need to move them at the root

calebpollman (commented):
Thanks for taking care of this! Left a question

package.json
0618: Yep, see the PR description....same reason as this PR https://github.com/aws-amplify/amplify-ui/pull/1686

0618: I kept it in `amplify-ui-angular-mono`, but moved those from `angular-example` and `angularcli` to root

packages/angular/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
-    "@angular-devkit/build-angular": "~0.1102.3",
+    "@angular-devkit/build-angular": "^13.3.7",
```
zchenwei, line 15: I think this is supposed to be made on canary?

canary/apps/angular/angularcli/package.json
```diff
@@ -27,7 +27,6 @@
     "zone.js": "~0.11.3"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "~0.1102.13",
```
zchenwei, line 30: You should add this back to canary?

package.json
0618: Alright! Thanks for the comment. I just updated the package on the original files 😁

zchenwei (dismissed):
LGTM. Just add a changeset! And I am not sure if this version bump will break any angular tests because it seems to me the previous one is very specific though old

packages/angular/package.json
reesscot: +1, would like @wlee221 to look at this change

reesscot: Also, I don't think we have officially tested Angular 13 yet, so I'm concerned about what changes this would make.

.changeset/moody-jobs-beg.md
```diff
@@ -0,0 +1,6 @@
+---
+'angular-example': patch
+'amplify-ui-angular-mono': patch
```
reesscot, line 3: We don't release either of these packages, so this will break changesets if it's included. we may want to include `@aws-amplify/ui-angular` if we really decide to bump the devkit version in packages/angular/package.json

0618: Done.

zchenwei (approved):

ErikCH (approved):
👍

calebpollman (approved):
🚢

PR #2010, opened 2022-05-31
Authenticator removes focus from input after Authenticated

ErikCH:
#### Description of changes
When users are following the Authenticator [getting started guide](https://docs.amplify.aws/start/getting-started/auth/q/integration/react/) in the official docs, they get stuck on the `Add authentication` page. Every typing in the input, it loses focus after every letter.

It seems to be related to #1451 

#### Issue #, if available
[#4306](https://github.com/aws-amplify/docs/issues/4306)

Wrote test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/e2e/cypress/integration/common/sign-in.ts
+++ b/packages/e2e/cypress/integration/common/sign-in.ts
@@ -34,6 +34,14 @@ When(
   }
 );
 
+When('I type {string} in the input field', (text: string) => {
+  cy.get('input').type(text);
+});
+
+When('I see input text {string}', (text: string) => {
+  cy.get('input').should('have.value', text);
+});
+
 When('I type my password', () => {
   cy.findInputField('Password').type(Cypress.env('VALID_PASSWORD'));
 });
--- a/packages/react/src/components/Authenticator/Router/Router.tsx
+++ b/packages/react/src/components/Authenticator/Router/Router.tsx
@@ -45,17 +45,17 @@ const getRouteComponent = (route: string) => {
   }
 };
 
-export function useRouterChildren(
-  children: RouterProps['children']
-): (props?: Omit<RouterProps, 'children'>) => JSX.Element {
+const isUnauthenticatedRoute = (route: string): boolean => {
+  return !(route === 'authenticated' || route === 'signOut');
+};
+
+export function useRouterChildren(): (
+  props?: Omit<RouterProps, 'children'>
+) => JSX.Element {
   const { route } = useAuthenticator(({ route }) => [route]);
 
   return React.useMemo(() => {
-    const isUnauthenticatedRoute = !(
-      route === 'authenticated' || route === 'signOut'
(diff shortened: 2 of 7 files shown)


packages/e2e/cypress/integration/common/sign-in.ts
```diff
@@ -34,6 +34,14 @@ When(
   }
 );
 
+When('I type {string} in the input field', (text: string) => {
+  cy.get('input').type(text);
+});
```
ErikCH, line 39: I couldn't find the right helper functions, so I wrote my own.

packages/react/src/components/Authenticator/Router/Router.tsx
```diff
@@ -79,9 +69,22 @@ export function Router({
...
+  // `Authenticator` might not have user defined `children` for non SPA use cases.
+  if (route === 'authenticated' || route === 'signOut') {
+    if (!children) {
+      return null;
+    }
+
+    return typeof children === 'function'
+      ? children({ signOut, user }) // children is a render prop
+      : children;
+  }
```
ErikCH, line 86: I moved this code out of the React.memo since it was causing a rerender whenever the input changed.

ErikCH: And that seemed to be causing the input to lose focus

calebpollman: Curious, why is `children` changing on input events?

ErikCH: I know the children have an input, that is tied to an `onChange` event, that updates `useState`. I'm not sure what's causing the `useMemo` to fire though and cause the issue.

calebpollman: Think we need to root cause the re-render before moving the code out of the `useMemo`. Not saying that this  is the wrong approach, but we should know for certain that this is the actual cause. 

If we are going to move forward with this approach we will want to remove the duplicate logic in the `useRouterChildren` hook, or consider removing the hook altogether

ErikCH: I can create a utility function that returns back `isUnauthenticatedRoute`. 

OK, I tested some more. From my testing `useMemo` is rerendering anytime the `children` change. That's causing the input to lose focus on every key press. It only occurs when the children have an input that is using a `useState` hook.

ErikCH: #5577559816d0be32fbbc96e70f0de8c8b5eece3b

calebpollman: Still a bit confused. The `useMemo` **should** update when the value of `children` changes, but not sure why `children` would change due to input events

PR #1984, opened 2022-05-26
docs: Getting Started CRA Tutorial

joebuono:
#### Description of changes

Screen Shot 2022-05-26 at 1 12 15 PM

This PR creates a new `Create React App` page under Getting Started after Usage. 

The page guides the user through creating a new CRA app, installing the Amplify UI library, and editing the files so that they can see the Amplify UI components in action. 

I chose to include both a simple example to get the customer up and running ASAP, and then a more complex example to show what's possible using Amplify UI. 

I annotated the files with my specific thoughts/questions. I am totally open to all suggestions. 

#### Description of how you validated changes

Ran the docs locally and also followed the tutorial. 

#### Checklist

- [X] PR description included
- [X] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2017, opened 2022-06-01
docs: Next.js Getting Started Tutorial

joebuono:
#### Description of changes

The Next.js getting started page is almost identical to the [CRA tutorial PR](https://github.com/aws-amplify/amplify-ui/pull/1984), with just some tiny changes to the description, file names, and example code. 

I also factored out some of the shared code into a `/shared` folder. 

Ran the docs locally. 

#### Checklist

- [X] PR description included


zchenwei (approved):
Good job! Just a quick question. Looks like we have a few props in the examples that are using fixed size. Have you tested those examples against different break points to ensure they are responsive? Same question for CRA examples but I think they are sharable.

joebuono:
> Good job! Just a quick question. Looks like we have a few props in the examples that are using fixed size. Have you tested those examples against different break points to ensure they are responsive? Same question for CRA examples but I think they are sharable. Non blocking though

Good idea! I just adjusted the examples so that they are responsive.

docs/src/pages/[platform]/getting-started/nextjs/examples/AdvancedExampleCode.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  step={1}
+                  labelHidden
+                  width="10rem"
+                  isDisabled={!currentPainting.inStock}
+                />
+              </Flex>
+              <Button
+                variation="primary"
+                onClick={handleAddToCart}
+                // width="10rem"
```
Comment (0618) [inline docs/src/pages/[platform]/getting-started/nextjs/examples/AdvancedExampleCode.tsx:197]: nit: what's this comment for?

docs/src/pages/[platform]/getting-started/shared/paintings.ts
```diff
@@ -0,0 +1,122 @@
+export const PAINTINGS = [
```
Comment (0618) [inline docs/src/pages/[platform]/getting-started/shared/paintings.ts:1]: should the file be name as `paintingData.ts` or put it in a data folder `data/paintings.ts`? I know it's for an example code, so we may want to have a higher bar even on naming

0618 (approved):
left small comments 😁

docs/src/pages/[platform]/getting-started/nextjs/index.page.mdx
```diff
@@ -0,0 +1,11 @@
+---
+title: Next.js
+metaTitle: Next.js
+description: How to get started using Amplify UI with Next.js
+metaDescription: How to get started using Amplify UI with Next.js
```
Comment (0618) [inline docs/src/pages/[platform]/getting-started/nextjs/index.page.mdx:5]:
```suggestion
description: How to get started using Amplify UI with Next.js
```

If they are the same, we only need to keep one

PR #1947, opened 2022-05-20
update collection docs, update existing examples, add grid example, a…

jacoblogan:
#### Description of changes
Updates to the collection primitive docs.  Use the new example format, update existing examples, add grid and styling examples.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated
- [x] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2021, opened 2022-06-02
Add design tokens

jacoblogan:
#### Description of changes
Add new design tokens to collection and searchfield primitives.  Previously these two primitives  had no design tokens so they couldn't be targeted for theming.  This pull request adds a few basic tokens

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/ui/src/theme/css/component/collection.scss
```diff
@@ -1,12 +1,52 @@
...
-.amplify-collection-search {
-  flex-direction: row;
-  justify-content: center;
-}
-
-.amplify-collection-pagination {
-  justify-content: center;
+  &-pagination {
+    justify-content: center;
+    --amplify-components-pagination-current-color: var(
```
jacoblogan, line 5: these tokens are set differently because we want to keep our 010 specificity and these components are composed of other components but we have no way of passing a class directly onto the pieces that we want to theme.  So we are setting the css variable of the pieces we want to theme within the local scope of the component.

packages/ui/src/theme/css/component/searchField.scss
```diff
@@ -0,0 +1,26 @@
+.amplify-searchfield {
+  --amplify-components-fieldcontrol-color: var(
+    --amplify-components-searchfield-color
+  );
+  &__search {
+    color: var(--amplify-components-searchfield-search-color);
```
jacoblogan, line 6: it seems like this would create a circular dependency but this works out because the values are resolved higher up in the page (Thanks danny for this tip).

jacoblogan:
The theme example for the collection primitive will be added to the docs as part of https://github.com/aws-amplify/amplify-ui/pull/1947 once these tokens have been merged in.

packages/react/src/primitives/SearchField/SearchField.tsx
```diff
@@ -13,6 +13,7 @@ const SearchFieldPrimitive: Primitive<SearchFieldProps, 'input'> = (
   {
     autoComplete = 'off',
     className,
+    isDisabled,
```
jacoblogan, line 16: pulling out isDisabled so that it can be passed to both the input and search button

.changeset/clever-bulldogs-thank.md
```diff
@@ -0,0 +1,7 @@
+---
+"@aws-amplify/ui-react": patch
+"@aws-amplify/ui": patch
+---
+
```
zchenwei, line 5: It may worth a minor bump?

dbanksdesign: +1

packages/ui/src/theme/css/component/collection.scss
dbanksdesign: Wow what a super cool solution!

packages/ui/src/theme/tokens/components/collection.ts
```diff
@@ -0,0 +1,92 @@
...
+  };
+  button: {
+    color: DesignToken<ColorValue>;
+    _hover: {
+      backgroundColor: DesignToken<BackgroundColorValue>;
+      color: DesignToken<ColorValue>;
+    };
+    _disabled: {
+      color: DesignToken<ColorValue>;
+    };
```
dbanksdesign, line 20: Should these be `StateTokens` type? Also should we add `_focus` and `_active` states too?

packages/ui/src/theme/tokens/components/collection.ts
```diff
@@ -0,0 +1,92 @@
...
+  };
+}
+
+interface StateTokens {
```
dbanksdesign, line 24: Should we get this interface from buttons.ts?

packages/ui/src/theme/tokens/components/collection.ts
```diff
@@ -0,0 +1,92 @@
...
+  color: DesignToken<ColorValue>;
+  backgroundColor: DesignToken<BackgroundColorValue>;
+}
+
+export interface SearchTokens {
+  color: DesignToken<ColorValue>;
```
dbanksdesign, line 30: It is hard to see what this token would be applied to. The input? The label? The wrapper?

packages/ui/src/theme/tokens/components/collection.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+interface CollectionTokens {
+  pagination: PaginationTokens;
+  search: SearchTokens;
+}
+
+export const collection: CollectionTokens = {
+  pagination: {
+    current: {
```
dbanksdesign, line 46: Maybe we should add a note somewhere that we don't want to just import and reference the pagination tokens like 

```typescript
import {pagination} from './pagination'; 
//...
pagination: pagination
```

Because that will break reference. Also that we are explicitly exposing certain tokens and can add more in the future if necessary.

packages/ui/src/theme/tokens/components/searchField.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export interface SearchFieldTokens {
+  color: DesignToken<ColorValue>;
+  search: SearchTokens;
+}
+
+export const searchfield: SearchFieldTokens = {
+  color: { value: '{components.fieldcontrol.color.value}' },
+  search: {
```
dbanksdesign, line 26: Should this be `button`?

dbanksdesign (commented):
Just a few minor comments/questions, but overall looks good!

packages/ui/src/theme/tokens/components/collection.ts
jacoblogan: I've updated these to use the hover and disabled types declared in the pagination tokens however I did not add the focus and active states because they don't exist on the pagination component.

packages/ui/src/theme/tokens/components/collection.ts
jacoblogan: This token has been renamed to add more clarity

zchenwei (approved):
LGTM!

docs/src/pages/[platform]/components/searchfield/examples/SearchFieldThemeExample.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const SearchFieldThemeExample = () => {
+  const { tokens } = useTheme();
+
+  const theme: Theme = {
+    name: 'searchfield-theme',
+    tokens: {
+      components: {
+        searchfield: {
+          color: { value: 'red' },
+          search: {
```
Comment (dbanksdesign) [inline docs/src/pages/[platform]/components/searchfield/examples/SearchFieldThemeExample.tsx:17]: I think this should be `button: {` right? Looking at the searchField.ts file below

dbanksdesign (approved):
Conditional approval, I think just the example needs a one line update

reesscot (approved):
LGTM, thanks for fixing this!

docs/src/pages/[platform]/components/searchfield/examples/SearchFieldThemeExample.tsx
Comment (jacoblogan) [reply docs/src/pages/[platform]/components/searchfield/examples/SearchFieldThemeExample.tsx]: thanks that has been updated to `button`

PR #2054, opened 2022-06-08
feat: adding colors theme key support in style props

zchenwei:
#### Description of changes

This PR will be adding colors theme key support in style props. Styling with theme key right away and it will be converted into CSS variable reference under the hood. Responsive value support as well. No need to call `useTheme` hook to get the tokens first anymore!

**Example Code**

```jsx
import { View } from '@aws-amplify/ui-react';

export const Demo = () => {
  return <View backgroundColor="pink.10" color="red.40"/>
}
```

For more colors, please reference [here](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/tokens/colors.ts#L100-L283).

Kapture 2022-06-08 at 14 45 06

Kapture 2022-06-08 at 14 38 28

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/primitives/shared/responsive/__tests__/utils.test.ts
+++ b/packages/react/src/primitives/shared/responsive/__tests__/utils.test.ts
@@ -1,4 +1,7 @@
-import { Breakpoints } from 'src/primitives/types/responsive';
+import { renderHook } from '@testing-library/react-hooks';
+
+import { useTheme } from '../../../../hooks/useTheme';
+import { Breakpoints } from '../../../types/responsive';
 import { getValueAtCurrentBreakpoint } from '../utils';
 
 const breakpoints: Breakpoints = {
@@ -15,8 +18,31 @@ describe('getValueAtCurrentBreakpoint', () => {
     const result = getValueAtCurrentBreakpoint('20px', 'base', breakpoints);
     expect(result).toBe('20px');
   });
+
+  it('should return number values directly', () => {
+    const result = getValueAtCurrentBreakpoint(0, 'base', breakpoints);
+    expect(result).toBe(0);
+  });
+
+  it('should return design token toString vaule directly', () => {
+    const { result } = renderHook(() => useTheme());
+    const { tokens } = result.current;
+    const value = getValueAtCurrentBreakpoint(
+      tokens.colors.red[10],
+      'base',
+      breakpoints
+    );
+    expect(value).toBe(tokens.colors.red[10].toString());
+  });
+
   it('should support object values syntax', () => {
-    const values = { base: '20px', medium: '40px' };
+    const { result } = renderHook(() => useTheme());
+    const { tokens } = result.current;
+    const values = {
+      base: 'red',
+      medium: tokens.colors.red[10],
+      large: 'red.60',
+    };
 
     const baseResult = getValueAtCurrentBreakpoint(values, 'base', breakpoints);
     const smallResult = getValueAtCurrentBreakpoint(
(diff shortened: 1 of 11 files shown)


packages/react/src/primitives/types/theme.ts
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * Though both 'white' and 'black' are valid CSS values
+ * We still want to define them as keys here and convert them into CSS variable reference afterwards
+ * This will make the default dark mode override work, where we flip these two
+ */
+type WhiteKey = 'white';
+
+type BlackKey = 'black';
+
+type RedKeys =
```
zchenwei, line 12: These are for type intellisense. No worries, they are no more than syntactic sugar in Typescript and will not increase complied javascript size.

packages/react/src/primitives/shared/utils.ts
```diff
@@ -193,3 +196,27 @@ export const classNameModifierByFlag = (
...
+  }
+  const path = value.split('.');
+  let { tokens } = defaultTheme;
+  tokens = tokens[stylePropsToThemeKeys[propKey]];
+  for (let i = 0; i < path.length; i++) {
+    if (tokens) {
+      tokens = tokens[path[i]];
+      continue;
+    }
+    break;
```
joebuono, line 215: Are `continue` and `break` necessary here?

packages/react/src/primitives/shared/responsive/__tests__/utils.test.ts
```diff
@@ -15,8 +18,31 @@ describe('getValueAtCurrentBreakpoint', () => {
     const result = getValueAtCurrentBreakpoint('20px', 'base', breakpoints);
     expect(result).toBe('20px');
   });
+
+  it('should return number values directly', () => {
+    const result = getValueAtCurrentBreakpoint(0, 'base', breakpoints);
+    expect(result).toBe(0);
+  });
+
+  it('should return design token toString vaule directly', () => {
```
joebuono, line 27: nit typo: "value"

packages/react/src/primitives/shared/utils.ts
```diff
@@ -193,3 +196,27 @@ export const classNameModifierByFlag = (
 ): string => {
   return flag ? `${base}--${modifier}` : null;
 };
+
+export const getCSSVariableIfValueIsThemeKey = (
+  propKey: string,
+  value: string | number
```
joebuono, line 202: Is it possible for value to be `null`, `undefined` or `boolean`?

packages/react/src/primitives/shared/utils.ts
```diff
@@ -193,3 +196,27 @@ export const classNameModifierByFlag = (
...
+) => {
+  if (typeof value === 'number') {
+    return value;
+  }
+  const path = value.split('.');
+  let { tokens } = defaultTheme;
+  tokens = tokens[stylePropsToThemeKeys[propKey]];
```
joebuono, line 209: I'm having trouble understanding what this line is doing. Can you explain?

packages/react/src/primitives/shared/utils.ts
```diff
@@ -193,3 +196,27 @@ export const classNameModifierByFlag = (
...
+  for (let i = 0; i < path.length; i++) {
+    if (tokens) {
+      tokens = tokens[path[i]];
```
joebuono, line 212: I'm having trouble understanding what exactly this line is doing. Would you mind explaining?

packages/react/src/primitives/shared/responsive/__tests__/utils.test.ts
zchenwei: oops, good catch

packages/react/src/primitives/shared/utils.ts
zchenwei: Nil value has been filter out - https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/shared/styleUtils.ts#L130

zchenwei: For `boolean`, I don't if people would supply a boolean to props because they will get syntax errors. But we can filter that out as well

packages/react/src/primitives/shared/utils.ts
zchenwei: You can run a test case to check it. Like if I have `pink.10`, it will be split into `['pink', '10']` and I want to loop through the array until the end, so `continue` is necessary cause otherwise it will break in advance. For invalid key like `invalid.xxx.xxxx`, definitely get nothing in `tokens` and so I and break the loop ahead.

packages/react/src/primitives/shared/utils.ts
zchenwei: This will go deep into the theme `tokens` and get its `value` and also works as a check to see if the key is valid because users can provide any string key. If a key is invalid, then you will get nothing but `undefined` from the `tokens` we defined

dbanksdesign (approved):
drool

PR #2057, opened 2022-06-09
fix(angular): Use partial ivy compilation mode

wlee221:
#### Description of changes
Sets `angularCompilerOptions` to use partial compilation mode, as suggested in https://angular.io/guide/creating-libraries#ensuring-library-version-compatibility.

This has been blocking publish to `@next`.

Confirmed NPM publish works with verdaccio:

Screen Shot 2022-06-08 at 5 58 42 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/angular/projects/ui-angular/tsconfig.lib.prod.json
+++ b/packages/angular/projects/ui-angular/tsconfig.lib.prod.json
@@ -5,6 +5,6 @@
     "declarationMap": false
   },
   "angularCompilerOptions": {
-    "enableIvy": true
+    "compilationMode": "partial"
   }
 }


calebpollman (approved):

wlee221:
Going to merge this early to do some backwards compatibility testing tonight. Reading more on https://angular.io/guide/creating-libraries#ensuring-library-version-compatibility:

> The Angular version used to build an application should always be the same or greater than the Angular versions used to build any of its dependent libraries. For example, if you had a library using Angular version 13, the application that depends on that library should use Angular version 13 or later. Angular does not support using an earlier version for the application.

It is very likely we'll find some backward compatibility problems, and I'll have to revert my Angular 13 changes. I'll give an update tomorrow.

PR #2060, opened 2022-06-09
chore(ui-react): lint hooks directory

calebpollman:
#### Description of changes
Update code in _@aws-amplify/ui-react/src/hooks_ to pass linting rules

#### Issue #, if available
NA

`yarn lint && yarn test`
#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/react/src/primitives/shared/responsive/utils.ts
```diff
@@ -3,10 +3,10 @@ import { isDesignToken } from '@aws-amplify/ui';
 import { Breakpoint, Breakpoints } from '../../types/responsive';
 
 export const getValueAtCurrentBreakpoint = (
-  values: Record<string, any> | string[] | string,
+  values: Record<string, string> | string[] | string,
```
calebpollman, line 6: Had to add return types to `getValueAtCurrentBreakpoint` and `getClosestValueByBreakpoint` in order to allow `useBreakpointValue` to pass linting. Additionally, updated `any` to `string` based on the usage throughout the code base

packages/react/src/hooks/useBreakpointValue.ts
```diff
@@ -6,10 +6,10 @@ import { useTheme } from './useTheme';
 /**
  * [📖 Docs](https://ui.docs.amplify.aws/react/theming/responsive#usebreakpointvalue)
  */
-export function useBreakpointValue<T>(
-  values: Record<string, T> | T[],
+export function useBreakpointValue(
```
calebpollman, line 9: Removed the generic here as it our usage of this hook always takes in an argument of `Record<string, string>` and returns `string` (or `null`)

packages/react/src/hooks/useAuth.ts
```diff
@@ -27,10 +27,10 @@ export const useAuth = (): UseAuthResult => {
     user: undefined,
   });
 
-  const handleAuth = ({ payload }) => {
+  const handleAuth: HubCallback = ({ payload }) => {
     switch (payload.event) {
       case 'signIn':
-        return setResult({ user: payload.data, isLoading: false });
+        return setResult({ user: payload.data as AuthUser, isLoading: false });
```
zchenwei, line 33: Do we still need the type assertion after typing the function? That's annoying

packages/react/src/hooks/useBreakpointValue.ts
zchenwei: I did some updates on the get function because I found that `number` is allowed for some props. But I am agree on removing generic. I will update my PR based on yours afterwards

packages/react/src/hooks/useAuth.ts
calebpollman: Yeah unfortunately, `data` is typed as `any` in `HubCapsule`

zchenwei (approved):
LGTM!

reesscot (approved):
LGTM

packages/react/src/primitives/shared/responsive/utils.ts
```diff
@@ -3,10 +3,10 @@ import { isDesignToken } from '@aws-amplify/ui';
...
   breakpoint: Breakpoint,
   breakpoints: Breakpoints
-) => {
+): string | null => {
```
reesscot, line 9: Will this mean I can't do this anymore?
```
const showSomething = useBreakpointValue({
    base: false,
    medium: true,
  });

//...

{showSomething ? <Something /> : null}
```

zchenwei: Yeah on second thoughts, the value should allow to be anything when using this hook.

Issue #2001, opened 2022-05-30
Failed to compile - Module not found: Error: Can't resolve 'mapbox-gl'


rekaburmeister:
My build is failing, complaining about `mapbox-gl` not found.
My package version is:

```
"@aws-amplify/ui-react": "2.16.2"
```

This seems very similar to [this issue](https://github.com/aws-amplify/amplify-ui/issues/1647) in that it's complaining about the same package.

The odd thing is that I can build perfectly fine locally (MacBook) but I can't on my build agent (CentOS) they both have the same node and yarn versions. 

Any suggestions on fixing this?

ErikCH:
Hi @rekaburmeister !

That is unusual that it works on mac but not on CentOS? And there is no difference between yarn and npm?

Can you try again on the CentOS and try a `yarn why @aws-amplify/ui-react` ?

rekaburmeister:
Hi @ErikCH, thanks for getting back to me, sorry about delay, we had a long bank holiday weekend.

Indeed, very odd and frustrating - seemingly makes no sense!!

This is the output on CentOS:

```
yarn why v1.22.19
[1/4] Why do we have the module "@aws-amplify/ui-react"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "@aws-amplify/ui-react@2.16.2"
info Has been hoisted to "@aws-amplify/ui-react"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "28.34MB"
info Disk size with unique dependencies: "72.31MB"
info Disk size with transitive dependencies: "89.41MB"
info Number of shared dependencies: 50
```

Also, for good measure:

```
node --version
v14.19.1
```

On Mac:

```
node --version
v14.19.1

yarn --version 
1.22.18
```

For some reason, I couldn't do `yarn install` on Mac:

```
[2/4] 🚚  Fetching packages...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/@types/carbon__icons-react/-/carbon__icons-react-10.31.2.tgz: ESOCKETTIMEDOUT".
```

- this worked fine on the CentOS machine so (🤷 ). So, I have this instead:

```
npm list --depth=0
[...]

Issue #1736, opened 2022-04-21
CRA 3.4.1 and 4.0.3 prod build fails on `mapbox-gl` transitive dependency


reesscot:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

CRA or 3.4.1 or 4.0.3

### Please describe your bug.

```
Creating an optimized production build...
Failed to compile.
./node_modules/react-map-gl/dist/esm/components/map.js
Cannot find module: 'mapbox-gl'. Make sure this package is installed.
You can install this package by running: npm install mapbox-gl.
```

### What's the expected behaviour?

Should build without issues

### Help us reproduce the bug!

Do a prod build with CRA 3 or 4 and `@aws-amplify/ui-react` package.

### Code Snippet

```typescript
// Put your code below this line.

```

ErikCH:
This seems to be a duplicate of https://github.com/aws-amplify/amplify-ui/issues/1656. We have a workaround, and we are waiting for an upstream dependency fix.

reesscot:
@ErikCH No, it's a different error message. See code above.

slaymance:
I'm not able to reproduce this on either of the CRA versions mentioned. Which version of npm was used to install dependencies?

slaymance:
I propose we reform this issue to capture the error of `Cannot find module: 'mapbox-gl'` since multiple users have reported this problem across, potentially, multiple build tools ([here](https://github.com/aws-amplify/amplify-ui/issues/1647#issuecomment-1092821750) and [here](https://github.com/aws-amplify/amplify-ui/issues/1647#issuecomment-1094161657)). What I'll ask of users that encounter this error is to provide the following:

1. Which build tool and version you're using (Create React App v3.4.1, Next JS v11.1.4, etc.)
2. Which package manager and version you're using (npm v8.7.0, Yarn v1.22.17, etc.)

cwoolum:
[...]

PR #2061, opened 2022-06-09
fix: pin mapbox-gl to v1.13.2

reesscot:
#### Description of changes
This PR updates our `mapbox-gl` dependency from using `empty-npm-package` to using actual module version 1.13.2. We had previously tried to point this dependency to an empty npm package because it's not used in our Geo components, but is required at build time by `react-map-gl`. Pointing to the empty package caused issues (https://github.com/aws-amplify/amplify-ui/issues/2001 and https://github.com/aws-amplify/amplify-ui/issues/1736) for a small number of customers, so we are going to point at the actual module to ensure the problem is fully solved. This module is unused, so it will be tree-shaken out of final customer bundles.

#### Issue #, if available
Fixes https://github.com/aws-amplify/amplify-ui/issues/2001
Fixes https://github.com/aws-amplify/amplify-ui/issues/1736

- [ ] Ran integration tests
- [ ] Manually tested in CRA app

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


slaymance:
Worth noting that we pin to v1.13 and not v2 as [v1.13 uses BSD license.](https://github.com/mapbox/mapbox-gl-js/blob/main/LICENSE.txt)

slaymance (approved):

wlee221 (approved):

PR #2067, opened 2022-06-10
fix(docs): fix type error

zchenwei:
#### Description of changes

The docs on dev endpoint is broken. The error is coming from `HomePage` component.

```jsx
  const showEditor = useBreakpointValue({
    base: false,
    medium: true,
  });

```

The value used in `useBreakpointValue` can be `any`

#### Issue #, if available

[](https://dev.ui.docs.amplify.aws/)

Tested it out locally

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/react/src/primitives/shared/utils.ts
+++ b/packages/react/src/primitives/shared/utils.ts
@@ -197,9 +197,9 @@ export const classNameModifierByFlag = (
   return flag ? `${base}--${modifier}` : null;
 };
 
-export const getCSSVariableIfValueIsThemeKey = (
+export const getCSSVariableIfValueIsThemeKey = <Value>(
   propKey: string,
-  value: any
+  value: Value
 ) => {
   if (typeof value !== 'string') {
     return value;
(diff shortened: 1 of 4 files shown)


0618 (approved):

packages/react/src/primitives/shared/utils.ts
```diff
@@ -199,9 +199,9 @@ export const classNameModifierByFlag = (
 
 export const getCSSVariableIfValueIsThemeKey = (
   propKey: string,
-  value: string | number
+  value: any
```
0618, line 202: `any`?

calebpollman, line 202: Should this really be `any`? Maybe we could use a union type of `string` and a generic?

zchenwei: Yeah, I think generic makes sense

calebpollman: Follow up question not necessarily related to this PR, but does this util need to live in `ui-react`? Looks to be platform agnostic to me

0618: how about using `unknown`?

zchenwei: Yea, I mean. We can move it to ui package later

calebpollman (approved):
LGTM

PR #2074, opened 2022-06-11
chore(ui-react): lint components/Authenticator directory

calebpollman:
#### Description of changes
Updates to allow `Authenticator` and related utils to pass linting rules

#### Issue #, if available
NA

`yarn react lint && yarn react test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/react/.eslintrc.js
+++ b/packages/react/.eslintrc.js
@@ -68,12 +68,10 @@ module.exports = {
     {
       files: [
         // TODO Uncomment the below lines on an individual basis as the updates to pass linting are completed.
-        // If needed, these can be broken down further, e.g. 'src/components/Geo/**/*'
-        'src/components/Geo/**/*',
-        'src/components/ThemeProvider/**/*',
+        // If needed, these can be broken down further, e.g. 'src/primitives/Alert/**/*'
+        'src/components/**/*',
         'src/helpers/**/*',
         'src/hooks/**/*',
-        // 'src/components/**/*',
         // 'src/primitives/**/*',
       ],
       extends: [
--- a/packages/react/src/components/Authenticator/Authenticator.tsx
+++ b/packages/react/src/components/Authenticator/Authenticator.tsx
@@ -53,8 +53,11 @@ export function Authenticator({
   services,
   socialProviders,
   variation,
-}: AuthenticatorProps) {
-  const components = { ...defaultComponents, ...customComponents };
+}: AuthenticatorProps): JSX.Element {
+  const value = React.useMemo(
+    () => ({ components: { ...defaultComponents, ...customComponents } }),
+    [customComponents]
+  );
   const machineProps = {
     initialState,
     loginMechanisms,
@@ -65,7 +68,7 @@ export function Authenticator({
   };
   return (
     <Provider>
-      <CustomComponentsContext.Provider value={{ components }}>
+      <CustomComponentsContext.Provider value={value}>
         <InitMachine {...machineProps}>
           <Router
             className={className}
--- a/packages/react/src/components/Authenticator/ConfirmSignUp/ConfirmSignUp.tsx
+++ b/packages/react/src/components/Authenticator/ConfirmSignUp/ConfirmSignUp.tsx
@@ -1,3 +1,4 @@
+import React from 'react';
 import { translate } from '@aws-amplify/ui';
 
 import { Button } from '../../../primitives/Button';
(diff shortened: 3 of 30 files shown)


packages/react/src/components/Authenticator/ConfirmSignUp/ConfirmSignUp.tsx
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
reesscot, line 1: Can we turn off the rule that's requiring this?

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignIn.tsx
```diff
@@ -61,6 +62,7 @@ export function FederatedSignIn() {
               />
             );
           default:
+            // eslint-disable-next-line no-console
             console.error(
               `Authenicator does not support ${provider}. Please open an issue: https://github.com/aws-amplify/amplify-ui/issues/choose`
```
reesscot, line 67: While you're here, mind fixing the spelling?
```suggestion
              `Authenticator does not support ${provider}. Please open an issue: https://github.com/aws-amplify/amplify-ui/issues/choose`
```

packages/react/src/components/Authenticator/ResetPassword/ConfirmResetPassword.tsx
```diff
@@ -55,10 +56,12 @@ export const ConfirmResetPassword = ({
   );
 };
 
-ConfirmResetPassword.Header = () => {
+ConfirmResetPassword.Header = function Header(): JSX.Element {
```
reesscot, line 59: 💯

packages/react/src/components/Authenticator/Router/Router.tsx
```diff
@@ -11,13 +11,19 @@ import { ConfirmResetPassword, ResetPassword } from '../ResetPassword';
 import { isSignInOrSignUpRoute } from '../utils';
 import { RouterProps } from './types';
 
-const getRouteComponent = (route: string) => {
+type RouteComponent = (props: Omit<RouterProps, 'children'>) => JSX.Element;
+
+function RenderNothing(): JSX.Element {
```
reesscot, line 16: Best function name ever 🤣

packages/react/src/components/Authenticator/ConfirmSignUp/ConfirmSignUp.tsx
calebpollman: Yes but would like to handle it in a follow up PR since it will be applied to the rest of the _components_ directories that have already been updated to pass linting if that's cool?

reesscot (approved):

wlee221 (approved):
Thanks @calebpollman!

packages/react/.eslintrc.js
```diff
@@ -65,12 +65,10 @@ module.exports = {
     {
       files: [
         // TODO Uncomment the below lines on an individual basis as the updates to pass linting are completed.
-        // If needed, these can be broken down further, e.g. 'src/components/Geo/**/*'
-        'src/components/Geo/**/*',
-        'src/components/ThemeProvider/**/*',
+        // If needed, these can be broken down further, e.g. 'src/primitives/Alert/**/*'
+        'src/components/**/*',
```
ayush987goyal, line 69: This will be breaking change in the staging repo. Can we keep it scoped to per component?

ayush987goyal (changes requested):

zchenwei (approved):
LGTM

ayush987goyal (approved):
Thanks for making the change!

packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -53,8 +53,11 @@ export function Authenticator({
   services,
   socialProviders,
   variation,
-}: AuthenticatorProps) {
-  const components = { ...defaultComponents, ...customComponents };
+}: AuthenticatorProps): JSX.Element {
+  const value = React.useMemo(
```
ErikCH, line 57: Why do you think this should be in a useMemo?

calebpollman: The `react/jsx-no-constructed-context-values` rule caught this. Longer explanation is that `components` was getting created with a new reference on each render, placing `components` inside the `useMemo`  wrapped `value` ensures  a static reference between renders

PR #2081, opened 2022-06-11
fix: add isNested prop to ThemeProvider

zchenwei:
#### Description of changes

Add `isNested` prop to `ThemeProvider` so that only the one at top will be able to set attributes on root element versus relying on the cleanup function in `useEffect`, which will not work as expected if there are 2 or more nested providers on the same page.

#### Issue #, if available

A good example I can see is on the [Authenticator page](https://dev.ui.docs.amplify.aws/react/components/authenticator) where 2 ThemeProvider are in use. If you go to that page and then move over to other pages like [Installation](https://dev.ui.docs.amplify.aws/react/getting-started/installation), you will see `data-amplify-theme` attribute on root element is set to a wrong value `Auth Example Theme` instead of the expected original value `amplify-docs`.

Tested it out locally

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/ThemeProvider/index.tsx
+++ b/packages/react/src/components/ThemeProvider/index.tsx
@@ -11,6 +11,11 @@ interface ThemeProviderProps {
   colorMode?: ColorMode;
   theme?: Theme;
   nonce?: string;
+  /**
+   * Set it to true if this is a nested ThemeProvider
+   * @default false
+   */
+  isNested?: boolean;
 }
 
 export function AmplifyProvider({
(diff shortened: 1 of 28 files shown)


packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -29,31 +37,15 @@ export function AmplifyProvider({
   // the same data attributes are on a div down the DOM tree, the CSS variables
   // will apply to both.
   React.useEffect(() => {
-    if (document && document.documentElement) {
-      // Keep original data attributes to reset on unmount
```
joebuono, line 33: Do we no longer need to keep original data attributes to reset on unmount?

joebuono (approved):
Just one question. Thanks for making this change!

zchenwei:
> Let's pause on this, @calebpollman has some concerns

What is the concern?

zchenwei:
Close it in favor of #2102

PR #2102, opened 2022-06-14
fix: nested theme providers by removing menu's portal

dbanksdesign:
#### Description of changes

* Menu primitive uses `portalled={false}` now so it no longer renders in a React portal outside the DOM node of the ThemeProvider, meaning it now properly accepts the theme of the nearest ThemeProvider. 
* Removing all the junk in the ThemeProvider to get around that Portal issue. 

Potential alternative: we could make our own Portal (using the ThemeProvider's DOM node) potentially and render the menu content in that. Seems a bit more difficult. 

* Navigated around the docs site, things work fine. Examples with ThemeProvider work.

CleanShot 2022-06-14 at 11 10 15

CleanShot 2022-06-14 at 11 10 42

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
+++ b/packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
@@ -21,21 +21,15 @@ describe('ThemeProvider', () => {
   });
 
   it('wraps the App in [data-amplify-theme="default-theme"]', () => {
-    jest.spyOn(global.document.documentElement, 'setAttribute');
     const { container } = render(
       <ThemeProvider>
         <App />
       </ThemeProvider>
     );
 
-    expect(global.document.documentElement.setAttribute).toHaveBeenCalledWith(
+    expect(container.querySelector(`[data-amplify-theme]`)).toHaveAttribute(
       'data-amplify-theme',
       'default-theme'
-    );
-
-    expect(global.document.documentElement.setAttribute).toHaveBeenCalledWith(
-      'data-amplify-color-mode',
-      'undefined'
     );
   });
 
--- a/packages/react/src/components/ThemeProvider/index.tsx
+++ b/packages/react/src/components/ThemeProvider/index.tsx
@@ -24,38 +24,6 @@ export function AmplifyProvider({
     theme: { name, cssText },
   } = value;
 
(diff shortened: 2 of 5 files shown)


reesscot (approved):
This is great!

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -24,38 +24,6 @@ export function AmplifyProvider({
...
-          'data-amplify-theme',
-          originalName
-        );
-        document.documentElement.setAttribute(
-          'data-amplify-color-mode',
-          originalColorMode
-        );
-      };
-    }
-  }, [name, colorMode]);
```
calebpollman, line 58: ✨

calebpollman (approved):
🚢

joebuono (approved):

packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
```diff
@@ -21,22 +21,16 @@ describe('ThemeProvider', () => {
...
   it('wraps the App in [data-amplify-theme="default-theme"]', () => {
-    jest.spyOn(global.document.documentElement, 'setAttribute');
     const { container } = render(
       <ThemeProvider>
         <App />
       </ThemeProvider>
     );
 
-    expect(global.document.documentElement.setAttribute).toHaveBeenCalledWith(
+    expect(container.querySelector(`[data-amplify-theme]`)).toHaveAttribute(
```
zchenwei, line 30:
```suggestion
    expect(global.document.documentElement).toHaveAttribute(
```
No need to query? The root element is kinda special

packages/react/src/components/ThemeProvider/index.tsx
zchenwei: 🤩

jacoblogan (approved):
Awesome this makes things so much simpler!!!

zchenwei (approved):
This is looking good. I am going to close #2081. Regarding the potential solution, radix has a Portal util that can help us achieve it easily with `containerRef` to customize the container - https://www.radix-ui.com/docs/primitives/utilities/portal#root

PR #2121, opened 2022-06-16
ci(canary): new `install-with-retries` action

wlee221:
#### Description of changes

This PR adds new `install-with-retries` action that will try `yarn install` up to 3 times. This can be reused in other PRs, but I scoped it to just canary workflow because that's where `yarn install` failed the most. 

#### Bug Fix
This fixed a bug where `yarn install` failed silently after it failed all three retries. See "Install Cypress" step in this [workflow](https://github.com/aws-amplify/amplify-ui/runs/6914109019?check_suite_focus=true) for example.

#### Optimization
`install-with-retries` optionally accepts `skip-cypress-binary` argument. If this is set to true, it'll skip cypress binary installation step, which has [been really flimsy](https://github.com/aws-amplify/amplify-ui/runs/6914109019?check_suite_focus=true) recently.

Instead of installing cypress binary every time, canaries will cache cypress binary and share it across jobs. It'll only install the binary on cache misses.

POC'd on #2122

- First run (cache miss): https://github.com/aws-amplify/amplify-ui/runs/6914131760?check_suite_focus=true
- Second run (cache hit): https://github.com/aws-amplify/amplify-ui/runs/6914184443?check_suite_focus=true

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] New workflow files do not accept any user inputs

### DIFF
--- a/.github/actions/install-with-retries/action.yml
+++ b/.github/actions/install-with-retries/action.yml
@@ -17,7 +17,6 @@ runs:
   using: 'composite'
   steps:
     - name: Run yarn install
-      if: steps.restore-cypress-cache.outputs.cache-hit != 'true'
       shell: bash
       run: ./install-with-retries.sh
       env:
--- a/.github/actions/install-with-retries/action.yml
+++ b/.github/actions/install-with-retries/action.yml
@@ -22,4 +22,4 @@ runs:
       env:
         SKIP_CYPRESS_BINARY: ${{ inputs.skip-cypress-binary }}
         NO_LOCKFILE: ${{ inputs.no-lockfile }}
-      working-directory: ${{ inputs.working-directory }}
+        WORKING_DIRECTORY: ${{ inputs.working-directory }}
--- a/.github/actions/install-with-retries/action.yml
+++ b/.github/actions/install-with-retries/action.yml
@@ -9,6 +9,10 @@ inputs:
     description: 'Path to run `yarn install` at'
     required: false
     default: './'
+  no-lockfile:
+    description: 'Whether to bypass yarn.lock file'
+    required: false
+    default: 'false'
 runs:
   using: 'composite'
   steps:
@@ -18,4 +22,5 @@ runs:
       run: ./install-with-retries.sh
       env:
         SKIP_CYPRESS_BINARY: ${{ inputs.skip-cypress-binary }}
+        NO_LOCKFILE: ${{ inputs.no-lockfile }}
       working-directory: ${{ inputs.working-directory }}
--- a/.github/actions/install-with-retries/action.yml
+++ b/.github/actions/install-with-retries/action.yml
@@ -12,12 +12,10 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: Set environment var to skip cypress binary installation if applicable
-      if: ${{ inputs.working-directory == 'true' }}
-      shell: bash
-      run: echo "CYPRESS_INSTALL_BINARY=0" >> $GITHUB_ENV
     - name: Install cypress with binary
       if: steps.restore-cypress-cache.outputs.cache-hit != 'true'
       shell: bash
-      run: for n in {1..3}; do yarn install && break || sleep 5; done
+      run: ./install-with-retries.sh
+      env:
+        SKIP_CYPRESS_BINARY: ${{ inputs.skip-cypress-binary }}
       working-directory: ${{ inputs.working-directory }}
--- a/.github/actions/install-with-retries/action.yml
+++ b/.github/actions/install-with-retries/action.yml
@@ -18,7 +18,7 @@ runs:
   steps:
     - name: Run yarn install
       shell: bash
-      run: ./install-with-retries.sh
+      run: ./.github/actions/install-with-retries/install-with-retries.sh
       env:
         SKIP_CYPRESS_BINARY: ${{ inputs.skip-cypress-binary }}
         NO_LOCKFILE: ${{ inputs.no-lockfile }}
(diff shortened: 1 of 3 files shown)


.github/actions/install-with-retries/action.yml
```diff
@@ -0,0 +1,25 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: Run yarn install
+      shell: bash
+      run: ./.github/actions/install-with-retries/install-with-retries.sh
+      env:
+        SKIP_CYPRESS_BINARY: ${{ inputs.skip-cypress-binary }}
+        NO_LOCKFILE: ${{ inputs.no-lockfile }}
+        WORKING_DIRECTORY: ${{ inputs.working-directory }}
```
wlee221, line 25: Security note:

Instead of using those inputs in the script, we pass them as environment variables. This prevents possible script injection by keeping the script static!

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

.github/workflows/run-and-test-builds.yml
```diff
@@ -56,9 +58,23 @@ jobs:
       - name: Build example app
         run: yarn build
         working-directory: ./canary/apps/${{ matrix.path }}
-      - name: Install cypress
-        run: for n in {1..3}; do yarn install --no-lockfile && break || sleep 5; done
-        working-directory: ./canary/e2e
```
wlee221, line 61: Security note:

Using an action is more secure then using inline scripts. This is because any inputs gets passed to the action as argument, and will not be used to generate scripts.

wlee221: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-action-instead-of-an-inline-script-recommended

.github/actions/install-with-retries/install-with-retries.sh
```diff
@@ -0,0 +1,42 @@
...
+  return_value=$?
+  [ $return_value -eq 0 ] && break
+  echo "[ERROR]: yarn install failed with exit code $return_value, waiting to retry..."
+
+  # Sleep 5 seconds before retrying
+  sleep 5
+done
+
+# exit 0 if last `yarn install` was successful, non-zero otherwise
+exit $return_value
```
wlee221, line 42: This fixes silent yarn install failures.

cc @reesscot

.github/actions/install-with-retries/install-with-retries.sh
```diff
@@ -0,0 +1,42 @@
...
+if [ -n "$WORKING_DIRECTORY" ]; then
+  echo "[LOG]: Changing directory to $WORKING_DIRECTORY"
+  cd $WORKING_DIRECTORY
+fi
+
+if [ "$SKIP_CYPRESS_BINARY" = "true" ]; then
+  # Skip cypress binary installation if specified
+  # https://docs.cypress.io/guides/references/advanced-installation#Environment-variables
+  echo "[LOG]: Setting flag to skip cypress binary installation"
+  export CYPRESS_INSTALL_BINARY=0
```
zchenwei, line 17: ⭐

.github/workflows/run-and-test-builds.yml
```diff
@@ -56,9 +58,23 @@ jobs:
...
+        id: restore-cypress-cache
+        with:
+          path: ~/.cache/Cypress
+          key: ${{ runner.os }}-canary-cypress-${{ hashFiles('canary/e2e/yarn.lock') }}
+      - name: Install e2e packages
+        uses: ./.github/actions/install-with-retries
+        with:
+          working-directory: ./canary/e2e
+          skip-cypress-binary: ${{ steps.restore-cypress-cache.outputs.cache-hit }}
+      - name: Cache cypress runner
```
zchenwei, line 72: Wouldn't the above restore step create a new cache if miss or the old one isn't hit

wlee221: Yes, it's required (and I realize this is not the most readable sequence of steps). Restore cypress cache step will **not** create a new cache because cypress runner hasn't been downloaded yet and is unavailable. So the sequence is like this:

1. Workflow will try to restore cypress cache but fails.
2. Since it failed, `yarn install` will run **with** cypress binary installation.
3. Now that cypress binary is available, github workflow will create a new cache entry.

wlee221: Added [fdbe871](https://github.com/aws-amplify/amplify-ui/pull/2121/commits/fdbe871993ba88076c9fbf39e22ab0dde472ca64) to clarify this

zchenwei (approved):
LGTM

.github/actions/install-with-retries/action.yml
```diff
@@ -0,0 +1,25 @@
+name: 'Install with retries'
+description: 'Runs yarn install with up to 2 retries'
+inputs:
+  skip-cypress-binary:
+    description: 'Whether to skip binary installation for cypress'
+    required: false
+    default: 'true'
```
ErikCH, line 7: I find it weird that default can only be a string here. But I guess that's the way it is.

ErikCH (approved):
LGTM!

.github/actions/install-with-retries/action.yml
wlee221: You're correct, inputs are treated as strings and not boolean.

https://github.com/actions/runner/issues/1483

wlee221: Also: https://github.community/t/composite-actions-input-types-boolean-numeric/219254

PR #2135, opened 2022-06-17
Add JS/TS tabs for dark mode examples

wlee221:
#### Description of changes

Adds JS/TS tabs for dark mode examples. Previously, that was using `useState<ColorMode>('system')`, which is a TS only contruct and would throw TS errors.

We previously discussed we should just remove that TS types and just write `useState('system')`, but that actually breaks TS users with `Type 'string' is not assignable to type 'ColorMode'`.

So instead I decided to use tabs to implement this.

Screen Shot 2022-06-17 at 4 36 08 PM

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/src/pages/[platform]/theming/dark-mode/examples/CustomDarkMode.jsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
```
Comment (wlee221) [inline docs/src/pages/[platform]/theming/dark-mode/examples/CustomDarkMode.jsx:1]: Here's perhaps the weirdest part of this PR -- I created `*.jsx` files based off the existing `*.tsx` examples. They do not get used however and only exists for the code snippet.

This was because TypeScript will complain if you try to use `React.useState('system')` with 

```
Type 'string' is not assignable to type 'ColorMode'
```

Comment (wlee221) [reply docs/src/pages/[platform]/theming/dark-mode/examples/CustomDarkMode.jsx:1]: Any suggestions are appreciated 🙏

Comment (wlee221) [reply docs/src/pages/[platform]/theming/dark-mode/examples/CustomDarkMode.jsx:1]: Talked with @calebpollman, we'll just go ahead and hard code the JS snippets instead of creating `.jsx` files. We can decide on a more scalable approach as we make more JS/TS tabs, but for now let's keep it simple for now.

docs/src/pages/[platform]/theming/dark-mode/react.mdx
```diff
@@ -66,15 +115,87 @@ _Note: to see dark mode applied, change your OS preferences_
...
 
-<Example>
-  <CustomDarkModeExample />
+<Tabs>
+  <TabItem title="JavaScript">
+    <Example>
+      <CustomDarkModeExample />
+      <ExampleCode>
+        ```jsx
+        import * as React from 'react';
```
Comment (wlee221) [inline docs/src/pages/[platform]/theming/dark-mode/react.mdx:124]:
```suggestion
import * as React from 'react';
```

docs/src/pages/[platform]/theming/dark-mode/react.mdx
```diff
@@ -38,15 +38,64 @@ Amplify UI comes with a default dark mode that you can use.
...
+      <DefaultDarkMode />
 
-  <ExampleCode>
-    ```tsx file=./examples/DefaultDarkMode.tsx
-    ```
-    
-  </ExampleCode>
-</Example>
+      <ExampleCode>
+        ```jsx file=./examples/DefaultDarkMode.jsx
```
Comment (wlee221) [inline docs/src/pages/[platform]/theming/dark-mode/react.mdx:47]:
```suggestion
        ```jsx
```

docs/src/pages/[platform]/theming/dark-mode/react.mdx
```diff
@@ -38,15 +38,64 @@ Amplify UI comes with a default dark mode that you can use.
...
+        
+      </ExampleCode>
+    </Example>
+  </TabItem>
+  <TabItem title="TypeScript">
+    <Example>
+      <DefaultDarkMode />
+
+      <ExampleCode>
+        ```tsx file=./examples/DefaultDarkMode.tsx
```
Comment (calebpollman) [inline docs/src/pages/[platform]/theming/dark-mode/react.mdx:92]: Wondering if we should move the example code in _DefaultDarkMode.tsx_ into this file as well to make it things more clear down the road. Just thinking that a future reader may be confused why there is a separate file for the `tsx` example while the `jsx` sample is included in this file. What do you think?

Comment (wlee221) [reply docs/src/pages/[platform]/theming/dark-mode/react.mdx:92]: I agree with example codes being confusing.

We are however actually rendering the JSX component from `DefaultDarkMode.tsx` -- that is the `<DefaultDarkMode />` right above this line, so I don't think we want to hard code the `tsx` here as well.

Comment (wlee221) [reply docs/src/pages/[platform]/theming/dark-mode/react.mdx:92]: This is why `.jsx` file made more sense in terms of consistency; but I was worried about it confusing the `next` bundler.

Comment (wlee221) [reply docs/src/pages/[platform]/theming/dark-mode/react.mdx:92]: Maybe we could postfix the extension with `.jsx.snippet`, just to signal that this file does not intend to be get rendered?

Comment (wlee221) [reply docs/src/pages/[platform]/theming/dark-mode/react.mdx:92]: cc @dbanksdesign, do you have a suggestion?

Comment (dbanksdesign) [reply docs/src/pages/[platform]/theming/dark-mode/react.mdx:92]: I definitely want to keep the pattern (at least for React right now) of rendering the component right next to the code that gets shown, like this:

```
      <DefaultDarkMode />

      <ExampleCode>
        ```tsx file=./examples/DefaultDarkMode.tsx
```

To ensure the code we are showing is valid and accurate. Obviously that is not practical for all cases like some of our CSS ones and this one where it's a bit ridiculous to have JSX and TSX files for on minor difference. I agree with the comment above that for now rendering the TSX file and hardcoding the JSX in the MDX file is the most practical until we come up with a better long-term approach for JS/TS examples.

dbanksdesign (approved):
LGTM!

reesscot (approved):

PR #2137, opened 2022-06-18
docs(chore): Add copy button to installation fragment

joebuono:
#### Description of changes

This PR creates a reusable `<InstallScripts />` component, which reduces some redundant code across the Getting Started section. It also includes a "Copy" button for improved UX. 

Current:
Screen Shot 2022-06-17 at 9 00 01 PM

Updated:
Screen Shot 2022-06-18 at 12 20 06 PM

Ran the docs locally.

#### Checklist

- [X] PR description included

### DIFF
--- a/docs/src/styles/docs/home.scss
+++ b/docs/src/styles/docs/home.scss
@@ -33,6 +33,7 @@
 .install-code__container {
   flex-grow: 1;
   display: flex;
+  max-width: 42rem;
 }
 
 .install-code__content {
@@ -84,7 +85,8 @@
 }
 
 .docs-home-editor {
-  border-right: var(--amplify-border-widths-small) solid var(--amplify-colors-border-primary);
+  border-right: var(--amplify-border-widths-small) solid
+    var(--amplify-colors-border-primary);
   max-height: 30rem;
   overflow-y: auto;
   font-size: var(--amplify-font-sizes-medium);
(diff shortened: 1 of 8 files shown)


docs/src/styles/docs/home.scss
```diff
@@ -33,6 +33,7 @@
 .install-code__container {
   flex-grow: 1;
   display: flex;
+  max-width: 42rem;
 }
 
 .install-code__content {
```
hbuchel, line 39: For some reason github won't let me select all the lines to make a suggestion here, but can we do something like this to keep it left aligned (and keep the $ to make it look like a shell, which isn't a huge deal without either):

```
.install-code__content {
  margin: var(--amplify-space-small);
  overflow-x: auto;
  white-space: normal;
  &:before {
    content: '$ ';
  }
}

.install-code__button {
  margin-left: auto;
  // Hiding on smaller screens
  // because there's not enough space
  @media (max-width: 480px) {
    display: none;
  }
}
```
Screen Shot 2022-06-18 at 11 01 35 AM

joebuono: I like it! Adding in that change now

hbuchel (approved):
LGTM! Thanks for the updates.

ErikCH (approved):
LGTM

PR #2182, opened 2022-06-23
fix(docs): fix variation anchor link in Table docs

hbuchel:
#### Description of changes

Fix the #variation link on https://ui.docs.amplify.aws/react/components/table#variation  so it scrolls properly. This updates the id on the select dropdown which had the same #variation id as the heading.

#### Issue #, if available

N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


docs/src/pages/[platform]/components/table/TablePropControls.tsx
```diff
@@ -59,7 +59,7 @@ export const TablePropControls: TablePropControlsInterface = ({
       <option value="large">large</option>
     </SelectField>
     <SelectField
-      id="variation"
+      id="variationProp"
```
Comment (zchenwei) [inline docs/src/pages/[platform]/components/table/TablePropControls.tsx:62]: I guess we need to update id for `size` as well?

Comment (hbuchel) [reply docs/src/pages/[platform]/components/table/TablePropControls.tsx:62]: Oops good catch! Updated.

dbanksdesign (approved):

PR #2183, opened 2022-06-23
fix(react): hub listener fix

wlee221:
#### Description of changes

This PR ensures that Authenticator correctly attaches and reattaches hub listener on React 18 [StrictMode](https://reactjs.org/docs/strict-mode.html).

#### Problem

In React 18 StrictMode, React will always [simulate unmounting and remounting](https://reactjs.org/docs/strict-mode.html#ensuring-reusable-state) whenever components initializes. This persist states and refs between the remounts:

> On the second mount, React will restore the state from the first mount. This feature simulates user behavior such as a user tabbing away from a screen and back [...].

This broke the hub listener pattern we used in React, because we've been expecting `ref` value to be reset between remounts. 

#### Fix

This PR fixes by this by not using `useRef` at all. Even if component were to remount, it'll reattach immediately.
 

Verified manually on `@experimental` tag. Added some unit test as well.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/Authenticator/hooks/useAuthenticator/__tests__/Provider.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+import * as React from 'react';
+import { Hub } from 'aws-amplify';
+
+import { Provider } from '../';
+import * as UI from '@aws-amplify/ui';
+
+jest.mock('@aws-amplify/ui', () => ({
+  ...(jest.requireActual('@aws-amplify/ui') as {}),
+  defaultAuthHubHandler: jest.fn(),
+}));
```
wlee221, line 11: Without this, 

```ts
const hubHandlerSpy = jest.spyOn(UI, 'defaultAuthHubHandler');
```
fails with 
```
TypeError: Cannot redefine property: defaultAuthHubHandler
```

packages/react/src/components/Authenticator/hooks/useAuthenticator/__tests__/Provider.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <Provider>
+        <></>
+      </Provider>
+    );
+
+    Hub.dispatch('auth', {
+      event: 'tokenRefresh',
+    });
+
+    expect(hubHandlerSpy).toHaveBeenCalledTimes(1);
```
wlee221, line 38: This has been failing due to 

```
    Expected number of calls: 1
    Received number of calls: 0
```

but I could verify the function does get called. Looks like jest is not correctly mocking it?

packages/react/src/components/Authenticator/hooks/useAuthenticator/__tests__/Provider.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+jest.mock('@aws-amplify/ui', () => ({
+  ...(jest.requireActual('@aws-amplify/ui') as {}),
+  defaultAuthHubHandler: jest.fn(),
+}));
+
+const hubListenSpy = jest.spyOn(Hub, 'listen');
+const hubHandlerSpy = jest.spyOn(UI, 'defaultAuthHubHandler');
+
+describe('Authenticator.Provider', () => {
+  it('listens to Auth Hub events on init', () => {
```
wlee221, line 17: This test runs well as expected, but the next one doesn't

packages/ui/src/helpers/authenticator/utils.ts
```diff
@@ -68,9 +68,13 @@ export const listenToAuthHub = (
...
 ) => {
-  return Hub.listen('auth', (data) => {
-    handler(data, service);
-  });
+  return Hub.listen(
+    'auth',
+    (data) => {
+      handler(data, service);
+    },
+    'authenticator-hub-handler'
```
wlee221, line 76: This is the hub handler name; added mostly for debugging purposes.

calebpollman (approved):
LGTM 🚢

Issue #2027, opened 2022-06-02
Cleanup Vue Code


ErikCH:
### On which framework/platform would you like to see this feature implemented?

Vue

### Which UI component is this feature-request for?

Authenticator

### Please describe your feature-request in detail.

Vue's code needs a bit of cleanup including

* Add in JSDoc for Vue Components similar to #2024 
* Fix issue with Code Editor error `Default export of the module has or is using private name ` in any `script setup` file that has an interface
* Add aria-described by to be in parity with #1945 (this might be added to a separate issue)
*  Add missing types (i.e. `useAuth`) (this might be added to a separate issue)
* Refactor Vue to internally use `useAuthenticator` instead of the deprecated `send`

### Please describe a solution you'd like.

Fix issues

### We love contributors! Is this something you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

ErikCH:
Closing this issue

PR #2184, opened 2022-06-23
chore: Cleaned up Vue code

ErikCH:
#### Description of changes

- [x] Moved Jest to latest
- [x] Added in JSX - For future tests or primitives
- [x] Fixed issue with Code Editor error Default export of the module has or is using private name in any script setup file that has an interface

#### Issue #, if available
#2027 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/vue/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "strict": true,
     "jsx": "preserve",
     "importHelpers": true,
-    "declaration": true,
+    "declaration": false,
```
calebpollman, line 8: Will this prevent type generation?

ErikCH: Our typescript d.ts generation still occurs with our vite `typeScript2` plugin settings.

```
// vite.config.js
    tsconfigOverride: {
        compilerOptions: {
          sourceMap: true,
          declaration: true,
          declarationMap: true,
        },
```

That way we still have typescript definition files being created.

package.json
```diff
@@ -53,15 +53,13 @@
     "eventsource": "~2.0.2",
     "fs-extra": "^10.0.0",
     "glob-parent": "^6.0.2",
-    "jest": "^26.6.3",
```
ErikCH, line 56: We used to need to pin jest, because of an issue with different jest versions breaking different tests being run on our frameworks. However, our tests no longer need this. 

And I won't be able to use the latest version of the vue3-jest library unless we update to jest 27.

packages/vue/tsconfig.json
calebpollman: Nice! Can we leave a comment here for future readers who may not have context in to `vite` type generation?

ErikCH: Sure! I'll add a note

packages/vue/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "strict": true,
     "jsx": "preserve",
     "importHelpers": true,
-    "declaration": true,
+    "declaration": false, // this is overriden in the vite.config.ts typescript2 settings
```
slaymance, line 8: Having a comment in JSON won't break anything, right?

slaymance (approved):

reesscot (approved):
LGTM

slaymance (approved):
LGTM!

Issue #2191, opened 2022-06-26
Authenticator (React) remounts the app after refreshing the access token


neil-a-thomson:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Webpack

### Please describe your bug.

The components provided to `withAuthenticator` are remounted when the access token is refreshed from the userpool.

### What's the expected behaviour?

I expect the app to be rerendered when a new token is fetched, I do not expect the whole app to remount.

### Code Snippet

```typescript
import React, { useCallback, useEffect } from 'react';
import { withAuthenticator } from '@aws-amplify/ui-react';
import { Auth } from 'aws-amplify';

function App({ signOut, user }) {
  useEffect(() => {
    // this gets called every time the token expires because the app remounts
    // I'm pretty sure this isn't the intended behaviour
    console.log('App mounting');
  }, []);

  const apiCall = useCallback(async () => {
    console.log('Making API call');
    const session = await Auth.currentSession(); // refreshes session if expired
    // don't need to do anything more, the line above is enough to cause a remount
  }, []);

  return (
    <div>
      <div>App</div>
      <div>{user.username}</div>
      <div>
        <button onClick={() => apiCall()}>Click me</button>
      </div>
    </div>
  );
}

export default withAuthenticator(App);
```

HudsonAkridge:
We're also running into this exact same issue. When performing a call to:
```JavaScript
let cognitoSession = await Auth.currentSession();
let idToken = cognitoSession.idToken.jwtToken;
```

in the background to capture our user's JWT's so we can issue API calls directly to our app, and the JWT needs to be refreshed, the entire app under the
```JavaScript
        <Authenticator
            loginMechanisms={["username"]}
            hideSignUp={true}
        >
            <Router>
                <Etc/>
            </Router>
        </Authenticator>
```
wrapper is forced to re-mount, losing state across the entire SPA.

[...]

PR #2228, opened 2022-06-30
fix(ui-react): directly return children from Authenticator

calebpollman:
#### Description of changes
In `Authenticator`, directly return user defined `children` argument when user is authenticated. Currently, `children` is rendered from within the `Router` component, which may be remounted due to `Authenticator` life cycle events.

#### Issue #, if available
Potentially fixes https://github.com/aws-amplify/amplify-ui/issues/2191

Manual testing, ran integ tests

#### Checklist
NA

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/react/src/components/Authenticator/Authenticator.tsx
+++ b/packages/react/src/components/Authenticator/Authenticator.tsx
@@ -51,7 +51,7 @@ function useInitMachine(data: AuthenticatorMachineOptions) {
 
 // `AuthenticatorInternal` exists to give access to the context returned via `useAuthenticator`,
 // which allows the `Authenticator` to just return `children` if a user is authenticated.
-// Once the `Provider` is removed from the `Authentcator` component and exported as
+// Once the `Provider` is removed from the `Authenticator` component and exported as
 // `AuthenticatorProvider`, this component should be renamed to `Authenticator`.
 export function AuthenticatorInternal({
   children,
--- a/packages/react/src/components/Authenticator/Authenticator.tsx
+++ b/packages/react/src/components/Authenticator/Authenticator.tsx
@@ -1,7 +1,14 @@
 import * as React from 'react';
-import { AuthenticatorMachineOptions } from '@aws-amplify/ui';
+import {
+  AuthenticatorMachineOptions,
+  CognitoUserAmplify,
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -1,7 +1,14 @@
...
+import {
+  AuthenticatorMachineOptions,
+  CognitoUserAmplify,
+} from '@aws-amplify/ui';
 
-import { Provider, useAuthenticator } from './hooks/useAuthenticator';
+import {
+  Provider,
+  useAuthenticator,
+  UseAuthenticator,
```
reesscot, line 10: Should this be a type import to be extra explicit? I think we should start distinguishing our imports because it would make them easier to read.
```
import type {  UseAuthenticator } from './hooks/useAuthenticator';
```

wlee221 (approved):
✨

packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -1,7 +1,14 @@
 import * as React from 'react';
-import { AuthenticatorMachineOptions } from '@aws-amplify/ui';
+import {
```
reesscot, line 2: `import type` here?

packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -34,14 +47,13 @@ function InitMachine({
...
-  return <>{children}</>;
 }
 
-/**
- * [📖 Docs](https://ui.docs.amplify.aws/react/connected-components/authenticator)
- */
-export function Authenticator({
+// `AuthenticatorInternal` exists to give access to the context returned via `useAuthenticator`,
+// which allows the `Authenticator` to just return `children` if a user is authenticated.
+// Once the `Provider` is removed from the `Authentcator` component and exported as
```
reesscot, line 54: nit: typo
```suggestion
// Once the `Provider` is removed from the `Authenticator` component and exported as
```

reesscot (dismissed):
LGTM. Had a few non-blocking nits

wlee221 (approved):

PR #2267, opened 2022-07-08
chore: update cra canary app to use npm install

reesscot:
#### Description of changes
This updates one of our canary apps to be installed with `npm install` in order to catch differences between `npm install` and `yarn install`. This is inspired by a problem where `npm install` went into an infinite loop on our package but it wasn't caught by canary apps because they only use `yarn install`.

**Tested `yarn react:install` triggers `npm install` for the CRA app:**
image

#### Checklist

- [x] PR description included

### DIFF
--- a/canary/package.json
+++ b/canary/package.json
@@ -18,7 +18,7 @@
     "angular": "yarn angularcli",
     "angular:install": "yarn angularcli:install",
     "cra": "yarn --cwd apps/react/cra build ",
-    "cra:install": "echo 'NOTE: Testing using npm install instead of yarn' && npm install apps/react/cra install",
+    "cra:install": "echo 'NOTE: Testing using npm install instead of yarn' && npm install --prefix apps/react/cra",
     "cra-ts": "yarn --cwd apps/react/cra-ts build ",
     "cra-ts:install": "yarn --cwd apps/react/cra-ts install",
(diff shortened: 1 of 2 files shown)


0618 (dismissed):

thaddmt (approved):
LGTM. Out of curiosity which version of npm is this using?

canary/package.json
```diff
@@ -18,7 +18,7 @@
     "angular": "yarn angularcli",
     "angular:install": "yarn angularcli:install",
     "cra": "yarn --cwd apps/react/cra build ",
-    "cra:install": "yarn --cwd apps/react/cra install",
+    "cra:install": "echo 'NOTE: Testing using npm install instead of yarn' && npm install apps/react/cra install",
```
ErikCH, line 21: This works - it looks like it installs on the root canary folder and the `apps/react/cra` folder.

You could also use `npm --prefix apps/react/cra install` which is more precise

reesscot: Good idea, I will update to use the --prefix

ErikCH (approved):
🚢

0618 (approved):

canary/package.json
reesscot: Updated

Issue #2235, opened 2022-06-30
"username undefined" error in authenticator when user auto-confirmed


clintfoster:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Vue

### Which UI component?

Authenticator

### How is your app built?

Vite

### What browsers are you seeing the problem on?

Chrome

### Please describe your bug.

We are using `authenticator` in our Vue 3 frontend, and we have a pre-signup trigger Lambda function defined in our Amplify backend to auto-confirm username, email and phone. Thanks to authenticator's "zero config" behavior, when the user signs up it automatically skips the prompt to enter the code since it recognizes the user has been confirmed. This has been working for several months, and there have been no significant changes to our trigger or frontend code.

Recently we started getting an error after the user signs up and the authenticator dialog switches to the signin tab: `Cannot read properties of undefined (reading 'username')`. Screenshot:

![AuthenticatorUsernameUndefined](https://user-images.githubusercontent.com/1449737/176778831-116389a0-7816-4691-9f35-d1ce728e3f6e.jpg)

There are no errors in the javascript console.

Neither our frontend nor our trigger references `username`. There are no errors for the trigger function in the CloudWatch logs. The error seems to be harmless since the new user is able to sign in successfully.

If we remove the pre-signup trigger the problem goes away. But, of course, now Cognito sends a confirmation email and the authenticator dialog prompts for the confirmation code, which is not the desired behavior.

### What's the expected behaviour?

The authenticator dialog should not show a spurious error when the user is auto-confirmed by the pre-signup trigger.

### Help us reproduce the bug!

In Amplify set up a Cognito backend with a pre-signup trigger similar to [this example](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-sign-up.html#aws-lambda-triggers-pre-registration-example-2)  in the Amplify developer's guide to auto-confirm users, e.g.:

```js
exports.handler = async (evt) => {
[...]

PR #2269, opened 2022-07-08
fix: username undefined with autoconfirmed lambda

ErikCH:
#### Description of changes
When using auto confirmed lambda an error will show that the `username is not defined`

This PR sets the correct credentials, so the sign in can occur correctly. This was caused by a regression that no longer
set the credentials when you skip the confirmed page.

#### Issue #, if available

#2235 

Created new test to check values being sent to SignIn

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/.changeset/nervous-clocks-reply.md
+++ b/.changeset/nervous-clocks-reply.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ui': patch
+---
+
+Fixed issue when customer is using auto confirmed lambda. Now it will work again.
--- a/.changeset/nervous-clocks-reply.md
+++ b/.changeset/nervous-clocks-reply.md
@@ -2,4 +2,4 @@
 '@aws-amplify/ui': patch
 ---
 
-Fixed issue when customer is using auto confirmed lambda. Now it will work again.
+Fixed bug that displayed "usernamed undefined" when using a lambda that autoconfirms the user during sign up. The sign up would fail and redirect the user to sign in with the "username undefined" error. The user then would have to sign in again. This patch will now assign the correct credentials during sign up for users that are auto confirmed. 
--- a/packages/e2e/cypress/fixtures/sign-up-with-email-with-lambda-trigger.json
+++ b/packages/e2e/cypress/fixtures/sign-up-with-email-with-lambda-trigger.json
@@ -1,9 +1,4 @@
 {
-  "CodeDeliveryDetails": {
-    "AttributeName": "email",
-    "DeliveryMedium": "EMAIL",
-    "Destination": "a***@e***.com"
-  },
   "UserConfirmed": true,
   "UserSub": "••••••-••••-••••-••••-•••••••••••••"
 }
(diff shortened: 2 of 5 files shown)


packages/e2e/cypress/fixtures/sign-up-with-email-with-lambda-trigger.json
```diff
@@ -1,9 +1,4 @@
 {
-  "CodeDeliveryDetails": {
-    "AttributeName": "email",
-    "DeliveryMedium": "EMAIL",
-    "Destination": "a***@e***.com"
-  },
```
ErikCH, line 6: This is not normally sent, so I deleted it for this fixture.

packages/ui/src/machines/authenticator/signUp.ts
```diff
@@ -114,7 +114,7 @@ export function createSignUpMachine({ services }: SignUpMachineOptions) {
                       {
                         cond: 'shouldSkipConfirm',
                         target: 'skipConfirm',
-                        actions: ['setUser'],
+                        actions: ['setUser', 'setCredentials'],
```
ErikCH, line 117: We need to set the credentials and the user so the `authAttributes` are set. Otherwise the `Auth.signIn` will fail later on, since the credentials will be undefined.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -7,6 +7,7 @@ import { get, escapeRegExp } from 'lodash';
 
 let language = 'en-US';
 let window = null;
+let stub = null;
```
ErikCH, line 10: I tried adding stub as an alias instead. But it wasn't working correctly so a global stub variable seems OK in this instance.

.changeset/nervous-clocks-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/ui': patch
+---
+
+Fixed issue when customer is using auto confirmed lambda. Now it will work again.
```
reesscot, line 5: Can you please add more descriptive detail on what the error was and how it was fixed for the changelog?

reesscot (commented):
Looks good, I think it just need an update to the changeset description

.changeset/nervous-clocks-reply.md
ErikCH: OK, I'll add it.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -105,11 +106,16 @@ Given(
...
       console.info('`%s` mocked with %o', path, result);
-      cy.stub(obj, method).returns(result);
+      stub = cy.stub(obj, method);
+      stub.returns(result);
     });
   }
 );
 
+When('Sign in was called with {string}', (username: string) => {
+  expect(stub.calledWith(username, Cypress.env('VALID_PASSWORD'))).to.be.true;
```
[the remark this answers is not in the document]
ErikCH: OK updated

.changeset/nervous-clocks-reply.md
ErikCH: Ok updated

hbuchel (approved):
LGTM

PR #2202, opened 2022-06-28
chore(ui-react): lint src/primitives/shared

calebpollman:
#### Description of changes
Allow linting for _src/primitives/shared/_

#### Issue #, if available
NA

#### Description of how you validated changes
`yarn react lint && yarn react test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated
- [ ] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2277, opened 2022-07-11
chore(ui-react): lint src/primitives/shared

calebpollman:
#### Description of changes

Reopens https://github.com/aws-amplify/amplify-ui/pull/2202 (reverted in https://github.com/aws-amplify/amplify-ui/pull/2250)

Allow linting for _src/primitives/shared/_
> Commit https://github.com/aws-amplify/amplify-ui/commit/8b4a21171ebf5787230ee28f78df5d965defcb24 fixes the issue that caused the revert and adds a unit test to prevent the regression in the future.

#### Issue #, if available
NA

- manually tested
- `yarn react lint && yarn react test`
- added an additional unit test for `getClosestValueByBreakpoint` to prevent the issue that required the revert

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/primitives/shared/types.ts
```diff
@@ -0,0 +1,174 @@
...
+  | 'TextAreaField'
+  | 'TextField'
+  | 'ToggleButton'
+  | 'ToggleButtonGroup'
+  | 'VisuallyHidden';
+
+export type ComponentClasses = string;
+
+interface ComponentClassNameMetadata {
+  className: ComponentClasses;
```
zchenwei, line 163: Feel like we can just use `string` to reduce the complexity. Is there a particular reason to add one more abstract level here?

zchenwei: Looks like the original one is a union of all component classnames, more specific than `string`

packages/react/src/primitives/types/theme.ts
```diff
@@ -239,51 +239,67 @@ export type TransformKeys<PropertyType> =
...
+  | 'paddingTop'
+  | 'paddingBottom'
+  | 'right'
+  | 'rowGap'
+  | 'top'
+  | 'transform'
+  | 'width';
+
+export type ThemeTokenKey =
+  | 'components'
```
zchenwei, line 291: I think we don't have a `components` token key now

zchenwei (commented):
Overall, LGTM! Just left a few questions

packages/react/src/primitives/types/theme.ts
zchenwei: I see what this is.

packages/react/src/primitives/shared/types.ts
calebpollman: That's a legitimate call out, will update

zchenwei (dismissed):
LGTM

packages/react/src/hooks/useBreakpointValue.ts
```diff
@@ -2,6 +2,7 @@ import { Breakpoint } from '../primitives/types/responsive';
 import { getValueAtCurrentBreakpoint } from '../primitives/shared/responsive/utils';
 import { useBreakpoint } from '../primitives/shared/responsive/useBreakpoint';
 import { useTheme } from './useTheme';
+import { ThemeStylePropKey } from 'src/primitives/types/theme';
```
reesscot, line 5: I think this import will fail in a later build.
```suggestion
import { ThemeStylePropKey } from '../primitives/types/theme';
```

reesscot (approved):

zchenwei (approved):

PR #2289, opened 2022-07-13
docs(chore): Standard HTML attributes note

joebuono:
#### Description of changes

_Asana task:_
Docs Props table: update the note under the Props Heading to specify the HTML element's name and MDN link.

This PR is a refactor of https://github.com/aws-amplify/amplify-ui/pull/2278, per @0618's comments:
- Include `mdnUrl` and `htmlElement` metadata in the frontmatter for each component
- Use the above metadata to generate the note about standard HTML attributes

The only two files that need review are:
- docs/scripts/generate-props-tables.ts
- docs/scripts/util/getFrontmatter.ts

Screen Shot 2022-07-13 at 4 23 05 PM

Ran the docs locally and checked the MDN links for each page. 

#### Checklist

- [X] PR description included


docs/scripts/util/getFrontmatter.ts
```diff
@@ -0,0 +1,13 @@
+export function getFrontmatter(fileData: string) {
```
0618, line 1: Nice 👍

docs/scripts/generate-props-tables.ts
```diff
@@ -35,6 +33,10 @@ async function createAllPropsTables() {
         categoryName.toLowerCase() === componentPageName.toLowerCase()
     ) as ComponentName;
 
+    const { MDN, htmlElement } = getFrontmatter(
+      fs.readFileSync(componentFilepath, 'utf8')
```
joebuono, line 37: Do we need to handle a possible error case here?

docs/scripts/util/getFrontmatter.ts
```diff
@@ -0,0 +1,13 @@
...
+  // get MDN URL
+  const mdn = fileData.indexOf('MDN:');
+  const endOfLine1 = fileData.indexOf('\n', mdn);
+  const MDN = fileData.slice(mdn, endOfLine1).split(' ')[1];
+
+  // get htmlElement
```
0618, line 7: nit: how about making a function `getFrontmatterElement(element)` and the `element` can be `mdn`, `htmlElement`, etc, so that we can re-use it?

0618: Or, we can consider what we did for `preval`, just grab all the frontmatter elements, so that we don't have to keep modifying this function in the future if we need to grab more than `MDN` and `htmlElement`. Does that make sense?

0618 (dismissed):

docs/scripts/util/getFrontmatter.ts
joebuono: Yes, that makes sense. I'll modify the function to grab all the frontmatter elements

docs/scripts/util/getFrontmatter.ts
0618, line 1: nit: maybe we can add the error handler here to check if there's any `fileData`

docs/scripts/util/getFrontmatter.ts
```diff
@@ -0,0 +1,13 @@
+export function getFrontmatter(fileData: string) {
+  // get MDN URL
+  const mdn = fileData.indexOf('MDN:');
+  const endOfLine1 = fileData.indexOf('\n', mdn);
+  const MDN = fileData.slice(mdn, endOfLine1).split(' ')[1];
```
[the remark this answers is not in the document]
joebuono: Yeah, I wasn't sure how to name is since both MDN and URL are all caps. I'll change it to be `mdnUrl`

docs/scripts/generate-props-tables.ts
```diff
@@ -35,6 +33,20 @@ async function createAllPropsTables() {
         categoryName.toLowerCase() === componentPageName.toLowerCase()
     ) as ComponentName;
 
+    let mdn, element;
+    try {
+      const fileData = fs.readFileSync(componentFilepath, 'utf8');
+      const { mdnUrl, htmlElement } = getFrontmatter(fileData);
```
0618, line 39: Would it safe as some effort in future if we catch the error in `getFrontmatter`? So that we don't have to catch it every time when using the util

joebuono: Good idea

0618 (approved):
Thanks for making this change for props table!

jacoblogan (approved):

PR #2297, opened 2022-07-14
feat: add alert for publishing @next version

0618:
#### Issue #, if available

https://app.asana.com/0/1201736086077862/1201914591116096/f

This PR add metrics for success and failures of the after merge next tag publishing events in Prod account.

TODO:
- manually create an alarm to trigger Sev3 tickets
- Ideally, we need to migrate the metrics and alarms to CDK and deploy them to the canary accounts

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/.github/workflows/run-and-test-builds.yml
+++ b/.github/workflows/run-and-test-builds.yml
@@ -96,8 +96,8 @@ jobs:
           NEW_PASSWORD: ${{ secrets.NEW_PASSWORD }}
           VALID_PASSWORD: ${{ secrets.VALID_PASSWORD }}
 
-  trigger-failure-alarm:
-    # Triggers an alarm if any of builds failed.
+  log-failure-metric:
+    # Send data point 1 to metric RunTimeTestsFailure, if failure
     runs-on: ubuntu-latest
     environment: ci
     needs: build
@@ -111,8 +111,8 @@ jobs:
           aws-region: us-east-1
       - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsFailure --namespace GithubCanaryApps --value 1
 
-  trigger-success-alarm:
-    # Triggers an alarm if all the builds succeed
+  log-success-metric:
+    # Send data point 0 to metric RunTimeTestsFailure, if success
     runs-on: ubuntu-latest
     environment: ci
     needs: build
@@ -124,4 +124,4 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
-      - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsSuccess --namespace GithubCanaryApps --value 0
+      - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsFailure --namespace GithubCanaryApps --value 0
--- a/.github/workflows/run-and-test-builds.yml
+++ b/.github/workflows/run-and-test-builds.yml
@@ -111,7 +111,7 @@ jobs:
           aws-region: us-east-1
       - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsFailure --namespace GithubCanaryApps --value 1
 
-  trigger-sucess-alarm:
+  trigger-success-alarm:
     # Triggers an alarm if all the builds succeed
     runs-on: ubuntu-latest
     environment: ci
@@ -124,4 +124,4 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
-      - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsFailure --namespace GithubCanaryApps --value 0
+      - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsSuccess --namespace GithubCanaryApps --value 0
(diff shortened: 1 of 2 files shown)


.github/workflows/run-and-test-builds.yml
```diff
@@ -124,4 +124,4 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
-      - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsFailure --namespace GithubCanaryApps --value 0
+      - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsSuccess --namespace GithubCanaryApps --value 0
```
ErikCH, line 127: We don't have a RunTimeTestsSuccess metric setup with any alarms right now. Is the idea that we'll use this in the future to alarm if it's missing for a certain period of time?

.github/workflows/test-deploy-main.yml
```diff
@@ -167,3 +167,33 @@ jobs:
...
+    needs: build
+    if: ${{ success() }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - run: aws cloudwatch put-metric-data --metric-name publishSuccess --namespace GithubCanaryApps --value 0
```
ErikCH, line 199: Same here with these two alarms.

.github/workflows/test-deploy-main.yml
```diff
@@ -167,3 +167,33 @@ jobs:
...
+    needs: build
+    if: ${{ failure() }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - run: aws cloudwatch put-metric-data --metric-name publishFailure --namespace GithubCanaryApps --value 1
```
ErikCH, line 184: I'm assuming publishFailure will be setup to alarm right away?

.github/workflows/run-and-test-builds.yml
0618: From my understand, this line will create a RunTimeTestsSuccess metric and we'll need to manually set up the alarm from the metric. Is that right?

.github/workflows/test-deploy-main.yml
0618: Same. I think this line is only for metrics

.github/workflows/run-and-test-builds.yml
ErikCH: Right

.github/workflows/test-deploy-main.yml
wlee221: We configured canary to trigger alarm on missing data. 

I don't think we can do that here though, because publish to `@next` does not happen on a regular schedule. IMO should have `@next` trigger alarm only on failure data and ignore missing data.

wlee221: Right, my thought here is that we shouldn't even need to add success metrics here. We can safely delete `trigger-success-alarm` and only have `trigger-failure-alarm` job.

wlee221 (approved):
LGTM, let's make sure cloudwatch alarms are set up before we make this happen.

.github/workflows/run-and-test-builds.yml
ErikCH: Looking at this further this might be an issue. The way the alarm works is that if RunTimeTestsFailure > 0 for 2 data points within 40 minutes the alarm is triggered. When a success occurs the `RunTimeTestsFailure` is sent a 0. If a 0 is never sent won't the `RunTimeTestsFailure` be above 0 forever?

Perhaps you have to change the additional configuration to have missing data treated as being good.

image

ErikCH: Or change this back to `RunTimeTestsFailure`

wlee221: With @ErikCH on this one, is this change required? This would break the existing metric because we've been relying on the value 0 and 1 to determine the alarm state.

wlee221 (commented):
Added a comment on run-and-test-builds changes 🙏

.github/workflows/run-and-test-builds.yml
reesscot: +1

0618: Awesome point! Thanks for catching it. That means I need to add the `success` action back to the publish workflow as well.

Just pushed a new commit 😁

.github/workflows/test-deploy-main.yml
```diff
@@ -167,3 +167,33 @@ jobs:
       - name: Run yarn build on each sample app
         run: yarn build
         working-directory: ./canary
+
+  trigger-failure-alarm:
+    # Triggers an alarm if any of builds failed.
```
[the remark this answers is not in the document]
0618: updated

.github/workflows/run-and-test-builds.yml
```diff
@@ -111,7 +111,7 @@ jobs:
           aws-region: us-east-1
       - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsFailure --namespace GithubCanaryApps --value 1
 
-  trigger-sucess-alarm:
+  trigger-success-alarm:
```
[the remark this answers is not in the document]
0618: renamed

ErikCH (approved):

wlee221 (approved):

PR #2311, opened 2022-07-20
feat: add CloudWatch metric when publishLatest fails

0618:
#### Description of changes

In https://github.com/aws-amplify/amplify-ui/pull/2297, we added a metric for publishing next@, but we still need an alarm to let us know when the build fails after a PR merged to `main`, which is "Publish Latest", so add a metric for the Publish Latest failures.

- add PublishNMPFailure metric
- create reusable workflow `metric.yml`

TODO:
Once the metric is ready, we'll set up the alarms `Publish NPM Failure` and `Publish Next@ Failure` for `PublishNPM` or `PublishNext`. When there's a value `1` on either of the metrics, we'll be notified with a Sev2, so oncall can take a look at the issue in time.
 - in alarm. 1 datapoint value >= 1 within 5 minutes

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


.github/workflows/publish-latest.yml
```diff
@@ -87,3 +87,33 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  log-failure-metric:
+    # Send data point 1 to metric PublishLatestFailure, if success
+    runs-on: ubuntu-latest
+    environment: ci
+    needs: build-test
```
wlee221, line 95:
```suggestion
    needs: publish
```

.github/workflows/publish-latest.yml
```diff
@@ -87,3 +87,33 @@ jobs:
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - run: aws cloudwatch put-metric-data --metric-name PublishLatestFailure --namespace GithubCanaryApps --value 1
+
+  log-success-metric:
+    # Send data point 0 to metric PublishLatestFailure, if success
+    runs-on: ubuntu-latest
+    environment: ci
+    needs: build-test
```
wlee221, line 110:
```suggestion
    needs: publish
```

.github/workflows/publish-latest.yml
wlee221: I also wonder if we should just remove `needs: publish` line and only have `if: ${{ failure() }}`

This will ensure that this runs when even when jobs before `publish` has failed.

0618: so `if: ${{ failure() }}` catches any failure from `build-test` and `publish`, whatever in this workflow?

If that's the case, yea, let's remove `needs` 👍

.github/workflows/publish-latest.yml
```diff
@@ -87,3 +87,33 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  log-failure-metric:
+    # Send data point 1 to metric PublishLatestFailure, if success
```
reesscot, line 92: I think the comment is wrong?
```suggestion
    # Send data point 1 to metric PublishLatestFailure, if failure occurs
```

.github/workflows/publish-latest.yml
0618: Done!

.github/workflows/publish-latest.yml
```diff
@@ -87,3 +87,25 @@ jobs:
...
+      status: ${{ failure() }}
+      metric-name: PublishNPMFailure
+      value: 1
+    secrets:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+  log-success-metric:
+    # Send data point 0 to metric PublishLatestFailure, if success
+    uses: aws-amplify/amplify-ui/.github/workflows/reusable/metric.yml
```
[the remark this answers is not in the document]
wlee221: For example, we can have:

```yml
  log-metric:
    # Send data point 0 or 1 to metric PublicLatestFailure
    uses: aws-amplify/amplify-ui/.github/workflows/reusable/metric.yml
    with:
      is-success: ${{ success() }} # This will be true if publish succeeded, false otherwise.
      metric-name: PublishNPMFailure 
      success-value: 0 # this is the value that will be sent when status is 'true'
      fail-value: 1 # this is the value that will be sent when status is 'false'
    secrets:
      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```

.github/workflows/reusable/metric.yml
```diff
@@ -0,0 +1,27 @@
+on:
```
wlee221, line 1: Since this is a fairly simple action, we can consider using reusable action (formerly called [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action)) instead of reusable workflow. I have an example composite action [here](https://github.com/aws-amplify/amplify-ui/blob/main/.github/actions/install-with-retries/action.yml).

To describe briefly, workflow consists of multiple jobs; and jobs consists of multiple actions (steps). Since this is a fairly simple workflow, imo composite action is better used here.

wlee221 (commented):
Few nits! Thanks for tackling CI/CD!

.github/workflows/publish-latest.yml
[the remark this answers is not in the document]
0618: That was my first thought, but the reason I didn't do it was because it would be more flexible to only have 1 type of datapoint logged each time, because we may have more than failures and successes in the future. Potentially, we can other numbers, status, etc.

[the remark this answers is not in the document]
wlee221: that's a very good point actually! I'm convinced.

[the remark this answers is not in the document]
wlee221: We can abstract more later as needed, this is a good start.

.github/workflows/run-and-test-builds.yml
```diff
@@ -98,30 +98,22 @@ jobs:
...
-    if: ${{ failure() }}
-    steps:
-      - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v1
-        with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-east-1
-      - run: aws cloudwatch put-metric-data --metric-name RunTimeTestsFailure --namespace GithubCanaryApps --value 1
+    uses: aws-amplify/amplify-ui/.github/workflows/reusable/metric.yml
```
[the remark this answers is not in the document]
0618: reverted. No wonder I didn't get the syntax checks 🤔

.github/workflows/publish-latest.yml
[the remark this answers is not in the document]
0618: Yea. I though about it. It's good for now and easy to be changed later.

.github/workflows/reusable/metric.yml
0618: Done!

.github/actions/log-metric/action.yml
```diff
@@ -0,0 +1,26 @@
...
+  metric-name:
+    description: 'target metric name'
+    required: true
+  value:
+    description: 'the value to log to the metric'
+    required: true
+  AWS_ACCESS_KEY_ID:
+    description: 'target AWS account credentials'
+    required: true
+  AWS_SECRET_ACCESS_KEY:
```
wlee221, line 13: This is the first time we add secrets to composite actions; do you know what the recommended ways to pass secrets are?

wlee221: I'm looking up too! Looks like `configure-aws-credentials` use `inputs` to use secrets too: https://github.com/aws-actions/configure-aws-credentials/blob/8a84b07f2009032ade05a88a28750d733cc30db1/action.yml#L7-L18

0618: Good catch. VS Code linter didn't allow me to add `env`, but I found on the internet that people just did it anyway and it worked, so let me try it out.

wlee221: Looking at the log, it scraps off secrets plaintext too: https://github.com/aws-amplify/amplify-ui/runs/7436708485?check_suite_focus=true

wlee221: Thanks for looking! I imagine what you have is safe too, I'm just trying to find an official source to solidify this.

0618: New commit pushed

.github/actions/log-metric/action.yml
```diff
@@ -0,0 +1,20 @@
...
+    description: 'the value to log to the metric'
+    required: true
+runs:
+  using: 'composite'
+  steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
wlee221, line 17: Will this work? According to https://stackoverflow.com/questions/70098241/using-secrets-in-composite-actions-github:

> you need to pass these secrets via inputs. there's a good reason for that: imagine if an action you're calling could use your secrets without you passing them to the action.

wlee221: Looks like two main ways of passing secrets in composite actions are:

(1) Pass them as inputs (e.g. [configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials)):
```yml
      - name: Configure AWS credentials
        uses: aws-actions/configure-aws-credentials@v1
        with:
          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```


(2) Pass them as env and reuse it (e.g. changesets/action):

```yml
      - name: Publish to @latest
        with:
          publish: yarn publish:latest
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

Though (2) is harder to work with because we're just using a simple composite action. If (1) is safe to do so, then I'm fine with keeping it simple and have it passed as an input.

0618: Tested in my personal fork. Secrets are masked even after they are passed to `inputs`, so we can safely use (1) 🎉 

image

.github/actions/log-metric/action.yml
wlee221: Resolving as per MJ's research

wlee221 (approved):
Thanks for addressing my nits!

jacoblogan (approved):
LGTM

Issue #1810, opened 2022-05-02
Bug: PasswordField ShowPassword button context not announced to AT


hbuchel:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Any

### Please describe your bug.

Currently, the ShowPasswordButton does not announce its change in state when toggled. Additionally, the aria-label provided which switches back and forth between "Hide label" and "Show label" has somewhat confusing context.

In attached video, notice that when the button is toggled on and off, nothing is announced.

https://user-images.githubusercontent.com/376920/166312889-7b07ed39-a79e-4040-a8a6-f58571036ae6.mp4

.

### What's the expected behaviour?

The change in state of the password input should be announced. When toggled "on", it should announce "Password is shown." When toggled "off", it should announce "Password is hidden." Additionally, we should consider using a switch role for the button to include the context of it turning on and off, and keep a consistent label (as you would for a checkbox, the label stays the same). We have a spike completed to show acceptance criteria for this bug and the recommended fix.

### Help us reproduce the bug!

Any use of the PasswordField component. This can be demonstrated in the docs or in an external application instance using either the primitive or the Authenticator component.

### Code Snippet

```typescript
// Put your code below this line.

```

PR #2323, opened 2022-07-21
ui-react(fix): Update ShowPasswordButton

joebuono:
#### Description of changes

Fixes: https://github.com/aws-amplify/amplify-ui/issues/1810

https://user-images.githubusercontent.com/48109584/180318813-edf6acb9-294d-4a97-8ede-cf1d043432cf.mov

**User Story**

- As a screen reader user, I want the Show Password button to announce its current context, so that I know when it is toggled or not.

**Acceptance Criteria**

- ShowPasswordButton should use role=“switch” and have an accompanying aria-checked state.
- Keep consistent aria-label “Show password”. Remove the unnecessary “Hide password” label/translation.
- Add additional screen reader only context to shared/i18n.ts for “checked” states: “Password is shown”, “Password is hidden”
- Add a visually hidden aria-live region (polite) that updates based on the ShowPasswordButton checked state using our new context (see previous item).

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/1810

Wrote tests, ran docs locally and used a screen reader to confirm accessibility. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


.changeset/poor-pans-chew.md
```diff
@@ -0,0 +1,14 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+ui-react(fix): Update ShowPasswordButton to use role=“switch” and add additional SR context
+
+User Story
+- As a screen reader user, I want the Show Password button to announce its current context, so that I know when it is toggled or not.
+
+Acceptance Criteria
```
0618, line 10: nit: since it'll be a part of our release note, we may want to remove "Acceptance Criteria"

packages/react/src/primitives/types/passwordField.ts
```diff
@@ -50,14 +42,6 @@ export interface ShowPasswordButtonProps extends ButtonProps {
    */
   fieldType: PasswordFieldType;
 
-  /**
-   * @description
```
0618, line 54: nit: let's update the props table

packages/react/src/primitives/shared/constants.ts
```diff
@@ -539,8 +539,9 @@ export const ComponentText = {
     searchButtonLabel: 'Search',
   },
   PasswordField: {
-    showPasswordButtonLabel: 'Show password',
-    hidePasswordButtonLabel: 'Hide password',
+    passwordHidden: 'Password is hidden',
+    passwordShown: 'Password is shown',
+    showPasswordButton: 'Show password',
```
0618, line 544: nit: I know we've been using "...Label", "...Button" to name text strings, in this file, but when I review the code, I feel it's a little confusing and I'm not able to immidiately know it's the text. Should we just use the verbiage itself, like "showPassword", ""passwordIsHidden", ""passwordIsShown", etc?

packages/react/src/primitives/PasswordField/__tests__/ShowPasswordButton.test.tsx
```diff
@@ -21,33 +21,55 @@ describe('ShowPasswordButton component', () => {
...
+    await screen.findByRole('switch');
 
-    expect(ref.current.nodeName).toBe('BUTTON');
+    expect(ref.current?.nodeName).toBe('BUTTON');
   });
 
-  it('should set correct ariaLabel for fieldType password', async () => {
-    const ref = React.createRef<HTMLButtonElement>();
-    render(<ShowPasswordButton fieldType="password" ref={ref} />);
+  it('should toggle field type, screen reader context, and aria-checked when clicked', async () => {
```
0618, line 29: Nice!

Out of curiosity, I just searched if there's any a11y test api from react testing library, and I found this https://testing-library.com/docs/dom-testing-library/api-accessibility/. Maybe you have already been aware of it. I wonder if there's anything we can use?

0618 (commented):
Thanks for improving the a11y!

packages/react/src/primitives/shared/constants.ts
joebuono: Yes, I agree that just using the verbiage itself would add some clarify

packages/react/src/primitives/types/passwordField.ts
```diff
@@ -9,14 +9,6 @@ export interface PasswordFieldProps extends TextInputFieldProps {
    */
   hideShowPassword?: boolean;
 
-  /**
-   * @description
-   * Set the `aria-label` for hide password button
-   * @default
-   * "Hide password"
-   */
-  hidePasswordButtonLabel?: string;
```
calebpollman, line 18: Does removing `hidePasswordButtonLabel` and `hidePasswordButtonLabel` constitute a breaking change?

reesscot: Yes, I think we can probably ignore these props, but removing them from the interface is a breaking change since it will cause TS apps to thrown an error. Rather than removing them, @joebuono can you please mark them as `@deprecated` with a message saying they are no longer in use?

joebuono: Hmm good question. I think so?

calebpollman: IMO we should preserve the existing behavior to prevent breaking existing usage and add the deprecation warnings, along with supporting the improvements

joebuono: Tagging @hbuchel to get her thoughts when she's back

reesscot: I don't think it's a breaking change to no longer use the `hidePasswordButtonLabel` because it never worked before because it was never announced by the screen reader. With the new `role=switch` there's no reason to include the `hidePasswordButtonLabel` aria label because the state is communicated via the on/off state (see video in pr description). I do think we should leave the prop in the type and simply mark it as being deprecated and unused. 

I think we can wait for @hbuchel's feedback as well.

hbuchel (commented):
Overall the functionality looks good! Agree with others about keeping the props and marking as deprecated to avoid customers having type errors. 

Also agree that we cannot keep current  functionality (updating the  value of the aria label on the button) because it leads to a confusing context. If a user tabs away from the button, after enabling it (so show password is "on") then comes back to it, the button would announce a new name "Hide password" but still show as "on". So the aria-label should remain the same, similar to how a checkbox works.

@joebuono @reesscot This was missing from the acceptance criteria I wrote 🤦‍♀️  but should we expose these values (passwordIsShown, passwordIsHidden) as props like the old aria label values were (and fall back to our ComponentText.PasswordField values if not provided)? This would enable us to make it translatable for the Authenticator component as well.

.changeset/poor-pans-chew.md
```diff
@@ -0,0 +1,12 @@
...
+"@aws-amplify/ui-react": patch
+---
+
+ui-react(fix): Update ShowPasswordButton to use role=“switch” and add additional SR context
+
+As a screen reader user, I want the Show Password button to announce its current context, so that I know when it is toggled or not.
+
+- ShowPasswordButton should use role=“switch” and have an accompanying aria-checked state.
+- Keep consistent aria-label “Show password”. Remove the unnecessary “Hide password” label/translation.
+- Add additional screen reader only context to shared/i18n.ts for “checked” states: “Password is shown”, “Password is hidden”
```
hbuchel, line 11: minor nit:
```suggestion
- Add additional screen reader only context to shared/constants.ts for “checked” states: “Password is shown”, “Password is hidden”
```

hbuchel (approved):

hbuchel (approved):

0618 (approved):
:shipit:

Issue #2215, opened 2022-06-29
Bug: StepperField disabled buttons are hard to see


reesscot:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

NA

### Please describe your bug.

Disabled state is hard to see

![image](https://user-images.githubusercontent.com/6165315/176493472-a5dedfc2-a09a-4e8e-93a4-7bce03b112a9.png)

### What's the expected behaviour?

See above

### Help us reproduce the bug!

See above

### Code Snippet

```typescript
// Put your code below this line.

```

PR #2330, opened 2022-07-22
chore: Update StepperField disabled buttons style

ioanabrooks:
#### Description of changes

This changes fixes an issue where the StepperField disabled buttons are hard to see (Fixes: #2215).

| Before  | After |
|Screen Shot 2022-07-22 at 9 16 48 AM|Screen Shot 2022-07-21 at 8 38 56 PM  |
| Screen Shot 2022-07-22 at 9 17 27 AM  | Screen Shot 2022-07-21 at 8 39 35 PM  |

#### Issue #, if available

#2215 

Ran docs locally.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/src/theme/css/component/stepperField.scss
+++ b/packages/ui/src/theme/css/component/stepperField.scss
@@ -48,7 +48,9 @@
     --amplify-components-stepperfield-button-background-color
   );
   &--disabled {
-    background-color: var(--amplify-colors-background-disabled);
+    background-color: var(
+      --amplify-components-stepperfield-button-disabled-background-color
+    );
   }
   &--quiet {
     border-width: 0 0 var(--amplify-components-button-border-width) 0;
--- a/packages/ui/src/theme/css/component/stepperField.scss
+++ b/packages/ui/src/theme/css/component/stepperField.scss
@@ -42,31 +42,29 @@
   );
 }
 
-.amplify-stepperfield__button--decrease {
+.amplify-stepperfield__button--decrease,
+.amplify-stepperfield__button--increase {
   background-color: var(
     --amplify-components-stepperfield-button-background-color
   );
-  &[data-invalid='true'] {
-    border-inline-end: none;
+  &--disabled {
+    background-color: var(--amplify-colors-background-disabled);
   }
-
   &--quiet {
     border-width: 0 0 var(--amplify-components-button-border-width) 0;
     border-radius: 0;
   }
 }
 
+.amplify-stepperfield__button--decrease {
+  &[data-invalid='true'] {
+    border-inline-end: none;
+  }
+}
+
 .amplify-stepperfield__button--increase {
-  background-color: var(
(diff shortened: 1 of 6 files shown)


packages/ui/src/theme/css/component/stepperField.scss
```diff
@@ -42,31 +42,29 @@
   );
 }
 
-.amplify-stepperfield__button--decrease {
+.amplify-stepperfield__button--decrease,
+.amplify-stepperfield__button--increase {
```
ioanabrooks, line 46: Consolidated common styles for increase/decrease buttons, such as `background-color`, `disabled` and `quiet` to reduce duplication.

packages/ui/src/theme/css/component/stepperField.scss
```diff
@@ -42,31 +42,29 @@
...
   background-color: var(
     --amplify-components-stepperfield-button-background-color
   );
-  &[data-invalid='true'] {
-    border-inline-end: none;
+  &--disabled {
+    background-color: var(--amplify-colors-background-disabled);
```
0618, line 51: Do we need to add a variable like `----amplify-components-stepperfield-button-background-color-disabled` ?

0618 (commented):
🎉  You delivered so quick! Nice job!

packages/ui/src/theme/css/component/stepperField.scss
zchenwei, line 51: I think we should use `var(--amplify-components-stepperfield-button-disabled-background-color);` but update its token value instead of using the inherited value from `Button` - https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/tokens/components/stepperField.ts#L57

zchenwei (commented):
Looks good! Left a comment

packages/ui/src/theme/tokens/components/stepperField.ts
```diff
@@ -54,7 +54,7 @@ export const stepperfield: StepperFieldTokens = {
     _disabled: {
       color: { value: '{components.button._disabled.color}' },
       backgroundColor: {
-        value: '{components.button._disabled.backgroundColor}',
+        value: '{components.fieldcontrol._disabled.backgroundColor}',
```
ioanabrooks, line 57: Chose `components.fieldcontrol._disabled.backgroundColor` here such that when the entire stepper  component is disabled there is style consistency between the buttons and the field.

packages/ui/src/theme/css/component/stepperField.scss
ioanabrooks: Done!

zchenwei (approved):
LGTM. You probably need to update the snapshot in ui package since you've updated the tokens there

0618 (approved):
:shipit:

zchenwei (approved):
:shipit:

Issue #1631, opened 2022-04-06
No result found or custom component should be support in Collection's search Bar


jaythakor-ai:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Create React App

### Please describe your bug.

Currently, the search filters in the Collection component is not showing any message or allow the rendering of the custom component when the searched value is not present in the Collection.

### What's the expected behaviour?

There should be a way to render the custom component or "No result Found" message while searched value is not present in the Collections

### Help us reproduce the bug!

Create a collection of any components and try to search for value from that collection.

### Code Snippet

```typescript
// Put your code below this line.

```

dbanksdesign:
Hey @jaythakor-ai thanks for raising this issue! We definitely want to address this and we are working on what the right developer experience should be.

PR #2333, opened 2022-07-22
fix(ui-react): Collection search noResultsFound

joebuono:
#### Description of changes

Fixes: https://github.com/aws-amplify/amplify-ui/issues/1631

As an Amplify UI app customer, when my search in a Collection component returns nothing, I should see the message "No results found" or custom text.

- When no results are returned, we render the text "No results found"
- Customer should be able to pass their own `ReactNode` for `noResultsfound`
- "No results found" should be translated into all of our currently supported languages (pending Totoro, ETA July 27)

https://user-images.githubusercontent.com/48109584/180516033-1111c0bd-92ab-4ac3-a7ff-3eeefc5855f0.mov

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/1631

Ran the docs locally and wrote tests

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


packages/react/src/primitives/Collection/Collection.tsx
```diff
@@ -109,7 +111,15 @@ export const Collection = <Item,>({
         </Flex>
       ) : null}
 
-      {collection}
+      {items.length ? (
```
ioanabrooks, line 114: Is there a possibility of items being undefined?

joebuono: Good questions. I don't think so because of line 67:
```
  items = Array.isArray(items) ? items : [];
 ```

Comment (ioanabrooks) [reply packages/react/src/primitives/Collection/Collection.tsx:114]: Wondering if it worth adding a check, just to be safe, in the future the code could be refactored and we may end up with NPE.

File: .changeset/unlucky-beers-agree.md [hunk 2]
```diff
@@ -0,0 +1,11 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+fix(ui-react): Collection search noResultsFound
+
+As an Amplify UI app customer, when my search in a Collection component returns nothing, I should see the message "No results found" or custom text.
```
reesscot, line 7: I don't think our agile user story language should leak into our changelog notes. This should be short, sweet, to the point, and include an example.

See "feat: adding formatValue prop on SliderField to allow formatting the value" example in changelog:
https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CHANGELOG.md

packages/react/src/primitives/types/collection.ts
```diff
@@ -50,6 +51,12 @@ export interface CollectionWrapperProps extends BaseStyleProps {
    * Search field placeholder
    */
   searchPlaceholder?: string;
+
+  /**
+   * @description
+   * Controls what is displayed when no results are found from the Collection search
+   */
+  noResultsFound?: React.ReactNode;
```
reesscot, line 59: I know I proposed this name in the original ticket, but what if we name this more consistently with the other search related props?
```suggestion
  searchNoResultsFound?: React.ReactNode;
```

reesscot (commented):
Looking great, just had some suggestions below

packages/react/src/primitives/types/collection.ts
joebuono: Good idea

ioanabrooks (approved):
LGTM

jacoblogan (approved):
LGTM

PR #2334, opened 2022-07-22
chore(ui-react): lint primitives/utils

calebpollman:
#### Description of changes
Update _primitives/utils_ to pass linting rules
- remove extraneous memoization in `useTestId` and rename to `getTestId` as it no longer requires usage of a React hook internally

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/utils/testUtils.ts
+++ b/packages/react/src/primitives/utils/testUtils.ts
@@ -1,25 +1,23 @@
-import * as React from 'react';
-export const useTestId = (testId: string, component: string) => {
-  const newTestId = React.useMemo(
-    () => (testId && component ? `${testId}-${component}` : undefined),
-    [testId, component]
-  );
-  return newTestId;
-};
+export const getTestId = (testId: string, component: string): string =>
+  testId && component ? `${testId}-${component}` : undefined;
 
-export const errorMessageWrapper = (fn: () => void, message: string) => {
+export const errorMessageWrapper = (
+  fn: () => void,
+  message: string
+): void | Error => {
   try {
     fn();
   } catch (error) {
     // Formatting below is intentional
     // and displays below Jest error message
-    error.message += `
+    (error as Error).message += `
 
 -- Custom Error Message --
 ${message}
 
 `;
+    // eslint-disable-next-line no-console
     console.error(error);
-    throw new Error(error);
+    throw new Error((error as Error).message);
   }
 };
--- a/packages/react/src/primitives/utils/useLayoutEffect.ts
+++ b/packages/react/src/primitives/utils/useLayoutEffect.ts
@@ -8,8 +8,6 @@ import * as React from 'react';
  *
  * See: https://reactjs.org/docs/hooks-reference.html#uselayouteffect
  */
-const useLayoutEffect = Boolean(globalThis?.document)
-  ? React.useLayoutEffect
-  : () => {};
+const useLayoutEffect = globalThis?.document ? React.useLayoutEffect : () => {};
 
 export { useLayoutEffect };
--- a/packages/react/src/primitives/utils/useStableId.ts
+++ b/packages/react/src/primitives/utils/useStableId.ts
@@ -9,10 +9,14 @@ export type UseStableId = (id?: string) => string;
 
 // Create a local version of React.useId which will reference React.useId for React 18
 // and fallback to noop for React 17 and below
-// Note: We `toString()` to prevent bundlers from trying to `import { useId } from 'react';`
+// Note: We use `toString()` to prevent bundlers from trying to `import { useId } from 'react';`
 // since it doesn't exist in React 17 and below (prevents https://github.com/aws-amplify/amplify-ui/issues/1154)
 const useReactId: () => string | (() => undefined) =
-  (React['useId'.toString()] as () => string) || (() => undefined);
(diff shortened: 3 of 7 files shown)


packages/react/src/primitives/utils/testUtils.ts
```diff
@@ -1,25 +1,23 @@
-import * as React from 'react';
-export const useTestId = (testId: string, component: string) => {
-  const newTestId = React.useMemo(
-    () => (testId && component ? `${testId}-${component}` : undefined),
-    [testId, component]
-  );
-  return newTestId;
-};
```
calebpollman, line 8: Since `useTestId` always returns a string, there is no need to memoize as the reference will remain the same unless the arguments are updated (which would result in `useMemo` recomputing)

packages/react/src/primitives/utils/testUtils.ts
```diff
@@ -1,25 +1,23 @@
...
+    (error as Error).message += `
 
 -- Custom Error Message --
 ${message}
 
 `;
+    // eslint-disable-next-line no-console
     console.error(error);
-    throw new Error(error);
+    throw new Error((error as Error).message);
```
calebpollman, line 21: Updated to pass the value of `error.message` in place of the `Error` object

packages/react/src/primitives/utils/useLayoutEffect.ts
```diff
@@ -8,8 +8,6 @@ import * as React from 'react';
  *
  * See: https://reactjs.org/docs/hooks-reference.html#uselayouteffect
  */
-const useLayoutEffect = Boolean(globalThis?.document)
```
calebpollman, line 11: The `Boolean` here is redundant (this was picked up by [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast), want to make it clear that it was not just an opinion)

packages/react/src/primitives/utils/testUtils.ts
reesscot: Curious why the need to throw the `error.message` rather than the `Error` object?

packages/react/src/primitives/utils/useLayoutEffect.ts
```diff
@@ -8,8 +8,6 @@ import * as React from 'react';
...
-  ? React.useLayoutEffect
-  : () => {};
+const useLayoutEffect = globalThis?.document ? React.useLayoutEffect : () => {};
```
reesscot, line 11: Is there a linting rule against using Boolean?

packages/react/src/primitives/utils/testUtils.ts
calebpollman: Error expects a `string` argument

packages/react/src/primitives/utils/useLayoutEffect.ts
calebpollman: No just redundant usage ([see this comment](https://github.com/aws-amplify/amplify-ui/pull/2334#discussion_r927973321))

packages/react/src/primitives/utils/useStableId.ts
```diff
@@ -12,7 +12,7 @@ export type UseStableId = (id?: string) => string;
 // Note: We `toString()` to prevent bundlers from trying to `import { useId } from 'react';`
 // since it doesn't exist in React 17 and below (prevents https://github.com/aws-amplify/amplify-ui/issues/1154)
 const useReactId: () => string | (() => undefined) =
-  (React as any)['useId'.toString()] || (() => undefined);
+  (React['useId'.toString()] as () => string) || (() => undefined);
```
reesscot, line 15: Doesn't TS complain that `useId` isn't available in React, or is the `toString` getting around that?

packages/react/src/primitives/utils/testUtils.ts
calebpollman: The reason this was not throwing a TS error previously was that `error` was implicitly `any`

reesscot: Right, I was misreading the code

reesscot (dismissed):
LGTM

packages/react/src/primitives/utils/useStableId.ts
```diff
@@ -9,10 +9,14 @@ export type UseStableId = (id?: string) => string;
 
 // Create a local version of React.useId which will reference React.useId for React 18
 // and fallback to noop for React 17 and below
-// Note: We `toString()` to prevent bundlers from trying to `import { useId } from 'react';`
+// Note: We use `toString()` to prevent bundlers from trying to `import { useId } from 'react';`
 // since it doesn't exist in React 17 and below (prevents https://github.com/aws-amplify/amplify-ui/issues/1154)
 const useReactId: () => string | (() => undefined) =
-  (React as any)['useId'.toString()] || (() => undefined);
+  // disable eslint below to allow usage of casting React to `any`, which ensures that TS
+  // does not get confused about the existence of `useId` in React 17 and below
```
wlee221, line 16: TIL what this code meant!

wlee221 (dismissed):

wlee221 (approved):

reesscot (approved):

Issue #1982, opened 2022-05-25
Pagination: Consider keeping current item a focusable button, use aria-current to specify current item


hbuchel:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

N/A

### Please describe your bug.

The current pagination component switches the "active" page to a span. This makes it a little awkward when navigating via screenreader (when tabbing through focusable items) because it won't ever read out the visually hidden text for that element. You could guess that the "missing" element is the current, and verify by switching to how you're navigating the page, but it seems like an improvement to just have it be a focusable button like the others and add some additional attributes.

https://user-images.githubusercontent.com/376920/170329437-9f2806b2-0e1f-4d8f-9ca7-eaadae314ea3.mp4

### What's the expected behaviour?

You could expect the current item button to be focusable, and for a screenreader to announce if the item is current or not. In that case it might look generally like:

```
<button aria-current="true"><span class="amplify-visually-hidden">Current page</span> 1<button>
```

We can check out some other similar examples:
https://designsystem.digital.gov/components/pagination/
https://mui.com/material-ui/react-pagination/
https://design-system.w3.org/components/pagination.html

### Help us reproduce the bug!

Current pagination examples https://ui.docs.amplify.aws/react/components/pagination

### Code Snippet

```typescript
// Put your code below this line.

```

PR #2336, opened 2022-07-22
fix(ui-react): Pagination current item a11y

joebuono:
#### Description of changes

Fixes: https://github.com/aws-amplify/amplify-ui/issues/1982

To improve the accessibility of the Pagination component:
- current item button is focusable
- screenreader announces if the item is current or not

https://user-images.githubusercontent.com/48109584/180571531-8cb72f58-70c6-4fbf-bb77-2ab775f5a9d0.mov

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/1982

Ran the docs locally and used a screen reader. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Pagination/PaginationItem.tsx
+++ b/packages/react/src/primitives/Pagination/PaginationItem.tsx
@@ -60,16 +60,12 @@ export const PaginationItem: React.FC<PaginationItemProps> = ({
         <View as="li">
           {page === currentPage ? (
             <Flex
-              aria-current="true"
+              aria-current="page"
               as="button"
               className={ComponentClassNames.PaginationItemCurrent}
               testId={PAGINATION_CURRENT_TEST_ID}
               {...rest}
             >
-              {/**
-               * Use markup to indicate the current item of a menu, such as the current page on a website, to improve orientation in the menu.
-               * @link https://www.w3.org/WAI/tutorials/menus/structure/#indicate-the-current-item
-               */}
               <VisuallyHidden>{currentPageLabel}:</VisuallyHidden>
               {page}
             </Flex>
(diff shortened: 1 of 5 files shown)


hbuchel (commented):
Tiny tiny nit: What do you think about updating our default `currentPageLabel` (and docs under Accessibility section) to be:
```
currentPageLabel: 'Page:',
```
instead of `Current Page: `? It reduces the verbosity a little since the aria-current already reads out the word "current item". It would then announce:

**"Page 2, current item, button, group"**  instead of **"Current Page 2, current item, button group"**

hbuchel (approved):
LGTM! Thanks for the updates!

hbuchel:
> LGTM! Thanks for the updates!

Oops spoke  too soon on this. Looks like there is a failing unit test.

hbuchel (approved):

packages/react/src/primitives/Pagination/PaginationItem.tsx
```diff
@@ -60,7 +60,8 @@ export const PaginationItem: React.FC<PaginationItemProps> = ({
         <View as="li">
           {page === currentPage ? (
             <Flex
-              as="span"
+              aria-current="true"
```
reesscot, line 63: @hbuchel Should this be set to `aria-current="page"`? 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#values

reesscot (approved):
LGTM pending my question below

packages/react/src/primitives/Pagination/PaginationItem.tsx
hbuchel: Oops sorry, yeah that is better, thanks! @joebuono there is a comment we can remove just below it too that references the menu aria pattern that isn't relevant anymore.

hbuchel (approved):

0618 (approved):
:shipit:

Issue #2284, opened 2022-07-12
close-toast does not have discernible button text


jssuttles:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Vue

### Which UI component?

Authenticator

### How is your app built?

vue cli

### What browsers are you seeing the problem on?

Chrome

### Please describe your bug.

<img width="375" alt="image" src="https://user-images.githubusercontent.com/8527741/178573036-eef248f2-7fd4-4681-8f7b-4a1e8a604e3e.png">
Button does not have discernible button text that would indicate to a screen-reader what this button is for.

### What's the expected behaviour?

The axe test passes.

### Help us reproduce the bug!

Add aria-label?
https://dequeuniversity.com/rules/axe/4.4/button-name?application=AxeChrome

### Code Snippet

```typescript
// Put your code below this line.

```

hbuchel:
Hi @jssuttles! Thanks for opening this issue. We are indeed missing a mechanism for adding an accessible label to the Alert primitives. We have similar work noted for the React package https://github.com/aws-amplify/amplify-ui/issues/2125 to fix this issue there, but will make sure a similar fix is applied for the Vue/Angular packages.

ErikCH:
Hi @jssuttles ! I fixed this in Angular and Vue and if you use the latest version it should have the correct accessible label. If you have any other problems, please let me know, or you can reopen this issue.

jssuttles:
Would it be possible to get another v1 build for Vue 2?

ErikCH:
Hi! 

Right now we are only updating V1 release for security issues. And since Vue 2 is in maintenance mode we're recommending our customers to upgrade to Vue 3, if they want to use the latest Authenticator component. Sorry about that!

PR #2339, opened 2022-07-25
fix(Angular + Vue): Add accessible label to Alert dismiss button Authenticator

ErikCH:
#### Description of changes
The alert dismiss button is missing the accessible label for Vue and React Authenticator. This fix will add this text, and also make it translatable, so it can be updated in the future.

#### Issue #, if available

#2284

Validated by manual testing, and Vue test.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/angular/projects/ui-angular/src/lib/primitives/error/error.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/primitives/error/error.component.html
@@ -20,7 +20,8 @@
   </div>
   <button
     amplify-button
-    class="amplify-field-group__control"
+    class="amplify-field-group__control amplify-alert__dismiss"
+    [ariaLabel]="dismissAriaLabel"
     variation="link"
     [fullWidth]="false"
     (click)="close()"
--- a/packages/angular/projects/ui-angular/src/lib/primitives/error/error.component.html
+++ b/packages/angular/projects/ui-angular/src/lib/primitives/error/error.component.html
@@ -21,7 +21,7 @@
   <button
     amplify-button
     class="amplify-field-group__control amplify-alert__dismiss"
-    [ariaLabel]="dismissAriaLabel"
+    [attr.aria-label]="dismissAriaLabel"
     variation="link"
     [fullWidth]="false"
     (click)="close()"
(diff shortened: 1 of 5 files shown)


packages/angular/projects/ui-angular/src/lib/primitives/error/error.component.html
```diff
@@ -20,7 +20,8 @@
   </div>
   <button
     amplify-button
-    class="amplify-field-group__control"
+    class="amplify-field-group__control amplify-alert__dismiss"
```
ErikCH, line 23: Also added in the `amplify-alert__dismiss` class, that was missing

hbuchel (dismissed):
LGTM. I ran the angular/vue example apps in a SR to test. There are a couple minor accessibility related tweaks we can make that I found while testing but they are outside of this change (and not introduced by it). Will create a task to track and we can follow up in a different PR.

packages/vue/src/components/primitives/__tests__/base-alert.spec.ts
```diff
@@ -0,0 +1,43 @@
+import BaseAlert from '../base-alert.vue';
+import { components } from '../../../../global-spec';
+import { render } from '@testing-library/vue';
+
+import { translations } from '@aws-amplify/ui-vue';
```
wlee221, line 5: Unused export here 🙏

wlee221 (commented):
One random nit, but actual pr lgtm!

packages/vue/src/components/primitives/__tests__/base-alert.spec.ts
wlee221: Oops, I meant unused import. 
```suggestion
```

ErikCH: Oops! I'll get rid of that!

ErikCH:
> LGTM. I ran the angular/vue example apps in a SR to test. There are a couple minor accessibility related tweaks we can make that I found while testing but they are outside of this change (and not introduced by it). Will create a task to track and we can follow up in a different PR.

Sounds good!

wlee221 (approved):

hbuchel (approved):

wlee221 (approved):

PR #2340, opened 2022-07-25
feat(react): add useMessageProps hook for In-App Messaging

cshfang:
This PR adds a hook for use with In-App Messaging components (to be added in future PRs)

Tested with sample app and `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/InAppMessaging/hooks/useMessageProps/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+import { MessageComponentStyles, MessageStyleParams } from './types';
+
+/**
+ * Utility for extracting message payload style
+ *
+ * @param props message props
+ *
+ * @returns message payload-specific styles
+ */
+
```
reesscot, line 15: nit: extra space
```suggestion
```

packages/react/src/components/InAppMessaging/hooks/useMessageProps/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    primaryButton: { backgroundColor: 'seagreen', color: 'black' },
+    secondaryButton: { backgroundColor: 'sienna', color: 'orchid' },
+  };
+
+  it('returns the expected output in the happy path', () => {
+    const output = getMessageStyles({
+      styleParams: { payloadStyle, overrideStyle },
+    });
+
+    expect(output).toMatchSnapshot();
```
reesscot, line 47: nit: rather than rely on comparing the snapshot to the test code and relying on humans not to just update the snapshot, what do you think of expecting it to match an `expectedStyle` object we can read right here. That would make me more confident in the test.

packages/react/src/components/InAppMessaging/hooks/useMessageProps/types.ts
```diff
@@ -0,0 +1,50 @@
+import { MessagePayloadStyle } from '@aws-amplify/ui-react-core';
+import React from 'react';
+
+// import { ImageDimensions } from '../useMessageImage';
+
+/**
+ * Resolved message types
+ */
+export interface MessageComponentStyles {
```
reesscot, line 9: These are the styles provided from pinpoint, right? Should we add a comment describing why these styles are the only ones that can be overwritten?

packages/react/src/components/InAppMessaging/hooks/useMessageProps/types.ts
```diff
@@ -0,0 +1,50 @@
+import { MessagePayloadStyle } from '@aws-amplify/ui-react-core';
+import React from 'react';
+
+// import { ImageDimensions } from '../useMessageImage';
```
reesscot, line 4: nit: unused import

packages/react/src/components/InAppMessaging/hooks/useMessageProps/types.ts
```diff
@@ -0,0 +1,50 @@
...
+export interface UseMessageProps {
+  hasButtons: boolean;
+  hasPrimaryButton: boolean;
+  hasRenderableImage: boolean;
+  hasSecondaryButton: boolean;
+  shouldRenderMessage: boolean;
+  styles: MessageComponentStyles | null;
+}
+
+export interface MessageStyleParams {
```
reesscot, line 28: Is this extra interface necessary, or could `getMessageStyles` just take a single argument of type `StyleParams`? I guess the question is, are you expecting to receive add more properties here?

reesscot (approved):
LGTM, just had a few miscellaneous comments

packages/react/src/components/InAppMessaging/hooks/useMessageProps/types.ts
cshfang: These aren't all coming from Pinpoint (e.g. image and close button styles aren't coming from payload). I think they represent (almost?) everything that is displayed in a message component, however, so we are actually allowing a lot of overriding via styles. Maybe the comment can be shored up (admittedly, it was copied from the React Native version of this hook)

packages/react/src/components/InAppMessaging/hooks/useMessageProps/__tests__/utils.spec.ts
cshfang: I think part of why we just opted for snapshots here was it was already the case with the RN versions of these utils. Imho, we *should* lean on snapshot tests for a quick (and often large) tree output and trust that our contributors respect them and make updates only after they have validated the changes are intentional. Personally, I find that to be the exact same exercise as just updating some `expectedStyle` object but I *do* understand your concern. But I will add a task to re-evaluate this approach holistically and swap out snapshot tests all at once if the team is unconvinced 👍

packages/react/src/components/InAppMessaging/hooks/useMessageProps/types.ts
cshfang: The biggest reason I left it this way was because I wanted it to have this type maintain some parity with its React Native counterpart. But, in practice, it actually seems I've stripped the hook down to a point where this type isn't even used 😓 so I will remove it altogether!

cshfang: A large part of doing this was simply to maintain parity with the React Native counterpart. We could scrap it and just inline the `StyleParams` typing in the function arguments but it is probably cleaner and more easily extensible (should we decide we do need to add more arguments especially as we add more complex components e.g. Carousel - even though I don't anticipate it yet).

packages/react/src/components/InAppMessaging/hooks/useMessageProps/__tests__/utils.spec.ts
reesscot: That's fair.

katiegoines (approved):
lgtm. qq - should the useMessageImage commented code be included in this PR?

cshfang:
> lgtm. qq - should the useMessageImage commented code be included in this PR?

Ideally, probably not. But we know that the PR for that is in flight also and this code needs to be added anyway. There's no real reason this commit would be all that useful on its own.

calebpollman (approved):
Left some suggestions but LGTM overall

PR #2341, opened 2022-07-26
feat(react): add backdrop component for In-App Messaging

cshfang:
This PR adds a Backdrop component to be used by the In-App Messaging feature.

Tested with a local app and `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
+++ b/packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+
+import { Flex } from '../../../primitives/Flex';
+
+import { BackdropProps } from './types';
+
+export default function Backdrop({
+  children,
+  ...rest
+}: BackdropProps): JSX.Element {
+  return (
+    <Flex className="amplify-backdrop" {...rest}>
+      {children}
+    </Flex>
+  );
+}
--- a/packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
+++ b/packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
@@ -4,12 +4,9 @@ import { Flex } from '../../../primitives/Flex';
 
 import { BackdropProps } from './types';
 
-export default function Backdrop({
-  children,
-  ...rest
-}: BackdropProps): JSX.Element {
+export function Backdrop({ children, ...rest }: BackdropProps): JSX.Element {
   return (
-    <Flex className="amplify-backdrop" {...rest}>
+    <Flex className="amplify-inappmessaging-backdrop" {...rest}>
       {children}
     </Flex>
   );
--- a/packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
+++ b/packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
@@ -1,13 +1,18 @@
 import React from 'react';
 
-import { Flex } from '../../../primitives/Flex';
+import { View } from '../../../primitives/View';
 
 import { BackdropProps } from './types';
 
-export function Backdrop({ children, ...rest }: BackdropProps): JSX.Element {
+export const BACKDROP_TEST_ID = 'inappmessaging-backdrop';
+
+export function Backdrop({ onClick, ...rest }: BackdropProps): JSX.Element {
   return (
-    <Flex className="amplify-inappmessaging-backdrop" {...rest}>
-      {children}
-    </Flex>
+    <View
+      className="amplify-inappmessaging-backdrop"
+      data-testid={BACKDROP_TEST_ID}
+      onClick={onClick}
+      {...rest}
+    />
   );
 }
--- a/packages/react/src/components/InAppMessaging/Backdrop/index.ts
+++ b/packages/react/src/components/InAppMessaging/Backdrop/index.ts
@@ -1,0 +1,1 @@
+export { default as Backdrop } from './Backdrop';
--- a/packages/react/src/components/InAppMessaging/Backdrop/index.ts
+++ b/packages/react/src/components/InAppMessaging/Backdrop/index.ts
@@ -1,1 +1,1 @@
-export { default as Backdrop } from './Backdrop';
+export { Backdrop } from './Backdrop';
--- a/packages/react/src/components/InAppMessaging/Backdrop/index.ts
+++ b/packages/react/src/components/InAppMessaging/Backdrop/index.ts
@@ -1,1 +1,2 @@
 export { Backdrop } from './Backdrop';
+export { withBackdrop } from './withBackdrop';
--- a/packages/react/src/components/InAppMessaging/Backdrop/withBackdrop.tsx
+++ b/packages/react/src/components/InAppMessaging/Backdrop/withBackdrop.tsx
@@ -15,11 +15,7 @@ export function withBackdrop<Props>(
       <>
         <Backdrop {...options} />
         <Flex className="amplify-inappmessaging-backdrop-content-container">
-          <View
-            aria-modal="true"
-            role="dialog"
-            className="amplify-inappmessaging-backdrop-content"
-          >
+          <View className="amplify-inappmessaging-backdrop-content">
             <Content {...props} />
           </View>
         </Flex>
--- a/packages/react/src/components/InAppMessaging/Backdrop/withBackdrop.tsx
+++ b/packages/react/src/components/InAppMessaging/Backdrop/withBackdrop.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+
+import { Flex } from '../../../primitives/Flex';
+import { View } from '../../../primitives/View';
+
+import { Backdrop } from './Backdrop';
+import { BackdropProps } from './types';
+
+export function withBackdrop<Props>(
+  Content: (props: Props) => JSX.Element,
+  options: BackdropProps = {}
+): (props: Props) => JSX.Element {
+  return function ContentWithBackdrop(props: Props) {
+    return (
+      <>
+        <Backdrop {...options} />
+        <Flex className="amplify-inappmessaging-backdrop-content-container">
+          <View
+            aria-modal="true"
+            role="dialog"
+            className="amplify-inappmessaging-backdrop-content"
+          >
+            <Content {...props} />
+          </View>
+        </Flex>
+      </>
+    );
+  };
+}
--- a/packages/react/src/components/InAppMessaging/index.ts
+++ b/packages/react/src/components/InAppMessaging/index.ts
@@ -1,3 +1,4 @@
+export { Backdrop } from './Backdrop';
 export {
   InAppMessageDisplay,
   MessageComponents,
--- a/packages/react/src/components/InAppMessaging/index.ts
+++ b/packages/react/src/components/InAppMessaging/index.ts
@@ -1,4 +1,3 @@
-export { Backdrop } from './Backdrop';
 export {
   InAppMessageDisplay,
   MessageComponents,
--- a/packages/ui/src/theme/css/component/inAppMessaging/backdrop.scss
+++ b/packages/ui/src/theme/css/component/inAppMessaging/backdrop.scss
@@ -1,11 +1,19 @@
 .amplify-inappmessaging-backdrop {
-  align-items: center;
   background-color: var(--amplify-colors-overlay-50);
-  bottom: 0;
-  justify-content: center;
-  left: 0;
+  inset: 0;
   position: fixed;
-  right: 0;
-  top: 0;
   z-index: 1000;
+}
+
+.amplify-inappmessaging-backdrop-content-container {
+  align-items: center;
+  inset: 0;
+  justify-content: center;
+  pointer-events: none;
+  position: fixed;
+  z-index: 1001;
+}
+
+.amplify-inappmessaging-backdrop-content {
+  pointer-events: auto;
 }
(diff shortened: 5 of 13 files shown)


packages/ui/src/theme/tokens/components/inAppMessaging/backdrop.ts
```diff
@@ -0,0 +1,9 @@
+import { BackgroundColorValue, DesignToken } from '../../types/designToken';
+
+export interface BackdropTokens {
```
reesscot, line 3:
```suggestion
export interface InAppMessagingTokens {
```

See authenticator as example: https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/tokens/components/authenticator.ts

packages/ui/src/theme/tokens/components/inAppMessaging/backdrop.ts
```diff
@@ -0,0 +1,9 @@
...
+  backgroundColor: DesignToken<BackgroundColorValue>;
```
reesscot, line 4: What about nesting the different components inside of InAppMessagingTokens instead?
```suggestion
  backdrop: {
     backgroundColor: DesignToken<BackgroundColorValue>;
  }
```

packages/ui/src/theme/tokens/components/inAppMessaging/index.ts
```diff
@@ -0,0 +1,8 @@
+// Exporting directly as backdrop ought to eventually become a fully fledged
```
reesscot, line 1: I think it would be better to put inside the `inAppMessaging` namespace rather than pre-emptively add it as a primitive. I don't think it would be a breaking change if/when we expose it as a primitive.

packages/react/src/components/InAppMessaging/index.ts
```diff
@@ -1,3 +1,4 @@
+export { Backdrop } from './Backdrop';
```
reesscot, line 1: I don't think this should be exported out of this folder, but just made available to to use internally within the InAppMessaging UI components. By exposing it this way it may get accidentally exported as a public API before we are ready. It should be internal only for now.

packages/ui/src/theme/tokens/components/index.ts
```diff
@@ -69,6 +71,7 @@ export interface ComponentTokens {
   heading: HeadingTokens;
   icon: IconTokens;
   image: ImageTokens;
+  // inAppMessaging: InAppMessagingTokens;
```
reesscot, line 74: I'd prefer this namespace for the all the inAppMessaging tokens.

packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import { Flex } from '../../../primitives/Flex';
+
+import { BackdropProps } from './types';
+
+export default function Backdrop({
```
reesscot, line 7: nit: just make this a named export

packages/ui/src/theme/css/component/inAppMessaging/backdrop.scss
```diff
@@ -0,0 +1,11 @@
+.amplify-backdrop {
+  align-items: center;
```
reesscot, line 2: Might customers want to customize some of these other properties besides background-color? I don't have enough context to make a conclusion either way.

packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
```diff
@@ -0,0 +1,16 @@
...
+  children,
+  ...rest
+}: BackdropProps): JSX.Element {
+  return (
+    <Flex className="amplify-backdrop" {...rest}>
```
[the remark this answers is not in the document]
cshfang: Hey @reesscot, the reason I did it this way was because
1. It is currently not a primitive, so I didn't want to add it to the primitives enum
2. I believe that this component has every potential to become a primitive (with much more robust features leveraging `backdrop-filter`)

It's a weird place for this to be in, admittedly, so given this rationale, does my proposed approach make more sense? I am open to suggestions on how to approach this as there certainly is not a precedence for doing this right now in the repo (i.e. Add a component scoped to a composed component with the *intent* of primitive-izing it later)

packages/react/src/components/InAppMessaging/Backdrop/index.ts
```diff
@@ -0,0 +1 @@
+export { default as Backdrop } from './Backdrop';
```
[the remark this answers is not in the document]
cshfang: No reason at all. I was just trying to maintain the conventions of the `InAppMessageDisplay` component that was already in here. But I can update

packages/ui/src/theme/tokens/components/index.ts
```diff
@@ -51,6 +52,7 @@ import {
 export interface ComponentTokens {
   alert: AlertTokens;
   authenticator: AuthenticatorTokens;
+  backdrop: BackdropTokens;
```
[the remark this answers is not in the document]
cshfang: I think so too, but for reasons I mentioned above, I felt doing this would present the least amount of friction if Backdrop were to become a primitive. But I'm open to refactoring this and scoping it solely to InApp if that is what you guys think is best.

packages/ui/src/theme/css/component/inAppMessaging/backdrop.scss
cshfang: We have no real intent/requirements/guidance one way or another. I felt this was a "bare minimal" starting point and future customizations would be additive

packages/react/src/components/InAppMessaging/index.ts
cshfang: Good call will do 👍

packages/ui/src/theme/css/component/inAppMessaging/backdrop.scss
calebpollman: IMO we shouldn't be offering more than a background color override here for the time being. The use case of this component is create a full screen overlay, and until we need more from it agree with the approach @cshfang has taken here

packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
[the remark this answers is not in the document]
calebpollman: Thinking we should hesitate towards adding anything new to `ComponentClassName` for the time being that is not a primitive. Currently, the `Authenticator` related classnames are not part of the `ComponentClassName` enum either, and we may want to keep separate enums for connected components and primitives

packages/react/src/components/InAppMessaging/Backdrop/index.ts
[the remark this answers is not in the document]
calebpollman: FWIW prefer default exports when the function matches the file name. No need to change it back

packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { Flex } from '../../../primitives/Flex';
+
+import { BackdropProps } from './types';
+
+export function Backdrop({ children, ...rest }: BackdropProps): JSX.Element {
+  return (
+    <Flex className="amplify-inappmessaging-backdrop" {...rest}>
```
calebpollman, line 9: We can discuss offline and update before release if we need to, but thinking that maybe it would be cool not to have the feature name be a part of the `className`

calebpollman (approved):
LGTM 🚢  Thanks @cshfang

packages/ui/src/theme/css/component/inAppMessaging/backdrop.scss
```diff
@@ -0,0 +1,11 @@
+.amplify-inappmessaging-backdrop {
+  align-items: center;
+  background-color: var(--amplify-colors-overlay-50);
```
hbuchel, line 3: Can we create a backdrop token for background-color and use that here? (which would reference --amplify-colors-overlay-50)

packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
hbuchel, line 9: Just some general questions:

What are the expected children of Backdrop? Just wondering what is going to be flexed.

Additionally, I'm not sure the exact usage of Backdrop (besides creating the overlay background) so these questions might be more clear when it's put together and ready to review:
- Are the rest of the InAppMessaging components (essentially the modal content) going to be nested inside Backdrop?
- Is there any click-off functionality (click outside to close) the InAppMessage? I'm asking this because a common pattern I've seen is to add a click functionality to the backdrop element, in which case you probably wouldn't want it to have children. i.e. the markup would generally be:
```
<div class="modal-wrapper">
  <div class="backdrop"></div>
  <div class="modal">here is all the modal content with the white background and buttons, and probably modal-wrapper has some flex properties to position me</div>
</div>
```
That would set this up to be re-usable in the future for more general modal functionality, but not necessarily required for this component.

hbuchel: Also, if this is the element that wraps the entire component that is overlayed onto the UI (and the rest of the component is nested inside) there are some accessibility requirements missing. Essentially we are mimicking a dialog pattern https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/dialog.html Happy to review the implementation details of that!

packages/ui/src/theme/css/component/inAppMessaging/backdrop.scss
```diff
@@ -0,0 +1,11 @@
...
+  bottom: 0;
+  justify-content: center;
+  left: 0;
+  position: fixed;
+  right: 0;
+  top: 0;
+  z-index: 1000;
```
hbuchel, line 10: I think I'd like to see z-index as token as well. Just thinking that 1000 is a common number people use to "win" the z-index wars, which means it could potentially conflict more often.

packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
cshfang: > What are the expected children of Backdrop? Just wondering what is going to be flexed.

> Are the rest of the InAppMessaging components (essentially the modal content) going to be nested inside Backdrop?

The expected children of the Backdrop is essentially modal content. Currently, the plan for executing on a messaging component spec which
* has a backdrop when screen size is non-mobile
* fills the screen when screen size is mobile
```tsx
// non-mobile
<Backdrop>
  <Message />
</Backdrop>


// mobile
<Message />
```

> Is there any click-off functionality (click outside to close) the InAppMessage? I'm asking this because a common pattern I've seen is to add a click functionality to the backdrop element, in which case you probably wouldn't want it to have children.

For this feature, there won't be click-off functionality added to the backdrop. This was something I considered, but the I think the specs call for explicitly using the close button to close the message. For now, the backdrop is sort of considered part of the message - but you raise a good point that this would probably not be as useful for a standalone Backdrop component.

I *think* that, given the fact that we are now narrowly scoping this component to the In-App Messaging feature, it is convenient that the Backdrop also acts as a container  so the message component itself can just try to fill its available space but I am open to changing this. @calebpollman thoughts?

> Also, if this is the element that wraps the entire component that is overlayed onto the UI (and the rest of the component is nested inside) there are some accessibility requirements missing.

Thanks for this callout - it seems like much of what is missing applies to the dialog itself which would be the content. Would you suggest that the backdrop automatically applies some of these missing `aria` properties to its children?

calebpollman (approved):
Left a non-blocking question but LGTM 🚢

packages/react/src/components/InAppMessaging/Backdrop/withBackdrop.tsx
```diff
@@ -0,0 +1,29 @@
...
+  options: BackdropProps = {}
+): (props: Props) => JSX.Element {
+  return function ContentWithBackdrop(props: Props) {
+    return (
+      <>
+        <Backdrop {...options} />
+        <Flex className="amplify-inappmessaging-backdrop-content-container">
+          <View
+            aria-modal="true"
+            role="dialog"
```
hbuchel, line 20: Are all in app messages going to use withBackdrop? If not, these should live on whatever is essentially passed as <Content />. For mobile-ey sizes, I think I saw the intent is to have a full screen view of the message. That would still match this aria pattern since it is content overlaid on top of the UI.

packages/react/src/components/InAppMessaging/Backdrop/Backdrop.tsx
```diff
@@ -1,13 +1,18 @@
...
-export function Backdrop({ children, ...rest }: BackdropProps): JSX.Element {
+export const BACKDROP_TEST_ID = 'inappmessaging-backdrop';
+
+export function Backdrop({ onClick, ...rest }: BackdropProps): JSX.Element {
   return (
-    <Flex className="amplify-inappmessaging-backdrop" {...rest}>
-      {children}
-    </Flex>
+    <View
+      className="amplify-inappmessaging-backdrop"
```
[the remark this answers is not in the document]
cshfang: Given that it's internal at this time, I think minimizing props is probably better for now

packages/react/src/components/InAppMessaging/Backdrop/withBackdrop.tsx
cshfang: Not all of them will be using it, but the ones that don't wouldn't fit the dialog/modal pattern anyway. But I can see your point about the pattern still being applicable to full screen non-backdropped versions of this as well. I'll remove this and move it into a prop for the shareable layout.

hbuchel (approved):
LGTM, thanks for the updates!

calebpollman (approved):

PR #2342, opened 2022-07-26
feat(react): add close button component for In-App Messaging

cshfang:
This PR adds a CloseIconButton component to be used by the In-App Messaging feature.

Tested with a local app and yarn test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/InAppMessaging/CloseIconButton/CloseIconButton.tsx
+++ b/packages/react/src/components/InAppMessaging/CloseIconButton/CloseIconButton.tsx
@@ -1,19 +1,27 @@
 import * as React from 'react';
 
-import { Link } from '../../../primitives';
+import { Button } from '../../../primitives';
 import { IconClose } from '../../../primitives/Icon/internal';
 
 import { CloseIconButtonProps } from './types';
 
 export function CloseIconButton({
   className,
+  dismissButtonLabel = 'Dismiss message',
   onClick,
   style,
   ...rest
 }: CloseIconButtonProps): JSX.Element {
   return (
-    <Link className={className} onClick={onClick} style={style} {...rest}>
-      <IconClose size="1.5rem" />
-    </Link>
+    <Button
+      ariaLabel={dismissButtonLabel}
+      className={className}
+      onClick={onClick}
+      style={style}
+      variation="link"
+      {...rest}
+    >
+      <IconClose aria-hidden="true" size="1.5rem" />
+    </Button>
   );
 }
--- a/packages/react/src/components/InAppMessaging/CloseIconButton/CloseIconButton.tsx
+++ b/packages/react/src/components/InAppMessaging/CloseIconButton/CloseIconButton.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react';
+
+import { Link } from '../../../primitives';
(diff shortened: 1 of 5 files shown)


packages/react/src/components/InAppMessaging/CloseIconButton/CloseIconButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { CloseIconButtonProps } from './types';
+
+export function CloseIconButton({
+  className,
+  onClick,
+  style,
+  ...rest
+}: CloseIconButtonProps): JSX.Element {
+  return (
+    <Link className={className} onClick={onClick} style={style} {...rest}>
```
hbuchel, line 15: Can we use a Button primitive here? Additionally, can we:

- add a string prop `dismissButtonLabel`
- add it as an aria-label attribute to this Button
- maybe default the prop to 'Dismiss message'? 

This way the button will have an accessible label. Our Alert primitive uses [a similar pattern](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/Alert/Alert.tsx#L19-L60)

packages/react/src/components/InAppMessaging/CloseIconButton/types.ts
```diff
@@ -0,0 +1,8 @@
+import { LinkProps } from '../../../primitives';
```
hbuchel, line 1: See note about preferring Button primitive.

packages/react/src/components/InAppMessaging/CloseIconButton/CloseIconButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+      <IconClose size="1.5rem" />
```
hbuchel, line 16: We can add an `aria-hidden="true"` here assuming we add the button label noted in the comment above.  Otherwise, screenreaders will read out redundant text; VoiceOver for instance adds to word "group" to the button name if it has an unlabeled image inside.

packages/react/src/components/InAppMessaging/CloseIconButton/types.ts
hbuchel: @cshfang Should this be: 
```suggestion
import { ButtonProps } from '../../../primitives';
```
now that we're using a Button?

packages/react/src/components/InAppMessaging/CloseIconButton/types.ts
```diff
@@ -0,0 +1,15 @@
...
+   * @description
+   * Configures the accessible label for the close iconButton
+   */
+  dismissButtonLabel?: string;
+
+  /**
+   * @description
+   * Function called when close iconButton is clicked
+   */
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
```
hbuchel, line 14: I think if we are extending ButtonProps we no longer need to declare this.

cshfang: Yep, missed updating from Link to Button props

calebpollman (approved):
LGTM 🚢

hbuchel (approved):

PR #2343, opened 2022-07-26
fix: Update Authenticator dictionary with key and value for 'Dismiss alert'

ErikCH:
#### Description of changes
The dictionary for Authenticator was missing the 'Dismiss alert' key and value

We already have a test on this in Vue

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/ui/src/i18n/dictionaries/authenticator/en.ts
```diff
@@ -14,6 +14,7 @@ export const enDict = {
   Confirming: 'Confirming',
   'Create Account': 'Create Account',
   'Creating Account': 'Creating Account',
+  'Dismiss alert': 'Dismiss alert',
```
calebpollman, line 17: Do we need to add key/values for our other dictionaries (example: _i18n/dictionaries/authenticator/id.ts_) to make sure we are covering international customers?

ErikCH: It would be nice, but we don't have the translations for other languages yet. We could add them in a future PR. If you look at the translations now, there is no parity between the translations we offer from one language to another.

ErikCH: Perhaps I can look into translations of those other texts, in this PR.

joebuono: I'll request that Totoro translate this text

joebuono (dismissed):

jacoblogan (dismissed):

zchenwei (approved):
LGTM!

joebuono (approved):

zchenwei (approved):

PR #2353, opened 2022-07-27
ci: remove `--no-lockfile` flag in e2e test

wlee221:
#### Description of changes

Previously in #1174, we added `--no-lockfile` to `yarn install` commands to test against latest dependency. 

But this has been problematic because:
- this creates discrepancy between local dev environment and the ci/cd environment, 
- ci/cd has been breaking to different hoisting structures ([example](https://github.com/aws-amplify/amplify-ui/runs/7548985338?check_suite_focus=true))

#### Minor improvement 
Also updates `reusable-e2e` test to use `install-with-retries` action that I created previously.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/.github/workflows/reusable-e2e.yml
+++ b/.github/workflows/reusable-e2e.yml
@@ -62,7 +62,7 @@ jobs:
         uses: ./.github/actions/install-with-retries
         with:
           # only install binary if cypress cache missed
-          no-lockfile: true
+          no-lockfile: false
           skip-cypress-binary: ${{ steps.restore-cypress-cache.outputs.cache-hit }}
       - name: Build ui package
         run: yarn ui build
@@ -132,7 +132,7 @@ jobs:
 
       - name: Install packages
         if: steps.restore-cache.outputs.cache-hit != 'true'
-        run: yarn --no-lockfile
+        run: yarn
 
       - name: Build ui package
         if: ${{ steps.restore-ui-cache.outputs.cache-hit != 'true' && matrix.package != 'ui' }}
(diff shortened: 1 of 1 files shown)


calebpollman (approved):
🛳

.github/workflows/reusable-e2e.yml
```diff
@@ -132,7 +131,7 @@ jobs:
 
       - name: Install packages
         if: steps.restore-cache.outputs.cache-hit != 'true'
-        run: yarn --no-lockfile
+        uses: ./.github/actions/install-with-retries
```
wlee221, line 134: Using the reusable action i created in #2121. It doesn't use `--no-lockfile` by default.

0618 (approved):
🤞

.github/workflows/reusable-e2e.yml
```diff
@@ -239,7 +238,7 @@ jobs:
 
       - name: Install packages
         if: steps.restore-cache.outputs.cache-hit != 'true' || steps.restore-cypress-cache.outputs.cache-hit != 'true'
-        run: yarn --no-lockfile
+        uses: ./.github/actions/install-with-retries
```
0618, line 241: why would the retries help here?

jacoblogan (approved):
LGTM Thanks!

.github/workflows/reusable-e2e.yml
wlee221: It won't help our specific problem, but I just added it here as a minor improvement.

wlee221:
Going to force merge this, as this needs to be merged first before it takes effect.

PR #2354, opened 2022-07-27
feat(ui-react-native): add React Native Home to docs

calebpollman:
#### Description of changes
Add React Native Home page to docs

[POC Docs site](https://ui-react-native-add-react-native-docs-home.d2wnvax2ctai8z.amplifyapp.com/)

#### Issue #, if available
NA

Manually tested
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated


docs/src/components/home/sections/HeroSection.tsx
```diff
@@ -22,6 +22,15 @@ import { FRAMEWORKS } from '@/data/frameworks';
 import { FrameworkLogo } from '@/components/Logo';
 import { TerminalCommand } from '@/components/InstallScripts';
 
+const INSTALL_SCRIPTS = {
```
calebpollman, line 25: Moved this outside the scope of `HeroSection` to keep a static reference

docs/src/components/home/sections/HeroSection.tsx
```diff
@@ -22,6 +22,15 @@ import { FRAMEWORKS } from '@/data/frameworks';
...
+  react: `npm i @aws-amplify/ui-react aws-amplify`,
+  vue: `npm i @aws-amplify/ui-vue aws-amplify`,
+  angular: `npm i @aws-amplify/ui-angular aws-amplify`,
+  flutter: 'flutter pub add amplify_authenticator',
+  // add other deps here
+  'react native': `npm i @aws-amplify/ui-react-native aws-amplify`,
```
calebpollman, line 31: Will be adding more deps here that we rely on that have native dependencies (for example `react-native-safe-area-context`)

docs/src/utils/isDev.ts
```diff
@@ -0,0 +1 @@
+export const isDev = process.env.NODE_ENV === 'development';
```
0618, line 1: Since we are testing it as a feature flag, should we create a new env var, like `ff-react-native`? 


Also, we can wrap up all the code in some comment like 

```js
/* Start React-Native */ {code} /* End React-Native */
```

Then, it's easier to tell which code is a part of the feature flag, in case we need to clean it up.

We may not have to, but just some A/B testing experience. Open to discuss 😁 

You might already found it, but just ref it here. Some env vars we are using https://github.com/aws-amplify/amplify-ui/blob/main/docs/.env.example

0618 (commented):
Thanks for the quick PR! Would be nice if you can make a PoC for us to check 😁

docs/src/components/Layout/FrameworkChooser.tsx
```diff
@@ -14,6 +14,17 @@ interface FrameworkLinkProps extends FrameworkChooserProps {
...
+  // if framework is more than one word, capitalize all words
+  if (framework?.includes(' ')) {
+    return framework
+      .split(' ')
+      .map((item) => capitalize(item))
+      .join(' ');
+  }
+  return capitalize(framework);
+};
+
```
ioanabrooks, line 27: Should we memoize this function via useCallback?

calebpollman: Since `getFrameworkDisplayName ` is declared outside the scope of `FrameworkChooser` its reference does not change so pretty sure we do not need to wrap in a `useCallback`, and because it always returns a string (and therefore will not trigger re-renders) _think_ we are okay with not memoizing the output with `useMemo`. Thoughts?

docs/src/components/home/sections/HeroSection.tsx
calebpollman: Ln 30 is mostly a note for me 😄

docs/src/pages/[platform]/index.reactnative.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { useBreakpointValue } from '@aws-amplify/ui-react';
+
+import { useCustomRouter } from '@/components/useCustomRouter';
+import { AmplifySection, CompatibleSection } from '@/components/home/sections';
+
+export default function ReactNativeHomePage({ colorMode }) {
+  const {
+    query: { platform = 'react' },
```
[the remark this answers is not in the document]
Comment (calebpollman) [reply docs/src/pages/[platform]/index.reactnative.tsx:9]: I was just following the convention in the other _index.\<PLATFORM\>.tsx_ files but open to changing it. What do you think @0618 ?

docs/src/components/Layout/FrameworkChooser.tsx
calebpollman: @ioanabrooks  We could also just have use an object to look up the current framework name:
```
const FRAME_WORK_DISPLAY_NAMES = {
  react: 'React',
  'react native': 'React Native',
  flutter: 'Flutter',
  // ...other values
}
```

Not sure if we really need the complexity of the function I added when we know exactly what the `framework` values are

docs/src/pages/[platform]/index.reactnative.tsx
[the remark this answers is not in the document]
Comment (0618) [reply docs/src/pages/[platform]/index.reactnative.tsx:9]: If it's a "react-native" specific component,  we don't have to set the default value as `react`

0618:
Did you get a chance to test it in a PoC and see how it works for both "dev" and "prod"?

docs/src/pages/[platform]/index.reactnative.tsx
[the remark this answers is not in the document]
Comment (calebpollman) [reply docs/src/pages/[platform]/index.reactnative.tsx:9]: Should we update the rest of the "Home" page components? [Example](https://github.com/aws-amplify/amplify-ui/blob/326e258d058db247a14f190e37083b635170035d/docs/src/pages/%5Bplatform%5D/index.vue.tsx#L15)

0618:
The preview picture command line hasn't updated?

Should it be the following ?

```sh
npm i @aws-amplify/ui-react-native aws-amplify
```

docs/src/utils/featureFlags.ts
```diff
@@ -0,0 +1 @@
+export const isReactNativeEnabled = process.env.FF_REACT_NATIVE === 'TRUE';
```
0618, line 1: Nice!

docs/src/pages/[platform]/index.reactnative.tsx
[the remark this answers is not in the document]
Comment (0618) [reply docs/src/pages/[platform]/index.reactnative.tsx:9]: Yep, good catch. Ideally we should.

0618 (approved):
The preview picture need to be updated. Ideally we should use canvas to auto generate the pictures using the background image

reesscot (approved):
LGTM

PR #2356, opened 2022-07-28
fix(ui): update types for ColorModeOverride and BreakpointOverride to allow strings

hbuchel:
#### Description of changes

Add `| (string & {})` to ColorModeOverride and BreakpointOverride.

**Context**
Previously, given the following example where a Theme object is created, with overrides for breakpoint and colorMode, without casting the Theme type to it, Typescript would complain (see below). This lets TS chill; you still need to cast as Theme (`const  theme: Theme`) if you want typing suggestions.

Screen Shot 2022-07-28 at 11 27 07 AM

```
import { ThemeProvider, Tabs, TabItem } from '@aws-amplify/ui-react';

const theme = {
  name: 'tabs-theme',
  overrides: [
    {
      colorMode: 'dark',
      tokens: {
        colors: {
          border: {
            secondary: { value: 'pink'}
          }
        }
      }
    },
    {
      breakpoint: 'large',
      tokens: {
        space: {
          small: { value: '40px' },
          medium: { value: '80px' },
          large: { value: '120px' },
        },
      },
    },
  ]
};

  return (
    <ThemeProvider theme={theme}>
      <Tabs>
        <TabItem title="Tab 1">Tab 1 Content</TabItem>
        <TabItem title="Tab 2">Tab 2 Content</TabItem>
      </Tabs>
    </ThemeProvider>
  )
};

```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/ui/src/theme/types.ts
+++ b/packages/ui/src/theme/types.ts
@@ -39,7 +39,7 @@ interface BaseOverride {
  * ```
  */
 interface BreakpointOverride extends BaseOverride {
-  breakpoint: keyof Breakpoints['values'];
+  breakpoint: keyof Breakpoints['values'] | (string & {});
 }
 
 /**
@@ -86,7 +86,7 @@ type ColorMode = 'light' | 'dark';
  * ```
  */
 export interface ColorModeOverride extends BaseOverride {
-  colorMode: ColorMode;
+  colorMode: ColorMode | (string & {});
 }
 
 /**
--- a/.changeset/calm-books-cry.md
+++ b/.changeset/calm-books-cry.md
@@ -1,0 +1,7 @@
+---
+"@aws-amplify/ui": patch
+---
+
+fix(ui): update types for ColorModeOverride and BreakpointOverride to allow generic strings
+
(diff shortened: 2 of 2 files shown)


joebuono:
My only concern here is that the `breakpoint` and `colorMode` types become too broad by adding `| (string & {})`

If a project is using TypeScript, perhaps we should expect the customer to cast the theme object as a `Theme`? Intellisense hints are another benefit of casting to `Theme`.

packages/ui/src/theme/types.ts
```diff
@@ -39,7 +39,7 @@ interface BaseOverride {
  * ```
  */
 interface BreakpointOverride extends BaseOverride {
-  breakpoint: keyof Breakpoints['values'];
+  breakpoint: keyof Breakpoints['values'] | (string & {});
```
calebpollman, line 42: Forgive me if i do not fully understand the need for this change, but was hoping you could clarify a couple things for me:
1. Doesn't `(string & {})` always enforce just a `string` making the `{}` unnecessary?
image

2. Is there a specific reason (that I am missing) for wanting to remove type strictness from `breakpoint`? It seems like allowing any `string` value for `breakpoint` kind of defeats the purpose of type safety?

PR #2363, opened 2022-07-28
fix(ui-react): move role to parent element of Alert, update docs

hbuchel:
#### Description of changes

This moves the `role` attribute to the parent element of the Alert component. This means you can override the role, if needed:

```
<Alert role="none">This is a presentation alert.</Alert>
```

**Context:**  We default the Alert component to the ARIA alert role, however there are big accessibility implications for this that do not always match the use case of the component. For example on the docs site, we often use it as a visual callout to show additional information, but not as one that requires the user's immediate attention as the ARIA alert role implies.

This change allows users to override the role attribute if needed to role="none" while not breaking the current use of Alert for customers who might currently rely on it actually being an ARIA alert.

This PR also:
- Adds two tests for the alert role existing by default, and one for overriding it as 'none'.
- Fixes the Alert icon section that references  passing an `iconSize` prop; this does not exist on the Alert component so I removed it and left the example for `hasIcon`
- Updates the example in the Accessibility section to show a dynamic Alert which is more fitting for using the default `role="alert"` and one for overriding the role. Also removed the part about it NOT using the alert role.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


docs/src/pages/[platform]/components/alert/react.mdx
```diff
@@ -75,7 +79,7 @@ Use the `heading` prop to pass a heading to the Alert.
 
 ### Icon
 
-Use the `hasIcon` prop to change whether the Alert includes an icon. Defaults to `true` (includes icon). Use the `iconSize` prop to change the size of the icon. Available options include `small`, (default), and `large`.
```
Comment (zchenwei) [inline docs/src/pages/[platform]/components/alert/react.mdx:78]: 👍

docs/src/pages/[platform]/components/alert/react.mdx
```diff
@@ -276,23 +280,44 @@ _Using style props:_
...
+The Alert component in Amplify UI has the `alert` role by default.  The `alert` role is an assertive live region which means any changes to the content of the Alert or adding the Alert dynamically to the DOM will cause the Alert to be announced by a screen reader. **This can be disruptive** to screen reader users, so it is best used sparingly and only when the Alert content requires the user's immediate attention.
 
-The most important thing to know about the `alert` role is that it is for dynamic content. It is perfect for situations such as when a user fills out a form and JavaScript is used to add an error message - the alert would immediately read out the message. It should not be used on HTML where the user hasn't interacted with it."
+Please see the <Link isExternal={true} href="https://www.w3.org/WAI/ARIA/apg/example-index/alert/alert.html">ARIA Authoring Practices Guide for the alert role</Link> for more information and use cases.
```
Comment (zchenwei) [inline docs/src/pages/[platform]/components/alert/react.mdx:285]:
```suggestion
Please see the <Link isExternal={true} href="https://www.w3.org/WAI/ARIA/apg/example-index/alert/alert.html">ARIA Authoring Practices Guide for the `alert` role</Link> for more information and use cases.
```

packages/react/src/primitives/Alert/__tests__/Alert.test.tsx
```diff
@@ -146,6 +146,24 @@ describe('Alert: ', () => {
     );
   });
 
+  it('renders as an aria alert', async () => {
+    render(<Alert testId="ariaAlertID">Alert with an aria role</Alert>);
+
+    const alert = await screen.findByTestId('ariaAlertID');
```
zchenwei, line 152:
```suggestion
    const alert = await screen.findByRole('role');
```

packages/react/src/primitives/Alert/__tests__/Alert.test.tsx
```diff
@@ -146,6 +146,24 @@ describe('Alert: ', () => {
...
+    expect(alert.getAttribute('role')).toBe('alert');
```
zchenwei, line 153:
```suggestion
    expect(alert).toBeDefined();
```

zchenwei (approved):

slaymance (approved):
Looks great!

PR #2364, opened 2022-07-28
feat(react): add in-app messaging shared layout

cshfang:
#### Description of changes

This PR adds a MessageLayout component which is intended to be shared across all In-App Messaging message types.

Note: This PR brings in a small [library](https://github.com/bgrins/TinyColor) for color detection. Since it is only available as a CJS module, Rollup requires a CJS plugin to handle it.

Tested in local app. `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
+++ b/packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
@@ -11,17 +11,17 @@ import {
 } from '../../../primitives';
 import { CloseIconButton } from '../CloseIconButton';
 import {
-  blockClass,
-  buttonClass,
-  closeButtonClass,
+  BLOCK_CLASS,
+  BUTTON_CLASS,
+  CLOSE_BUTTON_CLASS,
+  CONTENT_CLASS,
   CONTENT_TEST_ID,
-  contentClass,
-  headerClass,
+  HEADER_CLASS,
+  IMAGE_CONTAINER_CLASS,
   IMAGE_CONTAINER_TEST_ID,
-  imageContainerClass,
   MESSAGE_LAYOUT_TEST_ID,
+  TEXT_CONTAINER_CLASS,
   TEXT_CONTAINER_TEST_ID,
-  textContainerClass,
 } from './constants';
 import { MessageLayoutProps } from './types';
 import { getButtonModifier } from './utils';
@@ -50,7 +50,7 @@ export function MessageLayout({
   const isHorizontal = orientation === 'horizontal';
   const closeButton = (
     <CloseIconButton
-      className={closeButtonClass}
+      className={CLOSE_BUTTON_CLASS}
       onClick={onClose}
       style={styles.closeIconButton}
     />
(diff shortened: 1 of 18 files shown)


hbuchel (commented):
Thanks @cshfang Left a few questions/comments to start

packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
```diff
@@ -0,0 +1,128 @@
...
+        {hasRenderableImage && (
+          <Flex
+            className={classNames(
+              imageContainerClass,
+              `${imageContainerClass}--${orientation}`
+            )}
+            data-testid={IMAGE_CONTAINER_TEST_ID}
+          >
+            <Image
+              alt="In-App Message Image"
```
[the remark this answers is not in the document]
cshfang: No there's no way to set that via dashboard

packages/ui/src/theme/css/component/inAppMessaging/messageLayout.scss
```diff
@@ -0,0 +1,88 @@
...
+    flex-direction: column;
+    gap: var(--amplify-space-xxxs);
+    justify-content: center;
+    &--horizontal {
+      flex-grow: 1;
+    }
+    &--vertical {
+      flex-grow: 0;
+    }
+    h6 {
```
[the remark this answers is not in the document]
cshfang: The Heading by default is a h6

packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
```diff
@@ -0,0 +1,128 @@
...
+        )}
+        <Flex
+          className={classNames(
+            textContainerClass,
+            `${textContainerClass}--${orientation}`
+          )}
+          data-testid={TEXT_CONTAINER_TEST_ID}
+        >
+          {header?.content && (
+            <Heading style={styles.header}>{header.content}</Heading>
```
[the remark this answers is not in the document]
cshfang: Ah. I can set a level here, I guess it defaults to 6. If setting to 2 for semantics, we would have to override the default styles for h2 to make it essentially look like how h6 does now. 🤔

packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
```diff
@@ -89,7 +90,14 @@ export function MessageLayout({
           data-testid={TEXT_CONTAINER_TEST_ID}
         >
           {header?.content && (
-            <Heading style={styles.header}>{header.content}</Heading>
+            <Heading
+              className={headerClass}
+              isTruncated
```
cshfang, line 95: Need to merge in `main` to pull in [this change](https://github.com/aws-amplify/amplify-ui/pull/2371)

packages/react/src/components/InAppMessaging/index.ts
```diff
@@ -1,4 +1,3 @@
-export { MessageLayout } from './MessageLayout';
```
cshfang: Should not have been exported externally

calebpollman (approved):
Left a non-blocking question but LGTM

packages/react/src/components/InAppMessaging/MessageLayout/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param buttonStyles button styles which should contain the background color
+ *
+ * @returns the modifier - either 'light' or 'dark'
+ */
+export const getButtonModifier = (
+  buttonStyles: MessageComponentStyles['primaryButton' | 'primaryButton']
+): MessageLayoutButtonModifier => {
+  const { backgroundColor } = buttonStyles ?? {};
+  const color = tinycolor(backgroundColor);
+  return color.isDark() ? 'dark' : 'light';
```
ioanabrooks, line 20: Could you expand a bit on why we need to derive dark/light from bg color instead of relying on `Themeprovider` 's colorMode?

calebpollman: We receive the `backgroundColor` property from the `InAppMessage` payload, which does not currently support variants. Because the user is able to set the `backgroundColor` in Pinpoint, this utility allows us to know whether or not the payload `backgroundColor` translates to a light or dark color

ioanabrooks: I see, thanks, good to know. Would be nice to include this info in a comment for future reference. :)

hbuchel:
@cshfang @calebpollman is it possible to stand up a temp prototype in the next-example with a mocked payload? It'd be nice to see the intended usage. Are there other WIP components that would make that possible (i.e. do we need to get this merged first for that to be possible?)

hbuchel (approved):
Thanks @cshfang, I set up some time to chat about questions I had. But in the meantime I think this LGTM to move forward with!

PR #2379, opened 2022-08-01
Token naming

jacoblogan:
#### Description of changes
This pull request adds _state tokens to `Link`, `Table`, and `Pagination` and updates the theme examples to use the new state tokens.  Add filter to SD config as well as create theme to filter out values marked with the `deprecatedStateToken` flag.  Adds logic to create theme to copy over values from tokens marked with the `deprecatedStateToken` flag.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/ui/sd.config.ts
```diff
@@ -32,6 +32,13 @@ StyleDictionary.extend({
       files: [
         {
           destination: 'dist/theme.css',
+          // This filter function will remove any design tokens that are marked with a deprecatedStateToken flag
```
zchenwei, line 35: I do not have the entire context why we have to deprecate those state tokens without `_` and then copy them to the right buckets. I recall the generated CSS variable names should be the same because style dictionary will ignore `_`

packages/ui/sd.config.ts
```diff
@@ -32,6 +32,13 @@ StyleDictionary.extend({
...
+          filter: (token) => {
```
zchenwei, line 36: I am bit curious what would happen if we do not filter out the old tokens with `_`. Theoretically, the generated variable names should stay the same and should be fine?

packages/ui/sd.config.ts
jacoblogan: We are deprecating the old state tokens without `_` for consistency.The css variables created are the same between the state tokens with `_` and without, but when using the actual theme object it doesn't make much sense to have just a handful of state tokens not have `_` when all the rest of them do. The reason we are copying them into the new bucket is so that this isn't a breaking change, this allows us to introduce the new tokens without removing the old ones, the old tokens will be completely removed with the next major version release.

packages/ui/sd.config.ts
jacoblogan: If we don't filter them out style dicitionary will create two entries with the same variable, it doesn't actually break anything to have them both exist because they both have the same value, but this keeps the generated css a little cleaner.

packages/ui/src/theme/utils.ts
```diff
@@ -48,3 +51,93 @@ interface NameTransformProps {
...
+  }
+  return false;
+}
+
+/**
+ * This function takes in a theme object and a path array and returns the value of the design token at the end of the path or undefined if one is not found
+ */
+export function pathToValue(theme: Theme, pathArray: string[]): DesignToken {
+  let themeValue: DesignToken | PartialObjectDeep<Tokens> = theme.tokens || {};
+  pathArray.every((pathValue) => {
```
zchenwei, line 131: Do we want to guard the value `pathArray` to make sure it will never be `undefined` or `null`?

zchenwei (approved):
One more little comment! Otherwise, LGTM!

zchenwei (dismissed):

packages/ui/src/theme/__tests__/createTheme.test.ts
```diff
@@ -73,6 +73,58 @@ describe('@aws-amplify/ui', () => {
...
+                    backgroundColor: { value: '#0099ff' },
+                  },
+                },
+              },
+            },
+          },
+        });
+        const { tokens } = newTheme;
+        expect(
+          tokens.components.pagination.button._hover.backgroundColor.value
```
joebuono, line 123: What would we expect the value for `hover` to be, without the underscore?

packages/ui/src/theme/__tests__/defaultTheme.test.ts
```diff
@@ -382,8 +382,8 @@ describe('@aws-amplify/ui', () => {
         --amplify-components-image-height: auto;
         --amplify-components-image-object-fit: initial;
         --amplify-components-image-object-position: initial;
-        --amplify-components-link-active-color: var(--amplify-colors-font-active);
         --amplify-components-link-color: var(--amplify-colors-font-interactive);
+        --amplify-components-link-active-color: var(--amplify-colors-font-active);
```
joebuono, line 386: nit: Can we keep this alphabetized?

packages/ui/src/theme/__tests__/findDeprecatedStateTokens.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      const expectedPaths = [
+        ['components', 'button', 'hover', 'color'],
+        ['components', 'button', 'focus', 'fontSize'],
+      ];
+      expect(pathObjects.length).toEqual(2);
+      expectedPaths.forEach((path) => {
+        const containsExpected = pathObjects.some((pathObj) => {
+          return (
+            pathObj.path.length === path.length &&
+            path.every((val, idx) => val === pathObj.path[idx])
```
joebuono, line 39: Seems like a lot of nested iterations (`forEach`, `some`, `every`). Perhaps there's a more efficient way to do this?

packages/ui/src/theme/createTheme.ts
```diff
@@ -74,6 +82,53 @@ export function createTheme(
   theme?: Theme,
   baseTheme: BaseTheme = defaultTheme
 ): WebTheme {
+  /*
+1. find path to deprecated values in baseTheme
+2. see if values exist in theme that match depricated paths from base
+3. determine undepricated path
+4. see if values in theme exist for undepricated paths
+5. if no value of undepricated exists then replace it with value of depricated
```
joebuono, line 90: typos: should be "deprecated"

packages/ui/src/theme/createTheme.ts
```diff
@@ -74,6 +82,53 @@ export function createTheme(
...
+      const targetPath = path.map((pathValue) => {
+        if (STATE_NAMES.includes(pathValue)) {
+          return `_${pathValue}`;
+        }
+        return pathValue;
+      });
+
+      //follow the path to determine if a value already exists within the theme for the non deprecated version
+      let themeValue = theme.tokens || {};
+      targetPath.every((pathValue) => {
```
joebuono, line 110: What's the need for using `every` here? Why not just `forEach`?

packages/ui/src/theme/createTheme.ts
```diff
@@ -74,6 +82,53 @@ export function createTheme(
...
+        if (themeValue[pathValue]) {
+          themeValue = themeValue[pathValue];
+          return true;
+        } else {
+          themeValue = undefined;
```
joebuono, line 115: Is this step necessary?

packages/ui/src/theme/createTheme.ts
```diff
@@ -108,6 +169,12 @@ export function createTheme(
     overrides = mergedTheme.overrides.map((override) => {
       const tokens = setupTokens(override.tokens);
       const customProperties = flattenProperties(tokens)
+        .filter((token) => {
+          if (token.deprecatedStateToken) {
+            return false;
+          }
+          return true;
+        })
```
joebuono, line 177: nit: Could be simplified to return `!token.deprecatedStateToken`

joebuono (commented):
Good work! Just some various nits, nothing blocking though

packages/ui/src/theme/__tests__/createTheme.test.ts
jacoblogan: Here we would expect the `hover` without the underscore to be `#ff9900`.  I'll update the test to demonstrate this.  What this test is attempting to demonstrate is that if both `hover` and `_hover` is passed into create theme then `_hover` will not be overwritten by `hover`.  The way the logic works out is that create theme will now filter out the tokens marked with `deprecatedStateToken` and so what we are doing is copying the values over from the deprecatedStateTokens to the `_` version.  So if only `hover` exists in the passed in theme it will copy the value over to `_hover`, if only `_hover` exists then it will do nothing, and what this test demonstrates is that if both `hover` and `_hover` are passed in then `_hover` is the value being used.

packages/ui/src/theme/__tests__/defaultTheme.test.ts
jacoblogan: this ordering is based on the design token ordering within the tokens files, we don't seem to have a hard and fast rule yet, but I would prefer to keep those key/value pairs grouped by base tokens and then states
```
{
    color: { value: '{colors.font.interactive.value}' },
   _active: {
    color: { value: '{colors.font.active.value}' },
  },
  _focus: {
    color: { value: '{colors.font.focus.value}' },
  },
  _hover: {
    color: { value: '{colors.font.hover.value}' },
  },
}
```
to alphabetize this list we would either need to rearrange the key/value pairs within the tokens files or sort the output after creating all the variables.  Personally I'd like to just keep it the way it is grouped logically even if that means the results aren't alphabetized, but what does everyone think?

packages/ui/src/theme/__tests__/findDeprecatedStateTokens.test.ts
jacoblogan: The reason I'm doing it like this is so that the test is just reliant on the output and not the ordering of it.  And since this is just a test file the data set is small and know so I think the nested iteration is acceptable, but this could also be updated to just compare the first result value with the first expected value and skip the nested iteration but then if for whatever reason the ordering of the results every changed the test would fail even though the current ordering isn't a requirement for the algorithm to be correct.

packages/ui/src/theme/createTheme.ts
jacoblogan: every just gives us an easy way to drop out of the iteration once the condition fails.

packages/ui/src/theme/createTheme.ts
jacoblogan: This step is here because we are mutating themeValue as we iterate through the loop and expecting themeValue to either be the path that was found or undefined at the end.  This same operation is more commonly done using recursive functions where you pass in the sub object to the next recursive function call instead of mutating a value as your crawl through the object, but in this case I felt the loop was easier to read and that overall the algorithm was easier to follow keeping this logic in place instead of breaking it out into a recursive function call, but I can switch this to a recursive function call if you want.

joebuono (approved):
Thanks for your explanations and changes!

packages/ui/src/theme/createTheme.ts
```diff
@@ -74,6 +82,53 @@ export function createTheme(
...
+      const path: string[] = element.path;
+      const deprecatedThemeValue = pathToValue(theme, path);
+
+      //use the deprecated path to generate the new path by replacing any state names with _{state}
+      const targetPath = path.map((pathValue) => {
+        if (STATE_NAMES.includes(pathValue)) {
+          return `_${pathValue}`;
+        }
+        return pathValue;
+      });
```
0618, line 106: How about...?
```suggestion
      const targetPath = path.map((pathValue) => `${STATE_NAMES.includes(pathValue) ? "_" : ""}${pathValue}`
```

packages/ui/src/theme/createTheme.ts
0618: How about using `find`? 

something like...

```
themeValue = targetPath.find(pathValue => themeValue[pathValue])
```

packages/ui/src/theme/createTheme.ts
```diff
@@ -74,6 +82,53 @@ export function createTheme(
...
+      let themeValue = theme.tokens || {};
+      targetPath.every((pathValue) => {
+        if (themeValue[pathValue]) {
+          themeValue = themeValue[pathValue];
+          return true;
+        } else {
+          themeValue = undefined;
+          return false;
+        }
+      });
```
0618, line 118:
```suggestion
themeValue = themeValue(targetPath.find(pathValue => themeValue[pathValue]))
```

packages/ui/src/theme/createTheme.ts
```diff
@@ -93,6 +148,12 @@ export function createTheme(
   let cssText =
     `[data-amplify-theme="${name}"] {\n` +
     flattenProperties(tokens)
+      .filter((token) => {
+        if (token.deprecatedStateToken) {
+          return false;
+        }
+        return true;
+      })
```
0618, line 156:
```suggestion
      .filter((token) => !token.deprecatedStateToken )
```

calebpollman (commented):
Left some feedback mostly abt cleaning up naming, syntax and code style, additionally:

- Might be messing context here, but is there a strong reason we need to add these underscores? 
- The proposed solution to renaming the keys in question is adding a lot of complexity to theme generation that will need to be tested and maintained. Is there a simpler approach we can take?

packages/ui/src/theme/utils.ts
```diff
@@ -48,3 +51,95 @@ interface NameTransformProps {
...
+
+export function findDeprecatedStateTokens(baseTheme: BaseTheme): pathObject[] {
+  const baseEntry = baseTheme?.tokens || {};
+  return traverse(['', baseEntry], []);
+}
+
+/**
+ * This function will recursively traverse the tokens object from a theme and return a path array to
+ * tokens with the deprecatedStateToken flag. This will later be used to match up against the passed in theme object
+ * and write State token values that are being passed in using the old deprecated state tokens.
```
0618, line 76: nit: shall we use JS docs format? https://jsdoc.app/about-getting-started.html

packages/ui/src/theme/utils.ts
```diff
@@ -48,3 +51,95 @@ interface NameTransformProps {
...
+  'focus',
+  'active',
+  'loading',
+  'disabled',
+  'visited',
+];
+
+export function findDeprecatedStateTokens(baseTheme: BaseTheme): pathObject[] {
+  const baseEntry = baseTheme?.tokens || {};
+  return traverse(['', baseEntry], []);
```
0618, line 70: question: what's the first element, empty string for? 

Can we do this?

```suggestion
  return traverse([baseEntry], []);
```

packages/ui/src/theme/utils.ts
```diff
@@ -48,3 +51,95 @@ interface NameTransformProps {
...
+}
+
+/**
+ * This function will recursively traverse the tokens object from a theme and return a path array to
+ * tokens with the deprecatedStateToken flag. This will later be used to match up against the passed in theme object
+ * and write State token values that are being passed in using the old deprecated state tokens.
+ */
+function traverse(baseEntry, existingPath: string[]): pathObject[] {
```
0618, line 78: nit: is this name too generic? What about `tranverseForDepracated`?

also, instead of `existingPath`, would it be more understandable to name it as `childrenPath`?

packages/ui/src/theme/tokens/types/designToken.ts
```diff
@@ -18,6 +18,11 @@ export function isDesignToken(arg: unknown): arg is WebDesignToken {
  */
 export type DesignToken<ValueType = any> = {
   value: ValueType;
+  /**
+   * The deprecatedStateToken flag will mark tokens that should be filtered out of the theme variable output
+   * This is being used to create duplicate state entries for those states that are not conforming to our current naming convention i.e. hover vs _hover
+   */
+  deprecatedStateToken?: boolean;
```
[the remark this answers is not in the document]
jacoblogan: I started off using just `deprecated`, but decided to switch it to `deprecatedStateToken` because there is specific logic attached to these so the idea is that this flag would only be used on `state` tokens and not just on a general token that has been deprecated (which we don't currently have/support).

packages/ui/src/theme/utils.ts
```diff
@@ -48,3 +51,95 @@ interface NameTransformProps {
...
+      }
+      if (retValue.length) {
+        return retValue;
+      } else {
+        return [];
+      }
+    }
+  }
+  return [];
+}
```
0618, line 115: How about simplify it to be something like this? 

```suggestion
function traverseForDeprecated(baseEntry, childrenPath: string[]): pathObject[] {
  let path = [...childrenPath];
  const [key, obj] = baseEntry;
  if (typeof obj === 'object' && !Array.isArray(obj)) {
    if (key) {
      path.push(key);
    }
    if (obj.value) {
      return obj.deprecatedStateToken ? [{ path }] : [];
    } else {
      return Object.entries(obj).flatMap((entry) => traverseForDeprecated(entry, path));
    }
  }
  return [];
}
```

I didn't test it. It's just an idea 😁

packages/ui/src/theme/utils.ts
0618, line 78: What's the type of `baseEntry`?

packages/ui/src/theme/utils.ts
```diff
@@ -48,3 +51,95 @@ interface NameTransformProps {
...
+    pathArray.every((pathValue) => {
+      if (themeValue[pathValue]) {
+        themeValue = themeValue[pathValue];
+        return true;
+      } else {
+        themeValue = undefined;
+        return false;
+      }
+    });
+  }
```
0618, line 141: same here. This might have potential to be simplified

reesscot (changes requested):
I have some concerns about the approach in this PR and haven't finished reviewing yet. Please don't merge yet.

jacoblogan:
Changing approach to this issue

Issue #2041, opened 2022-06-06
Firefox does not visually indicate disabled SelectField options


joebuono:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Create React App

### Please describe your bug.

Firefox does not apply styling to `disabled` SelectField options. In the example below, the X-Large option is disabled and not clickable, but there is no visual indicator:

![Screen Shot 2022-06-06 at 2 03 19 PM](https://user-images.githubusercontent.com/48109584/172229290-5bbe4af3-c16c-4ab2-8302-b6ce16baf3dc.png)

### What's the expected behaviour?

In Chrome, the disabled X-Large option is clearly styled. Also, there should be "not-allowed" cursor styling for disabled options. 

![Screen Shot 2022-06-06 at 2 01 40 PM](https://user-images.githubusercontent.com/48109584/172229043-2b6a95ee-a30f-49b7-8e23-e3a5013a4e93.png)

### Help us reproduce the bug!

Make a CRA app, replace App.js with this code, and open up localhost in Firefox. 

```jsx
import { SelectField } from "@aws-amplify/ui-react";
import "@aws-amplify/ui-react/styles.css";

function App() {
  return (
    <SelectField label="sizes">
      <option value="small" label="small" />
      <option value="medium" label="medium" />
      <option value="large" label="large" />
      <option value="x-large" label="x-large" disabled />
    </SelectField>
  );
}

export default App;

```

### Code Snippet

https://github.com/aws-amplify/amplify-ui/blob/687a2302774eca887475e130d8359adb09e60016/docs/src/pages/%5Bplatform%5D/getting-started/create-react-app/examples/AdvancedExampleCode.tsx#L170

### Additional information and screenshots

@reesscot I think we need something like this:
```css
.amplify-select option[disabled=""] {
  background-color: var(--amplify-components-select-option-background-color);
}
```

PR #2386, opened 2022-08-02
fix(ui): Firefox SelectField disabled option styling

joebuono:
#### Description of changes

Fixes: https://github.com/aws-amplify/amplify-ui/issues/2041

This PR applies styling to disabled SelectField options in Firefox. 

Before (X-Large should be disabled):
Screen Shot 2022-08-02 at 11 42 11 AM

After: 
Screen Shot 2022-08-02 at 11 42 36 AM

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/2041

Ran the docs locally, validated changes on both Firefox and Chrome. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


hbuchel (commented):
Can we add a disabled attribute on one of the `<option>` in the default usage example (or some other example)? Just so we have something that demonstrates the styling on the docs site.

packages/ui/src/theme/css/component/select.scss
```diff
@@ -25,14 +25,18 @@
   option {
```
hbuchel, line 25: Currently Chrome and Edge on Windows don't show the disabled color for a disabled option. Can we fix that too? Applying the same fix here *should* work (and then maybe we don't need the firefox specific fix) though we should probably do a little more rigorous testing since there are always issues setting the color property for options/select cross browser :(

hbuchel: Looks good!

Screenshots from Windows in my own Browserstack account on latest of each browser. I guess this is a data point for needing automation for this :)

**Edge Dark**
EdgeDark
**Edge Light**
EdgeLight
**FF Dark**
FF Dark
**FF Light**
FF Light
**Chrome Dark**
Chrome Dark
**Chrome Light**
Chrome  Light

hbuchel (approved):
LGTM, thanks for the updates and for fixing the other browsers!

packages/ui/src/theme/css/component/select.scss
reesscot: Thank you for testing these @hbuchel! Definitely a datapoint for needing cross-platform automated visual regression tests.

reesscot (approved):
LGTM! Thanks for fixing this @joebuono

PR #2388, opened 2022-08-02
docs: Add callout for MapView margin reset

slaymance:
#### Description of changes
When using the `<MapView>` component, the map will sometimes not take up the entire viewport of the browser and the attribution footer will be cut off. To address this, users will need to reset the default `body` CSS margins. This pull request adds an info callout to reset css `body` margins.

### Without margin reset (notice attribution lower right)
Screen Shot 2022-08-02 at 11 38 32

### With margin reset
Screen Shot 2022-08-02 at 11 36 52

#### Issue #, if available
N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/docs/src/pages/[platform]/connected-components/geo/fragments/margin-reset.react.mdx
+++ b/docs/src/pages/[platform]/connected-components/geo/fragments/margin-reset.react.mdx
@@ -1,0 +1,5 @@
+```css
+body {
+  margin: 0;
+}
+```
--- a/docs/src/pages/[platform]/connected-components/geo/react.mdx
+++ b/docs/src/pages/[platform]/connected-components/geo/react.mdx
@@ -1,6 +1,6 @@
 import { Example } from '@/components/Example';
 import { Fragment } from '@/components/Fragment';
-import { Tabs, TabItem } from '@aws-amplify/ui-react';
+import { Alert, Tabs, TabItem } from '@aws-amplify/ui-react';
 
 Amplify UI Geo is powered by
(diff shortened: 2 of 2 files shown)


hbuchel:
I feel like this should be part of the [default ui package styles](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/css/styles.scss#L24). Do we know if there is context  for why it wasn't originally included?

slaymance:
> I feel like this should be part of the [default ui package styles](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/css/styles.scss#L24). Do we know if there is context for why it wasn't originally included?

The idea was that doing default margin resets was overly prescriptive for our UI library and that it's something better handled by application developers. What do you think @hbuchel?

thaddmt (approved):

joebuono (approved):
LGTM

zchenwei (approved):

reesscot:
> > I feel like this should be part of the [default ui package styles](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/theme/css/styles.scss#L24). Do we know if there is context for why it wasn't originally included?
> 
> The idea was that doing default margin resets was overly prescriptive for our UI library and that it's something better handled by application developers. What do you think @hbuchel?

+1, we felt like changing overall body defaults was too far since customers may be using it with other UI libraries.

reesscot (approved):

PR #2391, opened 2022-08-03
fix: exclude tracker from non prod environments

0618:
#### Description of changes

Exclude  tracker from non prod environments to prevent metrics pollution.

Also fixes an issue where the page visits are being counted again whenever the color mode switches.

Checklist

| Prod  | Dev  |
|---|---|
|   Screen Shot 2022-08-03 at 12 42 53 |  Screen Shot 2022-08-03 at 13 21 06  |

Validate fix to page visits:

1. Set my local environment to think it's prod
1. Added console.log("trackpage") to trackPageVisit function
1. Noted that the trackpage no longer fires for color switcher:

CleanShot 2022-08-03 at 22 10 27

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/src/utils/getStage.ts
```diff
@@ -0,0 +1,9 @@
+export const getStage = () => {
```
wlee221, line 1: Just a tiny TS nit -- what do you think about explicitly defining return types
```suggestion
type Stage = 'dev' | 'test' | 'prod';

export const getStage = (): Stage => {
```

docs/src/pages/_app.page.tsx
```diff
@@ -38,6 +39,7 @@ if (typeof window === 'undefined') {
 }
 
 function MyApp({ Component, pageProps }) {
+  const stage = getStage();
```
reesscot, line 42: Let's have `IS_PROD_STAGE`, `IS_DEV_STAGE` constants that call `getStage === 'prod'` once, then we don't have to do string checks on 75.

docs/src/pages/_app.page.tsx
```diff
@@ -70,8 +72,10 @@ function MyApp({ Component, pageProps }) {
...
 
-  configure();
-  trackPageVisit();
+  if (stage === 'prod') {
+    configure();
+    trackPageVisit();
+  }
 
   return (
     <>
```
reesscot, line 81: Let's also limit the following scripts to prod as well:
```
<Script src="https://a0.awsstatic.com/s_code/js/3.0/awshome_s_code.js" />
<Script src="/scripts/shortbreadv2.js" />
```

docs/src/utils/getStage.ts
reesscot: +1 to typing the stages, but it should be an enum instead:
```ts
enum STAGE {
  PROD: 'prod';
  DEV: 'dev';
  LOCAL: 'local';
}
```

Also, let's use `local` for our local environments rather than `test`?

0618: The reason why I use `test` is because we may test in either local or personal account

docs/src/utils/getStage.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const getStage = () => {
+  const url = process.env.SITE_URL;
+  if (url === 'https://ui.docs.amplify.aws') {
+    return STAGE.PROD;
+  } else if (url.includes('dev.ui.docs.amplify.aws')) {
+    return STAGE.DEV;
+  }
+  return STAGE.LOCAL;
+};
```
reesscot, line 15: Can we export IS_PROD_STAGE and use that instead?
```ts
const IS_PROD_STAGE = getStage() === 'prod'
```

docs/src/pages/_app.page.tsx
```diff
@@ -102,13 +106,17 @@ function MyApp({ Component, pageProps }) {
           </main>
         </ThemeProvider>
       </div>
-      <Script src="https://a0.awsstatic.com/s_code/js/3.0/awshome_s_code.js" />
-      {/* {process.env.NODE_ENV !== 'production' ? (
+      {isProd && (
```
reesscot, line 109:
```suggestion
      {IS_PROD_STAGE && (
```

docs/src/pages/_app.page.tsx
```diff
@@ -70,8 +72,10 @@ function MyApp({ Component, pageProps }) {
     );
   }, []);
 
-  configure();
-  trackPageVisit();
+  if (isProd) {
```
reesscot, line 75:
```suggestion
  if (IS_PROD_STAGE) {
```

docs/src/utils/getStage.ts
0618: Done

docs/src/pages/_app.page.tsx
0618: Done

docs/src/pages/_app.page.tsx
0618: Done

docs/src/pages/_app.page.tsx
0618: Done

reesscot (dismissed):

docs/src/utils/getStage.ts
```diff
@@ -0,0 +1,17 @@
+enum STAGE {
+  LOCAL = 'local',
+  DEV = 'dev',
+  PROD = 'prod',
+}
+
+export const getStage = () => {
+  const url = process.env.SITE_URL;
```
jacoblogan, line 8: can we have this fallback to an empty string so that the docs don't crash locally if your environment variables aren't set up?

reesscot: alternatively we can check for existence of url before using url.includes

0618: Done

jacoblogan (approved):

ioanabrooks (approved):

Milan-Shah (approved):

reesscot (approved):

Milan-Shah (approved):

ioanabrooks (approved):

PR #2404, opened 2022-08-03
build: override external dependencies warning messages and fix circua…

zchenwei:
#### Description of changes

1. Fixed a couple of circular dependencies in ui package
```
(!) Circular dependency
src/helpers/authenticator/index.ts → src/helpers/authenticator/formFields/index.ts → src/helpers/authenticator/formFields/formFields.ts → src/helpers/authenticator/formFields/defaults.ts → src/helpers/authenticator/index.ts
```
2. Clear up external dependencies warnings, which spam console output yet useless to us.
```
src/index.ts → dist...
(!) Unresolved dependencies
....
```
The output is only this much now

Screen Shot 2022-08-03 at 3 33 24 PM

Manually test the build in an sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


jacoblogan (approved):

packages/react/rollup.config.js
```diff
@@ -14,6 +15,7 @@ const config = defineConfig([
       sourcemap: false,
     },
     plugins: [
+      externals({ include: /^@aws-amplify/ }),
```
wlee221, line 18: For my knowledge, what does this do?

zchenwei: Ideally that plugin will take care of all types of dependencies declared in `package.json`. However, we do have a few `@aws-amplify` phantom dependencies used in hook. They are neither imported right from the `aws-amplify` nor  being declared in dependencies. So we have to include them explicitly

wlee221: Got it, lgtm 👍

wlee221 (approved):

reesscot:
Looking at the rollup docs, it looks like we might be able to do this by using the `/node_modules/` regex in our config like this:

```
// rollup.config.js
import path from 'path';

export default {
  ...,
  external: [
    /^@aws-amplify/,
    /node_modules/
  ]
};
```
See: https://rollupjs.org/guide/en/#core-functionality

It would be really nice if we didn't need to take another dependency for this.

packages/react/package.json
```diff
@@ -114,6 +114,7 @@
     "react-router-dom": "^6.2.1",
     "rimraf": "^3.0.2",
     "rollup": "^2.70.0",
+    "rollup-plugin-node-externals": "^4.1.1",
```
reesscot, line 117: Left a comment above that it would be nice not to have to add this dep.

zchenwei: What do you mean by not adding? I am leaning towards make it explicit instead of having a phantom dep

zchenwei: That's not the right way to go. You have to introduce a new one `@rollup/plugin-node-resolve` in order to use `/node_modules/` and that one will resolve the dependency imports to relative path in build artifacts (like `require('../../node_modules')`) which will break user stuff when you are not running in a mono repo

zchenwei: One more thing I want to point out here is `/node_modules/` will cannot properly handle `tslib` which we do want to include in the bundle

reesscot (approved):
LGTM

packages/react/package.json
reesscot: Makes sense, thanks for the explanation.

PR #2416, opened 2022-08-04
fix: add tslib as hard dependency in react and ui package

zchenwei:
#### Description of changes

Add `tslib` as hard dependency in both react and ui package so that it will not be included in the bundle but instead it will get pulled down from npm when users install the react/ui package.

#### Issue #, if available

This will fix the weird CI issue where `tslib` is unable to be resolved correctly

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


wlee221 (approved):

reesscot (approved):
Thanks for fixing this!

PR #2424, opened 2022-08-05
fix(docs): Add multiline support for underline animation on cards link

ioanabrooks:
#### Description of changes

This change add multi-line support for docs card link underline animation.

| Before  |
| ------------- | 
| BEFORE  | 

| After  |
| ------------- | 
| Kapture 2022-12-09 at 09 23 34 |
| Kapture 2022-12-09 at 09 28 39 |

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated


.changeset/dry-avocados-live.md
```diff
@@ -0,0 +1,7 @@
+---
+'docs': patch
```
hbuchel, line 2: Since we don't publish the docs site we don't need need a changeset for docs updates (can remove this file).

hbuchel (commented):
Nice! Just some general comments/questions:

- Is there anything we can do about descenders? I wonder if the text had a solid text shadow (or something)?
Screen Shot 2022-08-08 at 1 32 42 PM
Screen Shot 2022-08-08 at 1 31 06 PM

- Should we speed up the animation? It feels a little sluggish for a hover affect to me, but what do you think?

And just a note that there is another variation used on the docs pages (like /react/theming/default-theme and /react/connected-components/authenticator#next-steps) if you need to check that variation also.

ioanabrooks:
> Nice! Just some general comments/questions:
> 
> * Is there anything we can do about descenders? I wonder if the text had a solid text shadow (or something)?
> 
> Screen Shot 2022-08-08 at 1 32 42 PM
> 
> Screen Shot 2022-08-08 at 1 31 06 PM
> 
> * Should we speed up the animation? It feels a little sluggish for a hover affect to me, but what do you think?
> 
> And just a note that there is another variation used on the docs pages (like /react/theming/default-theme and /react/connected-components/authenticator#next-steps) if you need to check that variation also.

* I opted to have the underline completely clear of descenders for readability [before I added the animation it was just a text decoration underline].
* The screen capture is capped at 25 frames so may look more sluggish, pls check the ui.docs site, the animation is live atm, just without multi-line support, duration matches the card slide effect.

hbuchel:
> * I opted to have the underline completely clear of descenders for readability [before I added the animation it was just a text decoration underline].

That's fair. I was thinking of how browsers (or at least chrome?) deals with it by having some spacing around (to the left and right) of the descender so it doesn't get swallowed up by the underline (especially for letters like g). Not a blocker

> * The screen capture is capped at 25 frames so may look more sluggish, pls check the ui.docs site, the animation is live atm, just without multi-line support, duration matches the card slide effect.

Yeah that is what I was referencing is from running the local version. For some reason it seems slower than what is on live?

hbuchel (approved):
Nice! Checked it out locally and it LGTM. Thanks for the update!

reesscot (approved):
LGTM

Milan-Shah (approved):
Animations 🚀  - So good!

PR #2438, opened 2022-08-09
docs: remove tab in props control panel

zchenwei:
#### Description of changes

Remove props tab in demo control panel to reduce unneeded noise for SR users and keyboard users. We can add it back in once we have a theme section

Screen Shot 2022-08-09 at 1 39 20 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/docs/src/components/Demo.tsx
+++ b/docs/src/components/Demo.tsx
@@ -1,14 +1,6 @@
 import * as React from 'react';
-import { CopyToClipboard } from 'react-copy-to-clipboard';
 import Highlight, { defaultProps } from 'prism-react-renderer';
-import {
-  Tabs,
-  TabItem,
-  Flex,
-  View,
-  Button,
-  useTheme,
-} from '@aws-amplify/ui-react';
+import { Divider, Flex, View, useTheme } from '@aws-amplify/ui-react';
 import { CopyButton } from './CopyButton';
 
 interface DemoProps {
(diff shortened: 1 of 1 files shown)


docs/src/components/Demo.tsx
```diff
@@ -50,16 +42,17 @@ export const Demo = ({
           <View overflow="auto" padding="5px">
             {children}
           </View>
+          <Divider
+            label="Props"
+            margin="20px 0 0"
+            border={`2px solid ${tokens.colors.brand.primary[80]}`}
+            style={{
+              '--amplify-components-divider-label-color':
```
zchenwei, line 50: Have to use this trick to change the label color. Maybe we should consider add a `labelColor` prop on `Divider` ? Also, there is no prop way to change the `Divider` color

docs/src/components/Demo.tsx
```diff
@@ -50,16 +42,17 @@ export const Demo = ({
           <View overflow="auto" padding="5px">
             {children}
           </View>
+          <Divider
+            label="Props"
+            margin="20px 0 0"
+            border={`2px solid ${tokens.colors.brand.primary[80]}`}
```
hbuchel, line 48: This is a nit, but IMO, I think we should leave this as a gray or neutral color (border.color.secondary maybe?); I think the brand colors draw a little too much attention. Even removing the label would be fine since this is a pretty "busy" visual area
Screen Shot 2022-08-09 at 3 29 52 PM
.

docs/src/components/Demo.tsx
hbuchel: +1 it would be great to be able to customize those separately!

docs/src/components/Demo.tsx
hbuchel: or if keeping the label, whatever gray gives enough contrast against the background, might have to manually check what that is since we use an opacity to lighten the Divider. Maybe something else for us to update if we add tokens for customizing the label color and divider color; the opacity makes it a little harder to theme just how you want it.

joebuono: I agree with @hbuchel  about leaving it as a neutral gray color and removing the label

hbuchel (approved):
Thanks @zchenwei for the updates! LGTM

ioanabrooks (approved):
LGTM

PR #2442, opened 2022-08-09
ci: Handle `in-app-messaging` PR tests separately

wlee221:
#### Summary

Sets up `in-app-messaging` PR CI tests to run the workflow on `@in-app-messaging/main`, instead of one in `main`.

This **only** affects workflows related to in-app-messaging.

#### Description of changes

`in-app-messaging/main` has updated version of CI tests because it tests for `react-native` and `react-core` in addition to preexisting packages.

Now that its CI has diverged, we want to ensure the tests against `in-app-messaging` branches uses the new one.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/.github/workflows/test-in-app-messaging-prs.yml
+++ b/.github/workflows/test-in-app-messaging-prs.yml
@@ -1,11 +1,12 @@
-# Description: This workflow runs unit + e2e tests on PRs targeting `main` and
-#              other protected branches listed below.
+# Description: This workflow runs unit + e2e tests on PRs targeting
+#   - `in-app-messaging/main`
+#   - `in-app-messaging/release`
 #
 # Triggered by:
-#   (1) Internal PRs: contirubutor pushes a commit to PRs targeting protected branches.
-#   (2) Fork PRs: maintainer adds "run-test" label to PRs targeting protected branches.
+#   (1) Internal PRs: maintainer pushes a commit to PRs targeting those branches.
+#   (2) Fork PRs: maintainer adds "run-test" label to PRs targeting those branches.
 
-name: Test / PRs
+name: Test / in-app-messaging / PRs
 
 concurrency:
   group: e2e-${{ github.event.pull_request.id }}
@@ -13,7 +14,10 @@ concurrency:
 
 on:
   pull_request_target:
-    branches: [main, ui-svelte/main, ui-geo/main, ui-react@v2, studio-release]
+    branches: [
+        in-app-messaging/main
+        in-app-messaging/release,
+      ]
     types: [opened, synchronize, labeled]
 
 permissions:
@@ -40,7 +44,7 @@ jobs:
             const label = 'run-tests';
             github.issues.removeLabel({ owner, repo, issue_number, name: label });
   test:
-    uses: aws-amplify/amplify-ui/.github/workflows/reusable-e2e.yml@main
+    uses: aws-amplify/amplify-ui/.github/workflows/reusable-e2e.yml@in-app-messaging/main
     needs: setup
     with:
       commit: ${{ github.event.pull_request.head.sha }}


.github/workflows/test-in-app-messaging-prs.yml
```diff
@@ -0,0 +1,63 @@
...
+concurrency:
+  group: e2e-${{ github.event.pull_request.id }}
+  cancel-in-progress: true
+
+on:
+  pull_request_target:
+    branches: [
+        in-app-messaging/main
+        in-app-messaging/release,
+      ]
```
wlee221, line 20: This workflow is only for those two in app branches.

.github/workflows/test-in-app-messaging-prs.yml
```diff
@@ -0,0 +1,63 @@
...
+        if: always() && github.event.label.name == 'run-tests'
+        uses: actions/github-script@0.3.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const { issue: { number: issue_number }, repo: { owner, repo } } = context;
+            const label = 'run-tests';
+            github.issues.removeLabel({ owner, repo, issue_number, name: label });
+  test:
+    uses: aws-amplify/amplify-ui/.github/workflows/reusable-e2e.yml@in-app-messaging/main
```
wlee221, line 47: This is really the only change outside copy+paste. We're changing the reference:

```diff
- uses: aws-amplify/amplify-ui/.github/workflows/reusable-e2e.yml@main
+ uses: aws-amplify/amplify-ui/.github/workflows/reusable-e2e.yml@in-app-messaging/main
```

.github/workflows/test-prs.yml
```diff
@@ -13,15 +13,7 @@ concurrency:
...
 on:
   pull_request_target:
-    branches:
-      [
-        main,
-        ui-svelte/main,
-        ui-geo/main,
-        ui-react@v2,
-        studio-release,
-        in-app-messaging/release,
```
wlee221, line 23: Removed `in-app-messaging` since it's getting handled separately above

.github/workflows/test-in-app-messaging-prs.yml
wlee221: It references this file if you're curious: https://github.com/aws-amplify/amplify-ui/blob/in-app-messaging/main/.github/workflows/reusable-e2e.yml

calebpollman (approved):

.github/workflows/test-in-app-messaging-prs.yml
```diff
@@ -0,0 +1,63 @@
+# Description: This workflow runs unit + e2e tests on PRs targeting
+#   - `in-app-messaging/main`
+#   - `in-app-messaging/release`
+#
+# Triggered by:
```
reesscot, line 5: 💯

.github/workflows/test-in-app-messaging-prs.yml
```diff
@@ -0,0 +1,63 @@
...
+  # other permissions are defaulted to "none"
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    # We run tests only if it's:
+    #   1) pull request not from a fork (ie. internal PR), or
+    #   2) pull request from a fork (ie. external PR) that was added "run-tests" label
+    if: |
+      (github.event.pull_request.head.repo.full_name == github.repository) ||
```
reesscot, line 34: nit: is there a way to put this in a function? I think we use this conditional in multiple workflows.

.github/workflows/test-in-app-messaging-prs.yml
```diff
@@ -0,0 +1,63 @@
...
+      (github.event.action == 'labeled' && github.event.label.name == 'run-tests')
+    steps:
+      - name: Remove run-tests label, if applicable
+        if: always() && github.event.label.name == 'run-tests'
```
reesscot, line 38: for my own understanding, why is the `always()` call required here? Seems like only the latter conditional would be required.

reesscot (approved):

.github/workflows/test-in-app-messaging-prs.yml
wlee221: Ooh, that's a great question. 

This is a left over logic from when there was one monolithic testing workflow for PRs, `main`, etc. 

`always() && github.event.label.name == 'run-tests'` made sure this step always runs, even if the previous steps failed. I did that to feel 100% about security.

Now that we separated `test-prs` out with reusable workflows, this is no longer necessary. I'll create a separate PR that cleans it all up! No later than EOD.

wlee221:
I'll go ahead and merge this. I do have an action item from the comment, which I'll address by EOD.

PR #2445, opened 2022-08-10
test(react): Sort the exported keys of React package for snapshot testing

slaymance:
#### Description of changes
Sometimes there are inconsistencies with the order of exported keys when performing the React package exports snapshot test. To mitigate false negatives, this pull request sorts the keys of the exports to ensure that the order of keys is irrelevant to the test.

#### Issue #, if available
N/A

`yarn react test` passes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/__tests__/exports.ts
```diff
@@ -1,6 +1,3 @@
-import fs from 'fs';
-import path from 'path';
-import type { PrimitiveCatalogType } from '../src/types/catalog';
```
slaymance, line 3: These aren't used in the test file.

packages/react/__tests__/exports.ts
```diff
@@ -12,18 +9,8 @@ import * as internal from '../dist/internal';
 describe('@aws-amplify/ui-react', () => {
   describe('exports', () => {
     it('should match snapshot', () => {
-      expect(Object.keys(exported)).toMatchInlineSnapshot(`
+      expect(Object.keys(exported).sort()).toMatchInlineSnapshot(`
```
thaddmt, line 12: Can't comment on the lines outside of this PR but just calling out that we would need to add this to the other snapshot tests below as well!

thaddmt (approved):

packages/react/__tests__/exports.ts
calebpollman: Thanks for cleaning these up

packages/react/__tests__/exports.ts
```diff
@@ -1,29 +1,16 @@
...
 import { PrimitiveCatalog } from '@aws-amplify/ui-react/internal';
 
 // Jest doesn't support `exports` maps, so we have to reference `dist` directly.
 // See: https://github.com/facebook/jest/issues/9771
-import * as exported from '../dist';
-import * as legacy from '../dist/legacy';
-import * as internal from '../dist/internal';
+import * as exported from '../src';
+import * as legacy from '../src/legacy';
+import * as internal from '../src/internal';
```
calebpollman, line 7: ✨

packages/react/__tests__/exports.ts
calebpollman: Can we create "sorted" variables for the values used in the `expect` lines to make them a bit cleaner?
```
const sortedExports = Object.keys(exported).sort();
```

calebpollman: Also, since we are cleaning up this file already, what do you think of just using `toMatchSnapshot` in place of `toMatchInlineSnapshot` so we can remove the string arguments?

slaymance: I added all of the above suggestions.

calebpollman (approved):
LGTM 🚢

reesscot (approved):
shipit

PR #2464, opened 2022-08-12
chore(ui-react-native): Add storybook

ioanabrooks:
#### Description of changes

Add Storybook support to RN example app for rapid prototyping. This change adds Storybook server as an optional way to run Storybook for the added benefit of being able to control multiple simulators/emulators with one click.

#### Issue #, if available

n/A

Tested in ios simulator and android emulator.

With Storybook server:
```
yarn run react-native-example storybook
```

```
yarn run react-native-example dev
```
Kapture 2022-08-12 at 13 55 05

Without storybook server:

| Navigator tab  | Preview tab |
| Simulator Screen Shot - iPhone 13 - 2022-08-15 at 13 47 39  | Simulator Screen Shot - iPhone 13 - 2022-08-15 at 13 47 48  |

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -4,12 +4,12 @@ This directory contains example apps for each framework we support.
 
 ## Examples Development
 
-1. Create or Update an example at [`examples/{next,vue,angular}/...`](examples)
+1. Create or Update an example at [`examples/{next,react-native,vue,angular}/...`](examples)
 
    For your `aws-exports.js`, you can [reference an existing or create a new environment](../environments/README.md).
 
-1. Run your example at monorep root: `yarn {react,vue,angular}-example dev`
+1. Run your example at monorep root: `yarn {react,react-native,vue,angular}-example dev`
 1. Visit your example (e.g. <http://localhost:3000/ui/components/authenticator/sign-up-with-username>)
-1. Make changes to [`@aws-amplify/ui-{react,vue,angular}`](packages) & save.
+1. Make changes to [`@aws-amplify/ui-{react,react-native,vue,angular}`](packages) & save.
 
    Examples should automatically hot-reload your changes in the example.
--- a/examples/react-native/README.md
+++ b/examples/react-native/README.md
@@ -86,29 +86,31 @@ yarn react-native-example dev
 
 ## Storybook
 
-First boot Storybook Server, which will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators:
+1. With Storybook Server - which will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators.
+
+First boot Storybook Server:
 
 ```bash
 # terminal 1
 yarn run react-native-example storybook
 ```
 
-From a second terminal run ios simulator, android emulator or both:
+From a second terminal run iOS simulator, Android emulator or both (requires app to be installed):
 
 ```bash
 # terminal 2
-
-# ios simulator
-yarn run react-native-example ios
-
-# android emulator
-yarn run react-native-example android
-
-# both
 yarn run react-native-example dev
 ```
 
-Refresh the Storybook web interface.
+Refresh the Storybook web interface to load all the stories.
+
+1. On simulator only
+
+```bash
+yarn run react-native-example dev
+```
+
+In the Storybook app switch between `Navigator` and `Preview` tabs to select components/stories.
 
 ## Running Integration Tests - TODO
 
--- a/examples/react-native/README.md
+++ b/examples/react-native/README.md
@@ -86,7 +86,8 @@ yarn react-native-example dev
 
 ## Storybook
 
-1. With Storybook Server - which will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators.
+### With Storybook Server
+Will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators.
 
 First boot Storybook Server:
 
--- a/examples/react-native/README.md
+++ b/examples/react-native/README.md
@@ -84,6 +84,32 @@ yarn react-native-example dev
 
 1. Open the app on the iOS simulator or Android emulator.
 
+## Storybook
+
+First boot Storybook Server, which will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators:
+
+```bash
+# terminal 1
+yarn run react-native-example storybook
+```
+
+From a second terminal run ios simulator, android emulator or both:
+
+```bash
+# terminal 2
+
+# ios simulator
+yarn run react-native-example ios
+
+# android emulator
+yarn run react-native-example android
+
+# both
+yarn run react-native-example dev
+```
+
+Refresh the Storybook web interface.
+
 ## Running Integration Tests - TODO
 
 ## Linking an Internal Dependency
--- a/examples/react-native/README.md
+++ b/examples/react-native/README.md
@@ -105,7 +105,7 @@ yarn run react-native-example dev
 
 Refresh the Storybook web interface to load all the stories.
 
-1. On simulator only
+### On iOS simulator/Android emulator only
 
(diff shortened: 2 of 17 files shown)


packages/react-native/src/index.ts
```diff
@@ -3,3 +3,4 @@ export {
   InAppMessagingProvider,
   useInAppMessaging,
 } from '@aws-amplify/ui-react-core';
+export * from './primitives';
```
calebpollman, line 6: We cannot export these directly as that exposes them publicly, but you should be able to import the primitives from 
`@aws-amplify/ui-react-native/dist/primitives` in the example app

examples/react-native/README.md
```diff
@@ -84,6 +84,32 @@ yarn react-native-example dev
...
+yarn run react-native-example storybook
+```
+
+From a second terminal run ios simulator, android emulator or both:
+
+```bash
+# terminal 2
+
+# ios simulator
+yarn run react-native-example ios
```
calebpollman, line 102: Since this command builds and installs the app on the iOS simulator, do we need to run this every time we run storybook? Same question for line 105

examples/react-native/README.md
```diff
@@ -84,6 +84,32 @@ yarn react-native-example dev
...
+
+```bash
+# terminal 1
+yarn run react-native-example storybook
+```
+
+From a second terminal run ios simulator, android emulator or both:
+
+```bash
+# terminal 2
```
calebpollman, line 99: It would be really helpful to not have to run commands in a second terminal, and run the web storybook simulator optionally. Is there an approach we can take that would allow that?

examples/react-native/README.md
```diff
@@ -84,6 +84,32 @@ yarn react-native-example dev
...
+## Storybook
+
+First boot Storybook Server, which will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators:
+
+```bash
+# terminal 1
+yarn run react-native-example storybook
+```
+
+From a second terminal run ios simulator, android emulator or both:
```
calebpollman, line 96: Nit: follow existing casing for iOS and Android

calebpollman (changes requested):
Great start! Left a blocking comment on the `ui-react-native` exports and some general feedback/questions/nits

examples/react-native/README.md
ioanabrooks: Yes running Storyboook is optional, though I discourage it, ideally we should test these in multiple environments but I will add notes.

examples/react-native/README.md
ioanabrooks: No, just the first time, after which you just have to run metro, i'll update wording.

packages/react-native/src/index.ts
ioanabrooks: Ack

examples/react-native/README.md
calebpollman: Is running the storybook without the web server and testing the components on Android and iOS different then run testing on multiple platforms with the web server? Seems like with the web server the major difference is that it controls the emulator and/or simulator, am I missing something else?

ioanabrooks: Correct, Storybook server controls all emulators/simulators connected to it, that is the key difference. The Storybook app itself has a tab menu and you can freely switch between the Navigator and Preview tab (see updated PR description) to access the stories - more clicks though :)

examples/react-native/package.json
```diff
@@ -24,20 +28,31 @@
...
     "metro-react-native-babel-preset": "^0.70.2",
+    "react-dom": "17.0.2",
+    "react-native-storybook-loader": "^2.0.5",
     "react-test-renderer": "^17.0.2",
     "typescript": "^4.6.3"
   },
   "resolutions": {
-    "@types/react": "^17"
+    "@types/react": "^17",
+    "**/@types/react": "^17"
```
wlee221, line 55: To my knowledge, resolutions don't work unless it's in the monorepo root `package.json`: https://github.com/yarnpkg/yarn/issues/5039. 

Have you found this to work?

ioanabrooks: Yes..it works :)  and if you read the issue carefully, towards the end there's a post saying "Still happens to me with the version 1.21.1 and looks to be fixed with the last version 1.22.4". We use 1.2219.

wlee221: Cool, thanks for confirming! We'll just have to make sure we enforce minimum `yarn` across the team then.

examples/react-native/README.md
```diff
@@ -84,6 +84,34 @@ yarn react-native-example dev
 
 1. Open the app on the iOS simulator or Android emulator.
 
+## Storybook
+
+1. With Storybook Server - which will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators.
```
calebpollman, line 89:
```suggestion
### With Storybook Server
Will serve a web interface that allows switching between components (stories) and controlling multiple simulators/emulators.
```

examples/README.md
```diff
@@ -4,12 +4,12 @@ This directory contains example apps for each framework we support.
...
 ## Examples Development
 
-1. Create or Update an example at [`examples/{next,vue,angular}/...`](examples)
+1. Create or Update an example at [`examples/{next,react-native,vue,angular}/...`](examples)
 
    For your `aws-exports.js`, you can [reference an existing or create a new environment](../environments/README.md).
 
-1. Run your example at monorep root: `yarn {react,vue,angular}-example dev`
+1. Run your example at monorep root: `yarn {react,react-native,vue,angular}-example dev`
 1. Visit your example (e.g. <http://localhost:3000/ui/components/authenticator/sign-up-with-username>)
```
calebpollman, line 12: Wonder if we should update this line as well since it is specific to browser frameworks

examples/react-native/README.md
```diff
@@ -84,6 +84,34 @@ yarn react-native-example dev
...
+From a second terminal run iOS simulator, Android emulator or both (requires app to be installed):
+
+```bash
+# terminal 2
+yarn run react-native-example dev
+```
+
+Refresh the Storybook web interface to load all the stories.
+
+1. On simulator only
```
calebpollman, line 107:
```suggestion
### On iOS simulator/Android emulator only
```

examples/react-native/index.js
```diff
@@ -1,5 +1,11 @@
 import { AppRegistry } from 'react-native';
 import App from './App';
+import StorybookUIRoot from './storybook';
 import { name as appName } from './app.json';
 
-AppRegistry.registerComponent(appName, () => App);
+//TODO: replace with env var
+const isStorybook = true;
```
calebpollman, line 7: Pulled down your branch and ran the following:
```
yarn clean && yarn && yarn build
yarn react-native-example ios:pod-install
yarn react-native-example ios
```

Leading to:
Simulator Screen Shot - iPhone 13 - 2022-08-16 at 11 55 21

Think we may want to default `isStorybook` to `false` until we have wired up the env vars:

```suggestion
const isStorybook = false;
```

ioanabrooks: Yeah sure, good catch. By the way regarding the error you are getting, we are using [storybook-loader](https://github.com/elderfo/react-native-storybook-loader#quick-start) to dynamically generate imports for all stories [as opposed to manually adding each story as we create them]. Running storybook will generate that file.

calebpollman: Right, we should probably also build them in the scenario of running storybook without the web server

ioanabrooks: Ack will add that in follow-up PR when wiring the env var

calebpollman (approved):
LGTM 🚢

wlee221 (approved):

PR #2466, opened 2022-08-15
Token deprecation

jacoblogan:
#### Description of changes

This PR adds in consistent state design tokens that are missing and updates the create theme function to remove the duplicate css variables being created by having the duplicate tokens.

Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/ui/src/theme/__tests__/createTheme.test.ts
+++ b/packages/ui/src/theme/__tests__/createTheme.test.ts
@@ -31,6 +31,180 @@ describe('@aws-amplify/ui', () => {
           'var(--amplify-colors-background-primary)'
         );
       });
+
+      it('should filter out deprecated pagination design tokens', () => {
+        const theme = {
+          name: 'test',
+          tokens: {
+            components: {
+              pagination: {
+                button: {
+                  _hover: {
+                    backgroundColor: {
+                      value: 'red',
+                    },
+                    color: {
+                      value: 'blue',
+                    },
+                  },
+                  _disabled: {
+                    color: {
+                      value: 'grey',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        };
+        const { cssText, tokens } = createTheme(theme);
+        expect(cssText).toContain(
+          `--amplify-components-pagination-button-hover-background-color: ${tokens.components.pagination.button._hover?.backgroundColor.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-pagination-button-hover-background-color: ${tokens.components.pagination.button.hover?.backgroundColor.value}`
+        );
+
+        expect(cssText).toContain(
+          `--amplify-components-pagination-button-hover-color: ${tokens.components.pagination.button._hover?.color.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-pagination-button-hover-color: ${tokens.components.pagination.button.hover?.color.value}`
+        );
+
+        expect(cssText).toContain(
+          `--amplify-components-pagination-button-disabled-color: ${tokens.components.pagination.button._disabled?.color.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-pagination-button-disabled-color: ${tokens.components.pagination.button.disabled?.color.value}`
+        );
+      });
+
+      it('should filter out the deprecated link tokens', () => {
+        const theme = {
+          name: 'test',
+          tokens: {
+            components: {
+              link: {
+                _active: {
+                  color: {
+                    value: 'red',
+                  },
+                },
+                _focus: {
+                  color: {
+                    value: 'white',
+                  },
+                },
+                _hover: {
+                  color: {
+                    value: 'green',
+                  },
+                },
+                _visited: {
+                  color: {
+                    value: 'blue',
+                  },
+                },
+              },
+            },
+          },
+        };
+        const { cssText, tokens } = createTheme(theme);
+        expect(cssText).toContain(
+          `--amplify-components-link-active-color: ${tokens.components.link._active?.color.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-link-active-color: ${tokens.components.link.active?.color.value}`
+        );
+
+        expect(cssText).toContain(
+          `--amplify-components-link-focus-color: ${tokens.components.link._focus?.color.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-link-focus-color: ${tokens.components.link.focus?.color.value}`
+        );
+
+        expect(cssText).toContain(
+          `--amplify-components-link-hover-color: ${tokens.components.link._hover?.color.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-link-hover-color: ${tokens.components.link.hover?.color.value}`
+        );
+
+        expect(cssText).toContain(
+          `--amplify-components-link-visited-color: ${tokens.components.link._visited?.color.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-link-visited-color: ${tokens.components.link.visited?.color.value}`
+        );
+      });
+
+      it('should filter out the deprecated switchfield tokens', () => {
+        const theme = {
+          name: 'test',
+          tokens: {
+            components: {
+              switchfield: {
+                thumb: {
+                  _checked: {
+                    transform: {
+                      value: '{transforms.slideX.small.value}',
+                    },
+                  },
+                },
+                track: {
+                  _checked: {
+                    backgroundColor: {
+                      value: 'red',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        };
+        const { cssText, tokens } = createTheme(theme);
+        expect(cssText).toContain(
+          `--amplify-components-switchfield-thumb-checked-transform: ${tokens.components.switchfield.thumb._checked?.transform.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-switchfield-thumb-checked-transform: ${tokens.components.switchfield.thumb.checked?.transform.value}`
+        );
+
+        expect(cssText).toContain(
+          `--amplify-components-switchfield-track-checked-background-color: ${tokens.components.switchfield.track._checked?.backgroundColor.value}`
+        );
+        expect(cssText).not.toContain(
+          `--amplify-components-switchfield-track-checked-background-color: ${tokens.components.switchfield.track.checked?.backgroundColor.value}`
+        );
+      });
+
+      it('should filter out the deprecated table tokens', () => {
+        const theme = {
+          name: 'test',
+          tokens: {
+            components: {
+              table: {
+                row: {
+                  _hover: {
+                    backgroundColor: {
+                      value: 'red',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        };
+        const { cssText, tokens } = createTheme(theme);
+        expect(cssText).toContain(
+          `--amplify-components-table-row-hover-background-color: ${tokens.components.table.row._hover?.backgroundColor.value}`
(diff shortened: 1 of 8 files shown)


packages/ui/src/theme/createTheme.ts
```diff
@@ -63,6 +63,29 @@ function setupTokens(obj: any, path = []) {
...
+ * @param tokens
+ * @returns WebDesignTokens[]
+ */
+function removeDeprecated(tokens: WebDesignToken[]): WebDesignToken[] {
+  let tokenList = tokens;
+  DEPRECATED_TOKENS.forEach((duplicateToken) => {
+    const filteredTokens = tokenList.filter(
+      (token) => duplicateToken.tokenName === token.name
+    );
+    if (filteredTokens.length >= 2) {
```
calebpollman, line 78: Why should this block not run if `filteredTokens` has a length of `1`?

packages/ui/src/theme/tokens/components/link.ts
```diff
@@ -9,13 +9,48 @@ interface LinkSizeTokens {
...
+   * @deprecated to be removed in the next major version please use _visited
+   */
+  visited?: LinkStateTokens;
+  _active?: LinkStateTokens;
+  _focus?: LinkStateTokens;
+  _hover?: LinkStateTokens;
+  _visited?: LinkStateTokens;
 }
 
+export const DEPRECATED_LINK_TOKENS = [
```
calebpollman, line 35: Can we type this as `WebDesignToken[]` so we can ensure the shape of the deprecated tokens follows the standard tokens?

calebpollman (commented):
Did a short review, and left some questions. Let me know your thoughts, will continue reviewing after

packages/ui/src/theme/createTheme.ts
jacoblogan: Here we are running on 2 because we are expecting 1 to always exist and the point of this code block is to choose the desired token based on the static path value, and remove all the other duplicate tokens.

packages/ui/src/theme/tokens/components/link.ts
jacoblogan: I've added a type for these, but since they aren't `WebDesignTokens` I created a new type for them.

jacoblogan: Also this has been renamed to more clearly indicate that these are duplicate states and not just general purpose deprecated tokens

packages/ui/src/theme/__tests__/createTheme.test.ts
```diff
@@ -31,6 +31,180 @@ describe('@aws-amplify/ui', () => {
...
+                    },
+                  },
+                },
+              },
+            },
+          },
+        };
+        const { cssText, tokens } = createTheme(theme);
+        expect(cssText).toContain(
+          `--amplify-components-pagination-button-hover-background-color: ${tokens.components.pagination.button._hover?.backgroundColor.value}`
```
zchenwei, line 62: What do you think of using `const hoverStateTokens = tokens.components.pagination.button._hover`? That will make the entire test more concise and crisp

packages/ui/src/theme/tokens/components/switchField.ts
```diff
@@ -72,6 +81,18 @@ export interface SwitchFieldTokens {
   track: SwitchFieldTrackTokens;
 }
```
zchenwei, line 83: Do we want to comment out all the places we would like to remove after next major release?

packages/ui/src/theme/tokens/types/designToken.ts
```diff
@@ -43,6 +43,17 @@ export type WebDesignToken<ValueType = any> = {
   toString(): string;
 } & DesignToken<ValueType>;
 
+export type ReplacementStateToken = {
+  /**
+   * Name of the deprecated design token
```
zchenwei, line 48: nit: it seems to me `Name of duplicate design token` makes more sense?

packages/ui/src/theme/createTheme.ts
```diff
@@ -63,6 +63,35 @@ function setupTokens(obj: any, path = []) {
...
+ * next major version release.
+ * @param tokens
+ * @returns WebDesignTokens[]
+ */
+function removeDuplicateStateTokens(
+  tokens: WebDesignToken[]
+): WebDesignToken[] {
+  let tokenList = tokens;
+  DUPLICATE_STATE_TOKENS.forEach((duplicateToken) => {
+    const filteredTokens = tokenList.filter(
```
zchenwei, line 79: Is this filter necessary? Shouldn't we know that what inside `DUPLICATE_STATE_TOKENS` is guaranteed to be duplicate and safe to filter out?

jacoblogan: The reason we have this filter instead of just removing the tokens, is that we only want to remove duplicate tokens if more than one exists in the theme.  This happens because the users are currently using the deprecated tokens and once they switch over to the `_state` tokens two tokens will be generated (because the old deprecated version will still be added through the defaultTheme) so here we are checking that both exist and if they do we remove the old one, which we only want to do if the new token is being used.

dbanksdesign: I think we could optimize this function a bit. From what it looks like this will iterated over the entire tokenlist (~1k tokens) N times where N is the number in DUPLICATE_STATE_TOKENS, regardless of if there are duplicates or not. Instead could we try to filter the tokenlist once to check for duplicates, something like 

```javascript
tokenList.filter((token,i) => {
  if (tokenList.indexOf(token) !== i) {
    // there is a duplicate
  }
})
```

Would this be more efficient?

packages/ui/src/theme/tokens/components/pagination.ts
```diff
@@ -63,6 +72,22 @@ export interface PaginationTokens {
   itemShared: PaginationItemSharedTokens;
 }
 
+// To be removed in the next major version release
+export const DUPLICATE_PAGINATION_STATES: ReplacementStateToken[] = [
```
dbanksdesign, line 76: Based on my earlier comment, although it is less explicit, and maybe this is too much premature optimization, this could be a simple object where the key is tokenName and the value is path.

dbanksdesign (approved):
LGTM! Thinking out loud on the efficiency of the filtering logic, dunno if it would make a noticeable difference though.

packages/ui/src/theme/tokens/components/pagination.ts
reesscot: +1 on just using an object instead of an array

jacoblogan: This has been converted to an tokenName:path object instead

packages/ui/src/theme/createTheme.ts
jacoblogan: Thanks, this has been updated to make a single pass instead over the design tokens to find duplicates and a single filter to remove them

jacoblogan:
> LGTM! Thinking out loud on the efficiency of the filtering logic, dunno if it would make a noticeable difference though.

Good callout, I've updated the logic to make a single pass through the tokens to find all the duplicates and just do a single filter to remove them.

packages/ui/src/theme/createTheme.ts
```diff
@@ -63,6 +63,38 @@ function setupTokens(obj: any, path = []) {
...
+ * that are defined in the DUPLICATE_STATE_TOKENS list. This is designed to remove the duplicate tokens
+ * that were created due to inconsistent naming.  This should be removed when the tokens are removed in the
+ * next major version release.
+ * @param tokens
+ * @returns WebDesignTokens[]
+ */
+function removeDuplicateStateTokens(
+  tokens: WebDesignToken[]
+): WebDesignToken[] {
+  let tokenList = tokens;
```
0618, line 77: If you do `tokenList = tokens` hoping to keep the values in `tokens`, it doesn't work. Is that L77 for?

Try this
```
let a = [1,2,3];
let b = a;
b[0]++;
console.log(a); // [2,2,3]
```

0618: To keep `tokens` unchanged, you can do `let tokenList = [...tokens]`;

packages/ui/src/theme/createTheme.ts
```diff
@@ -63,6 +63,38 @@ function setupTokens(obj: any, path = []) {
...
+  });
+
+  // filter out the duplicate tokens that appear more than once in our token list
+  tokenList = tokenList.filter((token) => {
+    if (duplicateTokenCount[token.name] >= 2) {
+      return token.path.join('.') === DUPLICATE_STATE_TOKENS[token.name];
+    }
+    return true;
+  });
+  return tokenList;
```
0618, line 95:
```suggestion
  return tokenList.filter((token) => duplicateTokenCount[token.name] < 2 || token.path.join('.') === DUPLICATE_STATE_TOKENS[token.name];
  );
```

packages/ui/src/theme/createTheme.ts
jacoblogan: tokenList has been removed because it is no longer needed as we are just doing a single pass now

packages/ui/src/theme/createTheme.ts
```diff
@@ -92,7 +120,7 @@ export function createTheme(
   // that creates an array of all tokens.
   let cssText =
     `[data-amplify-theme="${name}"] {\n` +
-    flattenProperties(tokens)
+    removeDuplicateStateTokens(flattenProperties(tokens))
       .map((token) => `${token.name}: ${token.value};`)
       .join('\n') +
     `\n}\n`;
```
0618, line 126: nit: How about only use literal template?

0618 (approved):
LGTM. left a small comment

0618 (dismissed):
LGTM!

calebpollman (changes requested):

packages/ui/src/theme/duplicationUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+            duplicateTokens[current.name] = current;
+          }
+        } else {
+          prev.push(current);
+        }
+        return prev;
+      },
+      []
+    )
+    .concat(Object.values(duplicateTokens));
```
calebpollman, line 47: This reads a bit more complicated than it might need to be, is there anything we can do to simplify? Can discuss offline if needed

packages/ui/src/theme/createTheme.ts
```diff
@@ -63,6 +63,35 @@ function setupTokens(obj: any, path = []) {
...
+    }
+  });
+
+  // filter out the duplicate tokens that appear more than once in our list of tokens
+  return tokens.filter(
+    (token) =>
+      !duplicateTokenCount[token.name] ||
+      duplicateTokenCount[token.name] < 2 ||
+      token.path.join('.') === DUPLICATE_STATE_TOKENS[token.name]
+  );
```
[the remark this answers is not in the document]
jacoblogan: Thanks, updated to use a single reduce instead of iterating over the list twice

packages/ui/src/theme/tokens/components/index.ts
```diff
@@ -136,3 +144,11 @@ export const components: ComponentTokens = {
...
   togglebuttongroup,
 };
+
+export const DUPLICATE_STATE_TOKENS = Object.assign(
+  {},
+  DUPLICATE_LINK_STATES,
+  DUPLICATE_PAGINATION_STATES,
+  DUPLICATE_SWITCHFIELD_STATES,
+  DUPLICATE_TABLE_STATES
+);
```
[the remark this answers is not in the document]
jacoblogan: Moved all logic into a single file to facilitate removing, when the time comes.

calebpollman (approved):

reesscot:
Closing as the Token types have been refactored

PR #2468, opened 2022-08-16
chore(ui): Set `predictableActionArguments` to true

wlee221:
#### Description of changes

This sets `predictableActionArugments` to `true`, as [suggested](https://xstate.js.org/docs/guides/actions.html) in  `xstate@4.33.0+`:

> It is advised to configure predictableActionArguments: true at the top-level of your machine config.

#### `predictableActionArguments` Feature Flag
  
This flag fixes two things ([docs](https://xstate.js.org/docs/guides/actions.html)) to ensure that actions are executed in the right order with the right event:
- runs actions in order they're declared in. 
  - This does not affect us because we do not rely on action orders in auth machine, as [suggested](https://xstate.js.org/docs/guides/actions.html#action-order) by xstate best practices. 
- XState will always call an action with the event directly responsible for the related transition.
  - This is applicable for transient ("always") transitions that are triggered with [raise actions](https://xstate.js.org/docs/guides/actions.html#raise-action).
  - This does not affect us because we do not use this pattern in our state machine.
  - Also got response from XState team that this is not breaking.

Existing e2e tests. These test all the workflows, so I'm confident in this not being feature breaking.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated


ErikCH (dismissed):

hbuchel (dismissed):

reesscot (approved):

ioanabrooks (approved):
LGTM!

calebpollman (approved):

ErikCH (approved):

wlee221:
Closing until we find a better path forward.

PR #2472, opened 2022-08-17
fix(rn-example): update storybook behavior

calebpollman:
#### Description of changes

The current usage of `storybook` in the example app causes it to crash in the default (non-`storybook`) use case, which is for creating apps used in E2E testing, mirroring the existing sample apps. This is due to an import path that relies on an auto-generated file, _storybookLoader.js_, which is created created by the `prestorybook` command. 

While `prestorybook` could be run regardless of the use case, this PR implements a solution that decouples `storybook` from the example apps used for E2E testing.

- ~~add placeholder _storybookLoader.js_ file to prevent RN Example app from during non-storybook use cases of Example app~~
- ~~in `husky/pre-commit` check for updates to _storybookLoader.js_ and revert them to prevent auto-generated changes to _storybookLoader.js_ from being committed~~
- add `setupStorybook` utility

#### Issue #, if available
NA

Manual testing
#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


.husky/pre-commit
```diff
@@ -6,6 +6,14 @@ if [ "$branch" = "main" ]; then
...
 fi
 
+# Revert updates to storybookLoader.js autogenerated by `storybook` in the React Native Example app
+story_loader_path="examples/react-native/storybook/storyLoader.js"
+status="$(git status --porcelain)"
+if [[ $status == *$story_loader_path* ]]; then
+  git checkout HEAD -- $story_loader_path
+  echo "Revert updates to $story_loader_path"
+fi
+
```
ioanabrooks, line 16: This feels a little over-engineered, why not just add the `rnstl` to the dev script? it's not a time-consuming operation.

```
"dev": "rnstl && react-native start"
```

Also because you need to run storybook server in order to generate this file, so if you wanted for instance to change the env var and enable storybook, but just run the app without the web server, then you wouldn't see the stories, or risk not seeing updated stories.

examples/react-native/index.js
```diff
@@ -1,11 +1,29 @@
...
-AppRegistry.registerComponent(appName, () =>
-  isStorybook ? StorybookUIRoot : App
-);
+if (isStorybook) {
+  configure(() => {
+    loadStories();
+  }, module);
+}
+
+const Root = isStorybook
```
ioanabrooks, line 17: I kind of liked having Storybook initialization in a separate file, for better separation of concerns, instead of repeatedly checking if running storybook mode.

.husky/pre-commit
calebpollman: Valid concern but this approach is due to the fact that this App is not just for `storybook` usage and we need to isolate `storybook` outside of our e2e tests

examples/react-native/index.js
calebpollman: Prefer a separation of concerns in most scenarios but the initialization logic is here to prevent the need to have `isStorybook` bleed in to other files (potentially increasing the cognitive strain for engineers who may need to work in this file in the future) as well as prevent running of `storybook` specific code in non-`storybook` use cases.

This pattern is in line with the POC i initially created that added `storybook` and `detox` testing to the same app so RN could follow the e2e use case of its example app similar to other platforms and still allow `storybook` for UI development. 

In the future we may determine that it makes sense to just split out `storybook` from the example app which would  be best in terms of separating concerns, but prefer to discuss if this proves not to scale.

.husky/pre-commit
ioanabrooks: I added `react-native-storybook-loader` as a convenience tool, so we don't have to manually maintain a file with imports for all our stories. But if we need to add workarounds it's becoming inconvenient and maybe the right thing to do is to take a step back and remove this util. We can always add it back later, and since all our stories will be in one folder, adding the imports is not too big of a hassle for now. What do you think?

examples/react-native/index.js
ioanabrooks: I'm just worried that as we add more things to the storybook config/init, like decorators and knobs this file will become even harder to digest. Honestly I am having cognitive strain with this change, but won't block if others feel differently.

calebpollman: Just so i have a clear picture, what abt this change is causing you cognitive strain?

wlee221: For my own understanding, how does having `storybook` setup logic here prevent `isStorybook` from bleeding out? Can we create a storybook setup util file that `index.js` can import?

calebpollman: @wlee221 It restricts its usage to this file

calebpollman: That being said, not against just passing it as an arg to a setup util in _storybook/index.tsx_, but do want to prevent calling `getStorybookUI` if the example app is not running `storybook`. It triggers a require cycle warning in `LogBox` that i would prefer not to see if unnecessary. Definitely more of a nit, but still an ergonomic improvement for the non-`storybook` use case. @ioanabrooks would you find that approach to be agreeable?

Simulator Screen Shot - iPhone 13 - 2022-08-17 at 12 00 59

ioanabrooks: Yeah that sounds better imo.

examples/react-native/storybook/index.tsx
```diff
@@ -1,21 +1,29 @@
...
-
+import { getStorybookUI, configure } from '@storybook/react-native';
+import noop from 'lodash/noop';
 import { loadStories } from './storyLoader';
 
-// import stories
-configure(() => {
-  loadStories();
-}, module);
+// only initialize storbook resources when initStorybook is true
```
calebpollman, line 6:
```suggestion
// only initialize storybook resources when initStorybook is true
```

examples/react-native/package.json
```diff
@@ -34,7 +33,6 @@
     "@storybook/addon-ondevice-actions": "^5.3.23",
     "@storybook/addon-ondevice-knobs": "^5.3.26",
     "@storybook/react-native": "^5.3.25",
-    "@storybook/react-native-server": "^5.3.23",
```
ioanabrooks, line 37: We would want to remove the `react-native-storybook-loader` dependency at line 47 below, not the rn-server.

ioanabrooks (changes requested):

wlee221 (approved):

ioanabrooks (approved):
LGTM! Thanks for fixing this!

PR #2477, opened 2022-08-17
fix(docs): type error on fragment

0618:
#### Description of changes

fix TS syntax error

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/components/Fragment.tsx
+++ b/docs/src/components/Fragment.tsx
@@ -21,7 +21,7 @@ export interface FragmentProps {
    * Note: if this is true, platforms={['web']} should be used to enable web content instead of platforms={['react', 'vue', 'angular']}
    */
   useCommonWebContent?: boolean;
-  children: ({ platform: string }) => LoaderComponent;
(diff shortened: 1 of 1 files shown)


ErikCH (approved):

docs/src/components/Fragment.tsx
```diff
@@ -21,7 +21,7 @@ export interface FragmentProps {
    * Note: if this is true, platforms={['web']} should be used to enable web content instead of platforms={['react', 'vue', 'angular']}
    */
   useCommonWebContent?: boolean;
-  children: ({ platform: string }) => LoaderComponent;
+  children: ({ platform }: { platform: string }) => LoaderComponent;
```
ioanabrooks, line 24: Dumb question: why does `platform: string` not work?

0618: `{ platform: string }` means to rename `platform` to `string`, so the linter complains the var `string` is declared but never used 😂

ioanabrooks (approved):

reesscot (approved):

PR #2478, opened 2022-08-18
fix(docs): 404 routing

0618:
#### Description of changes

**Issue**: if a url has an invalid platform param, it doesn't return 404 directly, instead, it returns 200, then 404.
**Reason**: `useCustomRouter` detect invalid platform param when the page renders, which is too late.
**Fix**: use Next.js [`getStaticPaths`](https://nextjs.org/docs/basic-features/data-fetching/get-static-paths) instead, to validate the parameters.

Code changes:
- use `getStaticPaths` and `getCustomStaticPath`
- use frontmatter for the frameworks ([ref](https://mdxjs.com/guides/frontmatter/))
- remove `useCustomRouter`
- change next.config accordingly
- remove "figma" from the side nav and only keep it for react

**Before**

https://user-images.githubusercontent.com/11983489/185264225-f35370b3-dc5d-4248-9778-0a953c15fcc8.mov

**After**

https://user-images.githubusercontent.com/11983489/185264480-c157022f-f76a-4730-b1d8-c6fe975e827d.mov

#### Issue #, if available

https://app.asana.com/0/1201840995372760/1202709049052654/f

Local.

test app: https://add-getstaticpaths.d25cgjqwq86nv5.amplifyapp.com/

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/next.config.js
```diff
@@ -136,22 +136,22 @@ module.exports = withNextPluginPreval({
       },
       /**
        * source: a url has one of the folder's names (components, getting-started, guides, theming)
-       * destination: add '[platform]' to the the beginning
+       * destination: add '/react/' to the the beginning
```
0618, line 139: since we only allow the platform names to pass the validation, `[platform]` doesn't work anymore.

docs/src/components/Fragment.tsx
```diff
@@ -46,9 +46,10 @@ export const Fragment = ({
   platforms,
   useCommonWebContent,
 }: FragmentProps) => {
-  const { query } = useCustomRouter();
-  const framework = (query.platform as string) ?? 'react';
-  const platform = getPlatform(framework, useCommonWebContent);
+  const {
+    query: { platform: framework = 'react' },
+  } = useRouter();
```
ioanabrooks, line 51: These lines seem to be repeated a lot, may be worth having a `useReactRouter` hook for it

0618: hmmm.... 🤔  do we reduce much code that way? It's basically a one-liner destructure already.

reesscot:
The after screenhost seems to indicate that we show a blank page if there's a 404 error, is that the case? We would still like to render the 404 cats, but the return code should be changed to a 404.

docs/next.config.js
```diff
@@ -136,22 +136,22 @@ module.exports = withNextPluginPreval({
...
        */
       {
         source:
           '/:nav(legacy-components|connected-components|components|getting-started|guides|theming)/:page*',
-        destination: '/[platform]/:nav/:page*',
+        destination: '/react/:nav/:page*',
```
reesscot, line 144: Doesn't this break the other platforms?

0618: No it won't. Before it was 

"no platform" -> "[platform]" -> "default platform (react)", 

and now it's

"no platform" -> "react (default platform)"

0618:
> https://add-getstaticpaths.d25cgjqwq86nv5.amplifyapp.com/

Not blank page, you can test it out in the testing app https://add-getstaticpaths.d25cgjqwq86nv5.amplifyapp.com/

0618:
> The after screenhost seems to indicate that we show a blank page if there's a 404 error, is that the case? We would still like to render the 404 cats, but the return code should be changed to a 404.

Not a blank page, you can test it here https://add-getstaticpaths.d25cgjqwq86nv5.amplifyapp.com/

jacoblogan (dismissed):
LGTM, is the use case here that users are mistyping the [platform] param and therefore getting that 404? If so can we make this platform param case insensitive?

docs/src/components/Fragment.tsx
ioanabrooks: That's fair, I was also thinking maybe we'll want to add more in the future. won't block.

ioanabrooks (dismissed):

0618:
> LGTM, is the use case here that users are mistyping the [platform] param and therefore getting that 404? If so can we make this platform param case insensitive?

As it's mentioned in the PR description, this PR is to return 404 when it's supposed to, instead of returning 200 -> 404.

What you mentioned is a good point. It seems more user friendly if we have it  url insensitive. However, there are several reasons I prefer not to do that. 

1. I quickly searched on Google and seems URLs are supposed to be case sensitive ([ref](https://www.searchenginejournal.com/google-urls-are-case-sensitive/421264)). 
2. It's a very rare case since we only use lower cases in url
3. if we do make it case insensitive, `getStaticPaths` will have to accept many combinations of different paths and the number would increase once we add more frameworks.

jacoblogan (dismissed):
LGTM

reesscot (approved):
LGTM, thanks for refactoring out `getCustomStaticPath`

0618:
> LGTM, thanks for refactoring out `getCustomStaticPath`

Thanks! I also refactored the `getCustomStaticPath` to use framework info from frontmatter 🎉

ErikCH (dismissed):

docs/src/pages/[platform]/components/flex/index.page.mdx
```diff
@@ -8,7 +8,18 @@ supportedFrameworks: react
 ---
 
 import { Fragment } from '@/components/Fragment';
-import { PageTabLayout } from "@/components/Layout/PageTabLayout"
+import { PageTabLayout } from '@/components/Layout/PageTabLayout';
+import { getCustomStaticPath } from "@/utils/getCustomStaticPath";
+
+export async function getStaticPaths() {
+  return getCustomStaticPath(frontmatter.supportedFrameworks);
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/flex/index.page.mdx:15]: So clean! 💯

docs/src/utils/getCustomStaticPath.ts
```diff
@@ -0,0 +1,14 @@
+import { FRAMEWORKS } from '@/data/frameworks';
+
+export const getCustomStaticPath = (platforms: string[] | string) => {
+  const platformsArr = Array.isArray(platforms)
```
reesscot, line 4: nit: nested ternaries are a bit hard to read, can you use if statements here instead?

reesscot (approved):
LGTM, thanks for refactoring to use the frontmatter!

jacoblogan (approved):
LGTM

PR #2480, opened 2022-08-18
two examples of extending base with state tokens

jacoblogan:
Just a demonstration of proposed design token changes

### DIFF
--- a/packages/ui/src/theme/tokens/components/toggleButton.ts
+++ b/packages/ui/src/theme/tokens/components/toggleButton.ts
@@ -13,11 +13,11 @@ interface ToggleButtonStateTokens {
   color: DesignToken<ColorValue>;
 }
 interface ToggleButtonHoverTokens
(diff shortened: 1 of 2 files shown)


packages/ui/src/theme/tokens/components/toggleButton.ts
```diff
@@ -7,84 +7,85 @@ import {
...
+  extends Pick<
+    ToggleButtonPrimaryStateTokens,
+    'backgroundColor' | 'borderColor' | 'color'
+  > {}
 
-interface ToggleButtonPrimaryDisabledTokens {
-  color: DesignToken<ColorValue>;
-}
+interface ToggleButtonPrimaryDisabledTokens
+  extends Pick<ToggleButtonPrimaryStateTokens, 'color'> {}
```
0618, line 64: Why do we have 2 `ToggleButtonPrimaryDisabledTokens`? Can we combine them together?

PR #2487, opened 2022-08-18
feat(in-app-messaging): remove styles prop from InAppMessageDisplay

calebpollman:
#### Description of changes
- remove `styles` prop from `InAppMessageDisplay` components and related types
- add InApp components as properties of `inAppMessageDisplay` components
- add "exports" snapshot test for `@aws-amplify/ui-react-native`
- add NextJS example app for custom styling of InApp UI components

#### Issue #, if available
NA

Manually tested

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/examples/next/pages/ui/components/in-app-messaging/override-styling/index.page.tsx
+++ b/examples/next/pages/ui/components/in-app-messaging/override-styling/index.page.tsx
@@ -1,0 +1,47 @@
+import React, { useCallback, useEffect } from 'react';
+import { Amplify } from 'aws-amplify';
+import {
+  Button,
+  InAppMessageDisplay,
+  useInAppMessaging,
+  withInAppMessaging,
+} from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+
+import config from './aws-exports';
+
+Amplify.configure(config);
+
+const CustomModalMessage = (props) => (
+  <InAppMessageDisplay.ModalMessage
+    {...props}
+    style={{ container: { backgroundColor: 'antiquewhite' } }}
+  />
+);
+
+function App() {
+  const { displayMessage } = useInAppMessaging();
+
+  const displayCustomModalMessage = useCallback(
+    () =>
+      displayMessage({
+        content: [{ header: { content: 'Hello World!' } }],
+        id: 'custom message',
+        layout: 'MODAL',
+      }),
+    [displayMessage]
+  );
+
+  // display message component on initial render
+  useEffect(displayCustomModalMessage, [displayCustomModalMessage]);
+
+  return (
+    <Button margin="medium" onClick={displayCustomModalMessage}>
+      Display Custom Modal Message
+    </Button>
+  );
+}
+
+export default withInAppMessaging(App, {
+  components: { ModalMessage: CustomModalMessage },
+});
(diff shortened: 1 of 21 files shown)


examples/next/pages/ui/components/in-app-messaging/override-styling/index.page.tsx
```diff
@@ -0,0 +1,47 @@
...
+  <InAppMessageDisplay.ModalMessage
+    {...props}
+    style={{ container: { backgroundColor: 'antiquewhite' } }}
+  />
+);
+
+function App() {
+  const { displayMessage } = useInAppMessaging();
+
+  const displayCustomModalMessage = useCallback(
```
cshfang, line 25: This might actually end up being less performant since `displayMessage` is already memoized and isn't very likely to change in the provider right?

I think we can just stick the `displayMessage` call right inside the `useEffect`

packages/react-native/src/__tests__/index.spec.ts
```diff
@@ -0,0 +1,8 @@
+import * as exported from '..';
+
+describe('@aws-amplify/ui-react-native', () => {
+  it('exports should match snapshot', () => {
+    const exportedKeys = Object.keys(exported);
```
cshfang, line 5: Are keys guaranteed to be in order?

packages/react-native/src/__tests__/index.spec.ts
```diff
@@ -0,0 +1,8 @@
+import * as exported from '..';
+
+describe('@aws-amplify/ui-react-native', () => {
+  it('exports should match snapshot', () => {
```
cshfang, line 4: Nit: move "exports" into describe string and this would read more nicely

examples/next/pages/ui/components/in-app-messaging/override-styling/index.page.tsx
```diff
@@ -0,0 +1,47 @@
...
+import config from './aws-exports';
+
+Amplify.configure(config);
+
+const CustomModalMessage = (props) => (
+  <InAppMessageDisplay.ModalMessage
+    {...props}
+    style={{ container: { backgroundColor: 'antiquewhite' } }}
+  />
+);
```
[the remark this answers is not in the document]
calebpollman: Might be worth exploring, but as of right now we just use the single examples afaik

examples/next/pages/ui/components/in-app-messaging/override-styling/index.page.tsx
calebpollman: Generally prefer to keep memoization out of examples, but We are reusing `displayCustomModalMessage` in the `onClick` handler of the `Button` as well, and went with memoization over repetition

packages/react-native/src/__tests__/index.spec.ts
calebpollman: They should ordered be according to order of assignment, which should change unless we change the exports order. Cam change if you feel strongly else will update if needed in the future

cshfang (approved):
LGTM None of my comments are really blockers

examples/next/pages/ui/components/in-app-messaging/override-styling/index.page.tsx
calebpollman: To elaborate on this, in order to use `displayCustomModalMessage` inside the `useEffect`  it needs to be memoized, else the function reference is new on each render causing the `useEffect` to run on each render

0618 (approved):

PR #2497, opened 2022-08-22
chore: remove stale chmod line in package.json

wlee221:
#### Description of changes
Removes stale `chmod +x` commands.

#### Details 

We have stale `chmod +x` commands for `pull-environments.sh` and `pull-environment.sh`. But these shell files are already given `+x` access (you can run `ls -l` to verify). So this doesn't do anything, and is safe to remove. 

Running `chmod +x` before commit is a common pattern in GitHub Actions too -- see step 5 in [docs](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#prerequisites) for example.

#### Security

This does not the threat model because these files were getting `chmod`'d anyway.

Verified pulling works as expected in [poc workflow](https://github.com/aws-amplify/amplify-ui/actions/runs/2906417996) against this PR.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


reesscot (approved):
Makes sense to me

zchenwei (approved):

PR #2501, opened 2022-08-23
chore(environments): move environments to their own folder

wlee221:
### Note

Since this PR required repeated file changes, I split these off to three PRs:

- [x] #2501 (this PR)
- [x] #2502 
- [x] #2504 

The last two need to be merged into this branch first (else checks will fail). 

#### Description of changes

This PR splits `environments` by categories:
```
environments/
├── auth/ (13 backends)
├── datastore/ (1 backend)
└── geo/ (1 backend)
```

This just moves directories and files and do not make code changes. 

#### Description of how you validated changes

#2505 is running test against the changes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2502, opened 2022-08-23
chore(environments): Update scripts for environments split

wlee221:
#### Description of changes

Updates `package.json` `scripts` and `pull-environments.sh` to reflect directory changes in #2501.

#2505 is running test against all environment split changes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/environments/pull-environments.sh
+++ b/environments/pull-environments.sh
@@ -25,7 +25,7 @@ fi
 # (1) bash_source[0] contains the filename of this shell relative to cwd
 #     (e.g. `../pull-environments.sh`)
 # (2) dirname gets rid of filename and returns the relative path to the 
-#     directory is in (e.g. ../)
+#     directory is in (e.g. ..)
 #
 # source: https://stackoverflow.com/a/24112741
 shell_path="$(dirname "${BASH_SOURCE[0]}")" # under normal use, this points to `../`
--- a/environments/pull-environments.sh
+++ b/environments/pull-environments.sh
@@ -2,6 +2,18 @@
 set -e
 IFS='|'
 
+# CLI input
+WORKING_DIRECTORY=$1;
+echo $WORKING_DIRECTORY;
+
+if [ -n "$WORKING_DIRECTORY" ]; then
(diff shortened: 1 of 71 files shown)


environments/pull-environments.sh
```diff
@@ -21,5 +21,5 @@ if [ "$NODE_ENV" = "test" ]; then
   numParallelTasks=1; # GitHub actions has trouble handling parallel executions
 fi
 
-echo $dirs | xargs -P $numParallelTasks -I {} sh -c "./pull-environment.sh {}"
+echo $dirs | xargs -P $numParallelTasks -I {} sh -c "../pull-environment.sh {}"
```
slaymance, line 24: I understand this is relative to the cwd of each environment folder, but it is a little confusing that we're referencing a file in the same directory as this script using `..`. I wonder if it's easy to reference `pull-environment.sh` from the path of `pull-environments.sh`.

wlee221: That's a good point; we can get the directory that `pull-environments.sh` is in with:

```bash
shell_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
```

and then we can do:

```suggestion
echo $dirs | xargs -P $numParallelTasks -I {} sh -c ""$shell_path"/pull-environment.sh {}"
```

wlee221: (source: https://stackoverflow.com/a/24112741)

wlee221: actually that shell command can prob be simplified... taking a look.

slaymance: I this that approach is a bit cleaner and allows us to run the `pull-environments.sh` script from any directory, but I'd say timebox it as I think it's just a nice-to-have :)

wlee221: Agreed, done! [d506c78](https://github.com/aws-amplify/amplify-ui/pull/2502/commits/d506c7871cc97c404366ff67ccdab135918d9224)

slaymance (approved):
🚀

reesscot (approved):
LGTM

PR #2504, opened 2022-08-23
chore(environments): update `awsExports` to point to new paths

wlee221:
#### Description of changes

Updates `examples` to point to new environments paths. No actual code changes other than that.

tests are getting run on #2505.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/examples/angular/src/pages/ui/components/authenticator/custom-sign-up-fields/aws-exports.js
+++ b/examples/angular/src/pages/ui/components/authenticator/custom-sign-up-fields/aws-exports.js
@@ -1,2 +1,2 @@
-import awsExports from '@environments/auth-with-email-and-custom-attributes/src/aws-exports';
+import awsExports from '@environments/auth/auth-with-email-and-custom-attributes/src/aws-exports';
 export default awsExports;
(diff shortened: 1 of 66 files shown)


slaymance (approved):
Clean and organized!

examples/angular/src/pages/ui/components/authenticator/custom-sign-up-fields/aws-exports.js
```diff
@@ -1,2 +1,2 @@
-import awsExports from '@environments/auth-with-email-and-custom-attributes/src/aws-exports';
+import awsExports from '@environments/auth/auth-with-email-and-custom-attributes/src/aws-exports';
```
slaymance, line 1: Just wanted your thoughts on this: how do you anticipate handling environments which may have multiple categories? For example, a Geo category requires Auth to be added as well. Obviously, that use-case makes the most sense in Geo since Auth doesn't make use of the geo category, but it's just something I'm thinking about with this approach.

wlee221: Good question.

As with the geo example, I anticipate examples to test one primary category like `Storage` or `Interactions` even if it has multiple categories configured. In that case, we'll add them under the primary category.

wlee221: In the future though, if we do add multi category examples (e.g. Storage + Auth + Interactions), i think those can be a folder of its own. Maybe `multi-category/`. 

But those complex examples might make more sense to be a guide 🤔

slaymance: I like the idea of using a `multi-category` section, especially if we want to explore integration testing our components working together in the future. But we don't have to make any decision about that now of course.

hbuchel (approved):

reesscot (approved):

PR #2544, opened 2022-08-31
chore(rwa): convert AuthChallengeNames enum to string union

calebpollman:
#### Description of changes
Replace `AuthChallengeNames` enum with `AuthChallengeName` string union (set to the value of Cognito supplied `ChallengeName`) to prevent type conflicts between `CognitoUserAmplify['challengeName']` and usage of `AuthChallengeName` in utils

Additionally, added some typed `challengeName` utils in _packages/ui/src/machines/authenticator/actors/signIn.ts_

#### Issue #, if available
NA

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/.changeset/gentle-ads-hang.md
+++ b/.changeset/gentle-ads-hang.md
@@ -1,0 +1,9 @@
+---
+"amplify-ui-angular-mono": patch
+"@aws-amplify/ui-react": patch
+"@aws-amplify/ui": patch
+"@aws-amplify/ui-vue": patch
+"@aws-amplify/ui-angular": patch
+---
+
+chore(rwa): convert AuthChallengeNames enum to string union
--- a/.changeset/gentle-ads-hang.md
+++ b/.changeset/gentle-ads-hang.md
@@ -1,5 +1,4 @@
 ---
-"amplify-ui-angular-mono": patch
 "@aws-amplify/ui-react": patch
 "@aws-amplify/ui": patch
 "@aws-amplify/ui-vue": patch
--- a/packages/ui/src/machines/authenticator/actors/signIn.ts
+++ b/packages/ui/src/machines/authenticator/actors/signIn.ts
@@ -2,7 +2,12 @@ import { Auth } from 'aws-amplify';
 import get from 'lodash/get';
 import isEmpty from 'lodash/isEmpty';
 import { createMachine, sendUpdate } from 'xstate';
-import { AuthEvent, SignInContext } from '../../../types';
+import {
+  AuthChallengeName,
+  AuthEvent,
+  CognitoUserAmplify,
+  SignInContext,
+} from '../../../types';
 import { runValidators } from '../../../validators';
 import {
   clearAttributeToVerify,
@@ -33,6 +38,22 @@ import { defaultServices } from '../defaultServices';
 export type SignInMachineOptions = {
   services?: Partial<typeof defaultServices>;
 };
+
+const MFA_CHALLENGE_NAMES: AuthChallengeName[] = [
+  'SMS_MFA',
+  'SOFTWARE_TOKEN_MFA',
+];
+
+const getChallengeName = (event: AuthEvent): AuthChallengeName =>
+  get(event, 'data.challengeName');
+
+const isExpectedChallengeName = (
+  challengeName: AuthChallengeName,
+  expectedChallengeName: AuthChallengeName
+) => challengeName === expectedChallengeName;
+
+const isMfaChallengeName = (challengeName: AuthChallengeName) =>
+  MFA_CHALLENGE_NAMES.includes(challengeName);
 
 export function signInActor({ services }: SignInMachineOptions) {
   return createMachine<SignInContext, AuthEvent>(
@@ -461,10 +482,7 @@ export function signInActor({ services }: SignInMachineOptions) {
       },
       guards: {
         shouldConfirmSignIn: (_, event): boolean => {
-          const challengeName = get(event, 'data.challengeName');
-          const validChallengeNames = ['SMS_MFA', 'SOFTWARE_TOKEN_MFA'];
-
-          return validChallengeNames.includes(challengeName);
+          return isMfaChallengeName(getChallengeName(event));
         },
         shouldAutoSignIn: (context) => {
           return context?.intent === 'autoSignIn';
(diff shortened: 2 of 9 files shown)


packages/ui/src/types/authenticator/user.ts
```diff
@@ -1,13 +1,7 @@
...
-/** Enum of known challenge names */
-export enum AuthChallengeNames {
-  SMS_MFA = 'SMS_MFA',
-  SOFTWARE_TOKEN_MFA = 'SOFTWARE_TOKEN_MFA',
-  NEW_PASSWORD_REQUIRED = 'NEW_PASSWORD_REQUIRED',
-  RESET_REQUIRED = 'RESET_REQUIRED',
-  MFA_SETUP = 'MFA_SETUP',
-}
+/** Known challenge names */
+export type AuthChallengeName = ChallengeName;
```
calebpollman, line 4: Re-export `ChallengeName` as `AuthChallengeName` to prevent the need to reference `'amazon-cognito-identity-js'` in UI packages

packages/ui/src/types/authenticator/validator.ts
```diff
@@ -22,9 +21,3 @@ export type Validator = (
   touchData?: AuthFormData,
   passwordSettings?: PasswordSettings
 ) => ValidatorResult | Promise<ValidatorResult>;
-
-export type SignInTypes = (
-  user: string,
-  code: string,
-  mfaType: AuthChallengeNames.SMS_MFA | AuthChallengeNames.SOFTWARE_TOKEN_MFA
-) => SignInResult | Promise<SignInResult>;
```
calebpollman, line 30: This was unused 😃

0618:
Could you help me understand what conflict this PR is trying to prevent? 

> prevent type conflicts between CognitoUserAmplify['challengeName'] and usage of AuthChallengeName

calebpollman:
> Could you help me understand what conflict this PR is trying to prevent?
> 
> > prevent type conflicts between CognitoUserAmplify['challengeName'] and usage of AuthChallengeName

Sure MJ. The enum that was created for `AuthChallengeNames` does not align with the `CognitoUserAmplify` interface `challengeName` value, as it expects a string union, not an enum:

image

.changeset/gentle-ads-hang.md
```diff
@@ -0,0 +1,9 @@
+---
+"amplify-ui-angular-mono": patch
```
wlee221, line 2: You should `amplify-ui-angular-mono`: this is a dev-only workspace for Angular.

packages/ui/src/types/authenticator/stateMachine/context.ts
```diff
@@ -37,7 +37,7 @@ export interface AuthContext {
   username?: string;
   password?: string;
   code?: string;
-  mfaType?: AuthChallengeNames.SMS_MFA | AuthChallengeNames.SOFTWARE_TOKEN_MFA;
+  mfaType?: 'SMS_MFA' | 'SOFTWARE_TOKEN_MFA';
```
wlee221, line 40: What do you think about making this a separate type as well?

I understand this is a subset of external `ChallengeName`, so I get why that could be confusing. Just open to thoughts!

wlee221 (approved):
LGTM, just a non-blocking thought.

0618:
> > Could you help me understand what conflict this PR is trying to prevent?
> > > prevent type conflicts between CognitoUserAmplify['challengeName'] and usage of AuthChallengeName
> 
> Sure MJ. The enum that was created for `AuthChallengeNames` does not align with the `CognitoUserAmplify` interface `challengeName` value, as it expects a string union, not an enum:
> 
> image

I see...

What if we use something like this?

```ts
export type AuthChallengeName = {
    SMS_MFA :  "SMS_MFA",
    SOFTWARE_TOKEN_MFA : "SOFTWARE_TOKEN_MFA",
    NEW_PASSWORD_REQUIRED : "NEW_PASSWORD_REQUIRED",
    RESET_REQUIRED : "RESET_REQUIRED",
    MFA_SETUP : "MFA_SETUP",
    CUSTOM_CHALLENGE : 'CUSTOM_CHALLENGE',
    SELECT_MFA_TYPE : 'SELECT_MFA_TYPE'
}
```

.changeset/gentle-ads-hang.md
calebpollman: Got it. Just used the default changeset template provided :p 

```suggestion
```

calebpollman:
> > > Could you help me understand what conflict this PR is trying to prevent?
> > > > prevent type conflicts between CognitoUserAmplify['challengeName'] and usage of AuthChallengeName
> > 
> > 
> > Sure MJ. The enum that was created for `AuthChallengeNames` does not align with the `CognitoUserAmplify` interface `challengeName` value, as it expects a string union, not an enum:
> > image
> 
> I see...
> 
> What if we use something like this?
> 
> ```ts
> export type AuthChallengeName = {
>     SMS_MFA :  "SMS_MFA",
>     SOFTWARE_TOKEN_MFA : "SOFTWARE_TOKEN_MFA",
>     NEW_PASSWORD_REQUIRED : "NEW_PASSWORD_REQUIRED",
>     RESET_REQUIRED : "RESET_REQUIRED",
>     MFA_SETUP : "MFA_SETUP",
>     CUSTOM_CHALLENGE : 'CUSTOM_CHALLENGE',
>     SELECT_MFA_TYPE : 'SELECT_MFA_TYPE'
> }
> ```

@0618 This would require `challengeName: AuthChallengeName` arguments to be objects rather than just a typed `string` (you can read more info on union types [here](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)).

Additionally, this would result in our typing moving further away from what the `CognitoUser` interface uses under the hood for `challengeName`, which is ultimately the `string` union `ChallengeName` provided by Amplify JS, making it more difficult to maintain a contract between the consumed types and our own typing.

ErikCH (dismissed):
👍

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -569,9 +579,9 @@ export function signInActor({ services }: SignInMachineOptions) {
             return Promise.reject(err);
           }
         },
-        async verifyTotpToken(context, event) {
```
calebpollman, line 572: Removed unused function args throughout this file

reesscot (dismissed):
LGTM. Thanks for improving our alignment with the JS types here.

wlee221 (approved):

PR #2546, opened 2022-09-01
feat(docs): added mockServices and mockComponents to the POC branch

avirajdh:
#### Description of changes
Added the mockServices to the docs folder.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
+import { Amplify} from 'aws-amplify';
```
wlee221, line 1: Formatting error -- i'd ensure your prettier is installed correctly.

1. https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode
2. Make sure you ran `yarn install` at monorepo root

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const awsExports = {
```
wlee221, line 3: I'd add a comment above this that this is a fake awsexports that is needed to remove auth config errors, for future readers.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+  aws_user_pools_id: 'xx-xxxx-x_xxxxx', // (optional) -  Amazon Cognito User Pool ID
+  aws_user_pools_web_client_id: 'xxxxxxxxxxxxxx', // (optional) - Amazon Cognito App Client ID (App client secret needs to be disabled)
+};
+Amplify.configure(awsExports)
+
+let signUpPassword = "";
+const verifiedUsers = new Map();
+const fixed_username = "amplify-ui";
+const fixed_password = "connected-components";
```
wlee221, line 12: I wonder if we can genericize these values for easier reference. Maybe just username "test" and password "password"?

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+      return Promise.resolve(verifiedUsers.get(username));
+    } else {
+      return Promise.reject(new Error("Invalid Verification Code"));
+    }
+  },
+  async handleResendConfirmationCode({
+    username
+  }: {
+    username: string
+  }): Promise<any> {
```
wlee221, line 61:
```suggestion
  }) {
```

so we don't have `any` and let TS infer the return type for us.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+        return Promise.resolve();
+      } else {
+        return Promise.reject(new Error("Invalid Verification Code"));
+      }
+  },
+  async handleForgotPassword(formData): Promise<any> {
+    if((verifiedUsers.has(formData)) || formData === fixed_username) {
+      return Promise.resolve();
+    } else {
+      return Promise.reject(new Error("User does not exist!"))
```
wlee221, line 84: As discussed, I'd find the precise error message for this. I don't recall actual message having exclamation mark.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
+import { Amplify} from 'aws-amplify';
+
+const awsExports = {
+  aws_user_pools_id: 'xx-xxxx-x_xxxxx', // (optional) -  Amazon Cognito User Pool ID
+  aws_user_pools_web_client_id: 'xxxxxxxxxxxxxx', // (optional) - Amazon Cognito App Client ID (App client secret needs to be disabled)
+};
+Amplify.configure(awsExports)
```
wlee221, line 7: This file is really doing two things -- configure Amplify with fake data and declare mock services. Wondering if we should split this file into two.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+  },
+  async handleForgotPasswordSubmit({
+    username,
+    code,
+    password,
+  }: {
+    username: string;
+    code: string;
+    password: string;
+  }): Promise<any> {
```
wlee221, line 72: Same as above

docs/src/data/mockServices.tsx
zchenwei: +1

docs/src/data/mockServices.tsx
zchenwei: +1

docs/src/data/mockServices.tsx
zchenwei: I would say we should move the config call to `_app.tsx` because that's the NextJS way for global configuration

docs/src/data/mockServices.tsx
zchenwei: Yes, we need to use an easy one. And for constant, please use the `CONSTANT_CASE`

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+  },
+  async handleConfirmSignUp({
+    username,
+    code
+  }: {
+    username: string;
+    code: string;
+  }) {
+    if (code === '123456') {
+      verifiedUsers.set(username, {username : username, password : signUpPassword});
```
zchenwei, line 51: How come we need to store an object containing `username` as well?

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+const fixed_username = "amplify-ui";
+const fixed_password = "connected-components";
+
+export const mockServices = {
+  async getAmplifyConfig() {
+    return {};
+  },
+  async handleSignUp(formData) {
+    if(verifiedUsers.has(formData.username)) {
+      return Promise.reject(new Error("User already exists!"))
```
reesscot, line 20: Is this really necessary? I don't think we should build in too much service level logic that we then have to maintain. I think our demos should show off the UI screens, not perfectly match the service behaviors.

docs/src/data/mockServices.tsx
avirajdh: Done

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/data/mockServices.tsx
avirajdh: Done! I removed the error message.

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/data/mockServices.tsx
avirajdh: I did that because in handleSignIn, the resolve takes in the object containing the username and the password

docs/src/data/mockServices.tsx
avirajdh: Talked with Scott and fixed this!

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,89 @@
...
+const verifiedUsers = new Map();
+const fixed_username = "amplify-ui";
+const fixed_password = "connected-components";
+
+export const mockServices = {
+  async getAmplifyConfig() {
+    return {};
+  },
+  async handleSignUp(formData) {
+    if(verifiedUsers.has(formData.username)) {
```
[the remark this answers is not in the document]
avirajdh: Removed the formData check so won't be requiring this anymore.

docs/src/data/mockServices.tsx
wlee221: Curious about this too, what changes are we making?

avirajdh: Basically, if a user tries to sign up with an existing username, it won't throw any error now.

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/pages/_app.page.tsx
```diff
@@ -38,6 +40,15 @@ if (typeof window === 'undefined') {
   window['theme'] = defaultTheme;
 }
 
+//The following awsExports object contains fake user pool id and web client id values
+//which do not translate to any actual aws user information. The purpose of this awsExports
+//object is to mock Amplify.configure() and remove the Auth config erros from the console.
```
wlee221, line 45: Nit: use 
```js
/**
 * Your multiline
 * comment
 */
```
for multiline comments

docs/src/data/mockServices.tsx
```diff
@@ -32,33 +20,34 @@ export const mockServices = {
...
-        (username === fixed_username && password === fixed_password)) {
+  }) {
+    if (
+      verifiedUsers.get(username).password === password ||
+      (username === FIXED_USERNAME && password === FIXED_PASSWORD)
+    ) {
       return Promise.resolve(verifiedUsers.get(username));
     } else {
-      return Promise.reject(new Error("Invalid username or password"));
+      return Promise.reject(new Error('Invalid username or password'));
```
wlee221, line 30: Do you plan to remove these errors too?

avirajdh: I don't think so because that'll defeat the purpose of having fixed username and password combinations to preserve the authenticator flow.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
...
+    username,
+    code,
+    password,
+  }: {
+    username: string;
+    code: string;
+    password: string;
+  }) {
+    if (code === '123456') {
+      verifiedUsers.set(username, { username: username, password: password });
```
zchenwei, line 63:
```suggestion
      verifiedUsers.set(username, { username, password });
```

nit: when the object key has the same name as the value, this is a simple way to assign the value

docs/src/pages/_app.page.tsx
```diff
@@ -1,5 +1,7 @@
 import * as React from 'react';
 
+import { Amplify } from 'aws-amplify';
+
```
zchenwei, line 4: nit: we do not need a line break here, since both `aws-amplify` and `@aws-amplify/ui-react` are internal dependencies. They should be grouped together

docs/src/data/mockSocialProviders.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * mockComponents is an object containing the ovveride header function calls for
+ * the authenticator.
+ */
+export const mockComponents = {
+  SignIn: {
+    Header() {
+      return (
+        <Flex direction={'column'} paddingLeft={'2rem'} paddingRight={'2rem'}>
```
zchenwei, line 92: nit: do not need any parenthesis

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/pages/_app.page.tsx
avirajdh: Done!

docs/src/pages/_app.page.tsx
avirajdh: Done!

docs/src/data/mockSocialProviders.tsx
```diff
@@ -0,0 +1,176 @@
+import { Flex, Icon, Button, Text, Divider } from '@aws-amplify/ui-react';
+
```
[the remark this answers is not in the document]
zchenwei: On second thoughts, I think we should move these icons from `packages/react/components/authenticator/federatedSignIn` to internal icons directory and reference them from there. This is out of scope though

docs/src/data/mockSocialProviders.tsx
avirajdh: Done!

docs/src/data/mockSocialProviders.tsx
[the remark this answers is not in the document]
avirajdh: Yes. For now, as we discussed, we can copy the icons into the mockSocialServices.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+  },
+  async handleConfirmSignUp({
+    username,
+    code,
+  }: {
+    username: string;
+    code: string;
+  }) {
+    if (code === '123456') {
```
ErikCH, line 40: This is a magic number, and probably should be added to the top of the file. With the FIXED_USERNAME and FIXED_PASSWORD.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
...
+  async handleForgotPasswordSubmit({
+    username,
+    code,
+    password,
+  }: {
+    username: string;
+    code: string;
+    password: string;
+  }) {
+    if (code === '123456') {
```
ErikCH, line 62: Sam here

ErikCH (commented):
Just a couple of changes

docs/src/data/mockServices.tsx
avirajdh: Thank you so much for pointing that out!

docs/src/pages/_app.page.tsx
```diff
@@ -38,6 +39,17 @@ if (typeof window === 'undefined') {
...
+ * The following awsExports object contains fake user pool id and web client id values
+ * which do not translate to any actual aws user information. The purpose of this awsExports
+ * object is to mock Amplify.configure() and remove the Auth config erros from the console.
+ */
+const awsExports = {
+  aws_user_pools_id: 'xx-xxxx-x_xxxxx', // (optional) -  Amazon Cognito User Pool ID
+  aws_user_pools_web_client_id: 'xxxxxxxxxxxxxx', // (optional) - Amazon Cognito App Client ID (App client secret needs to be disabled)
+};
+Amplify.configure(awsExports);
+
```
calebpollman, line 52: Can we move this to into its own _aws-exports.js_ file please?

docs/src/data/mockSocialProviders.tsx
```diff
@@ -0,0 +1,176 @@
...
+        clipRule="evenodd"
+        d="M142.943 111.185C142.943 124.756 143.268 136.054 136.406 148.123C130.856 157.913 122.027 163.95 112.222 163.95C98.8288 163.95 90.9806 153.772 90.9806 138.693C90.9806 109.036 117.677 103.647 142.943 103.647V111.185ZM178.166 196.081C175.858 198.15 171.635 198.22 169.914 196.894C157.974 187.684 149.89 173.688 149.89 173.688C130.706 193.156 117.127 199 92.2879 199C62.8772 199 40 180.905 40 144.729C40 116.461 55.3552 97.2408 77.2563 87.823C96.2094 79.5256 122.684 78.0173 142.943 75.7517C142.943 75.7517 144.633 53.933 138.699 45.9806C134.098 39.8163 126.272 36.9329 119.089 36.9329C106.127 36.8829 93.61 43.9051 91.1262 57.4188C90.4136 61.2829 87.5533 64.5261 84.54 64.206L51.0823 60.5922C48.5156 60.2951 45.0381 57.6639 45.8636 53.3081C53.644 12.3684 90.7373 0 123.989 0C140.983 0 163.21 4.51651 176.608 17.3349C193.597 33.1648 191.969 54.2755 191.969 77.2722V131.51C191.969 147.835 198.768 154.987 205.151 163.775C207.376 166.953 207.886 170.714 205.04 173.032C197.902 178.999 178.166 196.081 178.166 196.081Z"
+        fill="#F2541B"
+      />
+    </svg>
+  );
+};
+
+/**
+ * mockComponents is an object containing the ovveride header function calls for
```
calebpollman, line 85: Typo: `ovveride` -> `override`

docs/src/data/mockServices.tsx
wlee221: I'm good with leaving some basic error in place for the purpose of demo.

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
...
+    code,
+  }: {
+    username: string;
+    code: string;
+  }) {
+    if (code === '123456') {
+      verifiedUsers.set(username, {
+        username: username,
+        password: signUpPassword,
+      });
```
calebpollman, line 44: To make this easier to read, we should move the second argument in its own variable, example:
```
const input =  { username, password: signUpPassword ];

verifiedUsers.set(username, input);
```

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
...
+  async handleSignIn({
+    username,
+    password,
+  }: {
+    username: string;
+    password: string;
+  }) {
+    if (
+      verifiedUsers.get(username).password === password ||
+      (username === FIXED_USERNAME && password === FIXED_PASSWORD)
```
calebpollman, line 26: Can we move this in to a meaningfully name util at the top of this file?

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
+let signUpPassword = '';
+const verifiedUsers = new Map();
+const FIXED_USERNAME = 'test';
+const FIXED_PASSWORD = 'password';
```
calebpollman, line 4: Constants should be place above other variable declarations

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
+let signUpPassword = '';
```
calebpollman, line 1: Can we add an inline comment explaining the need for a global mutable `string`?

packages/react/src/components/Authenticator/FederatedSignIn/FederatedSignInButtons/FederatedSignInButton.tsx
```diff
@@ -12,7 +12,7 @@ export interface FederatedSignInButtonProps {
   text: string;
 }
 
-const AppleIcon = (): JSX.Element => {
+export const AppleIcon = (): JSX.Element => {
```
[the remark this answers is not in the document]
avirajdh: I'm gonna undo these changes. Initially, we were planning to export these icons to use them in the socialProviders but we're not doing them anymore since it's out of scope for this POC. I will be removing these exports in the next commit. Thank you for the comment!

docs/src/pages/_app.page.tsx
avirajdh: Done!

docs/src/data/mockSocialProviders.tsx
avirajdh: Done!

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/data/mockServices.tsx
avirajdh: Done!

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,78 @@
...
+      return Promise.resolve();
+    } else {
+      return Promise.reject(new Error('Invalid Verification Code'));
+    }
+  },
+  async handleForgotPassword(formData): Promise<any> {
+    if (verifiedUsers.has(formData) || formData === FIXED_USERNAME) {
+      return Promise.resolve();
+    } else {
+      return Promise.reject(new Error('User does not exist!'));
```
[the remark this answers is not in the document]
avirajdh: Resolved!

docs/src/data/mockServices.tsx
wlee221: I think this is the remaining nit: util function for checking fixed username + pw

docs/src/pages/_app.page.tsx
```diff
@@ -8,6 +9,7 @@ import { Header } from '@/components/Layout/Header';
 import Script from 'next/script';
 import { baseTheme } from '../theme';
 import { useCustomRouter } from '@/components/useCustomRouter';
+import { awsExports } from '../aws-exports.js';
```
calebpollman, line 12: Where is this file 😃

avirajdh: It is in the root of the src folder

calebpollman: Maybe I'm missing something but I don't see it in the PR?

avirajdh: Yeah. That is strange because I have made multiple commits since that suggestion.

wlee221: Was it gitignored?

avirajdh: I'm not sure but I have added it now

docs/src/data/mockServices.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return Promise.resolve();
+    } else {
+      return Promise.reject(
+        new Error('Username/client id combination not found.')
+      );
+    }
+  },
+};
+
+export default mockServices;
```
calebpollman, line 90: We shouldn't need both to export `mockServices` as a named export and the default export (fwiw prefer default exports when the export matches the file name). Seeing the same issue with `mockComponents` in _mockSocialProviders.ts_ (which maybe should be renamed to _mockComponents.ts_)

calebpollman (commented):
Mostly LGTM, left some minor feedback. 

Additionally, consider nesting and renaming _mockServices_ and _mockSocialProviders_ in a _amplifyMock_ directory inside _data_ to colocate the feature files (thinking maybe the mock `awsExports` file should also be included) and create a separation of concerns from the rest of the _data_ contents. Example structure:
```
src/
  data/
    amplifyMock/
      components.ts
      exports.ts // maybe?
      index.ts // export the other files from here
      services.ts
```

docs/src/aws-exports.js
```diff
@@ -0,0 +1,10 @@
+/**
```
calebpollman, line 1: Looks like this file should be removed?

docs/src/pages/_app.page.tsx
```diff
@@ -8,6 +9,7 @@ import { Header } from '@/components/Layout/Header';
 import Script from 'next/script';
 import { baseTheme } from '../theme';
 import { useCustomRouter } from '@/components/useCustomRouter';
+import { awsExports } from '../data/amplifyMock/index.js';
```
calebpollman, line 12: Does this import work since it's pointing to a `.js` file that doesn't exist? Either way, should be able to just write this as 
```suggestion
import { awsExports } from '../data/amplifyMock';
```
Since we are exporting from the _index.ts_ file

wlee221 (approved):
LGTM, thanks @avirajdh!

docs/src/data/amplifyMock/components.tsx
```diff
@@ -0,0 +1,175 @@
...
+    Header() {
+      return (
+        <Flex direction="column" paddingLeft="2rem" paddingRight="2rem">
+          <Flex direction="column" paddingTop="2rem" paddingBottom="1rem">
+            <Button
+              className="federated-sign-in-button"
+              fontWeight="normal"
+              gap="1rem"
+            >
+              <AmazonIcon />
```
0618, line 99: I'm not very familiar with those SVGs, but I'm curious what's the difference between using them as components and using them directly as SVG images?

No matter components or images, does it make sense if we keep them in a separate file/folder so that we can re-use them in other parts of the codebase?

avirajdh: Chenwei actually pointed this out last week as well. These icons are derived from FederatedSignInButtons from the Authenticator. Chenwei and I had some ideas to add these icons to the public/svg folder to reduce duplicate code and reuse them at other places as well. However, since that is out of scope, I have just copied the code from FederatedSignInButtons.

calebpollman (approved):
LGTM 🚢

0618 (approved):
:shipit:

PR #2549, opened 2022-09-06
chore(e2e): Update testing dependencies and configurations

slaymance:
#### Description of changes
This pull request updates a few of the dependencies used for e2e testing.
This pull request updates the dependencies used for e2e testing.

- `@testing-library/cypress@^7.0.6` -> `@testing-library/cypress@^8.0.3`
- `cypress@^8.5.0` -> `cypress@^10.7.0`
- `cypress-cucumber-preprocessor` -> `@badeball/cypress-cucumber-preprocessor`: This package had a change in ownership and the original is no longer actively maintained (https://github.com/badeball/cypress-cucumber-preprocessor/issues/689)
- `dotenv-safe` -> `dotenv`: `dotenv-safe` is no longer actively maintained and is incompatible with the latest version of `cypress` (we should revisit and decide if `dotenv` is necessary for e2e testing)

Configuration files are also updated to be compatible with the latest versions of e2e testing dependencies.

Bundled features files are no longer supported as a performance improvement for `@badeball/cypress-cucumber-preprocessor`. Instead, the library recommends using esbuild for compiling `.feature` files which makes using `.features` files unnecessary (https://github.com/badeball/cypress-cucumber-preprocessor/issues/689). This PR removes the bundled `.features` files.

The new cucumber preprocessor is much stricter about duplicate step definitions so those have been resolved as well.

Verify that all e2e tests are passing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


examples/next/pages/ui/hooks/actions/NavigateActions.tsx
```diff
@@ -20,15 +20,9 @@ export const NavigateActions = () => {
...
-      </Button>
-      <Button testId="reload" onClick={() => reload()}>
-        Reload
-      </Button>
-      <Button testId="hash" onClick={() => handleHashClick()}>
-        Go to #notes
-      </Button>
+      <Button onClick={() => goToAmazon()}>Go to amazon.com</Button>
+      <Button onClick={() => reload()}>Reload</Button>
+      <Button onClick={() => handleHashClick()}>Go to #notes</Button>
```
slaymance, line 25: Corresponding tests have been updated to use accessible role queries instead of test IDs.

packages/e2e/cypress.config.ts
```diff
@@ -0,0 +1,28 @@
...
+        createBundler({ plugins: [createEsbuildPlugin(config)] })
+      );
+
+      Object.assign(config.env, process.env);
+
+      return config;
+    },
+  },
+  videoUploadOnPasses: false,
+});
```
slaymance, line 28: This is the new Cypress config file and `cypress.json` has been deprecated.

packages/e2e/cypress/integration/ui/hooks/actions-navigation/actions-navigation.steps.ts
```diff
@@ -1,13 +1,9 @@
...
+import { Then } from '@badeball/cypress-cucumber-preprocessor';
 
 Then('the page contains {string} section', (search: string) => {
   cy.findByRole('document').contains(search);
 });
 
-When('I click the {string} button', (testId: string) => {
-  cy.findByTestId(testId).click();
-});
-
```
slaymance, line 10: This is a duplicate step definition which are no longer allowed.

packages/e2e/features/ui/hooks/actions-navigation.feature
```diff
@@ -8,15 +8,15 @@ Feature: Navigate action hooks
...
   @react
   Scenario: Navigation reload action works correctly
-    When I click the "reload" button
+    When I click the "Reload" button
     Then My page should be reloaded
 
   @react
   Scenario: Navigation location change action works correctly
-    When I click the "locationChange" button
+    When I click the "Go to amazon.com" button
```
slaymance, line 21: These tests now use accessible role queries instead of test IDs.

packages/e2e/package.json
```diff
@@ -5,27 +5,28 @@
   "private": true,
   "scripts": {
     "clean": "rimraf node_modules",
-    "dev": "TZ=UTC cypress open",
-    "test:examples": "TZ=UTC cypress run --spec features/ui/hooks/*.features,features/ui/components/**/*.features",
-    "test:authenticator": "TZ=UTC cypress run --spec features/ui/components/authenticator/*.features",
-    "test:hooks": "TZ=UTC cypress run --spec features/ui/hooks/*.features",
-    "test:geo": "TZ=UTC cypress run --spec features/ui/components/geo/*.features",
-    "test:theme": "TZ=UTC cypress run --spec features/ui/theme/*.features",
-    "test:guides": "TZ=UTC cypress run --spec features/ui/guides/*.feature"
```
slaymance, line 14: `TZ=UTC` isn't necessary for our use cases since none of our tests are dealing with timezones.

packages/e2e/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "baseUrl": "http://localhost:3000/",
```
slaymance, line 3: Gets rid of warning that tsconfig doesn't specify a `baseUrl`.

packages/e2e/package.json
```diff
@@ -5,27 +5,28 @@
...
+    "dev": "cypress open",
+    "test:examples": "cypress run --spec 'features/ui/hooks/*.feature','features/ui/components/**/*.feature'",
+    "test:authenticator": "cypress run --spec 'features/ui/components/authenticator/*.feature'",
+    "test:hooks": "cypress run --spec 'features/ui/hooks/*.feature'",
+    "test:geo": "cypress run --spec 'features/ui/components/geo/*.feature'",
+    "test:theme": "cypress run --spec 'features/ui/theme/*.feature'",
+    "test:guides": "cypress run --spec 'features/ui/guides/*.feature'"
```
slaymance, line 14: All `.features` files have been removed in favor of the glob `*.feature` pattern. Cypress also now recommends using quotes when working with glob patterns.

packages/e2e/cypress.json
```diff
@@ -1,10 +0,0 @@
-{
-  "baseUrl": "http://localhost:3000/",
-  "defaultCommandTimeout": 15000,
-  "videoUploadOnPasses": false,
-  "env": {
-    "TAGS": "not @skip"
```
wlee221, line 6: Is `not @skip` something we would still need?

I thought this helped us skip tests when we do `yarn e2e dev`.

wlee221 (approved):
Just one non blocking question!

packages/e2e/features/ui/hooks/actions-navigation.feature
reesscot: There was a reason were using test IDs here though. These are testing the three different types of actions whereas not is says nothing about the types of navigations being tested. If there's a better way to do this, I'm all ears, but I think this makes the test less clear now.

examples/next/pages/ui/hooks/actions/NavigateActions.tsx
```diff
@@ -20,15 +20,9 @@ export const NavigateActions = () => {
   };
   return (
     <View>
-      <Button testId="locationChange" onClick={() => goToAmazon()}>
-        Go to amazon.com
-      </Button>
-      <Button testId="reload" onClick={() => reload()}>
```
reesscot, line 26: Not so sure about this change, thoughts detailed in comment on feature file.

reesscot (commented):
Thanks so much for doing these upgrades! I have some concerns about the actions hook test refactors, but otherwise looks solid

packages/e2e/features/ui/hooks/actions-navigation.feature
slaymance: The `Scenario` (the name of the test) blocks describe which action is being tested. I'd say that is more communicative than a test ID like "hash". What do you think?

packages/e2e/cypress.json
slaymance: The library has smart tagging and always skips `@skip` tags now: https://github.com/badeball/cypress-cucumber-preprocessor/blob/master/docs/tags.md#smart-tagging

reesscot (approved):

.github/workflows/publish-next.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       matrix:
         include:
           - path: ./guides/react/protected-routes
-            tags: '@guides and not (@skip or @todo-guides)'
+            tags: '@guides and not @todo-guides'
```
slaymance, line 45: Since `@skip` is supported by default with the latest version of the Cucumber preprocessor, we can remove the handlers for the tag in our Action workflows.

packages/e2e/package.json
```diff
@@ -5,27 +5,29 @@
...
   },
   "cypress-cucumber-preprocessor": {
-    "nonGlobalStepDefinitions": true,
-    "nonGlobalStepBaseDir": "cypress/integration",
-    "stepDefinitions": "features"
+    "filterSpecs": true,
```
slaymance, line 17: This option is nice as it will filter out files where there are no matching tests (e.g. the Geo test files won't be executed for Angular or Vue e2e tests, improving execution time a bit).

wlee221 (approved):
LGTM! 🚢

PR #2558, opened 2022-09-08
chore: upgrade radix to 1.0

### LINKED ISSUE #2089: React 18 peer dependency errors from radix dependencies

React

Other

We're trying to use `@aws-amplify/ui-react` with the React 18 and we're seeing this is not supported:

On `npm install`, with `"react": "^18.0.0",` and `"@aws-amplify/ui-react": "^2.19.1",` in our package.json dependencies, we get the following error:
```
npm ERR! Found: react@18.0.0
npm ERR! node_modules/react
npm ERR!   react@"^18.0.0" from the root project
npm ERR!   peer react@">= 16.8.0" from @aws-amplify/ui-react@2.19.1
npm ERR!   node_modules/@aws-amplify/ui-react
npm ERR!     @aws-amplify/ui-react@"^2.19.1" from the root project
```

[React 18](https://github.com/facebook/react/releases/tag/v18.0.0) is the latest supported release of React and we'd love to see support for it from Amplify!

This is a matter of upgrading the peerDependencies here: https://github.com/aws-amplify/amplify-ui/blob/28bf92ddbd893a3c2839d5215537f845db46b181/packages/react/package.json#L81

- [ ] ⚠️ This feature might incur a breaking change.

reesscot:
#### Description of changes
Upgrade `radix-ui` dependencies to `1.0` to eliminate warning messages about React 18 when using NPM 7. 
 
##### Problems to fix from upgrade to 1.0:

**ALL primitives using Radix (Menu, SliderField, Tabs, Expander)**
BREAKING CHANGE: Radix removed support for the `dir` html attribute and added a `DirectionProvider`. This PR puts the `DirectionProvider` into our `ThemeProvider` and adds a `direction` prop. This is a breaking change since customers may be using `dir="rtl"` and without the `ThemeProvider`, which would break their applications if they are using these four components. We will include this change in the 4.0 release
See: https://github.com/radix-ui/primitives/pull/1119

Note: While I'm not thrilled with exposing a new `direction` prop on the `ThemeProvider`, I feel it is the best solution available to us. This solution avoids affecting the majority of our customers who currently use the left to right languages, while not causing much inconvenience for right to left language apps (simply add the `direction` prop to `ThemeProvider`.

*Radix-based components with ThemeProvider with `direction="rtl"` prop value:*

_SliderField:_
image

_Tabs:_
image

_Expander:_
image

_Menu:_
image

*Non-Radix components with ThemeProvider with `direction="rtl"` prop value:*

_Alert:_
image

General Audit:

**Menu:**
_Problem: Menu dropdown content no longer had high enough z-index_
- Fixed by adding new class to popover wrapper to allow Radix Popper to [infer the correct zIndex](https://github.com/radix-ui/primitives/blob/8f75e5be679b9c269638bddf532a166d80edecb4/packages/react/popper/src/Popper.tsx#L210).

Screenshot showing fix:
CleanShot 2022-08-04 at 00 08 25

**SliderField**
Smoke tested all SlideField API's and no upgrade issues found

**Tabs**
Smoke tested all Tab component APIs and no upgrade issues found.

**Expander**
Smoke tested all Expander component APIs and no upgrade issues found.

#### Issue #, if available
Fixes #2089

### Documentation added:
image

Migration guide:
image

#### Checklist

- [x] Full smoke testing of `SliderField`, `Tabs`, `Menu`, `Expander`
- [x] PR description included
- [x] Add details to migration guide
- [x] Documentation added to use `direction` prop of `ThemeProvider` for language direction
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/package.json
```diff
@@ -57,10 +57,11 @@
   "dependencies": {
     "@aws-amplify/ui": "3.13.2",
     "@aws-amplify/ui-react-v1": "npm:@aws-amplify/ui-react@1.2.9",
-    "@radix-ui/react-accordion": "0.1.6",
-    "@radix-ui/react-dropdown-menu": "0.1.6",
-    "@radix-ui/react-slider": "0.1.4",
-    "@radix-ui/react-tabs": "0.1.5",
+    "@radix-ui/react-accordion": "^1.0.0",
+    "@radix-ui/react-direction": "^1.0.0",
```
reesscot, line 61: This dependency is needed to support the DirectionProvider required in 1.0

packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
```diff
@@ -34,6 +34,32 @@ describe('ThemeProvider', () => {
...
+      </ThemeProvider>
+    );
+
+    expect(container.querySelector(`[data-amplify-theme]`)).toHaveAttribute(
+      'dir',
+      'rtl'
+    );
+  });
+
+  it('takes the colorMode prop and sets [data-amplify-color-mode]', () => {
```
reesscot, line 50: This test is unrelated, but was missing from the suite, so I added it.

reesscot:
Temporarily changing base to main to run the tests, then will change back to next-release/main

docs/src/pages/[platform]/getting-started/internationalization/examples/ThemeProviderDirectionExample.tsx
```diff
@@ -0,0 +1,33 @@
+import { Alert, Text, ThemeProvider } from '@aws-amplify/ui-react';
+import { useState } from 'react';
+
+export const ThemeProviderDirectionExample = () => {
+  return (
+    <>
+      <ThemeProvider>
```
Comment (hbuchel) [inline docs/src/pages/[platform]/getting-started/internationalization/examples/ThemeProviderDirectionExample.tsx:7]: Have we tested how this behaves if you need to nest `dir` attributes? The Radix documentation only mentions wrapping the entire app, but [using the `dir` attribute on block elements](https://www.w3.org/International/questions/qa-html-dir#basicmarkup) is valid if you need to adjust for content that is translated in a language that is a different direction. This is a little weird use case given we suggest NOT nesting ThemeProvider.

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -1,23 +1,42 @@
 import * as React from 'react';
+import { DirectionProvider } from '@radix-ui/react-direction';
 
 import { createTheme, Theme } from '@aws-amplify/ui';
 
 import { AmplifyContext } from './AmplifyContext';
 
 export type ColorMode = 'system' | 'light' | 'dark';
+export type Direction = 'ltr' | 'rtl';
```
hbuchel, line 9: I think [auto](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) is a valid option, does this conflict with Radix's options?

hbuchel (commented):
Couple high level comments (+minor typo fix) to start:

packages/react/src/components/ThemeProvider/index.tsx
reesscot: Yeah, unfortunately radix only allows 'ltr' and 'rtl'
https://github.com/radix-ui/primitives/blob/main/packages/react/direction/src/Direction.tsx#L3

docs/src/pages/[platform]/getting-started/internationalization/examples/ThemeProviderDirectionExample.tsx
Comment (reesscot) [reply docs/src/pages/[platform]/getting-started/internationalization/examples/ThemeProviderDirectionExample.tsx:7]: That's a good callout. You can nest the ThemeProviders so long as you pass the same theme object to both, but it could be result in unexpected CSS variable cascading issues. Do you think this is a blocker? It feels like a bit of an edge case as I would think you would usually set the directionality for your whole app. And for all the other components other than the ones using `radix`, the dir="rtl" attribute would still work fine. For example for a text translation use case, a TextField or Text component would work correctly.

Is it worth calling out in internationalization that only these components need the direction prop on the ThemeProvider? It felts like too much information, I was trying to keep it simple.

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -1,23 +1,42 @@
...
 
 interface ThemeProviderProps {
   children: React.ReactNode;
+  /**
+   * Changes the color mode applied to the theme
```
calebpollman, line 14: Should we add `@description` tags to these props?
```suggestion
   * @description Changes the color mode applied to the theme
```

packages/react/src/components/ThemeProvider/index.tsx
```diff
@@ -1,23 +1,42 @@
...
+  theme?: Theme;
 }
 
 export function AmplifyProvider({
   children,
   colorMode,
-  theme,
+  direction = 'ltr',
   nonce,
+  theme,
```
calebpollman, line 39: 🙏 `A2Z`🙏

calebpollman (approved):
LGTM, just left a non-blocking question

dbanksdesign (approved):
LGTM!

PR #2571, opened 2022-09-09
chore(ci): cleanup environment secrets 

wlee221:
#### Background
We have two environments for GitHub Actions workflow: `CI`, and `deployment`. 

- `CI` holds secrets that are required to run e2e tests, and 
- `deployments` holds secrets related to publishing.

#### Description of changes

This PR cleans up the environment usage so that workflow references the correct environment, and only when they need it.

This also fixes a problem where `NPM_TOKEN` was duplicated between the two environments. Going onward, we will only rely on `deployment` environment to get `NPM_TOKEN` and remove the duplicated one on `CI`.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/.github/workflows/version-packages.yml
+++ b/.github/workflows/version-packages.yml
@@ -17,7 +17,6 @@ permissions:
 jobs:
   version:
     runs-on: ubuntu-latest
-    environment: deployment
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
--- a/.github/workflows/publish-latest.yml
+++ b/.github/workflows/publish-latest.yml
@@ -77,7 +77,7 @@ jobs:
 
   publish:
     runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


.github/workflows/version-packages.yml
```diff
@@ -17,7 +17,6 @@ permissions:
 jobs:
   version:
     runs-on: ubuntu-latest
-    environment: deployment
```
reesscot, line 20: This is removed because versioning the packages doesn't need access to any environment variables, right?

reesscot (approved):
LGTM

.github/workflows/version-packages.yml
wlee221: Right

joebuono (approved):

PR #2581, opened 2022-09-12
chore: remove unused postcssconfig

reesscot:
#### Description of changes
Remove `postcss.config.js` file since it's no longer used.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/.github/workflows/run-and-test-builds.yml
+++ b/.github/workflows/run-and-test-builds.yml
@@ -37,8 +37,6 @@ jobs:
         with:
           node-version: lts/*
           cache: 'yarn'
-      - name: Ignore monorepo postcss.config.js
-        run: mv postcss.config.js _postcss.config.js
       - name: Install canary package
         uses: ./.github/actions/install-with-retries
         with:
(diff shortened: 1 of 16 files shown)


.github/workflows/run-and-test-builds.yml
```diff
@@ -37,8 +37,6 @@ jobs:
         with:
           node-version: lts/*
           cache: 'yarn'
-      - name: Ignore monorepo postcss.config.js
-        run: mv postcss.config.js _postcss.config.js
```
ErikCH, line 41: I thought this was in there because without it the build failed. @wlee221 mentioned it. Maybe it's no longer an issue?

wlee221: Right, this will not be an issue after we move that monorepo postcss config.

wlee221: The problem was that canaries would find this `postcss.config.js` in the monorepo root and try to run postcss. Removing `postcss.config.js` will prevent this from happening at all.

wlee221 (approved):

calebpollman (approved):

.github/workflows/run-and-test-builds.yml
reesscot: With the `postcss.config.js` file gone, it's no longer an issue

PR #2413, opened 2022-08-04
deps: pin @types/react to 17.0.47 in next apps

wlee221:
#### Description of changes

This pins `@types/react` to `17.0.47` in next.js apps, because next.js@11 is incompatible with latest `@types/react@17.0.48`.

See minimal repro: 
- `@types/react@17.0.48` — [fails](https://codesandbox.io/s/next11-with-types-react-17-0-48-vq5fhr?file=/README.md)
- `@types/react@17.0.47` — [works](https://codesandbox.io/s/next11-with-types-react-17-0-47-dgrzuv?file=/README.md)

This has happened a few times before in https://github.com/vercel/next.js/issues/36085 and https://github.com/vercel/next.js/issues/36106, but the fix has been onto `next.js@12` and not 11. We can merge this until we upgrade to better supported version.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2586, opened 2022-09-13
fix(ui-react): update withAuthenticator typing

calebpollman:
#### Description of changes
Update `withAuthenticator` typing to allow more flexibility in `Component` argument and expose HOC `props` for render props pattern
- in `withAuthenticator` generic `Props` param, replace `extend object` with a default value of `{}` to closer align with React internal typing patterns
- create `AuthenticatorHOCProps` utility type, use to type render `props` passed to `Component` param
- replace `Component` typing of `(props: Props) => JSX.Element` to more flexible `React.ComponentType` to prevent type errors when user defined `Component` arg is typed as a `React.FC` or is a `Class` component
- add `@types/react@17.0.47` to _package.json_ resolutions due to a conflict between `React.ReactNode` in `@types/react@17` and `@types/react@18` ([more info here](https://github.com/facebook/react/issues/24304))

#### Issue #, if available
NA

- Manually tested for type safety in sample app:
  - ✅ Functional Component
  - ✅ Functional Component wrapped in `React.forwardRef`
  - ✅ Class Component
- tested `withAuthenticator` typing with `@types/react@18.0.18` to ensure compatibility

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     ]
   },
   "resolutions": {
+    "@types/react": "17.0.47",
     "ansi-regex": "5.0.1",
     "ansi-html": "0.0.8",
     "async": "^3.2.2",
--- a/packages/react/src/components/Authenticator/withAuthenticator.tsx
+++ b/packages/react/src/components/Authenticator/withAuthenticator.tsx
@@ -1,13 +1,18 @@
 import React from 'react';
 import { Authenticator, AuthenticatorProps } from './Authenticator';
+import { UseAuthenticator } from './hooks/useAuthenticator';
 
 export type WithAuthenticatorOptions = Omit<AuthenticatorProps, 'children'>;
+
+type AuthenticatorHOCProps = Partial<
+  Pick<UseAuthenticator, 'signOut' | 'user'>
+>;
 
 /**
  * [📖 Docs](https://ui.docs.amplify.aws/react/connected-components/authenticator)
  */
-export function withAuthenticator<Props extends object>(
-  Component: (props: Props) => JSX.Element,
+export function withAuthenticator<Props = {}>(
+  Component: React.ComponentType<Props & AuthenticatorHOCProps>,
   options: WithAuthenticatorOptions = {}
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     ]
   },
   "resolutions": {
+    "@types/react": "17.0.47",
```
reesscot, line 50: Does this mean that customers will have to resolve to this version in their apps as well?

packages/react/src/components/Authenticator/withAuthenticator.tsx
```diff
@@ -1,21 +1,28 @@
 import React from 'react';
 import { Authenticator, AuthenticatorProps } from './Authenticator';
+import { UseAuthenticator } from './hooks/useAuthenticator';
```
reesscot, line 3: Since this is only used as a type, could we make that clear by using `import type`?
```suggestion
import type { UseAuthenticator } from './hooks/useAuthenticator';
```

reesscot (commented):
Looks good to me, just a few minor questions below

package.json
calebpollman: The resolution only ensures that all of our packages and deps resolve to `@react/types@17.0.47`. Has no impact on users

reesscot (approved):

package.json
wlee221: I think I already asked this before, but why this specific version?

calebpollman: Mostly to align with [this PR](https://github.com/aws-amplify/amplify-ui/pull/2413)

wlee221 (approved):

package.json
zchenwei: The type files will be generated from 17. If we reference some types to build the library from React 17, customers still on 16 will run into missing types error. That's my concern. Actually we should pin it to 16.14 as part of the next breaking change

calebpollman: Discussed offline

PR #2593, opened 2022-09-14
chore(docs): Add fixed email and phone number credentials to mockServices

avirajdh:
#### Description of changes
The PR adds fixed email and phone number variables to handle the different login mechanisms. It also updates the checkFixedCredentials method to include a check for phone numbers with different country codes.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/src/data/amplifyMock/services.ts
```diff
@@ -8,7 +10,28 @@ const verifiedUsers = new Map();
...
+    }
+  } else if (username.length === 10) {
+    user = username;
+  }
+  return (
+    (username === FIXED_USERNAME ||
+      username === FIXED_EMAIL ||
+      user === FIXED_PHONE) &&
+    password === FIXED_PASSWORD
+  );
```
wlee221, line 34: I'd definitiely split this up, this is too complicated and will be hard to maintain later.

What I'd suggest is to break up into three cases: `isUsername`, `isPhone`, `isEmail`, and handle them separately.

```ts
if (isUsername(username)) {
  // some utility function here...
} else if (isPhone(username)) {
  // some utility function here... 
} else if (isEmail(username)) {
  // some utility function here...
 } else {
   // throw some error...
}

wlee221: Talked async, came up with simpler solution

wlee221 (approved):

Issue #2160, opened 2022-06-21
Type error using PasswordField and React 16


reesscot:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

CRA

### Please describe your bug.

When using the PasswordField with React 16 I receive the following error message:

```
Property 'enterKeyHint' is missing in type '{ label: string; defaultCountryCode: string; }' but required in type 'Pick<PrimitivePropsWithRef<PhoneNumberFieldProps, "input">, "as" | "area" | "form" | "label" | "slot" | "style" | "title" | "pattern" | ... 371 more ... | "enterKeyHint">'.ts(2741)
```

This occurs because React 16 doesn't include all the same HTML attributes in the [InputHTMLAttributes type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L2187) as are present in React [17](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v17/index.d.ts#L2194) and [18](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v18/index.d.ts#L2194) types.

### What's the expected behaviour?

No type error for non-required props

### Help us reproduce the bug!

See above

### Code Snippet

```tsx
      <PhoneNumberField label="hello" defaultCountryCode="1" />
```

reesscot:
Also see same error showing on StepperField!

reesscot:
Our type:
```ts
[...]

PR #2594, opened 2022-09-14
BREAKING CHANGE: bump the minimum React version support to 16.14.0

zchenwei:
#### Description of changes

1. As we discussed earlier, there is no reason for us to support React versions before 16.14.0. This is lowest React version Studio supports and has the new JSX transform backported by React team.

2. Update our dev build to align with this minimum React version. This can prevent our release from potentially breaking customers apps that are still running on React before 17 if anything is missing from React 16 but get included in our release. e.g, #2160 

#### Issue #, if available

1. Many of our React related dependencies require a higher React version(>16.9), like testing library. It would be a pain for our library to continue with 16.8

2. Plus, React 16.8 does not support the new JSX transform which generates clean code and ends up reducing the bundle size. 

Testing against a sample app running on React 16.14.0, React 17 and React 18

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/react/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "baseUrl": ".",
     "declaration": true,
     "esModuleInterop": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
zchenwei, line 7: This will turn on the new JSX transform for Typescript compiler.

packages/react/package.json
```diff
@@ -109,8 +109,8 @@
     "eslint-plugin-react-hooks": "^4.4.0",
     "jest": "^27.0.4",
     "jest-matchmedia-mock": "^1.1.0",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "16.14.0",
```
zchenwei, line 112: pin the dev build version to the minimum React version we support in ui react

reesscot:
Should we go ahead and make this change on the branches with ui-react-core as well?

calebpollman:
> Should we go ahead and make this change on the branches with ui-react-core as well?

That's a valid callout, needs to be done on the `in-app-messaging/main` branch. 

@zchenwei Should we also update the pinned version of `@types/react` in the root _package.json_ to point at a `react@16.14` compatible version as well?

zchenwei:
> > Should we go ahead and make this change on the branches with ui-react-core as well?
> 
> That's a valid callout, needs to be done on the `in-app-messaging/main` branch.
> 
> @zchenwei Should we also update the pinned version of `@types/react` in the root _package.json_ to point at a `react@16.14` compatible version as well?

That's a good callout. I am actually going back and forth with it. One the one hand, the tricky thing is there is no one to one match relationship between `types/react` and `react. `types/react` is normally a bit ahead of `react`, so literally we have no idea which version of `types/react` we should target to pin for `react@16.14.0`. Since `react@16.14.0` is the last release of react 16, I think no new types will be release into `types/react@~16.14.0` other than bug fixes or improvement, which still benefit us. Thoughts?

joebuono (dismissed):

zchenwei:
> LGTM! Will this fix the duplicate hooks issues we see building projects in stackblitz?

What was that duplicate hooks issues?

reesscot (commented):
Looks good, we just need to update the migration guide to callout the new minimum React version.

zchenwei:
> Looks good, we just need to update the migration guide to callout the new minimum React version.

Will do once I get the next-release/main branch into mine

docs/src/pages/_app.page.tsx
```diff
@@ -91,7 +91,7 @@ function MyApp({ Component, pageProps }) {
         <ThemeProvider theme={baseTheme} colorMode={colorMode}>
           {
             /* Remove after Global Nav launch START */
-            isGlobalNavEnabled ? (  /* Remove after Global Nav launch END */
+            isGlobalNavEnabled /* Remove after Global Nav launch END */ ? (
```
zchenwei, line 94: Prettier auto change

package.json
```diff
@@ -40,6 +40,7 @@
       "**/@radix-ui",
       "**/@radix-ui/**",
       "**/react",
+      "**/react-dom",
```
zchenwei, line 43: `react` and `react-dom` should be in pair.

package.json
```diff
@@ -48,7 +49,8 @@
     ]
   },
   "resolutions": {
-    "@types/react": "17.0.47",
+    "@types/react": "16.14.0",
+    "@types/react-dom": "16.9.0",
```
zchenwei, line 53: There is no `16.14.x` for `@types/react-dom`. This is the latest minor in 16.

reesscot: Hmm, does this mean we need to do additional testing?

zchenwei:
After discussion with @reesscot offline, we think it is risky to get this out because we do not fully understand how `@types/react@16` will affect users with apps running on a higher react version. It is not worth it taking the risk of making such changes that nobody complains about now

PR #2605, opened 2022-09-20
chore(ui-react-core): migrate useAuthenticator and AuthenticatorProvider

calebpollman:
#### Description of changes
Migrate `useAuthenticator` and `AuthenticatorProvider` to _@aws-amplify/ui-react-core_
- minor config updates to _@aws-amplify/ui-react-core_ to match in-app branch
- expose _@aws-amplify/ui-react-core_ as a public export
- abstract  `useAuthenticator` in to smaller functions and add unit tests
- add `useAuthenticator` and `AuthenticatorProvider` unit tests
- point `useAuthenticator` and `AuthenticatorProvider` imports in _@aws-amplify/ui-react_ at _@aws-amplify/ui-react-core_

#### Issue #, if available
NA

Manually tested, added unit tests, ran integ tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/useAuthenticator.spec.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
```
wlee221, line 1: ✨

wlee221, line 1: Should we keep the test extensions to `.test.tsx` for the sake of consistency? I don't have a preference, but imo we should use one.

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
```diff
@@ -1,162 +0,0 @@
...
-  AuthMachineState,
-  createAuthenticatorMachine,
-  getServiceFacade,
-  listenToAuthHub,
-  AuthenticatorServiceFacade,
-} from '@aws-amplify/ui';
-import { useSelector, useInterpret } from '@xstate/react';
-import isEmpty from 'lodash/isEmpty';
-
-import { areArrayValuesEqual } from '../../../../helpers';
```
wlee221, line 15: Should this helper be removed too?

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+  it('returns false when dep arrays have similar object values', () => {
+    const output = areSelectorDepsEqual([0, { id: 4 }], [0, { id: 4 }]);
+
+    expect(output).toBe(false);
+  });
+
+  it('returns false when nested object values are different', () => {
+    const output = areSelectorDepsEqual(
+      [0, { options: {} }],
+      [0, { options: {} }]
```
wlee221, line 27: LGTM, I wonder we should eventually do deep equality checks. But since that's gonna run on every xstate change, I'm good with avoiding deep equality check computations.

wlee221 (commented):
LGTM, just a few comment.

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/useAuthenticator.spec.tsx
calebpollman: Not super opinionated about aligning/not-aligning, but i do default to _.spec_ personally. Just for awareness:
- our linting and TS configs are set to allow both conventions
- unit tests on the in-app branch use _.spec_

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  it('returns true when arrays are equal length and have same deps', () => {
+    const output = areSelectorDepsEqual([{}, 0], [{}, 0]);
+
+    expect(output).toBe(true);
+  });
+});
+
+describe('getComparator', () => {
+  it('returns a camporator that compares arrays', () => {
```
calebpollman, line 41:
```suggestion
  it('returns a comparator that compares arrays', () => {
```

packages/react/src/components/Authenticator/hooks/useAuthenticator/index.tsx
calebpollman: Yes, good catch

packages/react-core/src/Authenticator/hooks/useAuthenticator/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const USE_AUTHENTICATOR_ERROR =
+  '`useAuthenticator` must be used inside an `Authenticator.Provider`.';
```
joebuono, line 2: Does this need to be added to our i18n translations?

ioanabrooks (approved):

packages/react-core/src/Authenticator/hooks/useAuthenticator/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    const nextDep = nextDeps[index];
+
+    if (
+      areEmptyArrays(currentDep, nextDep) ||
+      areEmptyObjects(currentDep, nextDep)
+    ) {
+      return true;
+    }
+
+    return currentDep === nextDep;
```
joebuono, line 25: Could `currentDep` or `nextDep` be a non-empty array or object?

packages/react-core/src/Authenticator/hooks/useAuthenticator/constants.ts
calebpollman: We don't translate console warning messages

packages/react-core/src/Authenticator/hooks/useAuthenticator/utils.ts
calebpollman: Without doing a deep comparison, comparing the non-empty arrays or objects with the same contents will always return `false`

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/useAuthenticator.spec.tsx
wlee221: I'm not opinionated either, it's just nice when you can search `.test.tsx` on vscode file search window. But anyway it's a very nit thing that I do not need to block on.

wlee221 (approved):

PR #2609, opened 2022-09-20
feat(environments): Add CLI options to pull-environments.sh

wlee221:
#### Description of changes

Adds three flags to `pull-environments.sh`:
- `-r`: aws region to pull from. Defaults to `us-east-2`.
- `-i`: regex pattern that matches environment folders to pull from. Defaults to all folder in current dir to match prev behavior.
- `-e`: regex pattern that matches environment folders to ignore.

It's worth noting that  won't use these flags directly, and we will instead use `yarn` scripts that abstract those flag usages. 

#### Why regex?

We used regex over glob because it's really difficult to specify multiple folder with globs (e.g. match `phone-and-sms-mfa` and `phone-number`). 

Admittedly, it is not intuitive to use regex over glob. But IMO this is the simplest solution to filter multiple environments, which is our purpose. Eventually, we can rewrite this script in Node.js so that it's more readable and extendable. Right now, we're restricted Linux/MacOS and would rather keep this flexible than add complexity.

**Example Usage**:
```bash
# Pull only the auth environments with phone or sms
./pull-environments.sh -r us-east-2 -i "\./auth/auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))"
```

Locally, plus POC workflows.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/.github/workflows/poc-test-environments.yml
+++ b/.github/workflows/poc-test-environments.yml
@@ -44,3 +44,16 @@ jobs:
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_DATASTORE }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_DATASTORE }}
+
+      # new pull-environments.sh tests
+      - name: Pull phone environments only
+        run: yarn environments auth pull:phone
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      - name: Pull all auth environments except phone
+        run: yarn environments auth pull:non-phone
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
--- a/environments/auth/package.json
+++ b/environments/auth/package.json
@@ -4,7 +4,7 @@
   "version": "0.0.1",
   "scripts": {
     "pull": "../pull-environments.sh -r us-east-1",
-    "pull:phone": "../pull-environments.sh -r us-east-1 -i \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\"",
-    "pull:non-phone": "../pull-environments.sh -r us-east-1 -e \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\""
+    "pull:phone": "yarn pull -i \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\"",
+    "pull:non-phone": "yarn pull -e \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\""
   }
 }
--- a/environments/auth/package.json
+++ b/environments/auth/package.json
@@ -3,6 +3,8 @@
   "name": "environments-auth",
   "version": "0.0.1",
   "scripts": {
-    "pull": "../pull-environments.sh"
+    "pull": "../pull-environments.sh -r us-east-1",
+    "pull:phone": "../pull-environments.sh -r us-east-1 -i \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\"",
+    "pull:non-phone": "../pull-environments.sh -r us-east-1 -e \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\""
   }
 }
--- a/environments/pull-environment.sh
+++ b/environments/pull-environment.sh
@@ -6,6 +6,8 @@ IFS='|'
 dir=$1
 region=$2
 
+echo find -regextype help
+
 # In development, AWS_PROFILE should be set. In CI, it's not.
 [ "$AWS_PROFILE" ] && useProfile="true" || useProfile="false"
 
--- a/environments/pull-environment.sh
+++ b/environments/pull-environment.sh
@@ -3,8 +3,8 @@ set -e
 IFS='|'
 
 # Get args
-dir=$1;
-region=$2;
+dir=$1
+region=$2
 
 # In development, AWS_PROFILE should be set. In CI, it's not.
 [ "$AWS_PROFILE" ] && useProfile="true" || useProfile="false"
--- a/environments/pull-environment.sh
+++ b/environments/pull-environment.sh
@@ -6,8 +6,6 @@ IFS='|'
 dir=$1
 region=$2
 
-find -regextype help
-
 # In development, AWS_PROFILE should be set. In CI, it's not.
 [ "$AWS_PROFILE" ] && useProfile="true" || useProfile="false"
 
--- a/environments/pull-environment.sh
+++ b/environments/pull-environment.sh
@@ -2,8 +2,12 @@
 set -e
 IFS='|'
 
+# Get args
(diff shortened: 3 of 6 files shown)


.github/workflows/poc-test-environments.yml
```diff
@@ -44,3 +44,16 @@ jobs:
...
+        run: yarn environments auth pull:phone
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      - name: Pull all auth environments except phone
+        run: yarn environments auth pull:non-phone
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
wlee221, line 59: Only for POC purposes, but we'll do something like this until our tfn is verified.

environments/auth/package.json
```diff
@@ -3,6 +3,8 @@
   "name": "environments-auth",
   "version": "0.0.1",
   "scripts": {
-    "pull": "../pull-environments.sh"
+    "pull": "../pull-environments.sh -r us-east-1",
+    "pull:phone": "../pull-environments.sh -r us-east-1 -i \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\"",
```
wlee221, line 7: While this is slightly ugly with escapings, this lets us filter just the sms/phone number environments.

environments/auth/package.json
```diff
@@ -3,6 +3,8 @@
...
+    "pull:non-phone": "../pull-environments.sh -r us-east-1 -e \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\""
```
wlee221, line 8: Same as above, but pulls everything except those three!

environments/pull-environment.sh
```diff
@@ -2,8 +2,12 @@
 set -e
 IFS='|'
 
+# Get args
+dir=$1
+region=$2
+
 # In development, AWS_PROFILE should be set. In CI, it's not.
-[ "$AWS_PROFILE" ] && useProfile="true" || useProfile="false";
```
wlee221, line 6: Ending line with semicolons in bash scripts are superfluous, because newline acts as command separator anyway. Removed these `;` while I was here.

environments/pull-environment.sh
```diff
@@ -25,12 +28,11 @@ AWSCLOUDFORMATIONCONFIG="{\
 \"profileName\":\"$AWS_PROFILE\",\
 \"accessKeyId\":\"$AWS_ACCESS_KEY_ID\",\
 \"secretAccessKey\":\"$AWS_SECRET_ACCESS_KEY\",\
-\"region\":\"us-east-2\"\
+\"region\":\""$region\""\
```
wlee221, line 31: Now we can control which region we pull from 😈

environments/pull-environments.sh
```diff
@@ -1,11 +1,79 @@
...
+#  -r (region): AWS region to pull environments from. Defaults to us-east-2.
+#  -i (include): Regex pattern that matches environment folders to pull from.
+#  -e (exclude): Regex pattern that matches environment folders to ignore.
+# If both -i and -e are specified, it'll match all dirs that match -i but not -e.
+
+# Examples:
+# - Include only sms/phone environments:
+#   ./pull-environments.sh -r us-east-2 -i "\./auth/auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))"
+
+while getopts ':r:i:e:' OPTION; do
```
wlee221, line 18: This is bash built-in utility that parses flag values. There are better alternatives, but these lack compatibility between different operating systems.

environments/pull-environments.sh
```diff
@@ -1,11 +1,79 @@
...
+
+# Check OS, because `find` implementation is slightly different between OS.
+uname="$(uname)"
+
+# We only support macOS and Linux.
+# TODO: We should rewrite this script in node.js for better cross-os support.
+if [[ "$uname" != "Darwin" && "$uname" != 'Linux' ]]; then
+  echo "ERROR: Unknown os: "$uname""
+  exit 1
+fi
```
wlee221, line 51: Here's the part I don't like. Bash scripts are not perfectly compatible, and even supporting both Linux and MacOS (ie. Darwin) is difficult. 

For now, I only tested against those two, which is imo okay because this is only for internal engineers who do not use windows.

wlee221: (even on linux, i assume wsl would work as it's essentially linux, but this is not something we need to worry about yet).

environments/pull-environments.sh
```diff
@@ -1,11 +1,79 @@
...
+    regexMatch=$(find -E . -regex "$include" -not -regex "$exclude")
+  fi
+else
+  # Linux has find syntax `find -E . [...]`.
+  if [ -z "$exclude" ]; then
+    regexMatch=$(find . -regextype posix-extended -regex "$include")
+  else
+    regexMatch=$(find . -regextype posix-extended -regex "$include" -not -regex "$exclude")
+  fi
+fi
```
wlee221, line 69: Here is where I look for all directories that matches given patterns. We have 4 cases based on two conditionals:

1. whether os is linux or macos (`find` has different flag spec, hence the separation).
2. whether user supplied `ignore` pattern with `-i`.

While we could abstract this further, I laid it out verbosely because abstracting CLI arguments are finicky (whitespace parsing hell).

environments/pull-environment.sh
```diff
@@ -2,8 +2,12 @@
 set -e
```
wlee221, line 2: Note that this is the child helper script that `pull-environment*s*.sh` uses. This file doesn't have code updates, other than accepting one more argument ("region").

environments/pull-environments.sh
```diff
@@ -1,11 +1,79 @@
 #!/bin/bash
 # using bin/bash for macOS and Linux compatibility
-set -e
-IFS='|'
```
wlee221, line 4: This is used for Amplify CLI usages, but is not needed in this file. Child script above handles it.

.github/workflows/poc-test-environments.yml
wlee221: These changes will be reverted when feature branch is merged.

0618 (approved):

environments/pull-environments.sh
```diff
@@ -16,22 +84,21 @@ for dir in ./*/ ; do
 done
 
 # max number of parallel tasks at a time
-numParallelTasks=8; # Future improvement: could set this to # of logical cores in localdevice
+numParallelTasks=8 # Future improvement: could set this to # of logical cores in localdevice
```
0618, line 87: I saw you change the number back in f7e32dccc933bb169fe42635a850e603eb76963e, curious what happened

0618 (approved):

environments/pull-environments.sh
wlee221: I was falling asleep and messed up the number while I was working locally 😅

environments/auth/package.json
slaymance, line 7: Would it make sense to reuse the `pull` script for the `phone` and `non-phone` scripts? So something like:
```
"pull-phone": "yarn pull -i \"\\./auth-with-((phone-and-sms-mfa)|(phone-number)|(totp-and-sms-mfa))\"",
```

slaymance (approved):
Looks excellent! Minor question that isn't blocking.

environments/auth/package.json
wlee221: Ooh, that's a great suggestion! Let me try that.

wlee221: Works like a charm! [07ac032](https://github.com/aws-amplify/amplify-ui/pull/2609/commits/07ac03231ab8844265a48e15b167beb849c427d1)

PR #2621, opened 2022-09-20
chore(ui-react-native): Migrate Checkbox primitive

ioanabrooks:
#### Description of changes

This change:
* migrates the checkbox primitive from react-native example package;
* adds a global storybook decorator and removes all occurences from primitives;
* refactors accessibilityRole for checkbox, heading and label primitives to allow users to specify it, while still providing defaults.

Kapture 2022-09-22 at 11 17 25

Ran storybook and example app in iOS and android.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


examples/react-native/storybook/index.tsx
```diff
@@ -15,6 +21,13 @@ export function setupStorybook(initStorybook: boolean) {
     configure(() => {
       loadStories();
     }, module);
+
+    // add decorators
+    addDecorator((Story: any) => (
```
joebuono, line 26: Thanks for doing this 👍

examples/react-native/storybook/stories/Checkbox.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { storiesOf } from '@storybook/react-native';
+import { Checkbox } from '@aws-amplify/ui-react-native/dist/primitives';
+
+const StatefulCheckbox = ({ ...props }: any) => {
+  const [value, setValue] = useState(props.value);
+  const onChange = (nextValue: string) => {
+    setValue(nextValue);
+  };
+
+  return <Checkbox {...props} value={value} onChange={onChange} size={20} />;
```
joebuono, line 11: Do we think the `Checkbox` should have a `size` prop that matches the [RWA](https://ui.docs.amplify.aws/react/components/checkboxfield#sizes)?

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -7,11 +7,8 @@ import {
...
-import {
-  IconButton,
-  Label,
-} from '@aws-amplify/ui-react-native/dist/primitives';
-
+import { IconButton } from '../IconButton';
+import { Label } from '../Label';
 import { CheckboxProps, CheckboxStyle } from './types';
 
 const styles: CheckboxStyle = {
```
joebuono, line 14: Should we put these styles in a separate file to match the pattern for other primitives in the RN package?

packages/react-native/src/primitives/Checkbox/__tests__/Checkbox.spec.tsx
```diff
@@ -0,0 +1,22 @@
...
+  [true, false].forEach((value) => {
+    it(`renders as expected when selected is ${value}`, () => {
+      const checkbox = create(
+        <Checkbox selected={value} value={value} onChange={onChange} />
+      );
+      expect(checkbox.toJSON()).toMatchSnapshot();
+    });
+  });
+
+  it('renders as expected when disabled', () => {
```
joebuono, line 18: Can we also check that the `selected` state cannot be toggled when the `Checkbox` is disabled?

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -34,12 +31,14 @@ export default function Checkbox<T>({
   value,
   ...rest
 }: CheckboxProps<T>): JSX.Element {
+  const [pressed, setPressed] = useState(selected ?? false);
```
joebuono, line 34: Should `Checkbox` be a controlled or uncontrolled component, or optionally both? ([RWA CheckboxField](https://ui.docs.amplify.aws/react/components/checkboxfield#controlled-component))

joebuono (commented):
Nice job! One global comment: should this component be named `Checkbox` or `CheckboxField`? In the [RWA](https://ui.docs.amplify.aws/react/components/checkboxfield), it's named `CheckboxField`, but there's also a `Checkbox` component [under the hood](https://github.com/aws-amplify/amplify-ui/tree/main/packages/react/src/primitives/Checkbox).

calebpollman (commented):
Great work! Let's sync offline about the handling of `selected` and `value`, not sure if my initial POC was well thought through in that terms of how those props should interact and the event cycle around them

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -72,14 +71,17 @@ export default function Checkbox<T>({
       {label && labelPrecedesIcon ? (
         <Label style={labelStyle}>{label}</Label>
       ) : null}
```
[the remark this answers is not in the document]
ioanabrooks: That takes care of top and bottom positioning but what about start/end?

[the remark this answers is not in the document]
calebpollman: We should be able to remove `labelPrecedesIcon` entirely and just use the `labelPosition` to derive the `flexDirection` value, example:

```
type LabelPosition = 'start' | 'end' | 'top' | 'bottom';

const FLEX_DIRECTIONS: Record<LabelPosition, ViewStyle['flexDirection']> = {
  start: 'row-reverse',
  end: 'row',
  top: 'column-reverse',
  bottom: 'column',
};

const FlexExample = ({ labelPosition }: { labelPosition: LabelPosition }) => {
  const flexDirection = FLEX_DIRECTIONS[labelPosition];

  return (
    <View style={{ flexDirection }}>
      <CheckboxIcon  />
      <Label />
    </View>
  );
};
```

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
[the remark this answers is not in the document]
ioanabrooks: `intermediate ` -> `indeterminate`. As per offline sync, we don't need indeterminate state.

examples/react-native/storybook/stories/Checkbox.stories.tsx
ioanabrooks: Currently the size prop lives on the IconButton but good call, will address this with the Checkbox refactor once we have Icon.

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -67,22 +61,20 @@ export default function Checkbox<T>({
     [buttonStyle]
   );
 
+  // TODO: replace View with Pressable and IconButton with icon once Icon primitive is added
```
calebpollman, line 64: Can we replace the `View` now and just pass `disabled={true}` to the `IconButton`?

packages/react-native/src/primitives/Checkbox/types.ts
```diff
@@ -1,14 +1,14 @@
 import { StyleProp, ViewStyle } from 'react-native';
 
-import { IconButtonProps } from '@aws-amplify/ui-react-native/src/primitives/IconButton';
+import { IconButtonProps } from '../IconButton';
 
 import { LabelProps } from '../Label';
 
 export interface CheckboxProps<T>
-  extends Omit<IconButtonProps, 'source' | 'style'> {
+  extends Omit<IconButtonProps, 'source' | 'style' | 'accessibilityRole'> {
```
calebpollman, line 8: Realize this is a pattern that we started doing earlier but not sure we should be omitting `accessibilityRole`. If we are planning on preventing users from overriding, it should be done on the base component props we are extending from, but ultimately I think we should allow users to define their own values for `accessibilityRole` and providing sensible defaults. Thoughts?

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -72,14 +71,17 @@ export default function Checkbox<T>({
       {label && labelPrecedesIcon ? (
         <Label style={labelStyle}>{label}</Label>
       ) : null}
+
       <IconButton
         {...rest}
+        accessibilityRole={'checkbox'}
```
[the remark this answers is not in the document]
ioanabrooks: Addressing this per offline sync, I misunderstood this comment.

packages/react-native/src/primitives/Checkbox/types.ts
joebuono: I agree with providing sensible defaults, but also allowing users to define their own values for `accessibilityRole`. 

Implementation detail: If we were to allow overrides, should we spread the `rest` props after our 
```jsx
    <View>
      accessibilityRole="checkbox"
      {...rest}
    </View>
```

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -34,24 +25,27 @@ export default function Checkbox<T>({
...
 
-  const containerStyle: ViewStyle = useMemo(
-    () => ({
+  const containerStyle: ViewStyle = useMemo(() => {
+    const FLEX_DIRECTIONS: Record<LabelPosition, ViewStyle['flexDirection']> = {
+      start: 'row-reverse',
+      end: 'row',
+      top: 'column-reverse',
+      bottom: 'column',
+    };
```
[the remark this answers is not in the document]
joebuono: I'm going to use this for `Radio` as well, so maybe we can place it in a `/util` folder?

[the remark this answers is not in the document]
ioanabrooks: We can move it, but we will need to wrap it into its own useMemo() hook, else will trigger re-enders of this memoized containerStyle.

packages/react-native/src/primitives/Checkbox/types.ts
```diff
@@ -19,3 +19,5 @@ export interface CheckboxProps<T>
 export interface CheckboxStyle {
   container: ViewStyle;
 }
+
+export type LabelPosition = 'start' | 'end' | 'top' | 'bottom';
```
joebuono, line 23: Since `Radio` will need the `LabelPosition` type as well, maybe we can put it in a shared types file?

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
ioanabrooks: I'm not really seeing the benefit of that, will need to come back and refactor this anyway.

packages/react-native/src/primitives/Checkbox/types.ts
ioanabrooks: Refactored!

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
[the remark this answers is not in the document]
calebpollman: Can you explain how moving it outside the scope of this component would cause re-renders?

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
calebpollman: IMO the benefit is that it is completed in this PR rather than a follow up, and the follow up is just swapping the `IconButton` out for the `Icon`. I don't feel strongly or anything though

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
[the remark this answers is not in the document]
ioanabrooks: Not moving outside the scope of the comp. I thought we were saying to remove it from the scope of the function but keep it in the scope of component. nvm.

packages/react-native/src/primitives/Checkbox/types.ts
ioanabrooks: Done!

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
ioanabrooks: sure. done!

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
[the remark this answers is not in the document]
ioanabrooks: moved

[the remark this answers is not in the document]
ioanabrooks: pressed -> checked

packages/react-native/src/primitives/Label/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { ViewStyle } from 'react-native';
+import { LabelPosition } from './types';
+
+const FLEX_DIRECTIONS: Record<LabelPosition, ViewStyle['flexDirection']> = {
+  start: 'row-reverse',
+  end: 'row',
+  top: 'column-reverse',
+  bottom: 'column',
+};
+
```
calebpollman, line 10:
```suggestion
 /**
 * The `flexDirection` value returned from `getFlexDirectionFromLabelPosition`
 * assumes a `Label` component follows the `Primary` component it labels inside
 * the `Container` that the `flexDirection` is applied to.
 *
 * Example Usage:
 * ```jsx
 *   const flexDirection = getFlexirectionFromLabelPostiion(labelPosition);
 *   <Container style={{ flexDirection }}>
 *     <Primary />
 *     <Label />
 *   </Container>
 * ```
 */
```

packages/react-native/src/primitives/Label/utils.ts
```diff
@@ -7,7 +7,20 @@ const FLEX_DIRECTIONS: Record<LabelPosition, ViewStyle['flexDirection']> = {
   top: 'column-reverse',
   bottom: 'column',
 };
-
+ /**
```
calebpollman, line 10:
```suggestion

 /**
```

packages/react-native/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -67,22 +57,23 @@ export default function Checkbox<T>({
...
+    >
       <IconButton
         {...rest}
-        disabled={disabled}
-        onPress={handleOnChange}
-        source={selected ? icons.checkboxFilled : icons.checkboxOutline}
+        disabled
+        source={checked ? icons.checkboxFilled : icons.checkboxOutline}
         size={size}
         style={iconButtonStyle}
```
calebpollman, line 74: Let's make sure we clean up the `buttonStyle` prop and `iconButtonStyle` when we replace the `IconButton`

calebpollman (approved):
Thanks for tackling this @ioanabrooks!

joebuono (approved):
Great job!

PR #2628, opened 2022-09-21
feat(ui-react-native): Radio

joebuono:
#### Description of changes

Updated PR in response to feedback. 

In order to allow the `size` prop to accept a `number` or a string union of `'small' | 'medium' | 'large'`, I wrote a few utility functions in `getRadioButtonStyles`. 

https://user-images.githubusercontent.com/48109584/192354990-ef5691a2-8704-445c-95d0-be734c6ecf25.mov

Ran locally in both iOS and Android. 

#### Checklist

- [X] PR description included
- [ ] `yarn test` passes
- [X] Tests are updated


examples/react-native/src/features/InAppMessaging/Demo/Demo.tsx
```diff
@@ -26,7 +26,7 @@ function DemoRadioGroup({ data, ...rest }: any) {
   return (
     <RadioGroup {...rest} labelStyle={[styles.radioGroupLabel, styles.label]}>
       {data.map((item: string) => (
-        <Radio key={item} label={item} size={20} value={item} />
+        <Radio key={item} label={item} size="small" value={item} />
```
joebuono, line 29: I changed the type for the `size` prop to match the [RWA](https://ui.docs.amplify.aws/react/components/radiogroupfield#sizes).

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+        styles.container,
+        containerDirection,
+        ...(disabled ? [styles._disabled] : []),
+        style,
+      ]}
+    >
+      {label && labelPrecedesButton ? (
+        <Label style={labelStyle}>{label}</Label>
+      ) : null}
+      <Pressable onPress={handleOnChange} hitSlop={5}>
```
joebuono, line 54: I still need to handle platform-specific touch target sizes for `hitSlop`

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+      borderRadius: 10,
+    },
+
+    small: {
+      height: 10,
+      width: 10,
+      borderRadius: 10,
+    },
+  },
+  _disabled: {
```
joebuono, line 49: I'm not sure what our naming convention is for different states.

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,44 @@
...
+interface RadioSizes {
+  large: RadioSize;
+  small: RadioSize;
+}
+
+interface DisabledState {
+  opacity: ViewStyle['opacity'];
+}
+
+export interface RadioStyles {
```
joebuono, line 39: Totally open to suggestions about how to type this better.

ioanabrooks:
Leaving a general note, I see the demo videos posted are iOS only, want to make sure that this is tested in android as well, to help uncover any potential issues early. By running `yarn run react-native-example storybook` in one terminal and `yarn run react-native-example dev` in a second one, you can have both simulators open and control stories through the storybook UI.

For styling related questions, I would defer to later when we have theming in place, we will go back and change styling related things when adding theming. But this is a great start!

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+    borderWidth: 2,
+    padding: 5,
+    alignItems: 'center',
+    justifyContent: 'center',
+    borderColor: '#838c95',
+
+    large: {
+      height: 40,
+      width: 40,
+      borderRadius: 40,
```
joebuono, line 22: I'm wondering if it'd be easier to just set an arbitrarily large `borderRadius` in the main `radio` so that it is always a circle regardless of its size. Same goes for `radioButton`.

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+                ...(size ? [styles.radioButton[size]] : []),
+                buttonStyle,
+              ]}
+            />
+          ) : null}
+        </View>
+      </Pressable>
+      {label && !labelPrecedesButton ? (
+        <Label style={labelStyle}>{label}</Label>
+      ) : null}
```
calebpollman, line 69: Maybe we can just have the `label` logic above and let the parent `flexDirection` value control the layout flow? For example when the `label` should render leading the Radio icon using `column-reverse`

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+  value,
+  ...rest
+}: RadioProps<T>): JSX.Element {
+  const handleOnChange = useCallback(() => {
+    if (!disabled) {
+      onChange?.(value);
+    }
+  }, [onChange, value, disabled]);
+
+  const containerDirection: ViewStyle = useMemo(
```
calebpollman, line 28: Can we consolidate all the container style logic here and rename as `containerStyle`?

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+import { RadioStyles } from './types';
+
+export const styles: RadioStyles = StyleSheet.create({
+  container: {
+    alignItems: 'center',
+  },
+  radio: {
+    height: 30,
+    width: 30,
+    borderRadius: 30,
```
calebpollman, line 12: `30` is really large for a default icon on mobile, maybe we can update the size base values to:
- small: 16
- medium: 20
- large: 24

Ultimately the theme will control but think these are sensible defaults. Thoughts?

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+        containerDirection,
+        ...(disabled ? [styles._disabled] : []),
+        style,
+      ]}
+    >
+      {label && labelPrecedesButton ? (
+        <Label style={labelStyle}>{label}</Label>
+      ) : null}
+      <Pressable onPress={handleOnChange} hitSlop={5}>
+        <View style={[styles.radio, ...(size ? [styles.radio[size]] : [])]}>
```
calebpollman, line 55: Do we need both a `View` and a `Pressable`? Thinking we can just add the `style` on the `Pressable`?

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { styles } from './styles';
+
+export default function Radio<T>({
+  buttonStyle,
+  disabled,
+  label,
+  labelPosition = 'end',
+  labelStyle,
+  onChange,
+  selected,
```
calebpollman, line 16: Just a reminder that we need to make sure we have logic in place to handle both controlled and uncontrolled use cases ([React docs](https://reactjs.org/docs/uncontrolled-components.html))

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useCallback, useMemo } from 'react';
+import { Pressable, View, ViewStyle } from 'react-native';
+
+import { Label } from '../index';
```
calebpollman, line 4: Should be able to leave off the `/index`

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import { RadioProps } from './types';
+import { styles } from './styles';
```
calebpollman, line 7: Nit: swap the order of these, `./styles` before `./types`

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+    borderWidth: 2,
+    padding: 5,
```
calebpollman, line 14: We want to avoid using odd numbers for assigning space related values, they prevent centering of layout items when used in theme structures. Our default spacing size will be `8`, and the rest of the values will be derived based on that. Suggest using `4` here

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,37 @@
+import { StyleProp, TextStyle, ViewProps, ViewStyle } from 'react-native';
+
+export interface RadioProps<T> extends ViewProps {
+  buttonStyle?: StyleProp<TextStyle>;
+  // direction can be a prop on RadioGroupField
+  // direction?: 'horizontal' | 'vertical';
```
calebpollman, line 6: Do we need this comment?

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <Label style={labelStyle}>{label}</Label>
+      ) : null}
+      <Pressable onPress={handleOnChange} hitSlop={5}>
+        <View style={[styles.radio, ...(size ? [styles.radio[size]] : [])]}>
+          {selected ? (
+            <View
+              style={[
+                styles.radioButton,
+                ...(size ? [styles.radioButton[size]] : []),
+                buttonStyle,
```
calebpollman, line 61: Should this be `buttonStyle`?

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+          : 'row',
+    }),
+    [labelPosition]
+  );
+
+  const labelPrecedesButton =
+    labelPosition === 'start' || labelPosition === 'top';
+
+  return (
+    <View
```
calebpollman, line 42: If we follow the UX of our React web `RadioGroupField` implementation, the entire `Radio` element should be handle press, so let's replace this `View` with a `Pressable`

calebpollman (commented):
Great start! Let me know if you need clarification on any of the feedback

examples/react-native/storybook/stories/Radio.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    </Screen>
+  ))
+  .add('default', () => <Radio label="Default" value="default" />)
+  .add('label', () => (
+    <>
+      <Radio label="Default" value="default" />
+      <Radio label="Start" labelPosition="start" value="start" />
+      <Radio label="End" labelPosition="end" value="end" />
+      <Radio label="Top" labelPosition="top" value="top" />
+      <Radio label="Bottom" labelPosition="bottom" value="bottom" />
```
ioanabrooks, line 20: Nice coverage of scenarios, stealing this for checkbox :)

packages/react-native/src/primitives/Radio/Radio.tsx
joebuono: As in [something like this](https://github.com/aws-amplify/amplify-ui/blob/in-app-messaging/main/packages/react-native/src/primitives/Checkbox/Checkbox.tsx#L30)?

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,38 @@
...
+export type Sizes = 'small' | 'medium' | 'large';
+
+export interface RadioProps<T> extends ViewProps {
+  disabled?: boolean;
+  label?: string;
+  labelPosition?: LabelPosition;
+  labelStyle?: StyleProp<TextStyle>;
+  onChange?: (value?: T) => void;
+  radioButtonContainerStyle?: StyleProp<ViewStyle>;
+  radioButtonStyle?: StyleProp<ViewStyle>;
```
calebpollman, line 15: Since these aren't really "button" styles, can we use names that are more indicative of their usage? Example:
```suggestion
  radioContainerStyle?: StyleProp<ViewStyle>;
  radioDotStyle?: StyleProp<ViewStyle>;
```

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,38 @@
+import { StyleProp, TextStyle, ViewProps, ViewStyle } from 'react-native';
+
+import { LabelPosition } from '../Label/types';
+
+// This type should live elsewhere
+export type Sizes = 'small' | 'medium' | 'large';
+
+export interface RadioProps<T> extends ViewProps {
```
calebpollman, line 8: Should `RadioProps` extend from `PressableProps`?

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,38 @@
+import { StyleProp, TextStyle, ViewProps, ViewStyle } from 'react-native';
+
+import { LabelPosition } from '../Label/types';
+
+// This type should live elsewhere
+export type Sizes = 'small' | 'medium' | 'large';
```
calebpollman, line 6: Union type naming should be singular form. Also let's add `number` to the union instead of declaring another union on ln 16

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  radioButtonContainer: {
+    alignItems: 'center',
+    borderColor: '#838c95',
+    borderRadius: 999,
+    justifyContent: 'center',
+  },
+
+  // radioButtonContainer sizes
+  large: {
```
calebpollman, line 27: What do you think of removing these "size" related objects and returning the style objects from the dimension creation util?

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,71 @@
...
+    [disabled, labelPosition]
+  );
+
+  return (
+    <Pressable
+      {...rest}
+      accessibilityRole={accessibilityRole}
+      // hitSlop will be platform-specific,
+      // and partially depends on how much spacing the RadioGroupField will apply
+      hitSlop={5}
```
calebpollman, line 47: Do we always want this to be `5`? What are the expected touch target sizes for the platforms?

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useCallback, useMemo } from 'react';
+import { Pressable, View, ViewStyle } from 'react-native';
+
+import { Label } from '../index';
```
calebpollman, line 4: To avoid potential circular dependencies, we should import directly from the `Label` directory

calebpollman (commented):
Getting closer, thanks for making the updates!

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,44 @@
...
+    borderWidth: 4,
```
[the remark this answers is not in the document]
joebuono: IMO, that would make the outlines a bit too thin:
Screen Shot 2022-09-26 at 5 35 14 PM

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,44 @@
...
+  container: {
+    alignItems: 'center',
+  },
+
+  disabled: {
+    opacity: 0.6,
+  },
+
+  radioButton: {
+    backgroundColor: '#317d95',
```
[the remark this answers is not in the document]
joebuono: I just copied the colors from the [React web Radio](https://ui.docs.amplify.aws/react/components/radiogroupfield), which has different colors for the border and the button:
Screen Shot 2022-09-26 at 5 41 52 PM

packages/react-native/src/primitives/Radio/styles.ts
calebpollman: Can we update these sizes?

packages/react-native/src/primitives/Radio/types.ts
joebuono: Ok. Since we are likely to have a `size` prop for other primitives as well, I think we'll eventually want to find another place to define the `Sizes` type.

packages/react-native/src/primitives/Radio/styles.ts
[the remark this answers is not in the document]
calebpollman: Not certain why that was done for web. Let's go ahead and align them to the same color

packages/react-native/src/primitives/Radio/styles.ts
joebuono: We could do that, but I'm wondering how customers would then be able to override the styles for a particular size (e.g., `large`).

packages/react-native/src/primitives/Radio/styles.ts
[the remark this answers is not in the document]
calebpollman: Maybe a default size of `1` is appropriate

packages/react-native/src/primitives/Radio/styles.ts
calebpollman: They would just pass the value they want in the `size` prop?

packages/react-native/src/primitives/Radio/Radio.tsx
joebuono: Looks like the minimum touch target size for [iOS](https://www.w3.org/WAI/GL/mobile-a11y-tf/wiki/M2#:~:text=iOS%20Human%20Interface%20Guidelines%3A%20Adaptivity,pixels%20wide%2044%20pixels%20tall.) is 44px wide and 44px tall. For [Android](https://www.w3.org/WAI/GL/mobile-a11y-tf/wiki/M2#:~:text=Android%20Developers%3A%20Metrics%20and%20Grids%20recommends%20targets%20be%20at%20least%2048dp%20high%20and%20wide%20so%20that%20they%20will%20never%20be%20smaller%20than%20a%20minimum%20recommended%20target%20size%20of%207mm.) it's 48dp X 48dp. 

In addition to taking the platform into consideration, do we want to dynamically calculate `hitSlop` based on the size of the element? For example, if the Radio is 12 high and 40 wide, then we would want the hitSlop to be something like:

```
hitSlop={{ top: 16, right: 2, bottom: 16, left: 2 }}
```
in order to increase the target size to the minimum. 

However, if the element is larger (e.g., 250 X 250), then adding `hitSlop` wouldn't be necessary. 

Thus I'm wondering, if `hitSlop` is something that we want to handle for all of our touchable/pressable primitives, perhaps we should consider writing a util for it? The util could take both the platform and the size of the element, and return the calculated `hitSlop`.

Thoughts?

examples/react-native/storybook/stories/Radio.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { storiesOf } from '@storybook/react-native';
+import { Radio } from '@aws-amplify/ui-react-native/dist/primitives';
+import { Screen } from '../ui';
+
+storiesOf('Radio', module)
+  .addDecorator((Story: any) => (
+    <Screen>
+      <Story />
+    </Screen>
+  ))
```
ioanabrooks, line 12: This is no longer needed.

examples/react-native/storybook/stories/Radio.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+  .add('default', () => <Radio label="Default" value="default" />)
+  .add('label', () => (
+    <>
+      <Radio label="Default" value="default" />
+      <Radio label="Start" labelPosition="start" value="start" />
+      <Radio label="End" labelPosition="end" value="end" />
+      <Radio label="Top" labelPosition="top" value="top" />
+      <Radio label="Bottom" labelPosition="bottom" value="bottom" />
```
ioanabrooks, line 20: If you wanted to, you could use storybook's addon -> knobs:
```
 <Radio labelPosition={select(
        'LabelPosition',
        ['start', 'end', 'top', 'bottom'],
        'end'
      )} />
```

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,37 @@
+import { PressableProps, StyleProp, TextStyle, ViewStyle } from 'react-native';
+
+import { LabelPosition } from '../Label/types';
+
+export type Size = number | 'small' | 'medium' | 'large';
```
ioanabrooks, line 5: This type could be shared among other components.

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface RadioProps<T> extends PressableProps {
+  disabled?: boolean;
+  label?: string;
+  labelPosition?: LabelPosition;
+  labelStyle?: StyleProp<TextStyle>;
+  onChange?: (value?: T) => void;
+  radioContainerStyle?: StyleProp<ViewStyle>;
+  radioDotStyle?: StyleProp<ViewStyle>;
+  size?: number | Size;
```
ioanabrooks, line 15: Size type already has number

joebuono: Good catch

packages/react-native/src/primitives/Radio/types.ts
joebuono: I agree. Is there a place we can put shared types?

packages/react-native/src/primitives/Radio/Radio.tsx
```diff
@@ -0,0 +1,83 @@
...
+  );
+
+  const radioStyle = useCallback(
+    ({ pressed }: PressableStateCallbackType): StyleProp<ViewStyle> => {
+      const pressedStateStyle =
+        typeof style === 'function' ? style({ pressed }) : style;
+      return [containerStyle, pressedStateStyle];
+    },
+    [containerStyle, style]
+  );
```
calebpollman, line 54: Can this be combined?

packages/react-native/src/primitives/Radio/types.ts
calebpollman: Let's move it in to the Icon types once that PR has been merged

packages/react-native/src/primitives/Radio/Radio.tsx
calebpollman, line 54: Can `containerStyle` and `radioStyle` be combined?

packages/react-native/src/primitives/Radio/styles.ts
calebpollman: Un-resolving this comment. Lmk if we need to discuss offline

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -0,0 +1,39 @@
...
+    borderWidth: 1,
+    justifyContent: 'center',
+  },
+  radioDot: {
+    backgroundColor: AMPLIFY_BLUE,
+    borderRadius: ROUNDED_BORDER_RADIUS,
+  },
+  // radioContainer sizes
+  large: {
+    height: 40,
```
calebpollman, line 28: These sizes are really large for mobile, for now let's use:
```
small: 16, 
medium: 20, 
large: 24
```

packages/react-native/src/primitives/Radio/Radio.tsx
joebuono: Yes

packages/react-native/src/primitives/Radio/types.ts
joebuono: But will Icon `size` accept `'small' | 'medium' | 'large'`?

packages/react-native/src/primitives/Radio/getRadioDimensions.ts
```diff
@@ -0,0 +1,40 @@
...
+  radioDotSize: ViewStyle;
+} => {
+  if (typeof size === 'number') {
+    return {
+      radioContainerSize: { height: size, width: size },
+      radioDotSize: { height: size * 0.5, width: size * 0.5 },
+    };
+  }
+
+  const sizeKey = capitalize(size); // 'small' => 'Small'
```
calebpollman, line 35:
```suggestion
const sizeKey = capitalize(size);
```

packages/react-native/src/primitives/Radio/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  radioDotMedium: RadioSize;
+  radioDotSmall: RadioSize;
+}
+
+export interface RadioStyles extends RadioSizes {
+  container: ViewStyle;
+  disabled: ViewStyle;
+  radioContainer: ViewStyle;
+  radioDot: ViewStyle;
+}
```
calebpollman, line 39: Thinking we can replace `ViewStyle` with `RadioDimensions` in the return of `getRadioDimensions`:

```suggestion
export type RadioDimensions = Pick<ViewStyle, 'height' | 'width'>;

export interface RadioStyles {
  container: ViewStyle;
  disabled: ViewStyle;
  radioContainer: ViewStyle;
  radioContainerLarge: RadioDimensions;
  radioContainerMedium: RadioDimensions;
  radioContainerSmall: RadioDimensions;
  radioDot: ViewStyle;
  radioDotLarge: RadioDimensions;
  radioDotMedium: RadioDimensions;
  radioDotSmall: RadioDimensions;
}
```

calebpollman (approved):
Left a couple suggestions but LGTM

ioanabrooks (approved):
LGTM!

calebpollman:
Please select the "squash and merge" option when merging in the future

PR #2651, opened 2022-09-27
feat(ui-react-native): Add icon primitive

ioanabrooks:
#### Description of changes

* Add icon primitve
* Refactor IconButton to use Icon primitive

Ran storybook for iOS and Android sims.

Kapture 2022-09-27 at 11 27 42

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react-native/src/primitives/Icon/Icon.tsx
```diff
@@ -0,0 +1,22 @@
...
+  animated = false,
+  size,
+  style,
+  ...rest
+}: IconProps): JSX.Element {
+  const { icon } = useMemo(() => getStyles(color, size), [color, size]);
+
+  if (animated) {
+    return (
+      <Animated.Image {...rest} style={[icon, { tintColor: color }, style]} />
```
calebpollman, line 18: Looks like `{ tintColor: color }` is already applied in `getStyles`?

packages/react-native/src/primitives/Icon/Icon.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { useMemo } from 'react';
+import { Animated, Image } from 'react-native';
+
+import { getStyles } from './styles';
+import { IconProps } from './types';
+
+export default function Icon({
+  color,
+  animated = false,
```
calebpollman, line 9: Since an `undefined` value for `animated` will evaluate to falsy, do we need the default value here?

calebpollman, line 9: Nit: alphabetize
```suggestion
  animated = false,
  color,
```

packages/react-native/src/primitives/Icon/Icon.tsx
ioanabrooks: Good catch!

packages/react-native/src/primitives/Checkbox/types.ts
```diff
@@ -1,13 +1,11 @@
...
-import { IconButtonProps } from '../IconButton';
-
+import { IconProps } from '../Icon/types';
 import { LabelProps } from '../Label';
 import { LabelPosition } from '../Label/types';
 
-export interface CheckboxProps<T>
-  extends Omit<IconButtonProps, 'source' | 'style'> {
-  buttonStyle?: IconButtonProps['style'];
+export interface CheckboxProps<T> extends Omit<IconProps, 'source' | 'style'> {
```
calebpollman, line 6: This is might be out of scope for this PR but maybe we should readdress the `CheckboxProps`. Thinking that they should extend from `PressableProps`, what do you think?

calebpollman (approved):
LGTM, left a non-blocking question abt future work 🚢

packages/react-native/src/primitives/Checkbox/types.ts
ioanabrooks: Thanks for bringing this up, I thought it looked weird. Addressing now.

packages/react-native/src/primitives/Icon/types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * Color to be applied to the icon
+   */
+  color?: ColorValue;
+
+  /**
+   * Icon size
+   */
+  size?: number;
```
joebuono, line 21: What happens if a customer does not pass something to the `size` prop?

Also, do we want to support `'small' | 'medium' | 'large'` sizes?

examples/react-native/storybook/stories/Icon.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useEffect } from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { Icon } from '@aws-amplify/ui-react-native/dist/primitives';
+import { icons } from '@aws-amplify/ui-react-native/dist/assets';
+import { Animated, Easing } from 'react-native';
+
+const source = icons.close;
+
+const StatefulAnimatedIcon = () => {
```
joebuono, line 9: Cool!

joebuono (approved):
LGTM, just a small non-blocking question

packages/react-native/src/primitives/Icon/types.ts
ioanabrooks: Yes. there is a Size type on the Radio currently, will follow up.

calebpollman (approved):

PR #2662, opened 2022-09-29
chore(ci): Migrate away from adminui environments

wlee221:
#### Description of changes

Previously, we created both `adminui-auth-with-totp-mfa` and `auth-with-totp-mfa` environments, because zero config was unstable back then and there were discrepancies between the two. 

Now these features are stable, and we can migrate away from adminui environments and just use the CLI one. 

I verified that this is the only remnant of `adminui-` environments.

Feature branch will have tests running against it before merge.

#### Checklist

- [x] PR description included


reesscot (approved):
LGTM, thanks for cleaning this up! 💯

ErikCH (approved):
👍

PR #2686, opened 2022-10-04
feat(ui-react-native): Add TextField component

ioanabrooks:
#### Description of changes

This change adds TextField component:
* base component
* unit test
* storybook

Notes:
* styling, size and variant will be added as follow-up with theming.
* iOS simulator is not bringing up the keyboard, should just be a simulator issue.

Ran Storybook. Tested accessibility with Xcode Accessibility Inspector.

Kapture 2022-10-04 at 10 44 30

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react-native/src/primitives/TextField/TextField.tsx
+++ b/packages/react-native/src/primitives/TextField/TextField.tsx
@@ -43,7 +43,9 @@ export default function TextField({
         editable={!disabled}
         style={[styles.input, inputStyle]}
       />
-      {error ? <Label style={errorMessageStyle}>{errorMessage}</Label> : null}
+      {error && errorMessage ? (
+        <Label style={errorMessageStyle}>{errorMessage}</Label>
+      ) : null}
     </View>
   );
 }
--- a/packages/react-native/src/primitives/TextField/TextField.tsx
+++ b/packages/react-native/src/primitives/TextField/TextField.tsx
@@ -6,18 +6,18 @@ import { styles } from './styles';
 import { TextFieldProps } from './types';
 
 export default function TextField({
-  accessible = true,
   accessibilityLabel,
   accessibilityRole,
   accessibilityState,
+  accessible = true,
   containerStyle,
   disabled,
   error,
   errorMessage,
   errorMessageStyle,
+  inputStyle,
   label,
   labelStyle,
-  inputStyle,
   ...rest
 }: TextFieldProps): JSX.Element {
   const inputContainerStyle: ViewStyle = useMemo(
--- a/packages/react-native/src/primitives/TextField/TextField.tsx
+++ b/packages/react-native/src/primitives/TextField/TextField.tsx
@@ -1,0 +1,39 @@
+import React, { useMemo, useState } from 'react';
+import { TextInput, View, ViewStyle } from 'react-native';
+import { Label } from '../Label';
+
+import { styles } from './styles';
+import { TextFieldProps } from './types';
+import { getFlexDirectionFromLabelPosition } from '../Label/utils';
+
+export default function TextField({
+  disabled,
+  label,
+  labelPosition = 'end',
+  labelStyle,
+  style,
+  ...rest
+}: TextFieldProps): JSX.Element {
+  const [value, setValue] = useState('');
+
+  const containerStyle: ViewStyle = useMemo(
+    () => ({
+      ...styles.container,
+      flexDirection: getFlexDirectionFromLabelPosition(labelPosition),
+      ...(disabled && styles.disabled),
+    }),
+    [disabled, labelPosition]
+  );
+
+  return (
+    <View style={[containerStyle, style]}>
+      <TextInput
+        editable={!disabled}
+        value={value}
+        onChangeText={(text) => setValue(text)}
+        {...rest}
+      />
+      {label ? <Label style={labelStyle}>{label}</Label> : null}
+    </View>
+  );
+}
--- a/packages/react-native/src/primitives/TextField/TextField.tsx
+++ b/packages/react-native/src/primitives/TextField/TextField.tsx
@@ -14,9 +14,7 @@ export default function TextField({
   errorMessageStyle,
   label,
   labelStyle,
-  password,
-  textStyle,
-  type = 'default',
+  inputStyle,
   ...rest
 }: TextFieldProps): JSX.Element {
   const inputContainerStyle: ViewStyle = useMemo(
@@ -31,13 +29,10 @@ export default function TextField({
     <View style={[inputContainerStyle, containerStyle]}>
       {label ? <Label style={labelStyle}>{label}</Label> : null}
       <TextInput
-        accessible
-        accessibilityLabel={accessibilityLabel ?? label}
-        style={[styles.text, textStyle]}
-        editable={!disabled}
-        secureTextEntry={password}
-        keyboardType={type}
         {...rest}
+        accessibilityLabel={accessibilityLabel ?? label}
+        editable={!disabled}
+        style={[styles.input, inputStyle]}
       />
       {error ? <Label style={errorMessageStyle}>{errorMessage}</Label> : null}
     </View>
--- a/packages/react-native/src/primitives/TextField/TextField.tsx
+++ b/packages/react-native/src/primitives/TextField/TextField.tsx
@@ -1,39 +1,45 @@
-import React, { useMemo, useState } from 'react';
+import React, { useMemo } from 'react';
 import { TextInput, View, ViewStyle } from 'react-native';
 import { Label } from '../Label';
 
 import { styles } from './styles';
 import { TextFieldProps } from './types';
-import { getFlexDirectionFromLabelPosition } from '../Label/utils';
 
 export default function TextField({
+  accessibilityLabel,
+  containerStyle,
   disabled,
+  error,
+  errorMessage,
+  errorMessageStyle,
   label,
-  labelPosition = 'end',
   labelStyle,
-  style,
+  password,
+  textStyle,
+  type = 'default',
   ...rest
 }: TextFieldProps): JSX.Element {
-  const [value, setValue] = useState('');
-
-  const containerStyle: ViewStyle = useMemo(
+  const inputContainerStyle: ViewStyle = useMemo(
     () => ({
       ...styles.container,
-      flexDirection: getFlexDirectionFromLabelPosition(labelPosition),
       ...(disabled && styles.disabled),
     }),
-    [disabled, labelPosition]
+    [disabled]
   );
 
   return (
-    <View style={[containerStyle, style]}>
+    <View style={[inputContainerStyle, containerStyle]}>
+      {label ? <Label style={labelStyle}>{label}</Label> : null}
       <TextInput
+        accessible
+        accessibilityLabel={accessibilityLabel ?? label}
+        style={[styles.text, textStyle]}
         editable={!disabled}
-        value={value}
-        onChangeText={(text) => setValue(text)}
+        secureTextEntry={password}
+        keyboardType={type}
         {...rest}
       />
-      {label ? <Label style={labelStyle}>{label}</Label> : null}
+      {error ? <Label style={errorMessageStyle}>{errorMessage}</Label> : null}
     </View>
   );
 }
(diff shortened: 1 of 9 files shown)


packages/react-native/src/primitives/TextField/types.ts
```diff
@@ -0,0 +1,76 @@
+import {
+  KeyboardTypeOptions,
+  StyleProp,
+  TextInputProps,
+  TextStyle,
+  ViewStyle,
+} from 'react-native';
+import { LabelProps } from '../Label/types';
+
+export interface TextFieldProps extends TextInputProps {
```
calebpollman, line 10: Can we omit any types that we are renaming (for example `editable`)?

packages/react-native/src/primitives/TextField/TextField.tsx
```diff
@@ -0,0 +1,45 @@
...
+      ...(disabled && styles.disabled),
+    }),
+    [disabled]
+  );
+
+  return (
+    <View style={[inputContainerStyle, containerStyle]}>
+      {label ? <Label style={labelStyle}>{label}</Label> : null}
+      <TextInput
+        accessible
```
calebpollman, line 34: Should this be configurable by users?

packages/react-native/src/primitives/TextField/types.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * @description
+   * Determines which keyboard to open: default or phone number.
+   */
+  type?: Extract<KeyboardTypeOptions, 'default' | 'phone-pad'>;
+}
+
+export interface TextFieldStyles {
+  container: ViewStyle;
+  text: TextStyle;
```
calebpollman, line 74: Can we rename this to `input`?

packages/react-native/src/primitives/TextField/types.ts
```diff
@@ -0,0 +1,76 @@
...
+  textStyle?: StyleProp<TextStyle>;
+
+  /**
+   * @description
+   * Determines which keyboard to open: default or phone number.
+   */
+  type?: Extract<KeyboardTypeOptions, 'default' | 'phone-pad'>;
+}
+
+export interface TextFieldStyles {
```
calebpollman, line 72: Will we add `error` specific styling in a follow up?

packages/react-native/src/primitives/TextField/TextField.tsx
```diff
@@ -0,0 +1,45 @@
...
+    () => ({
+      ...styles.container,
+      ...(disabled && styles.disabled),
+    }),
+    [disabled]
+  );
+
+  return (
+    <View style={[inputContainerStyle, containerStyle]}>
+      {label ? <Label style={labelStyle}>{label}</Label> : null}
```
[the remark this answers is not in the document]
ioanabrooks: I thought about it, we don't support it for web unless I missed it. If we do decide to support labelPosition, we need to consider supporting positioning of the error message as well.

packages/react-native/src/primitives/TextField/TextField.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <TextInput
+        accessible
+        accessibilityLabel={accessibilityLabel ?? label}
+        style={[styles.text, textStyle]}
+        editable={!disabled}
+        secureTextEntry={password}
+        keyboardType={type}
+        {...rest}
```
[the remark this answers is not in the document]
ioanabrooks: good catch, these got out of order when I alphabetically sorted them with vscode

packages/react-native/src/primitives/TextField/TextField.tsx
ioanabrooks: Upon further inspection of APIs, this is actually not needed, all components are accessible by default. Whether we allow configuration can be discussed as follow-up.

packages/react-native/src/primitives/TextField/TextField.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useMemo } from 'react';
+import { TextInput, View, ViewStyle } from 'react-native';
+import { Label } from '../Label';
+
+import { styles } from './styles';
+import { TextFieldProps } from './types';
+
+export default function TextField({
+  accessibilityLabel,
```
joebuono, line 9: Do we need `accessibilityRole` as well?

packages/react-native/src/primitives/TextField/types.ts
```diff
@@ -0,0 +1,76 @@
...
+  StyleProp,
+  TextInputProps,
+  TextStyle,
+  ViewStyle,
+} from 'react-native';
+import { LabelProps } from '../Label/types';
+
+export interface TextFieldProps extends TextInputProps {
+  /**
+   * @description
```
joebuono, line 12: I like that you're annotating the props with comments. Do you think we should do this for the other primitives as well?

ioanabrooks: Not in this PR though lol

packages/react-native/src/primitives/TextField/TextField.tsx
ioanabrooks: I haven't come across an accessibilityRole for input

packages/react-native/src/primitives/TextField/TextField.tsx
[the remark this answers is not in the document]
calebpollman: Agree. Do think there may be a good argument for allowing a leading label but can handled outside of this PR

packages/react-native/src/primitives/TextField/types.ts
ioanabrooks: Yes

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+describe('TextField', () => {
+  it('renders as expected', async () => {
+    const { toJSON, findAllByRole, getByPlaceholderText, getByText } = render(
+      <TextField {...defaultProps} />
+    );
+    expect(toJSON()).toMatchSnapshot();
+
+    expect(await findAllByRole('text')).toHaveLength(1);
+    const textInput = getByPlaceholderText(placeHolderText);
+    expect(textInput.props.editable).toBeTruthy();
```
calebpollman, line 25: Prefer `.toBe(true)` over `toBeTruthy()` for explicitness

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+    expect(textInput.props.secureTextEntry).toBeFalsy();
+    expect(textInput.parent?.props.accessible).toBeTruthy();
```
calebpollman, line 27: Curious abt the optional chaining here? Is it just a TS requirement?

packages/react-native/src/primitives/TextField/TextField.tsx
```diff
@@ -0,0 +1,49 @@
...
+    }),
+    [disabled]
+  );
+
+  return (
+    <View
+      accessible={accessible}
+      accessibilityLabel={accessibilityLabel ?? label}
+      accessibilityRole={accessibilityRole}
+      accessibilityState={{ disabled, ...accessibilityState }}
```
calebpollman, line 36: Just curious, why are we placing these on the container rather than the input?

ioanabrooks: The label gets announced twice by screen readers otherwise.

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
ioanabrooks: yeah ts req, parent may be null.

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
ioanabrooks: done!

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+  it('renders as expected', async () => {
+    const { toJSON, findAllByRole, getByTestId, getByText } = render(
+      <TextField {...defaultProps} />
+    );
+    expect(toJSON()).toMatchSnapshot();
+
+    expect(await findAllByRole('text')).toHaveLength(1);
+    const textInput = getByTestId(testID);
+    expect(textInput.props.editable).toBe(true);
+    expect(textInput.props.secureTextEntry).toBeFalsy();
```
calebpollman, line 26:
```suggestion
    expect(textInput.props.secureTextEntry).toBe(false);
```

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+    expect(label).not.toBeNull();
+  });
+
+  it('renders as expected when disabled', () => {
+    const { toJSON, getByPlaceholderText } = render(
+      <TextField {...defaultProps} disabled />
+    );
+    expect(toJSON()).toMatchSnapshot();
+    const textInput = getByPlaceholderText(placeHolderText);
+    expect(textInput.props.editable).toBeFalsy();
```
calebpollman, line 38:
```suggestion
    expect(textInput.props.editable).toBe(false);
```

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+    expect(getByText(message)).not.toBeNull();
+  });
+
+  it(`doesn't render the errorMessage if error prop is false`, () => {
+    const message = 'Error message';
+    const { toJSON, queryByText } = render(
+      <TextField {...defaultProps} errorMessage={message} />
+    );
+    expect(toJSON()).toMatchSnapshot();
+    expect(queryByText(message)).toBeFalsy();
```
calebpollman, line 62:
```suggestion
    expect(queryByText(message)).toBe(false);
```

packages/react-native/src/primitives/TextField/TextField.tsx
```diff
@@ -0,0 +1,49 @@
...
+      style={[inputContainerStyle, containerStyle]}
+    >
+      {label ? <Label style={labelStyle}>{label}</Label> : null}
+      <TextInput
+        {...rest}
+        accessible={accessible}
+        editable={!disabled}
+        style={[styles.input, inputStyle]}
+      />
+      {error ? <Label style={errorMessageStyle}>{errorMessage}</Label> : null}
```
calebpollman, line 46: Should we also check for an `errorMessage` before rendering the `Label`?

calebpollman (commented):
Just left a minor comment/suggestions

packages/react-native/src/primitives/TextField/__tests__/TextField.spec.tsx
ioanabrooks: This is undefined when not passed in, not false.

packages/react-native/src/primitives/TextField/TextField.tsx
ioanabrooks: Good catch, will add unit test as well.

packages/react-native/src/primitives/TextField/TextField.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { styles } from './styles';
+import { TextFieldProps } from './types';
+
+export default function TextField({
+  accessibilityLabel,
+  accessibilityRole,
+  accessibilityState,
+  accessible = true,
+  containerStyle,
```
joebuono, line 13: I just realized that for most of our primitives, we're passing the `style` prop to the outer container. However, I also like the specificity of naming it `containerStyle`. Maybe we should align on one of these approaches?

ioanabrooks: Not sure I follow, we pass in specific style props on all our primitives.

ioanabrooks: Synced offline, will allign when adding theming, w/ styles for different states and variants.

calebpollman (approved):
LGTM 🚢

PR #2687, opened 2022-10-04
chore(react-native): Migrate RadioGroup primitive

joebuono:
#### Description of changes

This PR migrates the RadioGroup primitive from InAppMessaging to the React Native package. 

https://user-images.githubusercontent.com/48109584/193913774-96d483c5-147e-42bf-9555-3095c2f039fd.mov

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


packages/react-native/src/primitives/RadioGroup/types.ts
```diff
@@ -0,0 +1,26 @@
+import { TextStyle, ViewStyle } from 'react-native';
+import { RadioProps } from '../Radio';
+
+export type Direction = 'vertical' | 'horizontal';
+
+export interface RadioGroupProps<T>
+  extends Pick<
```
calebpollman, line 7: Let's extend from `ViewProps` here as well

packages/react-native/src/primitives/RadioGroup/RadioGroup.tsx
```diff
@@ -1,46 +1,66 @@
...
+import { getFlexDirectionFromLabelPosition } from '../Label/utils';
+import { RadioProps } from '../Radio';
 
+import { styles } from './styles';
 import { RadioGroupProps } from './types';
 
 export default function RadioGroup<T>({
+  accessibilityRole = 'radiogroup',
   children,
+  defaultValue,
```
calebpollman, line 14: Why are there both `defaultValue` and `value` props?

packages/react-native/src/primitives/RadioGroup/__tests__/RadioGroup.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it.each(directions)('renders as expected when size is %s', (value) => {
+    const { toJSON } = render(<ControlledRadioGroup direction={value} />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it.each(sizes)('renders as expected when size is %s', (value) => {
+    const { toJSON } = render(<ControlledRadioGroup size={value} />);
+    expect(toJSON()).toMatchSnapshot();
```
ioanabrooks, line 47: This only tests that the rendered component matches against snapshot.
We should also test that the size prop of the rendered component matches expected value.

packages/react-native/src/primitives/RadioGroup/__tests__/RadioGroup.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <RadioGroup {...props} value={value} onChange={onChange}>
+      <Radio value="option-1" label="Option 1" />
+      <Radio value="option-2" label="Option 2" />
+      <Radio value="option-3" label="Option 3" />
+    </RadioGroup>
+  );
+};
+
+const onChange = jest.fn();
```
ioanabrooks, line 28: This doesn't seem to be used.

packages/react-native/src/primitives/RadioGroup/__tests__/RadioGroup.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+  it.each(sizes)('renders as expected when size is %s', (value) => {
+    const { toJSON } = render(<ControlledRadioGroup size={value} />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('renders as expected when disabled', () => {
+    const { toJSON } = render(<ControlledRadioGroup disabled />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+});
```
ioanabrooks, line 54: Some other unit test ideas to improve coverage:
* users can't interact with radios when component is disabled
* clicking on a radio sets the expected value on the radiogroup
* can we test horizontal/vertical direction?

packages/react-native/src/primitives/RadioGroup/types.ts
```diff
@@ -0,0 +1,26 @@
...
+    RadioProps<T>,
+    | 'accessibilityRole'
+    | 'disabled'
+    | 'label'
+    | 'labelPosition'
+    | 'labelStyle'
+    | 'onChange'
+    | 'size'
+    | 'value'
```
ioanabrooks, line 16: Why not just omit the props we don't want?

calebpollman: `RadioProps` extends from `PressableProps`

ioanabrooks: Ah, makes sense now, ty

examples/react-native/src/ui/index.ts
```diff
@@ -1,13 +1,5 @@
 /**
  * Re-export internal React Native UI primitives and types for use in example apps and storybook.
- * This file should be removed when primitves are exposed for external usage
+ * This file should be removed when primitives are exposed for external usage
```
calebpollman, line 3: 👀

packages/react-native/src/primitives/RadioGroup/RadioGroup.tsx
```diff
@@ -1,34 +1,50 @@
...
   labelStyle,
   onChange,
+  size,
   style,
   value,
   ...rest
 }: RadioGroupProps<T>): JSX.Element {
+  const containerStyle: ViewStyle = useMemo(
+    () => ({
+      style,
```
calebpollman, line 27: Does this need to be spread?

joebuono: When I spread it, TypeScript throws an error: 

```
Spread types may only be created from object types.ts(2698)
```

calebpollman: Ahhh, that's because `ViewProps['style']` can resolve to non-object types. So if i pass a style prop to a `RadioGroup`, does it get applied?

joebuono: Hmm okay, doesn't look like the `style` prop gets applied. I'll look into it some more

packages/react-native/src/primitives/RadioGroup/__tests__/RadioGroup.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <ControlledRadioGroup disabled onChange={onChangeMock} />
+    );
+
+    const option2 = getByText('Option 2');
+
+    fireEvent.press(option2);
+    expect(onChangeMock).not.toHaveBeenCalled();
+  });
+
+  it.each(directions)('renders as expected when size is %s', (value) => {
```
calebpollman, line 87: Should the description be `'renders as expected when direction is %s'`?

joebuono: Yes good catch

ioanabrooks (approved):

calebpollman (approved):

PR #2696, opened 2022-10-05
feat(react-native): ErrorMessage primitive

joebuono:
#### Description of changes

This PR adds the `ErrorMessage` primitive for use in the React Native Authenticator. 

https://user-images.githubusercontent.com/48109584/194184930-6c5944f3-2b2f-4d17-8830-ab511eef47d2.mov

Ran in both iOS and Android. In Android, I noticed that pressing the 'close' button does not always dismiss the `ErrorMessage`. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Text, View } from 'react-native';
+
+import { icons } from '../../assets';
+import { Icon } from '../Icon';
+import { IconButton } from '../IconButton';
+
+import { ErrorMessageProps } from './types';
+import { styles } from './styles';
+
+export default function Label({
```
calebpollman, line 11: Guessing this is meant to a have a different name 😏

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+  accessibilityRole = 'alert',
+  children,
```
calebpollman, line 13: Should this component have `children`?

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+  onDismiss,
+  style,
+  textStyle,
+  ...rest
+}: ErrorMessageProps): JSX.Element | null {
+  const [dismissed, setDismissed] = useState<boolean>(false);
```
calebpollman, line 19: Few things here:
- don't think we need to track whether or not the `ErrorMessage` has been dismissed internally
- should not need to return `null`, if there is not need to render the component the parent should just not render it
- no need to pass `boolean` to the `useState` generic, TS will infer from the default argument

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }, [dismissed, setDismissed, onDismiss]);
+
+  return dismissed ? null : (
+    <View
+      {...rest}
+      accessibilityRole={accessibilityRole}
+      style={[styles.container, style]}
+    >
+      <View style={styles.iconContainer}>
```
calebpollman, line 35: Can we just pass style directly to the `Icon`?

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <Icon source={icons.error} size={20} />
+      </View>
+      <View style={styles.textContainer}>
+        <Text style={[styles.text, textStyle]}>{children}</Text>
+      </View>
+      <View style={styles.iconContainer}>
+        <IconButton
```
calebpollman, line 42: Can we just pass style directly to the `IconButton`?

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
joebuono: Oh wow, embarrassing copy/paste mistake 🙈

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
joebuono: Instead of `children`, should `ErrorMessage` just accept an `errorMessage` prop of type `string`?

packages/react-native/src/primitives/ErrorMessage/styles.ts
```diff
@@ -0,0 +1,29 @@
...
+    flex: 1,
+    height: '100%',
+    justifyContent: 'center',
+  },
+  text: {
+    fontSize: 14,
+  },
+  textContainer: {
+    alignItems: 'flex-start',
+    flex: 4,
```
calebpollman, line 25: Just curious, how did you decide to use `4` here?

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
joebuono: `iconContainer` is being used for layout purposes, in order to set the flex ratios.

packages/react-native/src/primitives/ErrorMessage/styles.ts
joebuono: Arbitrary based on what looked proportional to me

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
joebuono: `iconContainer` is being used for layout purposes, in order to set the flex ratios and center the icon within it.

packages/react-native/src/primitives/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,51 @@
...
+          source={icons.close}
+          size={20}
+          onPress={dismissErrorMessage}
+          testID="dismissButton"
```
ioanabrooks, line 46: I was debating whether to use testIDs internally. It does make testing easier. But we should pick names that are less likely to be picked by our customers, maybe prefix with `rn-amplify-` or similar.

packages/react-native/src/primitives/ErrorMessage/__tests__/ErrorMessage.spec.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react-native';
+import ErrorMessage from '../ErrorMessage';
+
+describe('ErrorMessage', () => {
+  it('renders default ErrorMessage as expected', () => {
+    const { toJSON } = render(
+      <ErrorMessage>Default ErrorMessage</ErrorMessage>
+    );
+    expect(toJSON()).toMatchSnapshot();
```
ioanabrooks, line 10: Is there anything else we expect here, other than matching the snapshot?

packages/react-native/src/primitives/ErrorMessage/__tests__/ErrorMessage.spec.tsx
```diff
@@ -1,34 +1,54 @@
...
+    expect(queryByTestId(CLOSE_BUTTON_TEST_ID)).toBeFalsy();
   });
 
   it('handles an onDismiss callback', () => {
     const customOnDismiss = jest.fn();
     const { queryByRole, getByTestId } = render(
-      <ErrorMessage onDismiss={customOnDismiss}>
-        Test onDismiss handler
-      </ErrorMessage>
+      <OnDismissExample customOnDismiss={customOnDismiss} />
```
calebpollman, line 45: Could we just pass `customOnDismiss` to an `ErrorMessage` directly here?

packages/react-native/src/primitives/ErrorMessage/types.ts
```diff
@@ -1,13 +1,18 @@
...
   onDismiss?: () => void;
-  textStyle?: StyleProp<TextStyle>;
 }
 
 export interface ErrorMessageStyles {
   container: ViewStyle;
-  iconContainer: ViewStyle;
-  text: TextStyle;
-  textContainer: ViewStyle;
+  icon: StyleProp<ImageStyle>;
```
calebpollman, line 16: Does this need the `StyleProp`?

packages/react-native/src/primitives/ErrorMessage/types.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  ImageStyle,
+  StyleProp,
+  TextProps,
+  TextStyle,
+  ViewStyle,
+} from 'react-native';
+
+export interface ErrorMessageProps extends TextProps {
```
calebpollman, line 9: Should probably extend from `ViewProps`?

packages/react-native/src/primitives/ErrorMessage/__tests__/ErrorMessage.spec.tsx
joebuono: Yes, but we need a parent component to `ErrorMessage` in order to test that it gets removed from the screen

calebpollman: Why? 😄

joebuono: Yeah good point 😂. Life is easier when it's a controlled component

calebpollman (approved):

ioanabrooks (approved):

Issue #2598, opened 2022-09-16
@aws-amplify/ui-react@3.4.0 Authenticator component does not work with Gatsby 4.23.0


kota65535:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Gatsby 4.23.0

### What browsers are you seeing the problem on?

Chrome

### Please describe your bug.

This issue is maybe similar to https://github.com/aws-amplify/amplify-ui/issues/1226.

With the latest Gatsby 4.23.0, upgrading from @aws-amplify/ui-react 3.3.0 to 3.4.0 breaks UI interaction of Authenticator component.
I've confirmed the following buttons are not working (nothing happens).

- Sign in button
- Create Account button

### What's the expected behaviour?

- Sign in button should let the user sign in
- Create Account button should let the user create account

### Help us reproduce the bug!

1. Clone [gatsby-starter-default](https://github.com/gatsbyjs/gatsby-starter-default)
2. Rewrite index.js to use Authenticator according to this [tutorial](https://ui.docs.amplify.aws/react/connected-components/authenticator#quick-start)
3. Run `gatsby develop` to start server
4. Access to `localhost:8000` and then click Sign in button

![image](https://user-images.githubusercontent.com/10975834/190554439-b193d3ad-ad03-48ae-b323-8f9db3903c19.png)

### Code Snippet

index.js is as follows.

```javascript
// Put your code below this line.

import * as React from "react"
import awsExports from './aws-exports.json';
import {Amplify} from "aws-amplify";
import {Authenticator} from "@aws-amplify/ui-react";
import '@aws-amplify/ui-react/styles.css';

Amplify.configure(awsExports);

export default function App() {
  return (
    <Authenticator>
      {({signOut, user}) => (
        <main>
          <h1>Hello {user.username}</h1>
          <button onClick={signOut}>Sign out</button>
        </main>
      )}
[...]

PR #2702, opened 2022-10-06
fix(react): Align xstate dependency between UI and framework libraries

slaymance:
#### Description of changes
As uncovered in #2598, the `<Authenticator>` component will not work under certain circumstances when there is a pinned `xstate` dependency higher up the dependency tree. If that pinned dependency doesn't match the version of `xstate` used by `@aws-ampliy/ui` and `@xstate/react` (a dependency of `@aws-amplify/ui-react`), then the core `xstate` dependency is never de-duped and `@aws-amplify/ui` and `@xstate/react` point to different instances of the `xstate` package. When this 
occurs, the `<Authenticator>` won't function properly.

This is an issue we introduced by not ensuring that the instance of `xstate` used by our UI and React libraries are always in sync. This pull request addresses that issue by making `xstate` a `peerDependency` of `@aws-amplify/ui` and `@aws-amplify/ui-react-core` and a core dependency of our framework libraries. This will guarantee that the instance of `xstate` used by our libraries is always the one installed with the framework library added by users.

#### Issue #, if available
#2598 

### Existing problem
The example application below was built using the starter Gatsby app described in #2598 which includes a pinned `xstate` dependency.

https://user-images.githubusercontent.com/26472139/194409254-eafab5bd-7fbe-4ab5-b447-0abeadf234e0.mp4

Below is a screenshot demonstrating how the instances of `xstate` used by `@aws-amplify/ui` and `@xstate/react` are out of sync.
Screen Shot 2022-10-06 at 12 12 51

### Fixed with Changes
https://user-images.githubusercontent.com/26472139/194409811-3d74c9b6-87a5-4216-bb2c-28392f59cf98.mp4

Below you can see that `xstate` is de-duped and the same instance is used by `@aws-amplify/ui` and `@xstate/react`, even though there is a pinned `xstate` dependency higher up the dependency tree.
Screen Shot 2022-10-07 at 10 47 22

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/react-core/tsconfig.cjs.json
```diff
@@ -4,5 +4,11 @@
     "module": "CommonJS",
     "outDir": "dist"
   },
-  "exclude": ["node_modules", "src/**/*.test.ts", "src/**/*.test.tsx"]
+  "exclude": [
+    "node_modules",
+    "src/**/*.spec.ts",
+    "src/**/*.spec.tsx",
```
slaymance, line 10: Unrelated to this PR, but we also needed to exclude `.spec` files for building.

packages/react-core/package.json
```diff
@@ -34,7 +34,13 @@
...
     "aws-amplify": "3.x.x || 4.x.x",
     "react": ">= 16.8.0",
-    "react-dom": ">= 16.8.0"
+    "react-dom": ">= 16.8.0",
+    "xstate": "^4.33.6"
+  },
+  "peerDependenciesMeta": {
+    "xstate": {
+      "optional": true
+    }
```
slaymance, line 43: `xstate` also needs to be deduped between React and react-core libraries.

packages/react/package.json
```diff
@@ -72,7 +72,8 @@
     "qrcode": "1.5.0",
     "react-generate-context": "1.0.1",
     "react-map-gl": "7.0.15",
-    "tslib": "2.4.0"
+    "tslib": "2.4.0",
+    "xstate": "^4.33.6"
```
slaymance, line 76: `xstate` becomes a core dependency of `@aws-amplify/ui-react` and will be used by `@aws-amplify/ui`, `@aws-amplify/ui-react-core`, and `@xstate/react`.

packages/ui/package.json
```diff
@@ -66,7 +65,8 @@
     "type-fest": "^2.3.4"
   },
   "peerDependencies": {
-    "aws-amplify": "3.x.x || 4.x.x"
+    "aws-amplify": "3.x.x || 4.x.x",
+    "xstate": "^4.33.6"
```
slaymance, line 69: `xstate` is moved from a core dependency to a peer dependency of `@aws-amplify/ui`. This ensures the package will used the instance of `xstate` installed with the framework library of choice.

packages/vue/package.json
```diff
@@ -39,7 +39,8 @@
     "@vueuse/core": "7.5.5",
     "@xstate/vue": "0.8.1",
     "nanoid": "3.1.31",
-    "qrcode": "1.5.0"
+    "qrcode": "1.5.0",
+    "xstate": "^4.33.6"
```
slaymance, line 43: `xstate` is now included as a core dependency of `@aws-amplify/ui-vue`.

packages/react-core/package.json
calebpollman: Can we just put `xstate` and `xstate/react` in _ui-react-core_ as direct deps and leave them out of _ui-react_?

packages/react-core/tsconfig.cjs.json
calebpollman: Good catch

packages/react-core/package.json
slaymance: Yeah, you're right. Since react-core is handling all the xstate functionality, we can remove them from the react library.

slaymance: I need to test it since react and react-core both use the `@aws-amplify/ui` package.

packages/react-core/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@aws-amplify/ui": "3.14.0",
-    "@xstate/react": "3.0.0"
+    "@xstate/react": "3.0.1"
```
wlee221, line 32: Just curious, is there a reason for this bump?

packages/react/package.json
```diff
@@ -62,7 +62,6 @@
     "@radix-ui/react-dropdown-menu": "0.1.6",
     "@radix-ui/react-slider": "0.1.4",
     "@radix-ui/react-tabs": "0.1.5",
-    "@xstate/react": "3.0.0",
```
slaymance, line 65: We can safely remove this package since it's part of the _react-core_ library.

packages/react-core/package.json
```diff
@@ -29,7 +29,8 @@
   },
   "dependencies": {
     "@aws-amplify/ui": "3.14.0",
-    "@xstate/react": "3.0.0"
+    "@xstate/react": "3.0.1",
```
slaymance, line 32: This package needed to be bumped so that the minimum version of `xstate` is in-line with what we install.

packages/react-core/package.json
slaymance: It's safe to remove xstate from the react package, done!

packages/react-core/package.json
slaymance: The latest version of `@xstate/react` has a peer dependency on the same minimum version of `xstate` that we use. This is needed for proper deduping to happen and for `@xstate/react` to not use a different instance of the package than `@aws-amplify/ui` does.

packages/ui/package.json
```diff
@@ -42,9 +42,17 @@
...
-    "xstate": "^4.33.6",
     "tslib": "2.4.0"
   },
+  "peerDependencies": {
+    "aws-amplify": "3.x.x || 4.x.x",
+    "xstate": "^4.33.6"
+  },
+  "peerDependenciesMeta": {
+    "xstate": {
+      "optional": true
```
wlee221, line 53: QQ, why is this optional?

wlee221 (approved):
LGTM, just one q 👍

packages/ui/package.json
reesscot: I think this keeps NPM 8+ from automatically downloading it.

wlee221: Gotcha, it makes sense to do this then

slaymance: Also, we use the UI package in 2 places: once in the _react-core_ package and once in the _react_ package. The _react-core_ is the only one that uses the `xstate` functionality, while _react_ uses the styling aspects. Thus, it makes sense to treat `xstate` as an optional peer dependency since it isn't used by `@aws-amplify/ui-react`.

reesscot (approved):
LGTM

PR #2744, opened 2022-10-14
chore(readme): add badgesfor downloads and version info

0618:
#### Description of changes

- add Amplify UI logo
- Add this table
- add bundle size

| Package Name                                                                     | Downloads                                                                                       | Version                                                                 |
| -------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------- |
| [@aws-amplify/ui-react](https://www.npmjs.com/package/@aws-amplify/ui-react)     |  |           |
| [@aws-amplify/ui-vue](https://www.npmjs.com/package/@aws-amplify/ui-vue)         |               |  |
| [@aws-amplify/ui-angular](https://www.npmjs.com/package/@aws-amplify/ui-angular) |           |         |

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/vue/README.md
```diff
@@ -1,3 +1,6 @@
+![](https://img.shields.io/npm/dw/@aws-amplify/ui-vue?label=Download&logo=Amplify)  
+![](https://img.shields.io/npm/v/@aws-amplify/ui-vue/latest?style=flat)
```
hbuchel, line 2: These show on two lines for me when I look at the rich diff (assuming you mean for them to be single line like the others)

ioanabrooks: These should render in a single line in markdown, there would have to be another new line in between these two for them to render separately.

ioanabrooks (approved):

packages/vue/README.md
hbuchel: Hmm maybe just an issue with the preview then? Seems to be just for this one. 
Screen Shot 2022-10-14 at 11 39 07 AM

hbuchel (approved):
Left a comment about two of the badges showing on two lines in the preview, but otherwise LGTM!

packages/vue/README.md
0618: Those are the same format as in the PR description, so should be in a table

0618: Interesting, looks like the README can be rendered to the same line since that's how we do it in the current README [code](https://github.com/aws-amplify/amplify-ui/blob/194ba9fff4ea20fd87d6b40d4c98b19f77bad727/README.md?plain=1#L7-L11).

But to be safe, I can change the code to guarantee that

reesscot (commented):
Can't we use the Amplify UI logo that's already in the repo?

README.md
```diff
@@ -1,8 +1,11 @@
-<img src="https://s3.amazonaws.com/aws-mobile-hub-images/aws-amplify-logo.png" alt="AWS Amplify" width="225">
+<img src="https://s3.amazonaws.com/aws-mobile-hub-images/aws-amplify-logo.png" alt="AWS Amplify Logo" width="225">
```
joebuono, line 1: Not sure if it really makes a difference, but can we use the Amplify Logo in `./docs/public/amplify-logo.svg`?

0618: Those 2 are different. ./docs/public/amplify-logo.svg only has the logo, without text

reesscot:
image
Showing double logos

docs/public/svg/amplify-ui-logo.svg
```diff
@@ -0,0 +1,10 @@
+<svg class="docs-logo-svg" viewBox="0 0 644 125" version="1.1" xmlns="http://www.w3.org/2000/svg">
```
dbanksdesign, line 1: Did you mean to use this in the README? It does not appear to be used.

README.md
```diff
@@ -1,8 +1,11 @@
-<img src="https://s3.amazonaws.com/aws-mobile-hub-images/aws-amplify-logo.png" alt="AWS Amplify" width="225">
+<img src="./docs/public/amplify-logo.svg" alt="AWS Amplify Logo" width="25"> AWS Amplify
 
 ---
 
-# Amplify UI
+<p align="center">
+  <img src="docs/public/svg/favicon.svg" width="150vw" alt="Amplify UI logo" />
```
dbanksdesign, line 6: I don't think we want `150vw` for the width here? The `vw` unit is based on the width of the browser and this would be 150% of the browser's width. Also the `width` attribute on an image is a unitless attribute, this would just mean "150px".

docs/public/svg/amplify-ui-logo.svg
0618: I used it in a previous commit, but since it doesn't work with Github dark mode, I didn't use it in the new commit. Let me remove it.

README.md
0618: Good catch. Updated.

docs/public/svg/amplify-ui-logo.svg
0618: Removed.

dbanksdesign (approved):
LGTM

reesscot (approved):

PR #2751, opened 2022-10-17
chore(primitives): Move `ComponentClassName` to `@aws-amplify/ui`

wlee221:
#### Description of changes

Moving `ComponentClassName` variable to `@aws-amplify/ui`, so other frameworks can reference those directly rather than hardcoding them. 

Note that `ComponentClassName` was not exported `@aws-amplify/ui-react` in the first place, so this is not an API change. React was using `ComponentClassNames`, which is an extension from `ComponentClassName`, and that one has not been touched.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react/src/primitives/Collection/__tests__/Collection.test.tsx
```diff
@@ -6,9 +6,8 @@ import kebabCase from 'lodash/kebabCase';
 import { Collection } from '../Collection';
 import { Flex } from '../../Flex';
 import { Text } from '../../Text';
+import { ComponentClassNames, ComponentText } from '../../shared/constants';
 import { ComponentPropsToStylePropsMap } from '../../types';
-import { ComponentText } from '../../shared/constants';
-import { ComponentClassName } from '../../shared/types';
```
wlee221, line 11: This was the only test that was using `ComponentClassName` instead of `ComponentClassNames`. Updated to use the latter.

packages/ui/src/types/primitives/constants.ts
```diff
@@ -0,0 +1,117 @@
+export enum ComponentClassName {
```
calebpollman, line 1: Maybe consider naming the file _componentClassName.ts_? Thinking it might be confusing for future readers to find a _constants.ts_ nested inside a _types_ dir

calebpollman (commented):
LGTM, just left a question

packages/ui/src/types/primitives/constants.ts
wlee221: Good point! Done in [3c3e5f0](https://github.com/aws-amplify/amplify-ui/pull/2751/commits/3c3e5f059465c40bfac74e7c661726a99003505f).

wlee221: Looking forward, we might need a separate file for `Authenticator` class names. That's something we can strategize and adjust filenames later though, so don't want to worry about it now.

packages/react/src/primitives/Collection/__tests__/Collection.test.tsx
calebpollman: Might be out of scope for this PR, but pretty we can just delete `ComponentClassNames` and replace usage directly with `ComponentClassName`

calebpollman (approved):
LGTM, just left a non-blocking comment

packages/ui/src/types/primitives/constants.ts
reesscot: If we're going to expose this publicly via ui package, wouldn't we need to do that change now? Once this is exposed we can't change it.

zchenwei (approved):
LGTM

packages/react/src/primitives/Collection/__tests__/Collection.test.tsx
wlee221: +1, we have some non-official exports from `@aws-amplify/ui-react` that should be cleaned up. Out of scope for this PR however.

PR #2759, opened 2022-10-19
feat(rna): add Authenticator component

calebpollman:
#### Description of changes
Add base `Authenticator` component and unit tests

#### Issue #, if available
NA

Manually testing, unit testing
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react-native/src/Authenticator/Authenticator.tsx
+++ b/packages/react-native/src/Authenticator/Authenticator.tsx
@@ -1,0 +1,62 @@
+import React, { useMemo } from 'react';
+
+import {
+  AuthenticatorProvider as Provider,
+  AuthenticatorComponentDefaults,
+  AuthenticatorMachineContext,
+  resolveAuthenticatorComponents,
+  useAuthenticator,
+  useAuthenticatorRoute,
+  useAuthenticatorInitMachine,
+  UseAuthenticator,
+} from '@aws-amplify/ui-react-core';
+
+import { AuthenticatorProps, FormHandlers } from './types';
+
+const DEFAULTS = {
+  // TODO: remove cast after adding missing components
+} as unknown as AuthenticatorComponentDefaults<FormHandlers>;
+
+const isAuthenticatedRoute = (route: UseAuthenticator['route']) =>
+  route === 'authenticated' || route === 'signOut';
+
+const routePropSelector = ({
+  route,
+}: AuthenticatorMachineContext): AuthenticatorMachineContext['route'][] => [
+  route,
+];
+
+function Authenticator({
+  children,
+  components: overrides,
+  ...options
+}: AuthenticatorProps): JSX.Element | null {
+  useAuthenticatorInitMachine(options);
+
+  const { route } = useAuthenticator(routePropSelector);
+
+  const components = useMemo(
+    () => resolveAuthenticatorComponents(DEFAULTS, overrides),
+    [overrides]
+  );
+
+  const { Component, props } = useAuthenticatorRoute({ components });
+
+  if (isAuthenticatedRoute(route)) {
+    return children ? <>{children}</> : null;
+  }
+
+  return (
+    <Component
+      {...props}
+      onBlur={() => null}
+      onChangeText={() => null}
+      onSubmit={() => null}
+    />
+  );
+}
+
+// assign slot components
+Authenticator.Provider = Provider;
+
+export default Authenticator;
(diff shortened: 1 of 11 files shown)


packages/react-native/src/Authenticator/__tests__/Authenticator.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+        () => ({ route } as unknown as UseAuthenticator)
+      );
+
+      const { toJSON } = render(
+        <Authenticator>
+          <TestChildren />
+        </Authenticator>
+      );
+
+      expect(toJSON()).toMatchSnapshot();
```
ioanabrooks, line 77: Any other assertions we want to add, besides snapshot testing? Snapshots offer no guarantees that appearance or behavior is correct/as expected.

Can we also add some tests for unhappy path or any edge cases?

packages/react-native/src/Authenticator/Authenticator.tsx
```diff
@@ -0,0 +1,62 @@
...
+  UseAuthenticator,
+} from '@aws-amplify/ui-react-core';
+
+import { AuthenticatorProps, FormHandlers } from './types';
+
+const DEFAULTS = {
+  // TODO: remove cast after adding missing components
+} as unknown as AuthenticatorComponentDefaults<FormHandlers>;
+
+const isAuthenticatedRoute = (route: UseAuthenticator['route']) =>
```
ioanabrooks, line 20: Would this be better placed in a separate utils file?

calebpollman: If we need it in other places we can abstract it out but since it's such a minimal util think it's fine here for now

packages/react-native/src/Authenticator/__tests__/Authenticator.spec.tsx
calebpollman: Since we are just checking to make sure that the `children` argument is rendered for these `route` values, think we are okay with the snapshot but curious what other assertions/edge cases you are thinking?

ioanabrooks: I was thinking to give the Child component a testID:
```
<Text testId="testChild">Test Children</Text>
```
and then actually checking that it exists in the tree. snapshot testing is only good for checking if UI changed unexpectadly, it doesn't tell you the child rendered.

ioanabrooks: For the edge cases nothing particular off the top of my head, maybe like if you pass in an invalid component as child, does it blow up? Things liek that.

packages/react-native/src/Authenticator/Authenticator.tsx
ioanabrooks: How would one discover that this util exists here and not duplicate it in another component?

calebpollman: How would it living in a _util.ts_ as a sibling to _Authenticator.tsx_ component be any different? 

If we really want to think about whether this will be reused, we are not exposing `route` anywhere else in the UI besides in `Authenticator`, so unless we need to suddenly change the architecture of the functional components, there is really no chance that this util will ever be reused

packages/react-native/src/Authenticator/__tests__/Authenticator.spec.tsx
calebpollman: > snapshot testing is only good for checking if UI changed unexpectedly, it doesn't tell you the child rendered.

But the child is what the snapshot contains right?

> maybe like if you pass in an invalid component as child, does it blow up?

Is that a test we should write? Like at that point we are just testing the React Native library itself right?

packages/react-native/src/Authenticator/Authenticator.tsx
```diff
@@ -0,0 +1,62 @@
...
+  route === 'authenticated' || route === 'signOut';
```
joebuono, line 21: Are these the only two routes, or will we add more? If there's more, maybe we can put the routes in an array and use the `.includes` method

joebuono (approved):

packages/react-native/src/Authenticator/__tests__/Authenticator.spec.tsx
ioanabrooks: thanks for adding assertions!

ioanabrooks (approved):
LGTM

joebuono (approved):

PR #2765, opened 2022-10-19
feat(rna): FederatedProviderButton

joebuono:
#### Description of changes

Adds the `FederatedProviderButton` for the RNA. 

The `.png` files are placeholders for now until we decide on an official source for these assets. 

FederatedProviderButton

#### Checklist

- [X] PR description included
- [ ] `yarn test` passes
- [X] Tests are updated


packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+      const pressedStateStyle =
+        typeof style === 'function' ? style({ pressed }) : style;
+      return [styles.container, pressedStateStyle];
+    },
+    [style]
+  );
+
+  return (
+    <Button {...rest} onPress={onPress} style={containerStyle}>
+      <View style={styles.icon}>{Icon}</View>
```
ioanabrooks, line 33: Why not use Icon directly here and just pass in source?

packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+  View,
+  ViewStyle,
+} from 'react-native';
+
+import { Button, Label } from '..';
+
+import { styles } from './styles';
+import { FederatedProviderButtonProps } from './types';
+
+export default function FederatedProviderButton({
```
ioanabrooks, line 14: Wondering if this is a primitive or if it is Authenticator specific.  Thinking this should be an authenticator component.

packages/react-native/src/primitives/FederatedProviderButton/types.ts
```diff
@@ -0,0 +1,12 @@
+import { TextStyle, ViewStyle } from 'react-native';
+import { ButtonProps, IconProps } from '..';
+
+export interface FederatedProviderButtonProps extends ButtonProps {
+  Icon: React.ReactElement<IconProps>;
```
ioanabrooks, line 5: uppercase?

packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
calebpollman: This was part of the ask since we don't have a directory for common components yet. Now thinking we can just put it in _src/Authenticator/Defaults/common/_ since I've had more time to think through it

packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
calebpollman, line 33: Why does it need to be wrapped in a `View`? Can we just just apply the style to the `Icon`:
```
<Icon style={styles.icon} />
```

joebuono: IconCode

packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
joebuono: Yes, this will be moved into the `./Authenticator` folder once the [template PR](https://github.com/aws-amplify/amplify-ui/pull/2769) gets merged

packages/react-native/src/primitives/FederatedProviderButton/types.ts
joebuono: Because it is a component

packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
calebpollman: Thinking that might be because the prop is `React.ReactElement` and not `React.ComponentType`?

joebuono: Good question. @calebpollman?

joebuono: That fixes it, but then the `Icon` component complains that it's missing the `source` prop

packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children,
+  Icon,
+  onPress,
```
calebpollman, line 17: Nit: unnecessary destructure

packages/react-native/src/primitives/FederatedProviderButton/FederatedProviderButton.tsx
calebpollman: Maybe we should just pass the ` source` prop like @ioanabrooks suggested

packages/react-native/src/Authenticator/common/FederatedProviderButton/types.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { ButtonProps } from '../../../primitives';
+
+export interface FederatedProviderButtonProps extends ButtonProps {
+  source: ImageSourcePropType;
+}
+
+export interface FederatedProviderButtonStyles {
+  container: ViewStyle;
+  icon: StyleProp<ImageStyle>;
```
calebpollman, line 17: Just curious, why do we need the `StyleProp` here?

calebpollman (approved):
🚢

packages/react-native/src/Authenticator/common/FederatedProviderButton/types.ts
joebuono: We could just do `ImageStyle`

ioanabrooks (approved):

calebpollman (approved):

PR #2781, opened 2022-10-21
fix: ensure TextField, TextAreaField, StepperField show full width

reesscot:
#### Description of changes
When `TextField`, `TextAreaField`, `StepperField` are wrapped in a Flex with `alignItems` set to something other than the default `stretch` value, the input fields are not full width. This fixes that issue by setting default styling on the fieldgroup wrapper to `align-self: stretch` and moving the `alignItems` prop to the parent container (Flex), as is already planned in V5. This is not a breaking change since the default theme was designed for the input fields to take up the full width of their container. In addition, the `alignItems` prop has not affect on the input where it was placed before since it's not directly inside of a flex container.

Loaded all field primitives into an example app and tested different Flex container properties to ensure the expected behavior was shown when setting `alignSelf` and `alignItems` to the flex children. 

Before (with `<Flex direction="column" alignItems="start">` and `<[FieldComponent] alignSelf="stretch"/>`)
image

After (with `<Flex direction="column" alignItems="start">` and `<[FieldComponent] alignSelf="stretch"/>`):
image

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


dbanksdesign (approved):
LGTM!

awinberg-aws (approved):

zchenwei (approved):
LGTM

packages/react/src/primitives/StepperField/StepperField.tsx
```diff
@@ -43,6 +43,7 @@ const StepperFieldPrimitive: Primitive<StepperFieldProps, 'input'> = (
     variation,
     testId,
 
+    alignSelf, // @TODO: remove custom destructuring for 3.0 release
```
calebpollman, line 46: When is the 3.0 release?

calebpollman (approved):
🚢

dbanksdesign (approved):

packages/react/src/primitives/StepperField/StepperField.tsx
reesscot: I know, didn't want to update all the other lines since they also have the wrong release number 😱. This line will be completely removed in 5.x

PR #2791, opened 2022-10-24
chore(ui-react-native): Add default footer

ioanabrooks:
#### Description of changes

This change adds a default footer for Authenticator sub components.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react-native/src/Authenticator/Defaults/ConfirmResetPassword/ConfirmResetPassword.tsx
+++ b/packages/react-native/src/Authenticator/Defaults/ConfirmResetPassword/ConfirmResetPassword.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Text } from 'react-native';
 
 import { ConfirmResetPasswordComponent } from './types';
+import { DefaultFooter } from '../../common/DefaultFooter';
 import { DefaultHeader } from '../../common/DefaultHeader';
 
 const ConfirmResetPassword: ConfirmResetPasswordComponent = () => {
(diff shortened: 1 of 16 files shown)


packages/react-native/src/Authenticator/common/DefaultFooter/DefaultFooter.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { View } from 'react-native';
+
+import { DefaultFooterProps } from './types';
+
+export default function DefaultFooter({
+  children,
```
calebpollman, line 7: Can we add `rest` props here?

joebuono (approved):

packages/react-native/src/Authenticator/Defaults/ConfirmResetPassword/ConfirmResetPassword.tsx
```diff
@@ -2,17 +2,15 @@ import React from 'react';
 import { Text } from 'react-native';
 
 import { ConfirmResetPasswordComponent } from './types';
+import { DefaultFooter } from '../../common/DefaultFooter';
```
joebuono, line 5: nit: Can we just do:

```jsx
import { DefaultFooter, DefaultHeader } from '../../common';
```

ioanabrooks: I prefer to allow each export to be imported from its own module versus only through the main entry point (index file) as this reduces path resolution time.

packages/react-native/src/Authenticator/common/DefaultFooter/DefaultFooter.tsx
ioanabrooks: done!

calebpollman (approved):
🚢

PR #2795, opened 2022-10-24
feat(hooks): update `useAuth` to handle more auth hub events

wlee221:
#### Description of changes

Current `useAuth` hook only handles two of possible auth hub events. Updating this hook to handle other auth hub events that would change the `user` object.

Because this hook is being used internally, I'm keeping this change in backwards-compatible manner. But I did deprecate `fetch` from the return value, because there is no reason developer would need to call this manually.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/react/src/hooks/useAuth.ts
```diff
@@ -27,32 +23,64 @@ export const useAuth = (): UseAuthResult => {
...
-        break;
+  /**
+   * Hub events like `tokenRefresh` will not give back the user object.
+   * This util will be used to get current user after those events.
+   */
+  const fetchCurrentUser = React.useCallback(async () => {
+    setResult({ user: undefined, isLoading: true, error: undefined });
+
+    try {
+      const user = (await Auth.currentAuthenticatedUser()) as AmplifyUser;
```
wlee221, line 34: `Auth.currentAuthenticatedUser()` returns `any`, so I need to cast it.

packages/react/src/hooks/__tests__/useAuth.test.ts
```diff
@@ -21,23 +39,17 @@ describe('useAuth', () => {
   });
 
   it('should invoke Auth.currentAuthenticatedUser function', async () => {
-    const mockCurrentAuthenticatedUser = jest.fn(() => Promise.resolve());
-
-    (Auth.currentAuthenticatedUser as jest.Mock).mockImplementation(
-      mockCurrentAuthenticatedUser
-    );
+    currentAuthenticatedUser.mockResolvedValue({});
```
wlee221, line 42:
```suggestion
    currentAuthenticatedUser.mockResolvedValue(mockCognitoUser);
```

packages/react/src/hooks/useAuth.ts
calebpollman: Maybe we can put the above comment in the code 😆

packages/react/src/hooks/useAuth.ts
```diff
@@ -27,32 +23,64 @@ export const useAuth = (): UseAuthResult => {
...
+      setResult({ user, isLoading: false });
+    } catch (e) {
+      const error = e as Error;
+      setResult({ error, isLoading: false });
     }
-  };
+  }, [setResult]);
```
calebpollman, line 40: Since `setResult` is memoized, why do we need it here?

packages/react/src/hooks/useAuth.ts
```diff
@@ -27,32 +23,64 @@ export const useAuth = (): UseAuthResult => {
...
 
-  const fetch = () => {
-    setResult({ isLoading: true });
+  const handleAuth: HubCallback = React.useCallback(
+    ({ payload }) => {
+      switch (payload.event) {
```
calebpollman, line 44: Can we move the body of `handleAuth` to its own utility?

packages/react/src/hooks/useAuth.ts
wlee221: 😈 
```suggestion
     // casting the result because `Auth.currentAuthenticateduser` returns `any`
      const user = (await Auth.currentAuthenticatedUser()) as AmplifyUser;
```

packages/react/src/hooks/useAuth.ts
wlee221: Very fair, we don't. 8082476d26e4279990b99665cb10b4414fb065a7

packages/react/src/hooks/__tests__/useAuth.test.ts
```diff
@@ -67,65 +69,93 @@ describe('useAuth', () => {
...
 
-  it('should receive a Cognito user on Auth.signIn Hub event', async () => {
-    const mockCognitoUser = {
-      username: 'johndoe',
-      attributes: {
-        phone_number: '+1-234-567-890',
-        email: 'john@doe.com',
-      },
-    };
+  it.each(SUCCESS_EVENTS_WITH_USER)(
```
0618, line 72: TIL `it.each`

packages/react/src/hooks/useAuth.ts
```diff
@@ -27,32 +23,74 @@ export const useAuth = (): UseAuthResult => {
...
+   */
+  const fetchCurrentUser = React.useCallback(async () => {
+    setResult({ user: undefined, isLoading: true, error: undefined });
+
+    try {
+      // casting the result because `Auth.currentAuthenticateduser` returns `any`
+      const user = (await Auth.currentAuthenticatedUser()) as AmplifyUser;
+      setResult({ user, isLoading: false });
+    } catch (e) {
+      const error = e as Error;
```
0618, line 38: what's this line for?

wlee221: Ah, so I needed to cast `e` to an `error`, because `e` was `any`.

wlee221: I was hoping we could do something like
```ts
} catch (error: Error) {
  setResult({ error, isLoading: false })
}
```

but TS does [not support](https://github.com/microsoft/TypeScript/issues/20024) strongly typing `catch` statement, so left it like that.

0618: I see. Thanks!

0618 (approved):

packages/react/src/hooks/__tests__/useAuth.test.ts
```diff
@@ -3,13 +3,31 @@ import { Hub } from '@aws-amplify/core';
 import { act, renderHook } from '@testing-library/react-hooks';
 import { useAuth } from '../useAuth';
 
-jest.mock('@aws-amplify/auth');
+const currentAuthenticatedUser = jest.fn();
+
+jest
+  .spyOn(Auth, 'currentAuthenticatedUser')
+  .mockImplementation(currentAuthenticatedUser);
```
calebpollman, line 10: Nit: Since we are mocking the resolved value in each test case _think_ this can just be:
```
const currentAuthenticatedUserSpy = jest.spyOn(Auth, 'currentAuthenticatedUser');
```

calebpollman (approved):
LGTM, just left nit-ish suggestion

packages/react/src/hooks/__tests__/useAuth.test.ts
wlee221: Cleaned up in [9fb0d3f](https://github.com/aws-amplify/amplify-ui/pull/2795/commits/9fb0d3f48700cf6de047b851473069957d343fcf) 🙏

PR #2817, opened 2022-10-25
feat(angular): expose `subscribe` to subscribe for auth changes

### LINKED ISSUE #2324: Angular App Login functionality to work with guards, require a deprecated subscribe () method to be used in ui-angular

Angular

Authenticator

As per Issue [https://github.com/aws-amplify/amplify-ui/issues/2314#issuecomment-1191924995](https://github.com/aws-amplify/amplify-ui/issues/2314#issuecomment-1191924995) a workaround was created to use the deprecated subscribe method on AuthenticatorService. This issue is created to track the fix of this issue and prevent the deprecated to be removed until a solution is developed and documented.

Provide a non-deprecated way of making ui-angular work with an application which uses guards and to allow the user to navigate to  a guarded route after a successful login. 

- [X] ⚠️ This feature might incur a breaking change.

wlee221:
#### Description of changes

This PR exposes `subscribe` from `AuthenticatorService`. Previously, this was internal and returned just the internal xstate values. This PR changes it so that it exposes facade values instead, which they can subscribe to implement route guards. See changeset for usage.

Note that `subscribe` will return the same values as `useAuthenticator` from react + vue, so this is more of a parity improvement.

Also makes minor optimization while I was exposing facade values.

#### Issue #, if available

Fixes #2324.

WIP. Opened PR to demonstrate base code change needed.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/angular/projects/ui-angular/src/lib/common/types/auth-types.ts
```diff
@@ -1,6 +1,11 @@
...
+  DefaultFormFieldOptions,
+  AuthenticatorServiceFacade,
+} from '@aws-amplify/ui';
 
 // maps auth attribute to its repsective labels and placeholder
 export type AttributeInfoProvider = () => DefaultFormFieldOptions;
 
-export type AuthSubscriptionCallback = (state: AuthMachineState) => void;
+export type AuthSubscriptionCallback = (
+  state: AuthenticatorServiceFacade
```
wlee221, line 10: Now subscription returns facade data!

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
```diff
@@ -126,7 +126,7 @@ export class AuthenticatorComponent
 
         this.hasInitialized = true;
       }
-    }).unsubscribe;
```
wlee221, line 129: `AuthenticatorService.subscribe` now always returns `unsubscribe` directly, so `Authenticator` no longer need to access `.unsubscribe` to get the `unsubscribe` function.

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -166,10 +166,12 @@ export class AuthenticatorService implements OnDestroy {
...
 
-  /** @deprecated For internal use only */
-  public subscribe(callback: AuthSubscriptionCallback) {
+  public subscribe(callback: AuthSubscriptionCallback): () => void {
     if (this._authService) {
-      return this._authService.subscribe(callback);
+      const subscription = this._authService.subscribe(() => {
+        callback({ ...this._facade, ...this._sendEventAliases });
+      });
+      return subscription.unsubscribe;
```
wlee221, line 174: return `unsubscribe` function directly, so customers can just do 

```ts
const unsubscribe = this.authenitcator.subscribe(...);
```

like we do above.

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -166,10 +166,12 @@ export class AuthenticatorService implements OnDestroy {
     };
   }
 
-  /** @deprecated For internal use only */
-  public subscribe(callback: AuthSubscriptionCallback) {
+  public subscribe(callback: AuthSubscriptionCallback): () => void {
     if (this._authService) {
-      return this._authService.subscribe(callback);
```
wlee221, line 172: For context, this internal function was just directly using `xstate` subscription instead of facades.

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -24,21 +23,23 @@ const logger = new Logger('state-machine');
 export class AuthenticatorService implements OnDestroy {
   private _authState: AuthMachineState;
   private _authService: AuthInterpreter;
-  private _sendEventAliases: ReturnType<typeof getSendEventAliases>;
   private _machineSubscription: Subscription;
-  private _facade: ReturnType<typeof getServiceContextFacade>;
+  private _facade: ReturnType<typeof getServiceFacade>;
```
wlee221, line 27: While I was here, I made a small refactor to store facade values in one variable `_facade` instead of two variables.

For context, `AuthenticatorService` maintained two variables:
- `_facade`, which contains context facade values (e.g. `route`)
- `_sendEventAliases`, which contains transition methods like `toSignIn()`

Combining these two allows for performance boosts, because we can now do:

```diff
- return { ..._facade, ..._sendEventAliases };
+ return _facade;
```
to avoid unnecessary destructures.

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -155,26 +156,25 @@ export class AuthenticatorService implements OnDestroy {
...
+  public subscribe(callback: AuthSubscriptionCallback): () => void {
+    if (!this._authService) {
       logger.error(
         'Subscription attempted before machine was created. This is likely a bug on the library, please consider filing a bug.'
       );
     }
+
+    const subscription = this._authService.subscribe(() => {
+      callback(this._facade);
+    });
```
wlee221, line 176: Here's the actual feature; we return `_facade` instead of internal state machine state value.

packages/angular/projects/ui-angular/src/lib/services/__tests__/mockAuthService.ts
```diff
@@ -0,0 +1,22 @@
+// mock state machine service
```
wlee221, line 1: Unlike vue/angular, Angular was using `xstate` directly without any util library (`@xstate/react`), so had to do mocks at xstate level.

packages/angular/projects/ui-angular/src/lib/services/__tests__/authenticator.service.test.ts
```diff
@@ -0,0 +1,60 @@
+import { AuthenticatorService } from '../authenticator.service';
+
+import * as UIModule from '@aws-amplify/ui';
+import * as XState from 'xstate';
+
+// mock state machine service
+// based on https://github.com/statelyai/xstate/blob/main/packages/core/src/interpreter.ts
+export class MockAuthService {
```
wlee221, line 8: Unlike vue/angular, Angular uses `xstate` directly without any util library (`@xstate/react`), so had to do mocks at xstate level.

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -155,26 +156,23 @@ export class AuthenticatorService implements OnDestroy {
...
+    return subscription.unsubscribe;
```
wlee221, line 175: Here's the actual feature; we return `_facade` instead of internal state machine state value.

Note that we return `unsubscribe` directly so customers can just do `const unsubscribe = this.authenticator.subscribe()`.

packages/angular/projects/ui-angular/src/lib/services/__tests__/authenticator.service.test.ts
ErikCH: 👍

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
ErikCH: 👍

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
wlee221: Good point, and I see that this is much more [standard](https://github.com/aws-amplify/amplify-ui/pull/2935/files#diff-12783128521e452af0cfac94b99b8d250413c516ec71fe6d97dbea666ff7ba27R5-R7) in Angular ecosystem. I'll update this back to `unsubscribe()`!

.changeset/smooth-moons-occur.md
```diff
@@ -0,0 +1,25 @@
...
+class MyComponent implements OnInit, OnDelete {
+  private unsubscribe: () => void = null;
+  constructor(private authenticator: Authenticator, private route: Router) {}
+
+  ngOnInit() {
+    this.unsubscribe = authenticator.subscribe(({ authStatus }) => {
+      if (authStatus === 'authenticated') {
+        this.router.navigate(['/admin']);
+      }
+    });
```
wlee221, line 18:
```suggestion
    }).unsubscribe;
```

.changeset/smooth-moons-occur.md
```diff
@@ -0,0 +1,25 @@
+---
+'@aws-amplify/ui-angular': minor
+---
+
+Expose `subscribe` from `AuthenticatorService`.
+
+```ts
+@Component()
+class MyComponent implements OnInit, OnDelete {
+  private unsubscribe: () => void = null;
```
wlee221, line 10:
```suggestion
  private unsubscribe: () => void;
```

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
wlee221: Done in [e7a32a1](https://github.com/aws-amplify/amplify-ui/pull/2817/commits/e7a32a177a34ef1d3ccb1be9145dbc8a314797dc)

calebpollman (approved):
🚢

PR #2830, opened 2022-10-26
chore(authenticator): adding version to cognito user agent

dbanksdesign:
#### Description of changes

Adding the package name, version number, and component name to the Cognito user agent string. Right now I only did React because I don't know how to do this for Angular and Vue. I did also add a top-level script that creates a `version.ts` file in each top-level package with the current version number of the package. It runs as a prebuild step. 

Ran the next-example with our backends and verified the UA string:

CleanShot 2022-10-26 at 10 18 27@2x

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -109,6 +116,7 @@ export function AuthenticatorInternal({
  * [📖 Docs](https://ui.docs.amplify.aws/react/connected-components/authenticator)
  */
 export function Authenticator(props: AuthenticatorProps): JSX.Element {
+  appendToCognitoUserAgent(`@aws-amplify/ui-react/${VERSION}`);
```
calebpollman, line 119: Thinking we will want to wrap this in a `useEffect` with an empty dependency array to ensure it doesn't run on each render

packages/ui/src/helpers/authenticator/utils.ts
```diff
@@ -4,9 +4,14 @@
  */
 
 import { Hub } from 'aws-amplify';
+import { appendToCognitoUserAgent } from '@aws-amplify/auth';
 import { AuthInterpreter, HubHandler } from '../../types';
 import { ALLOWED_SPECIAL_CHARACTERS } from './constants';
 
+// Re-exporting this helper method so the public packages
+// don't need to import anything from '@aws-amplify/auth'
+export { appendToCognitoUserAgent };
```
calebpollman, line 13: Nitty, but maybe we can wrap this in a "configure" util in case we want to add other functionality in the feature outside appending to the user agent. Example:
```
type ConfigureOptions = { packageName: string; version: string }
const configureComponent = ({ packageName, version }: ConfigureOptions) => {
  // "@aws-amplify/ui-react" + "/" + "3.5.10"
  appendToCognitoUserAgent(`${packageName}/${version}`);
)

Comment (dbanksdesign) [reply packages/ui/src/helpers/authenticator/utils.ts:13]: oh yea thats a good idea!

**Comment by wlee221** [comment]
Wondering what triggered increase in React bundle size 🤔

**Comment by dbanksdesign** [comment]
Hmm maybe `@aws-amplify/auth` has some side effects?

File: packages/react/src/components/Authenticator/Authenticator.tsx [hunk 3]
```diff
@@ -18,6 +22,7 @@ import { SignUp } from './SignUp';
 import { ForceNewPassword } from './ForceNewPassword';
 import { ResetPassword } from './ResetPassword';
 import { defaultComponents } from './hooks/useCustomComponents/defaultComponents';
+import { VERSION } from 'src/version';
```
reesscot, line 25: I think this import is going to fail

dbanksdesign: You are right, fixed!

package.json
```diff
@@ -22,12 +22,13 @@
     "build:icons": "yarn react build:icons",
     "test": "yarn ui test && yarn react-core test && yarn react test && yarn react-native test && yarn e2e test",
     "prepare": "husky install",
-    "version:next": "yarn changeset version --snapshot next-$(git rev-parse --short=7 HEAD) && yarn angular build",
+    "version:next": "yarn changeset version --snapshot next-$(git rev-parse --short=7 HEAD) && yarn generateVersions && yarn angular build",
```
wlee221, line 25:
```suggestion
    "version:next": "yarn changeset version --snapshot next-$(git rev-parse --short=7 HEAD) && yarn generateVersions && yarn build",
```

as we discussed, same with other version scripts

dbanksdesign: done!

packages/angular/projects/ui-angular/src/lib/components/authenticator/components/authenticator/authenticator.component.ts
```diff
@@ -69,6 +71,11 @@ export class AuthenticatorComponent
       formFields,
     } = this;
 
+    configureComponent({
+      packageName: '@aws-amplify/ui-angular',
+      version: VERSION,
+    });
+
```
wlee221, line 78: Verified the change locally: 
image

.github/workflows/version-packages.yml
```diff
@@ -38,5 +38,7 @@ jobs:
       - name: Create or update Version Packages PR
         if: ${{ steps.has-changesets.outputs.has-changesets == 'true' }}
         uses: changesets/action@b98cec97583b917ff1dc6179dd4d230d3e439894
+        with:
+          version: yarn version:latest
```
wlee221, line 42: This will run our custom versioning command with `yarn generateVersions`. This will ensure "Version Packages" PR will update `version.ts` when it bumps up the version.

packages/ui/src/helpers/authenticator/utils.ts
```diff
@@ -4,12 +4,22 @@
...
 import { AuthInterpreter, AuthMachineHubHandler } from '../../types';
 import { ALLOWED_SPECIAL_CHARACTERS } from './constants';
 import { getActorState } from './actor';
 
+type ConfigureOptions = { packageName: string; version: string };
+export const configureComponent = ({
+  packageName,
+  version,
+}: ConfigureOptions) => {
+  // "@aws-amplify/ui-react" + "/" + "3.5.10"
```
wlee221, line 19: Is this comment necessary?

packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -96,6 +101,13 @@ export function AuthenticatorInternal({
  * [📖 Docs](https://ui.docs.amplify.aws/react/connected-components/authenticator)
  */
 export function Authenticator(props: AuthenticatorProps): JSX.Element {
+  React.useEffect(() => {
```
wlee221, line 104: Technically, this will run multiple times if this component remounts. Could that possibly mess up the data?

wlee221 (dismissed):
LGTM, just two non-blocking questions

packages/ui/src/helpers/authenticator/utils.ts
dbanksdesign: Not totally, but it is intentional to let future versions of ourselves remember what this is for

packages/react/src/components/Authenticator/Authenticator.tsx
reesscot: My understanding is this just sets up the correct header to be included with Cognito API calls, so it shouldn't have an negative impact.

wlee221: Oh yeah, I was worried this would add duplicate headers. But looks like Amplify JS handles duplicate for us! ([src](https://github.com/aws-amplify/amplify-js/blob/190bd501236a59b16318802b51a0494f7374d299/packages/amazon-cognito-identity-js/src/UserAgent.js#L13))

dbanksdesign: Just tested this by running our Next example app and adding a button to conditionally render the Authenticator, toggled it a few times, and tried to log in. The UA string seemed fine.

dbanksdesign:
RN is working!

CleanShot 2022-12-14 at 16 06 35@2x

calebpollman (approved):
LGTM 🚢

wlee221 (approved):

ErikCH (approved):

PR #2837, opened 2022-10-27
feat(ui-react-native): Add ThemeProvider and useTheme

ioanabrooks:
#### Description of changes

* Add ThemeProvider
* Add useTheme hook

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react-native/src/theme/createReactNativeTheme.ts
```diff
@@ -0,0 +1,55 @@
...
+    : (callback(v) as object);
+
+/**
+ * This will be used like `const myTheme = createReactNativeTheme({})`
+ * `myTheme` can then be passed to a Provider
+ * const myTheme = createReactNativeTheme({})
+ * const myOtherTheme = createReactNativeTheme({}, myTheme);
+ */
+export const createReactNativeTheme = (
+  theme?: Theme,
```
calebpollman, line 22: Since this is the param for a customer to pass a custom theme, it should be `ReactNativeTheme` right?

packages/react-native/src/theme/createReactNativeTheme.ts
```diff
@@ -0,0 +1,55 @@
...
+import mapValues from 'lodash/mapValues';
+
+import { defaultTheme } from './defaultTheme';
+import { ColorMode, Theme, ReactNativeTheme } from './types';
+import { Tokens } from './tokens';
+
+const mapValuesDeep = (v: object | string, callback: Function): object =>
+  typeof v == 'object'
+    ? mapValues(v, (v) => mapValuesDeep(v, callback))
+    : (callback(v) as object);
```
calebpollman, line 13: Since we are not going to reuse the util outside this file, let's write it in a way where it handles typing the output for us so we don't need to cast on ln 45

packages/react-native/src/theme/createReactNativeTheme.ts
```diff
@@ -0,0 +1,55 @@
...
+  // Setting up the tokens.
+  // At the end of this, each token should have a raw value
+  // All references to tokens will be replaced
+  const tokens = mapValuesDeep(mergedTheme.tokens, (value: string) => {
+    return typeof value == 'string' && value.startsWith('{')
+      ? (get(
+          mergedTheme.tokens,
+          value.substring(1, value.length - 2) // remove { }
+        ) as string)
+      : value;
```
calebpollman, line 44: A few things here:
- Can we move this callback in to a strongly typed util? 
- Think we can improve the  comment regarding removing `{` and `}` from strings. Future maintainers may not understand that in some scenarios a string is being passed and parsed to find the value.
- What happens when `value.length - 2` does not evaluate to `}`? In general, how are we handling strings that are malformed in context to what the lookup syntax expects?

packages/react-native/src/theme/tokens/index.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+type Colors = {
+  brand: {
+    primary: ColorScale;
+    secondary: ColorScale;
+  };
+};
+
+export const colors: Partial<Colors> = {
```
calebpollman, line 27: Why is this type with a `Partial`? Don't we expect these values to be fully provided?

packages/react-native/src/theme/tokens/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  brand: {
+    primary: {
+      10: 'hsl(190, 75%, 95%)',
+      20: 'hsl(190, 75%, 85%)',
+      40: 'hsl(190, 70%, 70%)',
+      60: 'hsl(190, 50%, 50%)',
+      80: 'hsl(190, 95%, 30%)',
+      90: 'hsl(190, 100%, 20%)',
+      100: 'hsl(190, 100%, 15%)',
+    },
```
calebpollman, line 37: Can we just import these directly from the `@aws-amplify/ui`?

packages/react-native/src/theme/tokens/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  colors: Partial<Colors>;
+}
+
+export interface Tokens extends BaseTokens {
+  components: ComponentStyles;
+}
+
+// TODO: remove color types and values and use the output generated by style-dictionary with correct types
+
+declare type ScaleKeys = 10 | 20 | 40 | 60 | 80 | 90 | 100;
```
calebpollman, line 15: Why do we need the `declare` keyword here?

packages/react-native/src/theme/createReactNativeTheme.ts
```diff
@@ -0,0 +1,55 @@
...
+    ? mapValues(v, (v) => mapValuesDeep(v, callback))
+    : (callback(v) as object);
+
+/**
+ * This will be used like `const myTheme = createReactNativeTheme({})`
+ * `myTheme` can then be passed to a Provider
+ * const myTheme = createReactNativeTheme({})
+ * const myOtherTheme = createReactNativeTheme({}, myTheme);
+ */
+export const createReactNativeTheme = (
```
calebpollman, line 21: Don't think we need to use `ReactNative` in this function name

packages/react-native/src/theme/tokens/index.ts
```diff
@@ -0,0 +1,53 @@
+import { components } from './components';
+import { ComponentStyles } from '../types';
+import { ColorValue } from '@aws-amplify/ui';
+
+interface BaseTokens {
```
calebpollman, line 5: Can we put `BaseTokens` and below types in to a _types.ts_?

packages/react-native/src/theme/defaultTheme.ts
```diff
@@ -0,0 +1,7 @@
+import { tokens } from './tokens';
+import { Theme } from './types';
+
+export const defaultTheme: Theme = {
```
calebpollman, line 4: Is this actually the default `Theme`? It doesn't contain the component default theme right?

packages/react-native/src/theme/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * The name of the theme.
+   */
+  name: string;
+  tokens?: PartialDeep<Tokens>;
+  /**
+   * Overrides allow you to change design tokens in different contexts, like
+   * light and dark mode.
+   */
+  overrides?: Array<Override>;
```
calebpollman, line 24: Prefer TS shorthand `Array` declaration, example `Override[]`

packages/react-native/src/theme/types.ts
```diff
@@ -0,0 +1,38 @@
...
+export interface ComponentStyles {
+  // TODO: add components
+}
+
+/**
+ * ReactNativeTheme is a fully built theme that has styling based
+ * on the design tokens and all design tokens have added fields
+ * to be used in Javascript/Typescript.
+ */
+export interface ReactNativeTheme extends Theme {
```
calebpollman, line 36: Seems like this should be our actual `Theme`, and that we should be extending from a `BaseTheme` (what is currently declared as `Theme`)

packages/react-native/src/theme/tokens/index.ts
```diff
@@ -0,0 +1,53 @@
+import { components } from './components';
```
calebpollman, line 1: This file can just be located in _theme/tokens.ts_. We can remove the _tokens_ directory entirely

packages/react-native/src/theme/types.ts
```diff
@@ -0,0 +1,38 @@
...
+ * A Theme just needs a name.
+ * Users can define any tokens they need, but they don't need a complete theme with all tokens.
+ */
+export interface Theme {
+  colorMode?: ColorMode;
+  /**
+   * The name of the theme.
+   */
+  name: string;
+  tokens?: PartialDeep<Tokens>;
```
calebpollman, line 19: Rather than using `PartialDeep`, let's build the `Theme` that we export publicly with all keys as optional by default, and then use that interface to create a strict/required theme.

packages/react-native/src/theme/tokens/components/index.ts
```diff
@@ -0,0 +1,5 @@
+import { ComponentStyles } from '../../../theme';
+
+export const components: ComponentStyles = {
```
calebpollman, line 3: This can just live in _theme/tokens.ts_

packages/react-native/src/hooks/__tests__/useTheme.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('should return a default theme if not provided through context', () => {
+    const { result } = renderHook(() => useTheme(), {
+      wrapper: ({ children }) => <ThemeProvider>{children}</ThemeProvider>,
+    });
+
+    expect(serializeTheme(result.current)).toBe(
+      serializeTheme(createReactNativeTheme())
+    );
```
calebpollman, line 44: Rather than serializing the output of `createReactNativeTheme`, can we just use `toStrictEqual` in place of `toBe` here and on line 50?

packages/react-native/src/theme/createReactNativeTheme.ts
```diff
@@ -0,0 +1,55 @@
+// Internal Style Dictionary methods
+import deepExtend from 'style-dictionary/lib/utils/deepExtend';
+import get from 'lodash/get';
+import mapValues from 'lodash/mapValues';
+
+import { defaultTheme } from './defaultTheme';
+import { ColorMode, Theme, ReactNativeTheme } from './types';
+import { Tokens } from './tokens';
+
+const mapValuesDeep = (v: object | string, callback: Function): object =>
```
calebpollman, line 10: Please use meaningfully named params for easier reading and maintenance in the future

packages/react-native/src/theme/createReactNativeTheme.ts
```diff
@@ -0,0 +1,55 @@
...
+  typeof v == 'object'
```
calebpollman, line 11: Any reason not use a strict comparison here?

packages/react-native/src/theme/createReactNativeTheme.ts
```diff
@@ -0,0 +1,55 @@
+// Internal Style Dictionary methods
```
calebpollman, line 1: Think we can remove this comment

packages/react-native/src/theme/tokens/index.ts
ioanabrooks: No, colors are not exported. This code will be removed once we generate the non-conponent tokens. This is just an example. I can remove entirely if too confusing.

packages/react-native/src/theme/tokens/components/index.ts
ioanabrooks: sure

packages/react-native/src/hooks/__tests__/useTheme.spec.tsx
ioanabrooks: Oh nice, looks like that works. Was following the pattern from ui pkg.

packages/react-native/src/theme/tokens/index.ts
ioanabrooks: Just wanted to bring in a couple colors for example purposes. This will be changed once we generate the comp. agnostic tokens.

packages/react-native/src/theme/createReactNativeTheme.ts
ioanabrooks: sure

packages/react-native/src/theme/defaultTheme.ts
ioanabrooks: yes this is the default Theme

calebpollman: Maybe I am misreading but the value of `tokens` here does not include the component tokens correct? So wouldn't that make this just the `baseTheme`?

packages/react-native/src/theme/createReactNativeTheme.ts
calebpollman: Unresolving this conversation, still not unclear on:
> What happens when value.length - 2 does not evaluate to }? In general, how are we handling strings that are malformed in context to what the lookup syntax expects?

packages/react-native/src/theme/tokens.ts
```diff
@@ -0,0 +1,55 @@
+import { ComponentStyles } from './types';
+import { ColorValue } from '@aws-amplify/ui';
+
+interface BaseTokens {
+  colors: Partial<Colors>;
+}
+
+export interface Tokens extends BaseTokens {
+  components: ComponentStyles;
+}
```
calebpollman, line 10: Why are these separate from the other types? Seems like they could just live in _types.ts_?

packages/ui/src/theme/types.ts
```diff
@@ -24,7 +23,7 @@ export type Override =
   | ColorModeOverride;
 
 interface BaseOverride {
-  tokens?: PartialDeep<Tokens>;
+  tokens?: Tokens;
```
calebpollman, line 26: Does this impact the existing `theme` usage?

packages/react-native/src/theme/createTheme.ts
```diff
@@ -0,0 +1,57 @@
...
+    : (callback(obj) as T);
+
+/**
+ * This will be used like `const myTheme = createTheme({})`
+ * `myTheme` can then be passed to a Provider
+ * const myTheme = createTheme({})
+ * const myOtherTheme = createTheme({}, myTheme);
+ */
+export const createTheme = (
+  theme?: BaseTheme,
```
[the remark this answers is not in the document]
ioanabrooks: Theme is strongly typed though so customer would need to provide complete theme, right?

packages/react-native/src/theme/createReactNativeTheme.ts
ioanabrooks: and if the string does not start/end with curlies then it gets returned as is

packages/ui/src/theme/types.ts
ioanabrooks: Good catch!

packages/react-native/src/theme/tokens.ts
ioanabrooks: Moved

packages/react-native/src/theme/createTheme.ts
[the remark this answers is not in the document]
ioanabrooks: Synced offline about Theme nomenclature and usage, renamed accordingly.

packages/react-native/src/theme/types.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface ComponentStyles {
+  // TODO: add components
+}
+
+/**
+ * Fully built theme that has styling based
+ * on the design tokens and all design tokens have added fields
+ * to be used in Javascript/Typescript.
+ */
+export interface StrictTheme extends Theme {
```
calebpollman, line 46: Should `overrides` and `colorMode` be optional still? Just verifying

packages/react-native/src/theme/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Tokens } from './types';
+
+/**
+ * Util similar to lodash mapValues but recurses into nested objects.
+ * @param obj
+ * @param callback
+ */
+const mapValuesDeep = <T>(obj: T, callback: Function): T =>
+  typeof obj === 'object'
+    ? (mapValues(obj, (obj) => mapValuesDeep(obj, callback)) as T) // cast back to T to appease TS
```
calebpollman, line 12: I see `obj` in three places in this line, is that expected?

packages/react-native/src/theme/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+    : (callback(obj) as T);
+
+/**
+ * Function that will walk down the token object
+ * and perform the mapValuesDeep function on each token.
+ */
+export const setupTokens = (tokens: Tokens): Tokens =>
```
calebpollman, line 19: Think maybe we have a disconnect on abstracting this callback in to it's own function. What I have been trying to express is that the below should be its own util:
```
const callback = (value: string) => {
    return typeof value === 'string' &&
      value.startsWith('{') &&
      value.endsWith('}')
      ? (get(tokens, value.substring(1, value.length - 2)) as string) // cast back to string to appease TS, as lodash get returns any
      : value;
  }
```

And still think we need to add explicit notes around the string parsing behavior, in particular due to how fragile it is

packages/react-native/src/theme/types.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface Theme {
+  colorMode?: ColorMode;
+  /**
+   * The name of the theme.
+   */
+  name: string;
+  /**
+   * Component and component agnostic tokens.
+   */
+  tokens?: PartialDeep<Tokens>;
```
calebpollman, line 29: Can we just declare `Tokens` in way that is loose and then wrap it with `Required` for the strict theme?

ioanabrooks: Required is not deep

packages/react-native/src/theme/utils.ts
ioanabrooks: Are you suggesting we should pass in `tokens` to the callback fn and also in `mapValuesDeep`?

packages/react-native/src/theme/utils.ts
ioanabrooks: I could rename the second and third, any suggestions?

packages/react-native/src/theme/types.ts
ioanabrooks: overrides for sure should be optional, unless we want to treat dark mode as an override [which is what react web does]. colormode no.

packages/react-native/src/theme/defaultTheme.ts
```diff
@@ -0,0 +1,8 @@
+import { tokens } from './tokens';
+import { StrictTheme } from './types';
+
+export const defaultTheme: StrictTheme = {
+  tokens,
+  name: 'default-theme',
+  colorMode: 'system',
```
calebpollman, line 7: Nice!

calebpollman (approved):
@ioanabrooks Thanks for all the hard work here! 🚢

packages/react-native/src/theme/utils.ts
dbanksdesign: Why are rewriting the string parsing behavior when we already have it in the UI package (which it gets from style dictionary)? That parsing/resolving behavior is thoroughly tested and used by lots of packages. Not a blocker to get started, but I would feel nervous about having different functions for React Native and React Web.

packages/react-native/src/theme/types.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Fully built theme that has styling based
+ * on the design tokens and all design tokens have added fields
+ * to be used in Javascript/Typescript.
+ */
+export interface StrictTheme extends Theme {
+  colorMode: ColorMode;
+  tokens: Tokens;
+}
+
+// TODO: refactor everything below [color types and values] and use the output generated by style-dictionary with correct types
```
dbanksdesign, line 51: We should just use the types we have defined in the UI package right? Right now a RN and RW theme are incompatible for the base tokens, and we are re-creating types and default values that already exist in a shared place.

packages/react-native/src/theme/createTheme.ts
```diff
@@ -0,0 +1,37 @@
...
+    {},
+    defaultTheme,
+    theme,
+  ]) as StrictTheme;
+  const { colorMode, name, overrides } = mergedTheme;
+
+  // Setting up the tokens.
+  // At the end of this, each token should have a raw value
+  // All references to tokens will be replaced; references to tokens are marked by curly braces
+  const tokens = setupTokens(mergedTheme.tokens);
```
dbanksdesign, line 27: Todo: in order to support how overrides work, we will need to conditionally merge override tokens in (if the colorMode is dark and the developer passes in a dark mode override). We don't need to do that for web because of CSS variables.

packages/react-native/src/theme/types.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface ComponentStyles {
+  // TODO: add components
+}
+
+/**
+ * Fully built theme that has styling based
+ * on the design tokens and all design tokens have added fields
+ * to be used in Javascript/Typescript.
+ */
+export interface StrictTheme extends Theme {
```
dbanksdesign, line 46: I like the name `StrictTheme`!

dbanksdesign (approved):
Nothing blocking for now. Just a few call-outs on where we should be leveraging what we have in the UI package for consistency.

PR #2844, opened 2022-10-27
test(docs): add E2E test to check the broken links at build time

0618:
#### Issue #, if available

**Issue**: We currently have a broken link check on CloudWatch Synthetic Canaries, but it has some problems:
1) it runs all the time which is expensive and unnecessary
2) it's heavy and takes a lot of lambda and S3 resources
3) it's only check the links on the homepage and component page

**fix**: In this PR, we move the broken link tests to E2E test at build time, so it'll run in Amplify Hosting when new commit appear in `main` or `docs-release`. It has the following benefits:
1) Cheap. Only run when there's a new commits.
2) Complete. Check every `href` on each page included on Sitemap.
3) Dynamic. When new links added or exist link changed, we don't have to update the tests.

**Code Changes**:
- add E2E test
- fix broken links

**NOTE**:
- `cy.request` is very memory-comsuming, so this PR tried different ways to 1) clean up/increase memory 2) make less `GET` requests 

**TODO**:
[ ] Need to add test for `#hash` links. Currently, we only check the page status.

image

- Tested in my personal account

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/cypress.config.ts
```diff
@@ -0,0 +1,38 @@
...
+          }
+        },
+        log: (message) => {
+          console.log(message);
+          return null;
+        },
+      });
+    },
+    baseUrl: BASE_URL,
+    numTestsKeptInMemory: 0,
```
0618, line 36: default is `50`, reduce it to save some memory. [docs](https://docs.cypress.io/guides/references/configuration#Global)

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+      function hrefWorks(htmlTag: JQuery<HTMLElement>): void {
+        const tagHref: string = htmlTag.prop('href');
+        const tagHash: string = htmlTag.prop('hash');
+        const tagText: string = htmlTag.prop('text');
+        const tagName: string = htmlTag.prop('tagName');
+        let pureHref: string;
+        if (tagHref) {
+          pureHref = tagHref.replace(tagHash, ''); // TODO: add test to validate links with a hash tag.
+          logMessage('CHECKING');
+
```
0618, line 39: The following `if...else...` statement skips as making `cy.request` as many as possible:
1) skip links from UI docs site
2) avoid testing a link twice

docs/scripts/generate-sitemap-robotstxt.ts
```diff
@@ -35,7 +35,7 @@ async function generateSitemap() {
     <?xml version="1.0" encoding="UTF-8"?>
     <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:mobile="http://www.google.com/schemas/sitemap-mobile/1.0" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1" xmlns:video="http://www.google.com/schemas/sitemap-video/1.1">
           <url>
-            <loc>${process.env.SITE_URL}</loc>
+            <loc>${process.env.SITE_URL}/</loc>
```
0618, line 38: link without `/` should be redirected to the link with `/`, so add `/` to avoid the redirection

docs/cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
dbanksdesign, line 1: Should we keep this file? It's just a big comment?

docs/cypress/support/e2e.ts
```diff
@@ -0,0 +1,20 @@
+// ***********************************************************
```
dbanksdesign, line 1: Same comment here

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+});
+
+describe('Local Sitemap', () => {
+  it('should have 119 links', () => {
+    expect(allLinks.length).to.eq(numberOfLinks);
+  });
+});
+
+for (let i = 0; i < numberOfLinks; i++) {
+  describe(`check page ${i}`, () => {
```
dbanksdesign, line 22: Nit: should we move this describe outside of the `for` loop? Seems odd to have 119 describe blocks each with only 1 test, instead of 1 describe with 119 test.

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+    it(`all links on page ${i} should work`, () => {
+      const link = allLinks[i];
+      cy.task('log', `🧪[TESTING...] page ${BASE_URL}/${link}`);
+      cy.visit(link || '/');
+      cy.get('a').each(hrefWorks);
+      cy.get('button').each(hrefWorks);
```
dbanksdesign, line 28: Do we have `<button>` elements with an `href`? I feel like those should be errors.

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,137 @@
...
+          | 'REQUESTING'
+          | 'RETURNING'
+          | 'NO_HREF';
+
+        function logMessage(evtName: EvtName, status?: number) {
+          switch (evtName) {
+            case 'CHECKING':
+              return cy.task(
+                'log',
+                `🔍[CHECKING...] ${pureHref} from ${tagName} tag ${
```
dbanksdesign, line 91: I like the very detailed logging!

docs/cypress/data/validatedLinks.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * VALIDATED_LINKS is a manually maintained allowlist.
+ * Any links on this list will not be tested, so KEEP THIS LIST AS SHORT AS POSSIBLE.
+ * Links added to this list because of one of the following reasons
+ * 1) they return unexpected status code
```
dbanksdesign, line 5: Thank you for this comment

docs/cypress.config.ts
```diff
@@ -0,0 +1,38 @@
+import { defineConfig } from 'cypress';
+import sitemapUrls from 'sitemap-urls';
+import { BASE_URL } from './cypress/data/constants';
+
+const fs = require('fs');
```
dbanksdesign, line 5: can this be `import`ed?

dbanksdesign (approved):
Nothing major, just a few non-blocking things/questions

docs/cypress/support/commands.ts
0618: Yea, I thought we would use the auto generated support files in the future, but yea, it looks cleaner without them.

Removed!

docs/cypress/support/e2e.ts
0618: Removed the file.

docs/cypress/e2e/links.cy.ts
0618: Good point. It looks better now.


image

docs/cypress.config.ts
0618: Good catch. Updated.

docs/cypress/e2e/links.cy.ts
0618: Right, we don't have `href` on `<button>`. I somehow remember we had, but, anyway, I removed the assertion to save more memory.

dbanksdesign (approved):

docs/cypress.config.ts
ioanabrooks: `The number of tests for which snapshots and command data are kept in memory` - does this mean that if we set it to 0 we won't have snapshots of failed tests? If so can we reduce to 5 or 10 instead?

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,136 @@
...
+
+describe('Local Sitemap', () => {
+  it('should have 119 links', () => {
+    expect(allLinks.length).to.eq(numberOfLinks);
+  });
+});
+
+describe(`All pages on Sitemap`, () => {
+  for (let i = 0; i < numberOfLinks; i++) {
+    it(`all links on page ${i} should work`, () => {
```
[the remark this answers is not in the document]
0618: If it works, it would look cleaner, but I tested it and it doesn't work because `allLinks` is generated at `before`, when `it.each(allLinks)` , it'd be `undefined`.

Or we can use something like ` it.each(new Array(numberOfLinks).fill(1))`, but I'm not sure how we can get the index 🤔

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,136 @@
...
+      const link = allLinks[i];
+      cy.task('log', `🧪[TESTING...] page ${BASE_URL}/${link}`);
+      cy.visit(link || '/');
+      cy.get('a').each(hrefWorks);
+
+      function hrefWorks(htmlTag: JQuery<HTMLElement>): void {
```
calebpollman, line 29: `JQuery`? 🤔

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,136 @@
...
+          logMessage('NO_HREF');
+        }
+
+        type EvtName =
+          | 'CHECKING'
+          | 'SKIPPING_SITEMAP'
+          | 'SKIPPING_VALIDATED'
+          | 'REQUESTING'
+          | 'RETURNING'
+          | 'NO_HREF';
```
calebpollman, line 83: Why are we declaring types within the body of a function?

calebpollman (commented):
Can we rethink the approach to avoid declaring so much logic within the tests? Concerned that we are going to need tests for the the test utils 😅

docs/cypress/e2e/links.cy.ts
0618: hahaha, yea, that's what `cy.get()` returns [ref](https://docs.cypress.io/api/commands/get)

image

docs/cypress/e2e/links.cy.ts
0618: Because we don't reuse the type, defining it close to where it's used might be the cleanest way.

docs/cypress.config.ts
0618: Right. We don't have the snapshot, but we still have the video...but might be good to keep the snapshot there. Let me try if we can change it to `1`. The memory issue makes it very flacky

docs/cypress/e2e/links.cy.ts
calebpollman: Types should be declared in a _types.ts_ files or at the top of a file

docs/cypress/e2e/links.cy.ts
calebpollman: But do we need to use `JQuery` types?

0618: I think so, since we use some `JQuery` methods in line30-33, like `htmlTag.prop('href')`

calebpollman: Are those only `JQuery` methods?

docs/cypress/e2e/links.cy.ts
0618: Done. Moved to top.

docs/cypress/e2e/links.cy.ts
0618: yes, it's actully from the type of `cy.each`

```ts
each<E extends Node = HTMLElement>(fn: (element: JQuery<E>, index: number, $list: E[]) => void): Chainable<JQuery<E>>
```

docs/cypress/e2e/links.cy.ts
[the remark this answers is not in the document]
0618: NVM, I figured it out. We just need to use [`cypress-each`](https://github.com/bahmutov/cypress-each#repeat-the-test-n-times).

Updated the PR.

docs/cypress/e2e/links.cy.ts
```diff
@@ -0,0 +1,132 @@
...
+});
+
+describe(`All pages on Sitemap`, () => {
+  it.each(numberOfLinks)('all links on page %k should work', (_, i) => {
+    const link = allLinks[i];
+    cy.task('log', `🧪[TESTING...] page ${BASE_URL}/${link}`);
+    cy.visit(link || '/');
+    cy.get('a').each(hrefWorks);
+
+    function hrefWorks(htmlTag: JQuery<HTMLElement>): void {
```
calebpollman, line 36: Can we move `hrefWorks` and `logMessage` outside the scope of `it.each`?

docs/cypress/e2e/links.cy.ts
calebpollman: Got it, thanks

docs/cypress/e2e/links.cy.ts
zchenwei: +1

zchenwei (approved):
LGTM

calebpollman:
The more i have looked at this PR and come to understand it's use case better, is `Cypress` the right tool to use here? If we only need a headless web scraper to check the validity of links are there cheaper options that we could consider?

docs/cypress/support/commands.ts
0618: Note:

after removing these support files in https://github.com/0618/amplify-ui/commit/9f1fbf14c9acc054812d5f2c0c32d34c429b3342, build time reduced from 45+ min to 40- min.

|  Before | After  |
|---|---|
| image  | image  |

docs/cypress/e2e/links.cy.ts
0618: It somehow uses a lot of memory

Before moving them out

image

After moving them out, I tried 3 times

image
image
image


They all end up with the memory issue

> 2022-11-01T01:22:44.848Z [WARNING]: <--- Last few GCs --->
                                    [5077:0x1788003cc000]   886175 ms: Scavenge 639.2 (687.9) -> 634.5 (690.1) MB, 10.3 / 0.0 ms  (average mu = 0.989, current mu = 0.992) allocation failure;
                                    [5077:0x1788003cc000]   886337 ms: Scavenge 640.8 (690.1) -> 635.3 (691.6) MB, 7.9 / 0.0 ms  (average mu = 0.989, current mu = 0.992) allocation failure;
                                    [5077:0x1788003cc000]   886667 ms: Scavenge 642.4 (691.6) -> 637.0 (692.6) MB, 6.1 / 0.0 ms  (average mu = 0.989, current mu = 0.992) task;
2022-11-01T01:22:44.848Z [WARNING]: <--- JS stacktrace --->
2022-11-01T01:22:44.849Z [WARNING]: [5077:1101/012244.849160:ERROR:v8_initializer.cc(730)] V8 process OOM (Oilpan: Reserving memory.).


Trying reverting.

0618:
> The more i have looked at this PR and come to understand the use case better, is `Cypress` the right tool to use here? If we only need a headless web scraper to check the validity of links are there cheaper options that we could consider?

If we want to use e2e test, Cypress might be the only tool supported by Amplify Hosting, according to the [docs](https://docs.aws.amazon.com/amplify/latest/userguide/running-tests.html)

calebpollman:
> > The more i have looked at this PR and come to understand the use case better, is `Cypress` the right tool to use here? If we only need a headless web scraper to check the validity of links are there cheaper options that we could consider?
> 
> If we want to use e2e test, Cypress might be the only tool supported by Amplify Hosting, according to the [docs](https://docs.aws.amazon.com/amplify/latest/userguide/running-tests.html)

Did we look in to other options? Just think it's important that we have considered all options and are certain that `Cypress` is the right tool

0618:
> > > The more i have looked at this PR and come to understand the use case better, is `Cypress` the right tool to use here? If we only need a headless web scraper to check the validity of links are there cheaper options that we could consider?
> > 
> > 
> > If we want to use e2e test, Cypress might be the only tool supported by Amplify Hosting, according to the [docs](https://docs.aws.amazon.com/amplify/latest/userguide/running-tests.html)
> 
> Did we look in to other options? Just think it's important that we have considered all options and are certain that `Cypress` is the right tool

Good question. There are 3 options so far I'm aware of:
1. AWS CloudWatch Synthetic Canary broken link checker. Our current broken link check. The problems are 1) it runs periodically, but not triggered by our recent code changes 2) synthetic Canary can only test about 10 links. It would fail if there are more links, so we have to create a lot of canaries for the check. We currently only check all the links on home page and the component page, which is why we are moving away from it.
2. Cypress `cy.request`. It's this PR. Pros: 1) supported by Amplify Hosting, so it runs at every build 2) it doesn't require more dependencies. Cons: 1) Amplify Hosting has a 7GB memory limitation for the build docker image 2) Have to write big test logic on our own (this PR) 3) takes about 40 min to build
3. [broken-link-checker](https://github.com/stevenvachon/broken-link-checker). Set up a lambda to use broken-link-checker to check on each deployment. Pros: we don't have to be limited by the Amplify Hosting build docker image size. Cons: 1) it's a 3rd party library, which last published 5 years ago 2) have to maintain a separate repo

So none of the solutions are very good at this moment. Option 2 seems the best. At least, it's the one we have the most control. 

Is there any other options that I'm not aware of?

calebpollman:
> > > > The more i have looked at this PR and come to understand the use case better, is `Cypress` the right tool to use here? If we only need a headless web scraper to check the validity of links are there cheaper options that we could consider?
> > > 
> > > 
> > > If we want to use e2e test, Cypress might be the only tool supported by Amplify Hosting, according to the [docs](https://docs.aws.amazon.com/amplify/latest/userguide/running-tests.html)
> > 
> > 
> > Did we look in to other options? Just think it's important that we have considered all options and are certain that `Cypress` is the right tool
> 
> Good question. There are 3 options so far I'm aware of:
> 
> 1. AWS CloudWatch Synthetic Canary broken link checker. Our current broken link check. The problems are 1) it runs periodically, but not triggered by our recent code changes 2) synthetic Canary can only test about 10 links. It would fail if there are more links, so we have to create a lot of canaries for the check. We currently only check all the links on home page and the component page, which is why we are moving away from it.
> 2. Cypress `cy.request`. It's this PR. Pros: 1) supported by Amplify Hosting, so it runs at every build 2) it doesn't require more dependencies. Cons: 1) Amplify Hosting has a 7GB memory limitation for the build docker image 2) Have to write big test logic on our own (this PR) 3) takes about 40 min to build
> 3. [broken-link-checker](https://github.com/stevenvachon/broken-link-checker). Set up a lambda to use broken-link-checker to check on each deployment. Pros: we don't have to be limited by the Amplify Hosting build docker image size. Cons: 1) it's a 3rd party library, which last published 5 years ago 2) have to maintain a separate repo
> 
> So none of the solutions are very good at this moment. Option 2 seems the best. At least, it's the one we have the most control.
> 
> Is there any other options that I'm not aware of?

Selenium is pretty straightforward for web scraping. Not going to block this PR, but in the future it would be good to discuss the plan for a feature like this as a team

0618:
> > > > > The more i have looked at this PR and come to understand the use case better, is `Cypress` the right tool to use here? If we only need a headless web scraper to check the validity of links are there cheaper options that we could consider?
> > > > 
> > > > 
> > > > If we want to use e2e test, Cypress might be the only tool supported by Amplify Hosting, according to the [docs](https://docs.aws.amazon.com/amplify/latest/userguide/running-tests.html)
> > > 
> > > 
> > > Did we look in to other options? Just think it's important that we have considered all options and are certain that `Cypress` is the right tool
> > 
> > 
> > Good question. There are 3 options so far I'm aware of:
> > 
> > 1. AWS CloudWatch Synthetic Canary broken link checker. Our current broken link check. The problems are 1) it runs periodically, but not triggered by our recent code changes 2) synthetic Canary can only test about 10 links. It would fail if there are more links, so we have to create a lot of canaries for the check. We currently only check all the links on home page and the component page, which is why we are moving away from it.
> > 2. Cypress `cy.request`. It's this PR. Pros: 1) supported by Amplify Hosting, so it runs at every build 2) it doesn't require more dependencies. Cons: 1) Amplify Hosting has a 7GB memory limitation for the build docker image 2) Have to write big test logic on our own (this PR) 3) takes about 40 min to build
> > 3. [broken-link-checker](https://github.com/stevenvachon/broken-link-checker). Set up a lambda to use broken-link-checker to check on each deployment. Pros: we don't have to be limited by the Amplify Hosting build docker image size. Cons: 1) it's a 3rd party library, which last published 5 years ago 2) have to maintain a separate repo
> > 
> > So none of the solutions are very good at this moment. Option 2 seems the best. At least, it's the one we have the most control.
> > Is there any other options that I'm not aware of?
> 
> Selenium is pretty straightforward for web scraping. Not going to block this PR, but in the future it would be good to discuss the plan for a feature like this as a team

Thanks for mentioning Selenium. It seems a very powerful tool and it works with Cypress. We'll definitly consider Selenium as an option for our future test. 

Since the broken link test is pretty straightforward. It might be better to introduce less 3rd party libraries.

calebpollman (approved):

dbanksdesign (approved):
LGTM!

PR #2860, opened 2022-10-30
Add scrubbing to verify user

ioanabrooks:
#### Description of changes

* Scrub email and phone fields for verify user screen
* Fixes for Phonenumber field - convert to text

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react-native/src/Authenticator/Defaults/VerifyUser/VerifyUser.tsx
+++ b/packages/react-native/src/Authenticator/Defaults/VerifyUser/VerifyUser.tsx
@@ -1,5 +1,9 @@
 import React from 'react';
-import { authenticatorTextUtil } from '@aws-amplify/ui';
+import {
+  authenticatorTextUtil,
+  censorAllButFirstAndLast,
+  censorPhoneNumber,
+} from '@aws-amplify/ui';
 
 import { Button, ErrorMessage, Radio, RadioGroup } from '../../../primitives';
 import { DefaultFooter } from '../../common/DefaultFooter';
@@ -14,6 +18,19 @@ const COMPONENT_NAME = 'VerifyUser';
 //TODO: add getVerifyingText util
 const { getSkipText, getVerifyText, getAccountRecoveryInfoText } =
   authenticatorTextUtil;
+
+const censorContactInformation = (name: string, value: string): string => {
+  let censoredVal = value;
+  if (name === 'email') {
+    const splitEmail = value.split('@');
+    const censoredName = censorAllButFirstAndLast(splitEmail[0]);
+
+    censoredVal = `${censoredName}@${splitEmail[1]}`;
+  } else if (name === 'phone') {
+    censoredVal = censorPhoneNumber(value);
+  }
+  return censoredVal;
+};
 
 const VerifyUser: DefaultVerifyUserComponent = ({
   error,
@@ -65,8 +82,13 @@ const FormFields: DefaultVerifyUserComponent['FormFields'] = ({
 }) => {
   return (
     <RadioGroup disabled={isPending}>
-      {fields.map(({ value, ...props }) => (
-        <Radio {...props} key={value} value={value} />
+      {fields.map(({ name, value, ...props }) => (
+        <Radio
+          {...props}
+          key={value}
+          value={value}
+          label={censorContactInformation(name, value)}
+        />
       ))}
     </RadioGroup>
   );
--- a/packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
+++ b/packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
@@ -25,7 +25,7 @@ export default function useFieldValues<FieldType extends TypedField>({
   const sanitizedFields = useMemo(() => {
     if (!Array.isArray(fields)) {
       logger.warn(
-        `Invalid fields type of ${typeof fields} passed to ${componentName}. Fields must be of type array.`
+        `Invalid fields type of ${typeof fields} passed to ${componentName}. \`fields\` must be of type array.`
       );
       return [];
     }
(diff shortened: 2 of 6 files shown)


packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -39,9 +39,20 @@ export const getSanitizedRadioFields = (
...
 
+    let censoredVal = value;
+    if (name === 'email') {
+      const splitEmail = value.split('@');
+      const censoredName = censorAllButFirstAndLast(splitEmail[0]);
+
+      censoredVal = `${censoredName}@${splitEmail[1]}`;
+    } else if (name === 'phone') {
+      censoredVal = censorPhoneNumber(value);
+    }
```
calebpollman, line 50: Have we tested the verify contact flow with the values handled this way? Think that it may fail since we use `value` in the submit handler?

packages/react-native/src/primitives/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -51,7 +51,6 @@ export default function PhoneNumberField({
       <TextField
         {...rest}
         disabled={disabled}
-        keyboardType="phone-pad"
```
calebpollman, line 54: Why are we removing the `keyboardType`?

ioanabrooks: Confirmed phone-pad has +, - characters on both platforms

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
ioanabrooks: Moved logic to subcomponent.

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
```diff
@@ -25,7 +25,7 @@ export default function useFieldValues<FieldType extends TypedField>({
   const sanitizedFields = useMemo(() => {
     if (!Array.isArray(fields)) {
       logger.warn(
-        `Invalid fields type of ${typeof fields} passed to ${componentName}. fields must be of type array.`
+        `Invalid fields type of ${typeof fields} passed to ${componentName}. Fields must be of type array.`
```
calebpollman, line 28: The variable name is `fields`, the lowercase is intentional 😅

packages/react-native/src/Authenticator/hooks/useRouteTypedFields/utils.ts
```diff
@@ -54,7 +54,7 @@ export const getTypedField = ({
       }
 
       if (isPhoneFieldType && key === 'dialCodeList') {
-        return { ...acc, dialCodes: value as string[] };
+        return { ...acc, dialCodes: undefined };
```
calebpollman, line 57: While this is helpful for the default `fields`, this won't prevent the picker from rendering when users pass in custom fields with a `dialCodes` prop

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
wlee221: Maybe 
```suggestion
        `Invalid fields type of ${typeof fields} passed to ${componentName}. \`Fields\` must be of type array.`
```

packages/react-native/src/Authenticator/Defaults/VerifyUser/VerifyUser.tsx
```diff
@@ -15,6 +19,19 @@ const COMPONENT_NAME = 'VerifyUser';
...
+  let censoredVal = value;
+  if (name === 'email') {
+    const splitEmail = value.split('@');
+    const censoredName = censorAllButFirstAndLast(splitEmail[0]);
+
+    censoredVal = `${censoredName}@${splitEmail[1]}`;
+  } else if (name === 'phone') {
+    censoredVal = censorPhoneNumber(value);
+  }
+  return censoredVal;
```
wlee221, line 32: Will this fallback `return` ever be called?

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
wlee221: (edited for typo)

packages/react-native/src/Authenticator/hooks/useRouteTypedFields/utils.ts
ioanabrooks: Right but this is a temporary change, as we will add dialcodes before dev preview.

packages/react-native/src/Authenticator/Defaults/VerifyUser/VerifyUser.tsx
ioanabrooks: Fallback return? how so? it's the only return.

wlee221: I'm sorry, I misread the code 😓  Please disregard!

packages/react-native/src/Authenticator/hooks/useRouteTypedFields/utils.ts
wlee221: I'm out of context here, is `dialCodes` not implemented yet?

calebpollman: @ioanabrooks Regardless of what happens in the future this will not prevent users currently testing the component from rendering the Picker in a `PhoneNumberField`. Please see: https://github.com/aws-amplify/amplify-ui/commit/7e544f77d98fd1643dbac76f07c0f9129fe4ee09

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
calebpollman: Since this change is not relevant to the PR, let's remove it. If we want to change the messages in this file they should all be done and discussed in its own PR

PR #2871, opened 2022-10-31
chore: upgrade maplibre-gl-js-amplify to v2.1.0 to support getSuggestions and getPlace

sreeramsama:
#### Description of changes
Upgrade maplibre-gl-js-amplify to v2.1.0 to support getSuggestions and getPlace
Docs PR: https://github.com/aws-amplify/docs/pull/4615

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/e2e/cypress/integration/common/geo.ts
+++ b/packages/e2e/cypress/integration/common/geo.ts
@@ -16,7 +16,7 @@ When('I clear the search results', () => {
   cy.findByRole('textbox', {
     name: /search/i,
   }).trigger('mouseenter');
-  cy.findByRole('button', { name: 'Clear' }).click();
+  cy.findByRole('button', { name: 'Clear' }).click({ force: true });
 });
 
 When('I click on a map marker', () => {
--- a/packages/e2e/cypress/integration/common/geo.ts
+++ b/packages/e2e/cypress/integration/common/geo.ts
@@ -16,6 +16,8 @@ When('I clear the search results', () => {
   cy.findByRole('textbox', {
     name: /search/i,
   }).trigger('mouseenter');
+  // Adding 'force' as the clear button is hidden until we hover on textbox,
+  // and the click action seems to happen before the hover thus failing the check for the clear button element.
   cy.findByRole('button', { name: 'Clear' }).click({ force: true });
 });
 
--- a/packages/e2e/cypress/integration/common/geo.ts
+++ b/packages/e2e/cypress/integration/common/geo.ts
@@ -16,8 +16,10 @@ When('I clear the search results', () => {
   cy.findByRole('textbox', {
     name: /search/i,
   }).trigger('mouseenter');
-  // Adding 'force' as the clear button is hidden until we hover on textbox,
-  // and the click action seems to happen before the hover thus failing the check for the clear button element.
+  /**
+   * Adding 'force' as the clear button is hidden until we hover on textbox,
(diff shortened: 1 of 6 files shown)


wlee221 (dismissed):

packages/react/package.json
```diff
@@ -67,7 +67,7 @@
     "lodash": "4.17.21",
     "mapbox-gl": "1.13.1",
     "maplibre-gl": "2.1.9",
-    "maplibre-gl-js-amplify": "2.0.4",
+    "maplibre-gl-js-amplify": "2.1.0",
```
wlee221, line 70: nit: we should still make a changeset because this changes customer surface area

sreeramsama: Done. Added changeset.

wlee221 (dismissed):

thaddmt (approved):

thaddmt (dismissed):

packages/e2e/cypress/integration/common/geo.ts
```diff
@@ -16,7 +16,7 @@ When('I clear the search results', () => {
   cy.findByRole('textbox', {
     name: /search/i,
   }).trigger('mouseenter');
-  cy.findByRole('button', { name: 'Clear' }).click();
+  cy.findByRole('button', { name: 'Clear' }).click({ force: true });
```
wlee221, line 19: Why is this necessary? Can we add a comment if this is not immediately resolvable?

sreeramsama: 'Clear' button somehow couldn’t be found by cypress when its display property is being toggled, so added {force: true} for the click. I see it is already being used in other places so thought maybe it is an existing Cypress issue.

thaddmt (approved):

packages/e2e/cypress/integration/common/geo.ts
wlee221: Understood -- ideally, we should avoid `force: true` if possible, or explain in comment why this is necessary. For example: https://github.com/aws-amplify/amplify-ui/blob/c11de75eaf0e0476cea56d7cd2da1ffa04d1c639/packages/e2e/cypress/integration/common/shared.ts#L178-L184

Not blocking, but I think it's worth looking at for test maintainability purposes.

thaddmt (approved):

wlee221 (dismissed):

thaddmt (dismissed):

packages/e2e/cypress/integration/common/geo.ts
```diff
@@ -16,7 +16,9 @@ When('I clear the search results', () => {
   cy.findByRole('textbox', {
     name: /search/i,
   }).trigger('mouseenter');
-  cy.findByRole('button', { name: 'Clear' }).click();
+  // Adding 'force' as the clear button is hidden until we hover on textbox,
+  // and the click action seems to happen before the hover thus failing the check for the clear button element.
```
wlee221, line 20: Last nit, I promise 🙏 :

```suggestion
  /** 
  /**
   * Adding 'force' as the clear button is hidden until we hover on textbox,
   * and the click action happens before the hover.
   */
```

sreeramsama: Updated

thaddmt (approved):

wlee221 (approved):

PR #2885, opened 2022-11-02
feat(account-settings): ChangePassword component

wlee221:
#### Description of changes

This implements `ChangePassword` component. 

#### What this implements:

- [x] Basic file structure for `@aws-amplify/ui-react` and `@aws-amplify/ui`
- [x] `Auth.changePassword` functionality
- [x] `onSuccess`, `onError` callbacks
- [x] Rendering Cognito error
- [x] Basic type structure 

#### What this does not implement (yet):

- [ ] confirm password and input validation (next PR)
- [ ] E2E test (next PR)
- [ ] Subcomponent Overrides (will be implemented after other components are done)

Unit test added for `ChangePassword`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/react/src/components/AccountSettings/types.ts
```diff
@@ -0,0 +1,38 @@
...
+ * These are primitive prop types that account settings component use.
+ *
+ * Note that `PrimitieProps` is used to get native html types, like `onSubmit`.
+ */
+type CommonPasswordFieldProps = PrimitiveProps<PasswordFieldProps, 'input'>;
+type CommonButtonProps = PrimitiveProps<ButtonProps, 'button'>;
+type CommonAlertProps = PrimitiveProps<AlertProps, 'div'>;
+
+/**
+ * These are component override types (e.g. submit button).
```
wlee221, line 19:
```suggestion
 * These are overridable component types (e.g. submit button).
```

packages/react/src/components/AccountSettings/ChangePassword/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface ChangePasswordProps {
```
wlee221, line 1: More `ChangePassword` types will be added to this file in later PRs, like validator types.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const currentPasswordLabel = translate('Current Password');
+  const newPasswordLabel = translate('New Password');
+  const updatePasswordText = translate('Update password');
+
+  /** Auth API Handler */
+  const handleChangePassword = React.useCallback(async () => {
+    const { currentPassword, newPassword } = formValues;
+    try {
+      await changePassword({ user, currentPassword, newPassword });
+      setError(null);
```
dbanksdesign, line 37: Should we move this above the `await` because it will block until it returns right? If there was a previous error should we hide the error if there is a previous error while we wait for a response?

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  // return null if Auth.getCurrentAuthenticatedUser is still in progress
+  if (isAuthUserLoading) {
+    return null;
+  }
+
+  if (!user) {
+    logger.warn('<ChangePassword /> requires user to be authenticated.');
+    return null;
+  }
```
dbanksdesign, line 79: Should we move these up to return earlier in the function?

packages/ui/src/helpers/accountSettings/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  user: AmplifyUser;
+  currentPassword: string;
+  newPassword: string;
+};
+
+export const changePassword = async ({
+  user,
+  currentPassword,
+  newPassword,
+}: ChangePasswordInput) => {
```
dbanksdesign, line 14: TS noob question, should we type the return of this function?

packages/react/src/components/AccountSettings/types.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Note that `PrimitieProps` is used to get native html types, like `onSubmit`.
+ */
+type CommonPasswordFieldProps = PrimitiveProps<PasswordFieldProps, 'input'>;
+type CommonButtonProps = PrimitiveProps<ButtonProps, 'button'>;
+type CommonAlertProps = PrimitiveProps<AlertProps, 'div'>;
+
+/**
+ * These are overridable component types (e.g. submit button).
+ */
+export type AccountSettingsPasswordField<Props = {}> = React.ComponentType<
```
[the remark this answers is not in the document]
dbanksdesign: We are not publicly exporting these types right? I'm not worried about the name either way

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  return (
+    <View as="form" className="amplify-changepassword" onSubmit={handleSubmit}>
+      <Flex direction="column">
```
dbanksdesign, line 83: We could simplify these 2 lines I think:
```suggestion
    <Flex as="form" direction="column" className="amplify-changepassword" onSubmit={handleSubmit}>
```

dbanksdesign (approved):
Nothing blocking, just a few minor nits. Great 'TODO' comments. LGTM!

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
calebpollman: They have to be below the hooks to guarantee the hooks are run in the same order on each render (and not conditionally)

packages/react/src/components/AccountSettings/ChangePassword/__tests__/ChangePassword.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    render(<ChangePassword onError={onError} />);
+
+    const submitButton = await screen.findByRole('button', {
+      name: 'Update password',
+    });
+
+    fireEvent.submit(submitButton);
+
+    // submit handling is async, wait for onError to be called
+    await waitFor(() => expect(onError).toBeCalledTimes(1));
```
calebpollman, line 85: Don't know _@testing-library/react_ well enough to have a full opinion here, but it seems odd to me to place an `expect` within the scope of `waitFor`. Is that the expected usage?

packages/react/src/components/AccountSettings/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  AlertProps,
+  ButtonProps,
+  PasswordFieldProps,
+  PrimitiveProps,
+} from '../../primitives/types';
+
+/**
+ * These are primitive prop types that account settings component use.
+ *
+ * Note that `PrimitieProps` is used to get native html types, like `onSubmit`.
```
calebpollman, line 12: Typo:
```suggestion
 * Note that `PrimitiveProps` is used to get native html types, like `onSubmit`.
 ```

File: packages/react/src/components/index.tsx [hunk 10]
```diff
@@ -1,3 +1,3 @@
-export * from './ThemeProvider';
```
calebpollman, line 1: Just scrapping the `ThemProvider` huh 😅

packages/react/src/components/AccountSettings/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ChangePassword';
```
calebpollman, line 1: Should we be exporting everything from `ChangePassword`?

packages/ui/src/helpers/accountSettings/utils.ts
wlee221: I personally like to have them be auto-inferred for simple functions like this, but it's debatable. 🤔 

Either way, it gets TS completion.

packages/react/src/components/AccountSettings/types.ts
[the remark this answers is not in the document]
wlee221: We're not. I'll leave it as-is for now then 👍

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: That's a fair point, adjusting.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Good point, I'll adjust accordingly.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+  DefaultError,
+  DefaultNewPassword,
+  DefaultSubmitButton,
+} from './defaultComponents';
+import { ChangePasswordProps } from './types';
+import { FormValues } from '../types';
+
+const logger = new Logger('ChangePassword');
+
+const ChangePassword: React.ComponentType<ChangePasswordProps> = (props) => {
```
calebpollman, line 19: Unless there is requirement to type `ChangePassword` with `React.ComponentType`, we don't generally need need to use when typing components:
```suggestion
function ChangePassword(props: ChangePasswordProps): JSX.Element | null {
```

Basically `React.ComponentType` is useful for declaring override component params on type interfaces due to how generic/flexible it is, but since we know exactly what `ChangePassword` is (a functional component), we can skip using it here

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
calebpollman: Agree with @dbanksdesign comment. Additionally would only call this if `error` is not `null`

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+  /** Auth API Handler */
+  const handleChangePassword = React.useCallback(async () => {
+    const { currentPassword, newPassword } = formValues;
+    try {
+      await changePassword({ user, currentPassword, newPassword });
+      setError(null);
+
+      if (onSuccess) {
+        onSuccess(); // notify success to the parent
+      }
```
calebpollman, line 42: Could just use optional chaining here (`onSuccess?.()`), same with the error handler below

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: I've initially tried this, but `onSubmit` handler for `Flex as="form"` had TS errors:

```
Type '(event: React.FormEvent<HTMLFormElement>) => void' is not assignable to type 'FormEventHandler<HTMLDivElement>'.
```

I'll create a repro and submit an issue. I think we can address this later after then.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+  DefaultNewPassword,
+  DefaultSubmitButton,
+} from './defaultComponents';
+import { ChangePasswordProps } from './types';
+import { FormValues } from '../types';
+
+const logger = new Logger('ChangePassword');
+
+const ChangePassword: React.ComponentType<ChangePasswordProps> = (props) => {
+  const { onSuccess, onError } = props;
```
calebpollman, line 20: Nit: destructure in function declaration

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Actually, these have to be placed below all the `React.useCallback` hooks due to rules of hooks. 

I'll keep it here, but clarify comments around these return codes.

packages/ui/src/helpers/accountSettings/utils.ts
wlee221: Actually, I think it's beneficial to have return annotation because so that other readers don't have to know what `Auth.changePassword` does or returns. I'll add a return type!

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+      }
+    } catch (e) {
+      const error = e as Error;
+      setError(error.message);
+
+      if (onError) {
+        onError(error); // notify error to the parent
+      }
+    }
+  }, [formValues, user, onSuccess, onError]);
```
calebpollman, line 51: Not sure if it makes sense to memoize this callback. Correct me if i'm wrong, but `formValues` is going to update on each call to `handleChange` and unless the reference to `user` is memoized in `useAuth` it will return a new object on each call, basically making memoization pointless and just additional memory overhead

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+      const { name, value } = event.target;
+      setFormValues({ ...formValues, [name]: value });
+    },
+    [formValues]
+  );
+
+  const handleSubmit = React.useCallback(
+    (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+      handleChangePassword();
```
calebpollman, line 67: We can probably just move the logic in `handleChangePassword` inside `handleSubmit`, unless there is a longer term plan to reuse `handleChangePassword`

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return <PasswordField {...props} />;
+};
+
+export const DefaultSubmitButton: AccountSettingsSubmitButton = (props) => {
+  const { children, ...rest } = props;
+  return <Button {...rest}>{children}</Button>;
+};
+
+export const DefaultError: AccountSettingsError = (props) => {
+  const { errorMessage, ...rest } = props;
```
calebpollman, line 25: Since errorMessage is just the `children` we pass to the underlying `Alert`, we can just remove the `errorMessage` prop and just `children` as a passthrough prop:
```
// usage example in ChangePassword
<DefaultError>{error}</DefaultError>
```

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+  DefaultSubmitButton,
+} from './defaultComponents';
+import { ChangePasswordProps } from './types';
+import { FormValues } from '../types';
+
+const logger = new Logger('ChangePassword');
+
+const ChangePassword: React.ComponentType<ChangePasswordProps> = (props) => {
+  const { onSuccess, onError } = props;
+  const [error, setError] = React.useState<string>(null);
```
calebpollman, line 21: Since `error` is just a string and not an `Error`, suggest renaming to something indicative of that like `errorMessage`/`setErrorMessage`

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: FYI created #2891

packages/ui/src/helpers/accountSettings/utils.ts
wlee221: Done in [05472b6](https://github.com/aws-amplify/amplify-ui/pull/2885/commits/05472b6582c90f19b53f6165062450415fb1a779)! Did a little bit more and added a wrapper around what `Auth.changePassword` returns.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done in [05472b6](https://github.com/aws-amplify/amplify-ui/pull/2885/commits/05472b6582c90f19b53f6165062450415fb1a779) 🙏

calebpollman (commented):
Great start!

packages/react/src/components/index.tsx
wlee221: Fixed xD

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const [formValues, setFormValues] = React.useState<FormValues>({});
+
+  const { user, isLoading: isAuthUserLoading } = useAuth();
```
calebpollman, line 24: Don't see a benefit in reassigning the `isLoading`?

examples/next/pages/ui/components/account-settings/change-password/index.page.tsx
```diff
@@ -0,0 +1,34 @@
...
+  Card,
+} from '@aws-amplify/ui-react';
+import '@aws-amplify/ui-react/styles.css';
+
+import awsExports from './aws-exports';
+Amplify.configure(awsExports);
+
+export default function AuthenticatorWithEmail() {
+  return (
+    <>
```
calebpollman, line 16: This fragment is extraneous i think?

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,109 @@
...
+  /** Translations */
+  // TODO: add AccountSettingsTextUtil to collect these strings
+  const currentPasswordLabel = translate('Current Password');
+  const newPasswordLabel = translate('New Password');
+  const updatePasswordText = translate('Update password');
+
+  /** Auth API Handler */
+  const handleChangePassword = React.useCallback(async () => {
+    const { currentPassword, newPassword } = formValues;
+    setError(null);
```
calebpollman, line 35: See: https://github.com/aws-amplify/amplify-ui/pull/2885#discussion_r1013416923

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: That's very fair, convinced. Adjusting.

packages/react/src/components/AccountSettings/ChangePassword/__tests__/ChangePassword.test.tsx
wlee221: It is, at least according to its documentation: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor.

```ts
// ...
// Wait until the callback does not throw an error. In this case, that means
// it'll wait until the mock function has been called once.
await waitFor(() => expect(mockAPI).toHaveBeenCalledTimes(1))
// ...
```

wlee221: I'll add a comment since this is not a familiar API:

`
```suggestion
    // https://testing-library.com/docs/dom-testing-library/api-async/#waitfor
    await waitFor(() => expect(onError).toBeCalledTimes(1));
```

examples/next/pages/ui/components/account-settings/change-password/index.page.tsx
wlee221: Oh shoot, I did not mean to add this yet. Please disregard [🤦](https://emojipedia.org/person-facepalming/)

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Understood! Adjusting.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: IMO I thought `isLoading` was somewhat ambiguous, because it sounds like the whole component is loading. What do you think?

calebpollman: Since we return `null` as the component if `isLoading` is `true`, in a way we could say the the whole component is loading right?

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
wlee221: I specifically had `errorMessage` as a prop so that customer may do any errorMessage handling (like rewording). That said, I do see this is an antipattern to other default component here. (and same argument can be made with `DefaultSubmitButton). 

What do you think?

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Fair, adjusting.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Made more explicit in 97936eec8e4fce7511c6c3c5deee9aed98b9984b: 

https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/index.ts#L1

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
calebpollman: Couldn't they just do something like:
```
// not sure what the component slot name is so using `Error`
<ChangePassword components={{ Error: ({ children, ...props }) => <ChangePassword.Error {...props}>{reformatError(children)}</ChangePassword.Error> />
```

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done in https://github.com/aws-amplify/amplify-ui/commit/97936eec8e4fce7511c6c3c5deee9aed98b9984b:

https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx#L61-L66

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done: https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx#L19-L22

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done, see https://github.com/aws-amplify/amplify-ui/commit/97936eec8e4fce7511c6c3c5deee9aed98b9984b.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done, https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx#L53-L59

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done: https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx#L53-L59

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx#L26

packages/react/src/components/AccountSettings/index.ts
wlee221: Made more explicit: https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/index.ts

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done: https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx#L19-L22

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
wlee221: You're right, adjusting.

wlee221: Done in 4ea9fa43c00ef82f77c574bf9d7c45a3ed55b435

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done: https://github.com/aws-amplify/amplify-ui/blob/97936eec8e4fce7511c6c3c5deee9aed98b9984b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx#L64

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: ---
After addressing Caleb's feedbacks, I ended up removing these `useCallback` hooks, so I moved these returns to the top.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { useAuth } from '../../../internal';
+import { View, Flex } from '../../../primitives';
+import {
+  DefaultCurrentPassword,
+  DefaultError,
+  DefaultNewPassword,
+  DefaultSubmitButton,
+} from './defaultComponents';
+import { ChangePasswordProps } from './types';
+import { FormValues } from '../types';
```
zchenwei, line 15: Maybe we can start using `import type xxx`?

wlee221: I'm definitely open to the idea. I'm not too familiar with the syntax though, what's the benefit of using `import type` over `import`?

zchenwei: As far as I know, it is an explicit way to tell TS compiler that it is safe to not emit a js file for that import since it is just importing typings.

calebpollman: My two cents on this:
- we should be writing our _types.ts_ so that they do only contain types
- if we are going to start using the `import type` syntax it should be enforced by eslint (see rule [here](https://typescript-eslint.io/rules/consistent-type-imports/))

packages/ui/src/helpers/accountSettings/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  user,
+  currentPassword,
+  newPassword,
+}: ChangePasswordInput): Promise<void> => {
+  try {
+    /**
+     * Auth.changePassword returns `Promise<"SUCCESS">`. We're not interested
+     * in its resolved string value, so we just return Promise.resolve() on success.
+     */
+    await Auth.changePassword(user, currentPassword, newPassword);
```
calebpollman, line 20: Not super important/non-blocking,  but what do you think of improving our logging for utilities that call Amplify JS?

calebpollman (approved):
LGTM. Just left a non-blocking comment

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: This sounds like non-trivial (but important!) refactor in React source code. What do you think about:
- I'll enforce the convention of `types.ts` to contain only type files
- Let's do a refactor after account settings and file uploader components are merged to clean up types and use `import type` with eslint rules enforced.

calebpollman: > I'll enforce the convention of types.ts to contain only type files for my upcoming PRs.

🎉

> Let's do a refactor after account settings and file uploader components are merged to clean up types and use import type with eslint rules enforced.

Cool! Good end of year item

packages/ui/src/helpers/accountSettings/utils.ts
wlee221: Added logger util in [6a9d385](https://github.com/aws-amplify/amplify-ui/pull/2885/commits/6a9d3850a7f8d85e67aed74ed876073934f16468)

packages/ui/src/helpers/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { Logger } from 'aws-amplify';
+
+type LoggerCategory = 'Auth' | 'Storage' | 'Geo';
```
calebpollman, line 3: Cause InApp 😅 :
```suggestion
type LoggerCategory = 'Auth' | 'Geo' | 'Notifications' | 'Storage';
```

packages/ui/src/helpers/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { Logger } from 'aws-amplify';
+
+type LoggerCategory = 'Auth' | 'Geo' | 'Notifications' | 'Storage';
+
+export const getCategoryLogger = (category: LoggerCategory) =>
```
ioanabrooks, line 5: super nit: `getCategoryLogger` -> `getLogger`

ErikCH (approved):
👍

packages/ui/src/helpers/utils.ts
wlee221: Done in [b1796f0](https://github.com/aws-amplify/amplify-ui/pull/2885/commits/b1796f0bdddecbf9af3266324edbbfcf0c25fd3b)

dbanksdesign (approved):
LGTM!

PR #2887, opened 2022-11-02
chore(rna): Improve Authenticator stories

joebuono:
#### Description of changes

- Consolidates all Authenticator subcomponent stories into one file
- Adds `getComponentSlots` and `getComponentFields` utils
- Replaces `Screen` component with `DefaultContainer`

It would be nice to figure out how we can use `knobs` or `addons` to toggle different types of input fields. For example, auth-with-username would render a TextField, and auth-with-phone-number would render a PhoneNumberField. 

https://user-images.githubusercontent.com/48109584/199615853-d435c280-45ac-424f-a015-d0273be3a85a.mov

Ran Storybook on both iOS and Android. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


examples/react-native/storybook/ui/InnerContainer.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { SafeAreaView, StyleSheet } from 'react-native';
+
+interface InnerContainerProps {
+  children: React.ReactNode;
+}
+
+export const InnerContainer = ({ children }: InnerContainerProps) => {
+  return <SafeAreaView style={[styles.container]}>{children}</SafeAreaView>;
```
ioanabrooks, line 9: Curious why we didn't just use the `InnerContainer` component from `DefaultContainer` pkg. Is it because we need SafeAreaView here? Concerned that if we make changes to the InnerContainer we'll have to make changes here as well [either props or styling].

joebuono: Good point! I'll import it from the `DefaultContainer` pkg

examples/react-native/storybook/utils/index.ts
```diff
@@ -0,0 +1,55 @@
+// how can we make this type more strict than `any`?
+export const getComponentSlots = (subcomponent: any) => {
```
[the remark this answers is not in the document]
joebuono: Good point

ioanabrooks (dismissed):

examples/react-native/storybook/stories/Authenticator.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+  fields: [],
+  skipVerification: noop,
+};
+
+storiesOf('Authenticator', module)
+  .addDecorator((Story: any) => (
+    <InnerContainer>
+      <Story />
+    </InnerContainer>
+  ))
```
calebpollman, line 159: We need the `DefaultContainer` too right?

joebuono: I don't think we need to because we're already using the `Screen` component as a decorator for each of our stories, and it has the same styles as `Default Container`:

```
  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  ```

Comment (calebpollman) [reply examples/react-native/storybook/stories/Authenticator.stories.tsx:159]: What if the `DefaultContainer` changes?

Comment (joebuono) [reply examples/react-native/storybook/stories/Authenticator.stories.tsx:159]: Hmm. Maybe we can replace the `Screen` component with `DefaultContainer`?

File: examples/react-native/storybook/index.tsx [hunk 4]
```diff
@@ -26,9 +26,9 @@ export function setupStorybook(initStorybook: boolean) {
     // add decorators
     addDecorator(withKnobs);
     addDecorator((Story: any) => (
-      <Screen>
+      <DefaultContainer>
```
calebpollman, line 29: Think this is fine for now but we should probably decouple `DefaultContainer` from our stories in the long term, which most likely means removing this decorator altogether

calebpollman (approved):
LGTM 🚢

examples/react-native/storybook/index.tsx
ioanabrooks: +1 to decouple decorators from our components. If we change  DefaultContainer  it's fine for our authenticator sub-component stories, but perhaps it shouldn't be applied to all primitives stories.

ioanabrooks (approved):

PR #2900, opened 2022-11-04
feat: add password validation to `ConfirmPassword`

wlee221:
### Description of changes

This adds password validation to `ConfirmPassword`. 

- [x] `<ChangePassword />` now renders `ConfirmPassword`
- [x] `<ChangePassword />` now optionally takes in `validators` prop, which overrides default password validation if specified
- [x] `<ChangePassword />` now handles blur event, and stores that info accordingly
- [x] added new utils and types for new validator impelementation

### Details 

Since `AccountSettings` components do not use xstate, much of validation had to be reimplemented.

#### Interface

Here's the new type for each validator: 

https://github.com/aws-amplify/amplify-ui/blob/6bc2edb5f8b46cf21966324ada78d5751788a11b/packages/ui/src/types/accountSettings/validator.ts#L7-L11

- `validationMode: 'onBlur' | 'onChange' | 'onTouched'` denotes when each validation will happen. 
  - Notably, `onTouched` happens on first blur event, then on every change event after that. We will default to this mode.
  - This is inspired from https://react-hook-form.com/api/useform
- `validate` function is the actual validator function.
- `message` denotes the error message.

For example, here's a minLength validator:

```ts
const minLength: FieldValidator = {
  validationMode: 'onBlur',
  valdiate: (field) => field.length > 4,
  message: 'your password must have length of 4 or greater'
}
```

New unit tests added.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,161 @@
...
+
+  const validateConfirmPassword = React.useCallback(
+    (
+      formValues: FormValues,
+      blurredFields: BlurredFields,
+      eventType: FormEventType
+    ): string[] => {
+      const { newPassword, confirmPassword } = formValues;
+
+      const confirmPasswordValidator = getConfirmPasswordValidator(newPassword);
```
wlee221, line 94: `confirmPasswordValidator` is getting generated dynamically, since it depends on `confirmPassword`.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,161 @@
...
   DefaultError,
   DefaultNewPassword,
   DefaultSubmitButton,
 } from './defaultComponents';
 import { ChangePasswordProps } from './types';
-import { FormValues } from '../types';
 
 const logger = new Logger('ChangePassword');
 
+const getIsDisabled = (
```
wlee221, line 29: Pulled this outside `ChangePassword` because it's really an util that doesn't need to be redefined everytime.

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
```diff
@@ -12,8 +13,28 @@ export const DefaultCurrentPassword: AccountSettingsPasswordField = (props) => {
...
-export const DefaultNewPassword: AccountSettingsPasswordField = (props) => {
-  return <PasswordField {...props} />;
+export const DefaultNewPassword: AccountSettingsPasswordField = ({
+  validationErrors,
+  ...props
+}) => {
+  return (
+    <>
+      <PasswordField {...props} />
+      <ValidationErrors errors={validationErrors} />
```
wlee221, line 23: Note that we're using `ValidationErrors` helper component because we need to render `string[]`. `PasswordField` only supports rendering just `string` ([docs](https://ui.docs.amplify.aws/react/components/passwordfield?tab=props)).

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,168 @@
...
+    return true;
+  } else if (
+    // if there are some password validation error, disable submit
+    validationError.newPassword?.length > 0 ||
+    validationError.confirmPassword?.length > 0
+  ) {
+    return true;
+  }
+
+  return false;
```
0618, line 46: What do you think if we make it more self-explained?

```suggestion
const noPasswordEntered = !currentPassword || !newPassword || !confirmPassword;
const invalidPassword = validationError.newPassword?.length > 0 || validationError.confirmPassword?.length > 0;

return noPasswordEntered ||  invalidPassword;
```

packages/react/src/components/AccountSettings/ChangePassword/__tests__/ChangePassword.test.tsx
```diff
@@ -98,7 +110,129 @@ describe('ChangePassword', () => {
 
     fireEvent.submit(submitButton);
 
-    // submit handling is async, wait for error to be displayed
-    await waitFor(() => expect(screen.findByText('Mock Error')).toBeDefined());
+    // findBy is already async, so we do not need to use `waitFor`
+    // https://testing-library.com/docs/queries/about#types-of-queries
```
0618, line 114: #TIL 

This comment makes a lot of sense when reading this PR, but if without the previous code, the reader may be confused why you need to mention `waitFor` here. Maybe move it to a PR comment instead?

0618 (approved):
Left nit comments

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
+import { FieldValidator, ValidationMode } from '../../../types';
+
+/**
+ * `shouldValidate` determines whether validator should be run, based on validation mode,
+ * input event type, and whether it has been blurred yet.
+ */
+export const shouldValidate = (
+  validationMode: ValidationMode,
+  eventType: 'change' | 'blur',
```
joebuono, line 9: nit: Can we use `FormEventType` here?

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+};
+
+/**
+ * `runFieldValidator` runs all validators, and returns error messages.
+ */
+export const runFieldValidators = (
+  field: string,
+  validators: FieldValidator[],
+  eventType: 'change' | 'blur',
```
joebuono, line 37: Can we use `FormEventType` here?

packages/ui/src/helpers/accountSettings/validators/password.ts
```diff
@@ -0,0 +1,84 @@
...
+/** Gets password setting from Amplify configuration */
+export const getPasswordSettings = () => {
+  // need to cast to any because `Amplify.configure()` isn't typed properly
+  const config = Amplify.configure() as any;
+  return config?.aws_cognito_password_protection_settings as PasswordSettings;
+};
+
+export const getMinLengthValidator = (minLength: number): FieldValidator => ({
+  validationMode: 'onTouched',
+  validate: (field) => field.length >= minLength,
```
joebuono, line 14: nit: I wonder if calling it `isValid` more clearly indicates that it returns a boolean?

packages/ui/src/helpers/accountSettings/validators/util.ts
wlee221: Thanks for the catch again, adjusting.

packages/react/src/components/AccountSettings/ChangePassword/__tests__/ChangePassword.test.tsx
wlee221: Agreed! https://github.com/aws-amplify/amplify-ui/pull/2900/commits/11593c3ef3679fcb4f485914940d2b3249592342

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: I love this suggestion, so much cleaner. Done in [52d5b02](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/52d5b021b566d63b98ed4a4d1573aa2b98f468ed)!

packages/ui/src/helpers/accountSettings/validators/util.ts
wlee221: Fixed in [c0403a1](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/c0403a12022bdfd6f63e8c6e9bd34000fc351831)!

wlee221: Thinking it should maybe named `InputEventType` instead of `FormEventType` to be clearer too... 🤔

packages/ui/src/helpers/accountSettings/validators/util.ts
wlee221: Done, see above comment

packages/react/src/components/AccountSettings/types.ts
```diff
@@ -33,3 +33,7 @@ export type AccountSettingsError<Props = {}> = React.ComponentType<
 
 /** Form specific types */
 export type FormValues = Record<string, string>;
+
+export type BlurredFields = Record<string, boolean>;
+
+export type ValidationError = Record<string, string[]>;
```
joebuono, line 39: nit: Wondering if it's possible/necessary to be more specific than `string[]`

joebuono (approved):

packages/ui/src/helpers/accountSettings/validators/password.ts
wlee221: That's fair. I personally think `isValid` sounds like a boolean than a function that returns boolean, so I'm hesitant on the change.

Wonder what others think too

packages/react/src/components/AccountSettings/types.ts
wlee221: I do not think so, since validation errors are really dynamic (especially with custom validators)

packages/ui/src/helpers/accountSettings/validators/util.ts
wlee221: Above renaming done in dc6411f0f28c0d0a653839288c74031f5b33d10d

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,163 @@
...
+
+  const validateConfirmPassword = React.useCallback(
+    (
+      formValues: FormValues,
+      blurredFields: BlurredFields,
+      eventType: InputEventType
+    ): string[] => {
+      const { newPassword, confirmPassword } = formValues;
+
+      const confirmPasswordValidator = getConfirmPasswordValidator(newPassword);
```
calebpollman, line 93: Maybe we should update `getConfirmPasswordValidator` to return an array so we don't have to wrap `confirmPasswordValidator` in an array on line 97

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,163 @@
...
+    []
+  );
+
+  const runValidation = React.useCallback(
+    (
+      formValues: FormValues,
+      blurredFields: BlurredFields,
+      eventType: InputEventType
+    ) => {
+      const newPasswordErrors = validateNewPassword(
```
calebpollman, line 111: `newPasswordErrors` kind of reads as though there were previous password errors

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -68,6 +178,17 @@ function ChangePassword({
     }
   };
 
+  /** Return null if Auth.getCurrentAuthenticatedUser is still in progress  */
```
calebpollman, line 181: Think this can just be an inline comment, same with other one line comments

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
```diff
@@ -12,8 +13,28 @@ export const DefaultCurrentPassword: AccountSettingsPasswordField = (props) => {
...
+    </>
+  );
+};
+
+export const DefaultConfirmPassword: AccountSettingsPasswordField = ({
```
calebpollman, line 28: Looks like we could just have one component, `DefaultPasswordField` and use it in place of `DefaultNewPassword` and `DefaultConfirmPassword`?

packages/react/src/components/AccountSettings/ChangePassword/types.ts
```diff
@@ -1,6 +1,10 @@
+import { FieldValidator } from '@aws-amplify/ui';
+
 export interface ChangePasswordProps {
   // callback once password is successfully updated
   onSuccess?: () => void;
   // callback when there's an error
   onError?: (error: Error) => void;
+  // custom password validations
```
calebpollman, line 8: This is a scenario where i think JSDoc syntax is helpful, just noting

packages/react/src/components/AccountSettings/types.ts
```diff
@@ -33,3 +33,7 @@ export type AccountSettingsError<Props = {}> = React.ComponentType<
 
 /** Form specific types */
 export type FormValues = Record<string, string>;
+
+export type BlurredFields = Record<string, boolean>;
```
calebpollman, line 37: Couple things here:
- Would it make sense to track more than just whether or not the field has been blurred? Might be helpful to be able to look at whether or not a `field` has been `touched`, `dirty`, etc
- From what i can tell from reading the usage, we never assign any `false` values to `BlurredFields`

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,163 @@
...
+
+    setFormValues(newFormValues);
+  };
+
+  const handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    event.preventDefault();
+
+    const { name } = event.target;
+    const newBlurredFields = { ...blurredFields, [name]: true };
+    runValidation(formValues, newBlurredFields, 'blur');
```
calebpollman, line 159: Wonder if this should be ran as an effect rather than called here, thoughts?

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,163 @@
...
+    setBlurredFields({ ...blurredFields, [name]: true });
```
calebpollman, line 160: Do we need to call `setBlurredFields` if the field has already been blurred? Why can't we just directly pass `newBlurredFields` here?

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
calebpollman: Can we just pass a `string` with line breaks (`\n`) to prevent the need for `ValidationErrors`?

calebpollman (commented):
Left some feedback/questions/nits, thanks for doing this!

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+/**
+ * `runFieldValidator` runs all validators, and returns error messages.
+ */
+export const runFieldValidators = (
+  field: string,
+  validators: FieldValidator[],
+  eventType: InputEventType,
+  hasBlurred: boolean
```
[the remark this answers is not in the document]
wlee221: `isDirty` is different, because `isDirty` just means that value has been modified even if input hasn't been blurred yet if I'm reading this correctly:

> Each [input](https://www.w3.org/TR/2011/WD-html5-20110525/the-input-element.html#the-input-element) element has a boolean dirty value flag. The [dirty value flag](https://www.w3.org/TR/2011/WD-html5-20110525/the-input-element.html#concept-input-value-dirty-flag) must be initially set to false when the element is created, and must be set to true whenever the user interacts with the control in a way that changes the [value](https://www.w3.org/TR/2011/WD-html5-20110525/association-of-controls-and-forms.html#concept-fe-value).

- https://www.w3.org/TR/2011/WD-html5-20110525/the-input-element.html

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+    case 'onChange': {
+      // only run validtor on change event
+      return eventType === 'change';
+    }
+    case 'onTouched': {
+      /**
+       * run validator on first blur event, and then every subsequent
+       * blur/change event.
+       */
```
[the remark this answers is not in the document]
wlee221: Good point, let me just make these plain comments (these are internal helpers anyway)

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
...
+      return eventType === 'blur' || hasBlurred;
+    }
+  }
+};
+
+/**
+ * `runFieldValidator` runs all validators, and returns error messages.
+ */
+export const runFieldValidators = (
```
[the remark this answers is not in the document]
wlee221: Addressed in [5058f22](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/5058f223c7a49d4aa6cd72f4abd50127f8f6a96d)

packages/ui/src/helpers/accountSettings/validators/password.ts
```diff
@@ -0,0 +1,84 @@
+import { Amplify } from 'aws-amplify';
+import { hasSpecialChars } from '../../../helpers/authenticator';
+import { FieldValidator, PasswordSettings } from '../../../types';
+
+/** Gets password setting from Amplify configuration */
+export const getPasswordSettings = () => {
+  // need to cast to any because `Amplify.configure()` isn't typed properly
+  const config = Amplify.configure() as any;
+  return config?.aws_cognito_password_protection_settings as PasswordSettings;
```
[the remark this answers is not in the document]
wlee221: Done in [1388d25](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/1388d25b982e33d77da4d1589897e4d0b6c9c5bc)

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: IMO no, because this also depends on what the inputEvent was beyond React states `formValues` and `newBlurredFields`. We'll have to keep track of last `inputEvent` type if we want to convert this to an effect. 

I personally think having this as util makes it easier to see when it is triggered from which event.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Made optimization, and directly passed `newBlurredFields` in [e187a92](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/e187a926bfedb42e0c9b131643548a4f77f901b7)

packages/react/src/components/AccountSettings/types.ts
wlee221: > Would it make sense to track more than just whether or not the field has been blurred? 

IMO no, because `hasBlurred` is really all we need to determine whether to run validation or not. This is true with larger scope validation libraries too, such as react-hook-forms: their [skipValidation.ts](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/skipValidation.ts).

> From what i can tell from reading the usage, we never assign any false values to BlurredFields
No, it really depends on `falsey` values then gets set to `true`. Do you have any suggestion to improve this? 

I can set all of them to `false` initially, but I'm not sure that's necessary when we really only care about whether or not it's `true`.

packages/react/src/components/AccountSettings/ChangePassword/types.ts
wlee221: Converted to jsdoc comment in [9a9fa07](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/9a9fa073a640b861a26ab719782a0d43fb78a6e3)

packages/ui/src/helpers/accountSettings/validators/util.ts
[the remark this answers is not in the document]
wlee221: This is done through other bigger commits, please see recently changed commits.

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
+import { FieldValidator, InputEventType, ValidationMode } from '../../../types';
+
+/**
+ * `shouldValidate` determines whether validator should be run, based on validation mode,
+ * input event type, and whether it has been blurred yet.
+ */
```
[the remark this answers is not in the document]
wlee221: Converted to inline comments for these internal functions.

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
...
+export const shouldValidate = (
```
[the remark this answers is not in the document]
wlee221: Done in https://github.com/aws-amplify/amplify-ui/pull/2900/commits/5058f223c7a49d4aa6cd72f4abd50127f8f6a96d

packages/ui/src/helpers/accountSettings/validators/password.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+export const getMinLengthValidator = (minLength: number): FieldValidator => ({
```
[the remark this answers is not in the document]
wlee221: Done in https://github.com/aws-amplify/amplify-ui/pull/2900/commits/ee476ae2ca075bae5368bff508a3ea6fcb462dcd

packages/ui/src/helpers/accountSettings/validators/util.ts
```diff
@@ -0,0 +1,56 @@
...
+  field: string,
+  validators: FieldValidator[],
+  eventType: InputEventType,
+  hasBlurred: boolean
+): string[] => {
+  if (!field) return [];
+
+  const errors = [];
+
+  validators?.forEach((validator) => {
```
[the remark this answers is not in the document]
wlee221: https://github.com/aws-amplify/amplify-ui/pull/2900/commits/77710d8fb7614b1c8ddc5dcc6a953c4c40294f2e

packages/ui/src/helpers/accountSettings/validators/index.ts
```diff
@@ -0,0 +1,3 @@
+export { getConfirmPasswordValidator } from './confirmPassword';
+export { getDefaultPasswordValidators } from './password';
```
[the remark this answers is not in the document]
wlee221: Converted them to `getDefaultPasswordValidators` and `getDefaultConfirmPasswordValidators`, to denote the field name they expect to validate.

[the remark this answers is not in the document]
wlee221: https://github.com/aws-amplify/amplify-ui/pull/2900/commits/c07c230b9bbf0c6cde9652607e684795627be2b1

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done in https://github.com/aws-amplify/amplify-ui/pull/2900/commits/c07c230b9bbf0c6cde9652607e684795627be2b1

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Changed to `passwordErrors`, think that's clear enough.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Done through other commits.

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
wlee221: That's fair -- done in https://github.com/aws-amplify/amplify-ui/pull/2900/commits/5c3906a2728771755681b511968e208cd84cfbf2

packages/ui/src/types/accountSettings/validator.ts
```diff
@@ -0,0 +1,9 @@
+export type InputEventType = 'blur' | 'change';
+
+export type ValidationMode = 'onBlur' | 'onChange' | 'onTouched';
+
+export interface FieldValidator {
+  validationMode: ValidationMode;
+  validate: (field: string) => boolean;
```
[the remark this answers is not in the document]
wlee221: Agree with suggested change, done in https://github.com/aws-amplify/amplify-ui/pull/2900/commits/ff5dc1081cc18e92350ccca7546c16c4b00f4210.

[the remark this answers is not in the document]
wlee221: I do not think we need to accept async validators yet, because these are field validators that run on every change/blur. 

I believe async validator typically happens in `onSubmit`, which we do not support yet. Wwe can add async support as an additive feature when we support `onSubmit` validators.

packages/react/src/components/AccountSettings/ChangePassword/defaultComponents.tsx
wlee221: I believe it's better to have them separate for accessibility purposes. If you look at `ValidationErrors`, each error becomes its own `<role="alert"` />` error message rather than one error string.

packages/ui/src/helpers/accountSettings/validators/password.ts
wlee221: Resolving, changed to `validator`.

packages/ui/src/helpers/accountSettings/validators/password.ts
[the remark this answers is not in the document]
wlee221: Talked offline, decided to proceed as this is an internal type.

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: Leaving this as-is, and open a separate ticket to refactor this later, as discussed offline.

packages/react/src/components/AccountSettings/types.ts
wlee221: Changed to an array in [78e73c2](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/78e73c27325e04b784154a116a8a85f9a99edde3).

packages/ui/src/types/accountSettings/validator.ts
```diff
@@ -0,0 +1,17 @@
+export type InputEventType = 'blur' | 'change';
+
+export type ValidationMode = 'onBlur' | 'onChange' | 'onTouched';
+
+export interface ValidatorSpec {
```
calebpollman, line 5: Nittish: Consider renaming `ValidatorOptions`

packages/ui/src/helpers/accountSettings/validator.ts
```diff
@@ -0,0 +1,168 @@
...
+    return null;
+  }
+
+  const {
+    passwordPolicyCharacters: characterPolicy = [],
+    passwordPolicyMinLength: minLength,
+  } = passwordSettings;
+
+  return {
+    minLength: minLength ? minLength : undefined,
```
calebpollman, line 28: This ternary looks it might be extraneous unless we're worried abt falsy values for `minLength` that aren't `undefined`?

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,165 @@
...
+      });
+    },
+    []
+  );
+
+  const runValidation = React.useCallback(
+    (
+      formValues: FormValues,
+      blurredFields: BlurredFields,
+      eventType: InputEventType
```
calebpollman, line 108: This is me getting in to the weeds a bit but could just replace these params with an object and shared interface for use between `runValidation`, `validateNewPassword` and `validateConfirmPassword`:
```
const runValidation = React.useCallback((params: ValidateParams) => {
  const passwordErrors = validateNewPassword(params);
  const confirmPasswordErrors = validateConfirmPassword(params);

File: packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx [hunk 31]
```diff
@@ -1,53 +1,165 @@
...
+
+  const handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    event.preventDefault();
+
+    const { name } = event.target;
+    // only update state and run validation if this is the first time blurring the field
+    if (!blurredFields.includes(name)) {
+      const newBlurredFields = [...blurredFields, name];
+      runValidation(formValues, newBlurredFields, 'blur');
+      setBlurredFields(newBlurredFields);
```
calebpollman, line 161: The more i've looked at this PR, have started thinking you could probably use a ref for `blurredFields` to avoid the re-renders that triggered by `setBlurredFields`

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -1,53 +1,165 @@
...
 
 const logger = new Logger('ChangePassword');
 
+const getIsDisabled = (
+  formValues: FormValues,
+  validationError: ValidationError
+): boolean => {
+  const { currentPassword, newPassword, confirmPassword } = formValues;
+
+  const hasEmptyField = !currentPassword || !newPassword || !confirmPassword;
```
calebpollman, line 33: This isn't super important or anything but for readability it would be a bit easier if we just early return from this line since we know we don't need to evaluate the lengths of the validation errors below

calebpollman (approved):
Left some more feedback, but looking good overall. Let me know your thoughts

packages/ui/src/types/accountSettings/validator.ts
wlee221: Agreed, done in [7ff2acb](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/7ff2acb46a95c1754c7e93bc9b7c98d3b527479f)

packages/ui/src/helpers/accountSettings/validator.ts
wlee221: +1, simplified in [2a1d469](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/2a1d4697f25d2ac57e036941880f21df26fc4248)

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: 👍, done in [6a7349b](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/6a7349b94bae719dff691bdc27c7fa937888d72a)

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: I agree, `blurredFields` by itself doesn't need a re-render; only blur validation does.

[39a189c](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/39a189cffc3c057e461d197f3da8e5aac6954a0a)

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
wlee221: [b38848d](https://github.com/aws-amplify/amplify-ui/pull/2900/commits/b38848da27d1979f72e70aaf79b8a59eb97ba27a)

calebpollman (approved):

PR #2902, opened 2022-11-05
chore(theme): allow non-nested custom theme values

calebpollman:
#### Description of changes
Updates `color` web custom theme values to be of type token objects or `string`. 

The required update to allow for the desired behavior in `createTheme` is very minimal, but required a rewrite of a large portion of the `Theme` types. 

Additionally, made some updates to React Native theming to keep types happy.

#### Issue #, if available
NA

Unit tests, checked `defaultDarkMode` in the docs
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/ui/.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     node: true,
   },
   extends: ['eslint:recommended'],
-  parser: 'babel-eslint',
+  parser: '@typescript-eslint/parser',
```
calebpollman, line 7: Got over seeing linting errors due to `bable-eslint` not understanding TS

packages/ui/src/theme/tokens/index.ts
```diff
@@ -59,7 +62,7 @@ export interface Tokens {
  * The fully setup theme tokens. It has the same shape as Tokens
  * but each token has added fields.
  */
-export interface WebTokens extends Tokens {
+export interface WebTokens {
```
calebpollman, line 65: Decoupling `WebTokens` from the `Tokens` interface allows for properties of `WebTokens` to only expect `WebDesignToken`, and `Tokens` to utilize `DesignToken`/`string` without getting TS errors on this line due to  `WebTokens` not extending `Tokens` correctly

packages/ui/src/theme/tokens/types/designToken.ts
```diff
@@ -1,20 +1,5 @@
 /**
- * Helper function to test if something is a design token or not.
- * Used in the React component style props.
- *
- * @param arg - thing to test if it is a design token or not
- * @returns boolean
- */
-export function isDesignToken(arg: unknown): arg is WebDesignToken {
```
calebpollman, line 8: Moved this to _tokens/utils.ts_ in an effort to begin reducing the amount of non-TS that is inside of types files

packages/ui/src/theme/createTheme.ts
```diff
@@ -43,16 +43,7 @@ function setupTokens(obj: any, path = []) {
...
-          // basically some part of the theme object that is not a design token, which is
-          // anything that is not an object with a value attribute
-          console.warn(
-            `Non-design token found when creating the theme at path: ${path.join(
-              '.'
-            )}.${name}\nDid you forget to add '{value:"${obj[name]}"}'?`
-          );
-          // Keep the users data there just in case
-          tokens[name] = obj[name];
+          tokens[name] = setupTokens({ value: obj[name] }, path.concat(name));
```
dbanksdesign, line 46: It's so simple.

packages/ui/src/theme/tokens/colors.ts
```diff
@@ -1,125 +1,101 @@
...
+    brand: {
+      primary: ColorScale<DesignTokenType>;
+      secondary: ColorScale<DesignTokenType>;
+    };
+
+    overlay: OverlayColors<DesignTokenType>;
+  } & Record<
+      string,
+      ColorTypes<DesignTokenType> | Record<string, ColorTypes<DesignTokenType>>
+    >;
```
dbanksdesign, line 89: Will this allow users to add colors that are not in the theme? That has been a request

packages/ui/src/theme/createTheme.ts
calebpollman, line 46: Basically this is the change to allow custom values not contained in objects. If `obj[name]` is not an `object` then we assume that it should be assigned to a design token and continue the recursion

dbanksdesign (dismissed):
its beautiful

packages/ui/src/theme/createTheme.ts
dbanksdesign: TBH I love this. I've been thinking about it for 30 mins and it just works.

packages/ui/src/theme/tokens/colors.ts
calebpollman: I _believe_ so, this is `[key: string]: WebColorTypes | Record<string, WebColorTypes>` from [what was line 97](https://github.com/aws-amplify/amplify-ui/pull/2902/files#diff-f3f72291b73e9cbe2298b56eec3eb54efb801d7c2d7086366d905563475fe08dL97). Will look in to it

calebpollman: Added a unit test asserting it: https://github.com/aws-amplify/amplify-ui/pull/2902/commits/95340ffa93bef598c96a01b7e426ec508231fb9d

packages/ui/src/theme/tokens/types/designToken.ts
```diff
@@ -41,7 +24,9 @@ export type WebDesignToken<ValueType = any> = {
    * `var(--amplify-colors-font-primary)`
    */
   toString(): string;
-} & DesignToken<ValueType>;
+
+  value: ValueType;
```
calebpollman, line 28: Include `value: ValueType` directly on `WebDesignToken` and remove the intersection type to prevent extension of `ValueType`

packages/react-native/src/theme/types.ts
```diff
@@ -1,55 +1,56 @@
...
- * A Theme just needs a name.
- * Users can define any tokens they need, but they don't need a complete theme with all tokens.
+ * A Theme just needs a name, all other properties are optional.
  */
 export interface Theme {
   colorMode?: ColorMode;
+  /**
+   * Custom component styles
+   */
+  components?: Components;
```
calebpollman, line 29: @dbanksdesign moved `components` outside of `tokens`

packages/ui/src/theme/__tests__/createTheme.test.ts
```diff
@@ -33,36 +34,52 @@ describe('@aws-amplify/ui', () => {
...
         expect(tokens.colors.background.primary.value).toEqual('#bada55');
+
+        expect(tokens.shadows.large.value).toEqual(
+          'offsetX offsetY blurRadius spreadRadius color'
+        );
+        expect(tokens.shadows.small.value).toEqual(shadows.small.value);
+
+        expect(
+          // TODO: link gthub issue
+          (tokens.components.alert.alignItems as WebDesignToken).value
```
calebpollman, line 73: For properties of the `components` key  of `BaseTheme` to accept both syntaxes, `DesignToken` was extended [here](https://github.com/aws-amplify/amplify-ui/pull/2902/files#diff-061a7577867c5583dc1ceeafba7707e260d62c775c6f950bc893f9f463f3182cR4). 

We need something along the lines of a `WebComponents` type (similar to our `WebTokens` and other `Web` prefixed types) that will be used as the output for `components` from `createTheme`. Right now there is only the `components` key on `Tokens` being shared between both `Tokens` and `WebTokens`, and it ultimately uses `DesignToken` for the style properties to allow both custom style syntaxes, but the output of `createTheme` creates the `components` style properties only as `WebDesignToken` but without the correct typing. 

Would prefer to address this issue in a follow up as creating a `WebComponents` type will touch every component type file at the very least, this issue is internal, and is not visible any where in the codebase except for this file.

packages/ui/src/theme/types.ts
```diff
@@ -117,7 +117,6 @@ export interface Theme {
 export interface BaseTheme extends Theme {
   tokens: Tokens;
   breakpoints: Breakpoints;
-  overrides?: Array<Override>;
```
calebpollman, line 120: This was unnecessary since `overrides` is optional on `Theme`

packages/ui/src/theme/__tests__/createTheme.test.ts
calebpollman: Actually think the issue described above needs to be addressed as a part of this PR since there is a potential edge case where a developer could be referencing a `component` style property in their own code resulting in a type error, example:
```
<View style={{ alignItems: theme.tokens.components.alert.alignItems.value }} />
```

packages/ui/src/theme/tokens/components/alert.ts
```diff
@@ -1,42 +1,38 @@
...
-}
+type AlertVariationTokens<OutputType> = DesignTokenProperties<
+  'backgroundColor' | 'color',
+  OutputType
+>;
 
-interface AlertIconTokens {
-  size: DesignToken<FontSizeValue>;
-}
+type AlertIconTokens<OutputType> = DesignTokenProperties<'size', OutputType>;
```
calebpollman, line 8: `DesignTokenProperties` ideally speeds up development and gives us type safety:

https://user-images.githubusercontent.com/15002885/200149487-e0951548-7eba-464b-8c32-9ef32775faa4.mov

packages/ui/src/theme/tokens/components/badge.ts
```diff
@@ -1,43 +1,38 @@
...
+export type BadgeTokens<OutputType = unknown> = DesignTokenProperties<
+  BadgeKey,
+  OutputType
+> & {
+  error?: BadgeVariationTokens<OutputType>;
+  large?: BadgeSizeTokens<OutputType>;
+  small?: BadgeSizeTokens<OutputType>;
+  info?: BadgeVariationTokens<OutputType>;
+  success?: BadgeVariationTokens<OutputType>;
+  warning?: BadgeVariationTokens<OutputType>;
```
calebpollman, line 34: Setting the properties in the base `BadgeTokens` will allow us to remove the usage of `PartialDeep` on `Tokens`, not sure if that's super important or anything but thought it would be nice to have

packages/ui/src/theme/tokens/types/designToken.ts
```diff
@@ -101,3 +86,93 @@ export type TransitionTimingFunctionValue = string;
...
+export type RequiredDesignTokenProperties<Keys extends PropertyKey> = {
+  [Key in Keys]: WebDesignToken<Properties[Key]>;
+};
+
+export type DesignTokenProperties<
+  Keys extends PropertyKey,
+  OutputType = unknown
+> = OutputType extends 'required'
+  ? RequiredDesignTokenProperties<Keys>
+  : OptionalDesignTokenProperties<Keys>;
```
calebpollman, line 178: Added this type utility to  `DesignTokenProperties` ideally to speed up development, cut down on imports and duplicate types,  and give us type safety for style property look up:

https://user-images.githubusercontent.com/15002885/200149487-e0951548-7eba-464b-8c32-9ef32775faa4.mov

packages/ui/src/theme/__tests__/createTheme.test.ts
calebpollman: [Pushed a commit](https://github.com/aws-amplify/amplify-ui/pull/2902/commits/a0c328f13e8f5de14ee4ade37dd887a08fcd4da6) with `Alert`, `Authenticator`, `Badge` and `Card`  refactored

packages/ui/src/theme/tokens/index.ts
```diff
@@ -73,7 +77,7 @@ export interface WebTokens extends Tokens {
   shadows: WebShadows;
   space: WebSpace;
   time: WebTime;
-  transform: WebTransforms;
+  transforms: WebTransforms;
```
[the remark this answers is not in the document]
dbanksdesign: That is a typo and I would say this is a bug fix. Also this is _only_ on the interface of the return type (what `createTheme` returns and not what a customer inputs). So if a customer was using the transforms part of the tokens they would still need to call `tokens.transforms` and not `token.transform` and ignore the TS error. I don't think this needs a major version bump

packages/ui/src/theme/tokens/types/designToken.ts
dbanksdesign: oohhh fancy

calebpollman:
Marking as draft until the `components` types can be completed

packages/ui/src/theme/tokens/fontWeights.ts
```diff
@@ -4,25 +4,40 @@ import {
...
+  FontWeight,
+  FontWeightValue,
+  OutputType
+>;
+// export type FontWeights<OutputType = unknown> = OutputType extends 'required'
+//   ? Record<FontWeight, WebDesignToken<FontWeightValue>>
+//   : Partial<Record<FontWeight, DesignToken<FontWeightValue>>>;
+
+// export type FontWeights<DesignTokenType = DesignToken<FontWeightValue>> =
+//   Record<FontWeight, DesignTokenType>;
```
ioanabrooks, line 36: Please remove commented out code.

packages/ui/src/theme/utils.ts
```diff
@@ -14,27 +34,19 @@ function referenceValue(value: string) {
...
-        referenceValue(offsetY),
-        referenceValue(blurRadius),
-        referenceValue(spreadRadius),
-        referenceValue(color),
-      ].join(' ');
+      return SHADOW_PROPERTIES.map((property) =>
+        // lookup property against `token` first for custom non-nested value, then lookup
+        // property against `value` for design token value, default to empty string
+        referenceValue(token[property] ?? value[property] ?? '')
+      ).join(' ');
```
ioanabrooks, line 49: Nice cleanup!

packages/ui/src/theme/utils.ts
```diff
@@ -48,3 +60,14 @@ interface NameTransformProps {
...
+
+/**
+ * Helper function to test if something is a design token or not.
+ * Used in the React component style props.
+ *
+ * @param value - thing to test if it is a design token or not
+ * @returns boolean
+ */
+export function isDesignToken(value: unknown): value is WebDesignToken {
+  return isObject(value) && has(value, 'value');
```
ioanabrooks, line 72: Is there any advantage to using lodash util vs just `Object.hasOwn(value, 'value')`. Also might be worth considering renaming the value prop. The check for whether a value has property value is less intuitive. Maybe `object` or `token`?

packages/ui/src/theme/createTheme.ts
```diff
@@ -16,51 +17,39 @@ import { DesignToken, WebDesignToken } from './tokens/types/designToken';
...
-            )}.${name}\nDid you forget to add '{value:"${obj[name]}"}'?`
-          );
-          // Keep the users data there just in case
-          tokens[name] = obj[name];
-        } else {
-          tokens[name] = setupTokens(obj[name], path.concat(name));
-        }
-      }
+function setupTokens(tokens: any, path = []) {
+  if (has(tokens, 'value')) {
```
ioanabrooks, line 37: Same question as below about lodash has vs just [Object.hasOwn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn)

packages/ui/src/theme/createTheme.ts
ioanabrooks: The TS is strong with this one.

packages/ui/src/theme/tokens/fontWeights.ts
calebpollman: It's a draft still! 😄

packages/ui/src/theme/utils.ts
calebpollman: Is `Object.hasOwn` available for use with `ESNext`?

ioanabrooks: I believe so, it has been added to the Object standard API a while back (2017?) and is supported by all major browsers: https://caniuse.com/?search=Object.hasOwn

packages/ui/src/theme/tokens/fontWeights.ts
ioanabrooks: Raising the bar on draft PRs. :)

PR #2913, opened 2022-11-08
feat(ui-react-native): Add theming for Radio primitive

joebuono:
#### Description of changes

In addition to adding theming, also refactored the logic for proportionally sizing the Radio. 

RadioTheming3

https://user-images.githubusercontent.com/48109584/200642886-a654ffc8-0b9f-4879-87b7-5592e2061337.mov

Ran in iOS and Android, and updated unit tests. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


ioanabrooks (commented):
Great work! Overall lgtm but I think styling needs a little more love, especially for the different sizes. We should have different border widths for each radio size, and the border seems to be getting further and further away from the dot as the size of the radio increases (look and feel should be consistent and proportional to radio size)

packages/react-native/src/primitives/Radio/getRadioDimensions.ts
```diff
@@ -16,23 +21,32 @@ import { RadioDimensions, RadioStyles, Size } from './types';
...
-  radioDotSize: RadioDimensions;
+  radioContainerDimensions: RadioContainerDimensions;
+  radioDotDimensions: RadioDotDimensions;
 } => {
   if (typeof size === 'number') {
     return {
-      radioContainerSize: { height: size, width: size },
-      radioDotSize: { height: size * 0.5, width: size * 0.5 },
+      radioContainerDimensions: {
+        borderWidth: Math.floor(size / 10) - 1,
```
calebpollman, line 36: Called this out in the thread, but i think the `borderWidth` should be stable across our `Size` variants. It is very easy for users to make changes if they choose to

calebpollman (changes requested):

packages/react-native/src/primitives/Radio/getRadioDimensions.ts
calebpollman: See https://github.com/aws-amplify/amplify-ui/pull/2913#discussion_r1017099842

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -1,50 +1,64 @@
...
+    },
+    disabled: {
+      opacity: opacities[60],
+      ...components?.radio.disabled,
+    },
+    radioContainer: {
+      alignItems: 'center',
+      borderColor: colors.teal[80],
+      borderRadius: ROUNDED_BORDER_RADIUS,
+      borderWidth: 2,
```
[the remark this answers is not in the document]
ioanabrooks: Any particular reason why it should be the same, 2 looks too thick to me for the small variant.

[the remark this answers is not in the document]
calebpollman: A `borderWidth` of `1` looks very thin for the `small` variant:



If we bump the `borderWidth` on the `large`  to `3`, think that might looks a bit better due to the decrease in white space around the dot, but looks too thick when the `Radio` is unselected and alos when compared to the default `fontWeight` of the `Label`:



Preference is to keep the `borderWidth` at `2`, and slightly increase the dot dimensions on the `large` variant:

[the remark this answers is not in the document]
joebuono: @dbanksdesign Thoughts?

packages/react-native/src/primitives/Radio/styles.ts
```diff
@@ -1,50 +1,65 @@
...
+    radioContainer: {
+      alignItems: 'center',
+      borderColor: colors.border.primary,
+      borderRadius: ROUNDED_BORDER_RADIUS,
+      borderWidth: space.xxs,
+      justifyContent: 'center',
+      ...components?.radio.radioContainer,
+    },
+    radioDot: {
+      backgroundColor: colors.teal[80],
```
ioanabrooks, line 30: This should reference `colors.brand.primary[80]`

packages/react-native/src/primitives/Radio/getRadioDimensions.ts
```diff
@@ -16,23 +16,31 @@ import { RadioDimensions, RadioStyles, Size } from './types';
  * (e.g., 'small' => { radioContainerSmall, radioDotSmall } )
  */
 
+export const RADIO_DOT_PROPORTION = 0.6;
```
ioanabrooks, line 19: Nice! Conjugate golden ratio.  Would be nice to add a comment on why we landed on this value or purpose to help future readers.

ioanabrooks: The Silver Ratio! :)

ioanabrooks (approved):

calebpollman (approved):

PR #2915, opened 2022-11-08
feat(ui-react-native): Add theming to RadioGroup primitive

joebuono:
#### Description of changes

https://user-images.githubusercontent.com/48109584/200676255-b57ba4ef-6ff2-4753-b4c0-c17d59c3b96e.mov

Ran in iOS and Android

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


packages/react-native/src/primitives/RadioGroup/styles.ts
```diff
@@ -1,12 +1,15 @@
...
-  },
-});
+export const getThemedStyles = (theme: StrictTheme): RadioGroupStyles => {
+  const { components } = theme.tokens;
+
+  return StyleSheet.create({
+    label: {
+      // The RadioGroup label inherits its styles from the Label primitive
+      ...components?.radiogroup.label,
+    },
```
calebpollman, line 13: Do we need this? Just thinking it might be confusing for users since `theme` style applied to the `Label` itself would override this? Example:
```
const customTheme = {
  // syntax might be off
  label: { color: 'orange' }, 
  radioGroup: { label: { color: 'red' } },
}
```

That said, think we may still want to provide users the ability to add custom styles for the `container` even if we don't apply any by default

joebuono: Right. Do we want to allow customers to specifically theme the `RadioGroup` label, or should they only be able to theme the `Label` itself?

calebpollman: Actually i was off abt my concern. Since these styles are passed via the `style` prop to the `Label` they will take precedence

packages/react-native/src/primitives/RadioGroup/styles.ts
```diff
@@ -1,12 +1,15 @@
...
-    fontSize: 18,
-    fontWeight: '500',
-    marginHorizontal: 4,
-    marginVertical: 2,
-  },
-});
+export const getThemedStyles = (theme: StrictTheme): RadioGroupStyles => {
+  const { components } = theme.tokens;
+
+  return StyleSheet.create({
```
calebpollman, line 9: We could add a `container` property to `RadioGroupStyle` for custom theme `container` styles, what do you think?

joebuono: I like that idea

packages/react-native/src/primitives/RadioGroup/RadioGroup.tsx
```diff
@@ -48,9 +52,10 @@ export default function RadioGroup<T>({
 
   const containerStyle: ViewStyle = useMemo(
     () => ({
+      ...themedStyle.container,
       flexDirection: getFlexDirectionFromLabelPosition(labelPosition),
     }),
-    [labelPosition]
+    [labelPosition, themedStyle]
```
calebpollman, line 58: `themedStyle` is going to cause this to recalculate on every render since it does not maintain a stable reference

ioanabrooks (approved):

calebpollman (approved):

PR #2916, opened 2022-11-08
feat(ui-react-native): Add theming for Icon primitive

ioanabrooks:
#### Description of changes

This change adds theming to the icon primitive and updates unit tests and story.

```
src/primitives/Icon                                |     100 |      100 |     100 |     100 |                   
  Icon.tsx                                         |     100 |      100 |     100 |     100 |                   
  index.ts                                         |       0 |        0 |       0 |       0 |                   
  styles.ts                                        |     100 |      100 |     100 |     100 |                   
  types.ts                                         |     100 |      100 |     100 |     100 |                  
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react-native/src/primitives/Icon/Icon.tsx
```diff
@@ -12,7 +13,11 @@ export default function Icon({
   style,
   ...rest
 }: IconProps): JSX.Element {
-  const { icon } = useMemo(() => getStyles(color, size), [color, size]);
+  const theme = useTheme();
+  const { icon } = useMemo(
```
joebuono, line 17: Just for consistency, can we rename `icon` to `themedStyle`?

packages/react-native/src/primitives/Icon/styles.ts
```diff
@@ -1,16 +1,22 @@
...
+export const getThemedStyles = (
+  theme: StrictTheme,
+  color?: IconProps['color'],
+  size?: IconProps['size']
+): IconStyles => {
+  const { components } = theme.tokens;
+
+  return StyleSheet.create({
     icon: {
       height: size,
```
joebuono, line 15: What happens if the `size` is `undefined`?

joebuono (commented):
LGTM, just a few nits

examples/react-native/storybook/stories/Icon.stories.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { useEffect } from 'react';
 import { storiesOf } from '@storybook/react-native';
+import { Animated, Easing, StyleSheet } from 'react-native';
```
calebpollman, line 3: ❤️

packages/react-native/src/primitives/Icon/styles.ts
calebpollman: That's a good call out. Should we add the same default size we use for the `IconButton` in the `Icon` and set `size` as required on line 9? Might make sense to do something for `color` as well. What do you think @ioanabrooks

ioanabrooks: Yup! Added tokens for icon sizes, the default value is medium coming through the default theme in case size is undefined. For color we should use one of the tokens, otherwise RN will use black if it's undefined, not great for dark mode. Not sure how we are supporting dark mode or which token I should use atm.

packages/react-native/src/primitives/Icon/styles.ts
```diff
@@ -1,16 +1,22 @@
...
+  theme: StrictTheme,
+  color?: IconProps['color'],
+  size?: IconProps['size']
+): IconStyles => {
+  const { iconSizes, components } = theme.tokens;
+
+  return StyleSheet.create({
     icon: {
-      height: size,
+      height: size ?? iconSizes.medium,
```
joebuono, line 15: 👍

joebuono (approved):

packages/react-native/src/primitives/IconButton/IconButton.tsx
```diff
@@ -14,7 +14,7 @@ export default function IconButton({
   accessibilityRole = 'button',
   color,
   iconStyle,
-  size = 20,
+  size,
```
calebpollman, line 17: Can we set `medium` as the default value?

packages/react-native/src/primitives/Icon/types.ts
```diff
@@ -1,5 +1,16 @@
 import { ColorValue, ImageProps, ImageStyle } from 'react-native';
 
+export type Size = 'xs' | 'small' | 'medium' | 'large' | 'xl';
+export type IconSizes = Record<Size, number>;
+
+export const iconSizes: IconSizes = {
+  xs: 16,
+  small: 20,
+  medium: 24,
```
calebpollman, line 9: Just want to confirm that `24` should be `medium`? Might be a bit large

packages/react-native/src/primitives/Icon/styles.ts
calebpollman: What happens if `size` is a `string`?

joebuono (dismissed):

packages/react-native/src/primitives/Icon/types.ts
```diff
@@ -1,5 +1,16 @@
 import { ColorValue, ImageProps, ImageStyle } from 'react-native';
 
+export type Size = 'xs' | 'small' | 'medium' | 'large' | 'xl';
+export type IconSizes = Record<Size, number>;
+
+export const iconSizes: IconSizes = {
```
calebpollman, line 6: Can we put this in a _constants.ts_ file rather than including it with the types?

ioanabrooks: sure

calebpollman (approved):

PR #2921, opened 2022-11-09
chore(ui-react-native): Add utility hook for pressable styles

ioanabrooks:
#### Description of changes

This change adds a utility hook for pressable components styling/theming.
Usage:

```
const pressableStyle = usePressableContainerStyles({
     overrideStyle: style,
     containerStyle: { ...baseStyle.container, ...themedStyle.container },
     pressedStyle: { ...themedStyle.pressed }
   });
```

Updated `Button`, `IconButton`, `FederatedProviderButton`, `Radio`, `Tab` components to use the hook.

Updated unit tests.

Ran demo app and storybook

| Button pressed style  | Icon pressed style |
| Simulator Screen Shot - iPhone 13 - 2022-12-06 at 10 12 13 | Simulator Screen Shot - iPhone 13 - 2022-12-06 at 10 13 02 |

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react-native/src/theme/hooks/usePressableStyles.ts
```diff
@@ -0,0 +1,40 @@
...
+} from 'react-native';
+
+type PressableStyle = ViewStyle | TextStyle;
+
+export type PressableStyleProps = {
+  disabled: PressableProps['disabled'];
+  disabledStyle?: PressableStyle;
+  style: PressableProps['style'];
+  themedStyle: PressableStyle;
+};
```
calebpollman, line 17: Think we can simplify this to just:
```ts
interface UsePressableStyleParams extends PressableProps<'style'> {}
```

Whether or not something is `disabled` should have no impact on `pressed` styles, which incidentally is also where we are using `themedStyle`

packages/react-native/src/Authenticator/Defaults/ConfirmResetPassword/__tests__/__snapshots__/ConfirmResetPassword.spec.tsx.snap
```diff
@@ -461,17 +459,13 @@ Array [
...
           "justifyContent": "center",
           "paddingHorizontal": 16,
           "paddingVertical": 12,
         },
-        Object {
-          "backgroundColor": "hsl(190, 95%, 30%)",
-        },
-        null,
-        null,
+        undefined,
```
ioanabrooks, line 468: Nothing got removed, container and disabled styles are just merged in one single object.

packages/react-native/src/Authenticator/common/FederatedProviderButton/FederatedProviderButton.tsx
```diff
@@ -13,18 +13,14 @@ export default function FederatedProviderButton({
...
-    [style]
-  );
+  const pressableStyle = usePressableOverrideStyle({
+    style,
+    themedStyle: styles.container,
+  });
 
   return (
-    <Button {...rest} style={containerStyle}>
-      {source ? <Icon source={source} style={styles.icon} size={20} /> : null}
```
ioanabrooks, line 27: `source` is not optional so removed ternary op., it's lowering our unit test threshold.

packages/react-native/src/Authenticator/common/FederatedProviderButton/__tests__/FederatedProviderButton.spec.tsx
```diff
@@ -1,20 +1,27 @@
...
 
 import { icons } from '../../../../assets';
 import FederatedProviderButton from '../FederatedProviderButton';
+import { useTheme } from '../../../../theme';
+import { getThemedStyles as getButtonThemedStyles } from '../../../../primitives/Button/styles';
+import { styles } from '../styles';
 
 const signInText = 'Sign In with Amazon';
 
-/* TODO: Add more tests */
```
ioanabrooks, line 9: Added more tests! 100% coverage

packages/react-native/src/primitives/Tabs/__tests__/Tab.spec.tsx
```diff
@@ -35,14 +36,21 @@ describe('Tab', () => {
 
     const { result } = renderHook(() => useTheme());
     const themedStyle = getThemedStyles(result.current);
+    const buttonThemedStyle = getButtonThemedStyles(result.current);
 
     const tab = getByRole('tab');
-    // select second index as `Button` applies its own style first
-    expect((tab.props.style as ViewStyle[])[4]).toStrictEqual([
```
ioanabrooks, line 41: Updated this test, this positional style testing is very brittle.

packages/react-native/src/Authenticator/common/FederatedProviderButton/FederatedProviderButton.tsx
calebpollman: IMO `source` should be optional in case we were to allow custom social providers in the future. It just provides more flexibility

packages/react-native/src/theme/hooks/usePressableOverrideStyle.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface UsePressableOverrideStyleProps
+  extends Pick<PressableProps, 'style'> {
+  themedStyle?: StyleProp<ViewStyle>;
+  pressedStyle?: StyleProp<ViewStyle>;
+}
+
+export const usePressableOverrideStyle = ({
+  style,
+  themedStyle,
```
calebpollman, line 17: Since `themedStyle` is passed through the hook and returned without having an impact on the functionality of the hook, it doesn't feel like we need it here. Ultimately this hook reads to me as a generic utility for handling a `Pressable['style']` passed as a prop to a component and applying that component's `pressed` state style. 

Additionally my thinking is that removing `themedStyle` will also make this hook easier to understand. In this commit we are passing theme related styles for `pressedStyle` while also having `themedStyle` which may be confusing when referencing its usage by future authors. For example, this is from the usage in the `Tabs` component:
```tsx
  const pressableStyle = usePressableOverrideStyle({
    style,
    themedStyle: containerStyle,
    pressedStyle: themedStyle.pressed,
  });
 ```

calebpollman (commented):
Thank you for adding this and doing all the clean up! Left some thoughts on the params, also can we move the _usePressableOverrideStyle_ directory inside the _hooks_ directory located inside _src_ (as a sibling to the _useDeviceOrientation_ directory)?

calebpollman (approved):
🚢  (just don't forget to squash 😆)

joebuono (approved):
Thank you for doing this!

PR #2925, opened 2022-11-09
feat: add resumable as a prop

dbanksdesign:
#### Description of changes

Added the ability for uploads to be resumable or not. In doing so had to write a hack to fix the Storage lib's code. 

Need some help on my Typescript. 

https://user-images.githubusercontent.com/321279/200904115-714aaf32-631a-4cf8-ad37-e909a9a960d2.mp4

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Storage/FileUploader/FileUploader.tsx
+++ b/packages/react/src/components/Storage/FileUploader/FileUploader.tsx
@@ -9,6 +9,9 @@ import { Previewer } from './Previewer';
 import { UploadDropZone } from './UploadDropZone';
 import { Tracker } from './Tracker';
 
+const isUploadTask = (value: unknown): value is UploadTask =>
+  typeof (value as UploadTask)?.resume === 'function';
+
 export function FileUploader({
   acceptedFileTypes,
   components = {},
@@ -128,9 +131,11 @@ export function FileUploader({
   const onPause = useCallback(
     (index: number): (() => void) => {
       return function () {
-        fileStatuses[index].uploadTask.pause();
-        const newFileStatuses = [...fileStatuses];
         const status = fileStatuses[index];
+        if (isUploadTask(status.uploadTask)) {
+          status.uploadTask.pause();
+        }
+        const newFileStatuses = [...fileStatuses];
 
         newFileStatuses[index] = { ...status, fileState: 'paused' };
         setFileStatuses(newFileStatuses);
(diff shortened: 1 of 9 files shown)


ErikCH (approved):
Looks good, other then what we chatted about offline about the file uploader pause bug. Which we can do in this PR or next.

dbanksdesign (commented):
Need some TS help on a function returning different types based on if a prop is true or false.

packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
```diff
@@ -19,21 +19,37 @@ export function uploadFile({
...
+  if (resumable === true) {
+    return Storage.put(fileName, file, {
+      level,
+      resumable,
+      progressCallback,
+      errorCallback,
+      // hack to remove the bucket name at the beginning
+      completeCallback: (event) => {
+        completeCallback({ key: event.key.split('/').slice(1).join('/') });
+      },
```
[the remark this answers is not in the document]
dbanksdesign: Added a comment to remove this hack once v5 is released and we pull it in. The JS team fixed this behavior: https://github.com/aws-amplify/amplify-js/pull/10651

packages/react/src/components/Storage/FileUploader/Tracker/FileState.tsx
```diff
@@ -12,12 +12,15 @@ export const FileState = ({
   percentage,
 }: FileStateProps): JSX.Element => {
   switch (fileState) {
-    case 'loading':
+    case 'loading': {
+      let text = translate('Uploading');
+      if (percentage > 0) text = text + `: ${percentage}%`;
```
0618, line 17: nit: 

```suggestion
      if (percentage > 0) text += `: ${percentage}%`;
```

packages/react/src/components/Storage/FileUploader/FileUploader.tsx
```diff
@@ -9,6 +9,9 @@ import { Previewer } from './Previewer';
 import { UploadDropZone } from './UploadDropZone';
 import { Tracker } from './Tracker';
 
+const isUploadTask = (value: unknown): value is UploadTask =>
+  typeof (value as UploadTask)?.resume === 'function';
```
0618, line 13: If we have to do `value as UploadTask`, why not define it at the beginning? 

```suggestion
const isUploadTask = (value: UploadTask): value is UploadTask =>
  typeof (value)?.resume === 'function';
```

0618 (approved):
LGTM. Left non-blocker comments

wlee221 (approved):
LGTM

packages/react/src/components/Storage/FileUploader/FileUploader.tsx
wlee221: I believe it's a standard practice to have `value` defined loosely in typeguards: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards.

This typeguard, for example, asserts whether this loosely typed `value` is actually of type `UploadTask`, and narrows the type to it if so.

PR #2927, opened 2022-11-09
feat(account-settings): add DeleteUser component

wlee221:
#### Description of changes

This adds `<DeleteUser />` component. 
- functionality to delete user
- prop support for `handleDelete`, `onSuccess`, and `onError`
- Confirmation / Warning component to confirm deletion

Component overrides, or sub-component specific customization are not implemented yet.

New unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


examples/next/pages/ui/components/account-settings/delete-user/aws-exports.js
```diff
@@ -0,0 +1,2 @@
+import awsExports from '@environments/auth/auth-with-email/src/aws-exports';
```
wlee221, line 1: Added example just for internal development purposes. E2e specs will follow with more complete example.

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,98 @@
...
+      }
+      setState('DONE');
+      onSuccess?.();
+    } catch (e) {
+      const error = e as Error;
+      setState('IDLE');
+      setErrorMessage(error.message);
+      onError?.(error);
+    }
+  }, [handleDelete, onError, onSuccess, user]);
```
wlee221, line 51: I pulled this to a separate async function, because the event handler for `onConfirmDelete` expects a synchronous function.

Advices to clean this up definitely appreciated.

packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return <Button {...rest}>{children}</Button>;
+};
+
+export const DefaultWarning = ({
+  onCancel,
+  onConfirmDelete,
+  isInProgress,
+}: DeleteUserWarningProps): JSX.Element => {
+  // translations
+  // TODO: consolodiate translations to accountSettingsTextUtil
```
joebuono, line 21: Will this be done in a follow-up PR?

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,98 @@
...
+    setState('IS_CONFIRMING');
+  };
+
+  const runDeleteUser = React.useCallback(async () => {
+    setState('IN_PROGRESS');
+    try {
+      if (handleDelete) {
+        await handleDelete(user);
+      } else {
+        await deleteUser();
```
joebuono, line 41: Just checking that we don't need to pass `user` to `deleteUser`?

packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx
wlee221: Yes — it will be a launch requirement item as well. Ill create a ticket to track this too.

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
wlee221: No, because `Auth.deleteUser` actually does not require underlying `user` object.

wlee221: On the other hand, if customer provided their own `handleDelete` function to override the default service, we will pass `user` object for their convenience so that they can do whatever clean up with the user.

wlee221: I do not think the code is clear here, i can add a comment (or let me know if you have any refactor suggestions)

packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx
```diff
@@ -0,0 +1,55 @@
...
+    'Deleting your account is not reversable. You will lose access to your account and all data associated with it.'
+  );
+  const cancelText = translate('Cancel');
+  const deleteMyAccountText = translate('Delete my account');
+
+  return (
+    <Card>
+      <Flex direction="column">
+        <Text color="font.error">{warningText}</Text>
+        <Flex direction="row">
```
joebuono, line 32: nit: `direction` is 'row' by default

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,98 @@
...
+      }
+      setState('DONE');
+      onSuccess?.();
+    } catch (e) {
+      const error = e as Error;
```
joebuono, line 46: Is this cast necessary to access `error.message`?

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  /** Return null if user isn't authenticated in the first place */
+  if (!user) {
+    logger.warn('<DeleteUser /> requires user to be authenticated.');
+    return null;
+  }
+
+  // Return null if delete user was successful
+  if (state === 'DONE') {
+    return null;
```
joebuono, line 75: Wondering if there is some UX that confirms that the user was successfully deleted?

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
wlee221: Right, and also because `onError(error)` expects `error` to be of `Error` type.

Having to type `error` from catch clause is a limitation with TS: https://github.com/microsoft/TypeScript/issues/20024

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
wlee221: cc @dbanksdesign. Default expectation is that users are expected to be signed out and navigated out immediately, and thus there will no confirmation messages. 

Developers can implement their own confirmation messages, if they wish, with an `alert` or `toast`. That will be up to customers to implement based on their use case.

packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx
wlee221: Good point!
```suggestion
        <Flex>
```

packages/react/src/components/AccountSettings/DeleteUser/types.ts
```diff
@@ -0,0 +1,23 @@
+import { AmplifyUser } from '@aws-amplify/ui';
+
+export interface DeleteUserWarningProps {
+  /** called when end user cancels account deletion */
+  onCancel: () => void;
+  /** called when user acknowledges account deletion */
+  onConfirmDelete: () => void;
```
calebpollman, line 7: Maybe `onConfirm` since its the actual action? And can it be optional?

packages/react/__tests__/__snapshots__/exports.ts.snap
```diff
@@ -16,6 +16,7 @@ Array [
   "ComponentClassObject",
   "ComponentPropsToStylePropsMap",
   "ComponentPropsToStylePropsMapKeys",
+  "DeleteUser",
```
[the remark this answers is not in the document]
wlee221: That's a nice idea actually, especially as we add more connected components.

Import from separate entry point could work too:
```ts
import { DeleteUser } from '@aws-amplify/ui-react/account-settings
```

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
wlee221: Comments added in https://github.com/aws-amplify/amplify-ui/pull/2927/commits/9df0d8c22af6f0570d3e554d7a49b872b482cf29, resolving.

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { useAuth } from '../../../internal';
+import { Flex } from '../../../primitives';
+import {
+  DefaultWarning,
+  DefaultError,
+  DefaultSubmitButton,
+} from './defaultComponents';
+import { DeleteUserProps, DeleteUserState } from './types';
+
+const logger = new Logger('DeleteUser');
```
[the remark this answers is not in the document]
wlee221: You're totally right 😅 

Using that util in [Use logger util](https://github.com/aws-amplify/amplify-ui/pull/2927/commits/204088ba0345ff9f8d313abf9cf69ee89ea12cf7)

packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx
```diff
@@ -0,0 +1,55 @@
...
+  children,
+  ...rest
+}) => {
+  return <Button {...rest}>{children}</Button>;
+};
+
+export const DefaultWarning = ({
+  onCancel,
+  onConfirmDelete,
+  isInProgress,
```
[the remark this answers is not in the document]
wlee221: Right, changed in https://github.com/aws-amplify/amplify-ui/pull/2927/commits/5d29c5fb030f4bde6246c227854c88a0616fd6c8

packages/react/src/components/AccountSettings/DeleteUser/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  /** whether account deletion is in progress */
+  isInProgress?: boolean;
+}
+
+export type DeleteUserState = 'IDLE' | 'IS_CONFIRMING' | 'IN_PROGRESS' | 'DONE';
```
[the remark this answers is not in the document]
wlee221: I like the readability of this, done in https://github.com/aws-amplify/amplify-ui/pull/2927/commits/12e973473687a42c708d8aa90d010d53fe4da500. 

I changed `CONFIRMATION` -> `CONFIRMING` just to align wording.

packages/react/src/components/AccountSettings/DeleteUser/types.ts
wlee221: +1, https://github.com/aws-amplify/amplify-ui/pull/2927/commits/627877cd862e9c10021e99f72c46d6777bbea02f

wlee221: I don't think it should be optional imo, this is essential to using `Warning` component. This will also block user if customer overrides this incorrectly.

packages/react/__tests__/__snapshots__/exports.ts.snap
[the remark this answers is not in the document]
calebpollman: My thinking is that we have already set a precedent for subcomponents with the `Authenticator` and `InAppMessageDisplay` components

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }, [handleDelete, onError, onSuccess, user]);
+
+  // called when end user cancels account deletion confirmation
+  const handleCancel = React.useCallback(() => {
+    setState('IDLE');
+  }, []);
+
+  const handleConfirmDelete = React.useCallback(() => {
+    runDeleteUser();
+  }, [runDeleteUser]);
```
calebpollman, line 64: Nit: **pretty** sure this could be simplified to 
```suggestion
const handleConfirmDelete = React.useCallback(runDeleteUser, [runDeleteUser]);

File: packages/react/src/components/AccountSettings/DeleteUser/types.ts (hunk 12 above)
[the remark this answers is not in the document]
Comment (calebpollman) [reply packages/react/src/components/AccountSettings/DeleteUser/types.ts]: Maybe i misread but are they "CONFIRMING" or is the component waiting for them to confirm in this state?

File: packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx [hunk 14]
```diff
@@ -0,0 +1,102 @@
...
+  // Return null if Auth.getCurrentAuthenticatedUser is still in progress
+  if (isLoading) {
+    return null;
+  }
+
+  /** Return null if user isn't authenticated in the first place */
+  if (!user) {
+    logger.warn('<DeleteUser /> requires user to be authenticated.');
+    return null;
+  }
```
calebpollman, line 75: Kind of just a thought (really more of a code style thing) and totally not required, but think it can be useful to order early return conditionals by whether a conditional excludes the next from running. Like, if `isLoading` is `true` it doesn't matter if `user` is falsy, same if state is `Done`, so the order of these conditionals would read like:
```
if (!user) {} // place before `DONE` check since it logs an error
if (state === 'DONE') {}
if (isLoading) {}
return ()

File: packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx [hunk 15]
```diff
@@ -0,0 +1,55 @@
+import { translate } from '@aws-amplify/ui';
+import React from 'react';
+
+import { Card, Flex, Text, Button, Alert } from '../../../primitives';
+import { AccountSettingsError, AccountSettingsSubmitButton } from '../types';
+import { DeleteUserWarningProps } from './types';
+
+export const DefaultSubmitButton: AccountSettingsSubmitButton = ({
```
calebpollman, line 8: Just curious, is this component overridable?

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,102 @@
...
+        await handleDelete(user);
+      } else {
+        // else, run default deleteUser function.
+        await deleteUser();
+      }
+      setState('DONE');
+      onSuccess?.();
+    } catch (e) {
+      const error = e as Error;
+      setState('IDLE');
```
calebpollman, line 51: Not certain it's super important, but have we considered whether we want to track `ERROR` as the state here in case we want to have differing component behavior post error in the future?

packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx
wlee221: It will be -- but on a later PR.

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
wlee221: I have a bit, IMO it doesn't need to be another state because there are no functional difference between `IDLE` and `ERROR` other than UI layer. I don't foresee needing to do so, but open to thoughts.

wlee221: Update, will make this change for clarity of state/code.

wlee221: [78dd9d0](https://github.com/aws-amplify/amplify-ui/pull/2927/commits/78dd9d003fe5a550deac98a633c5b69494ff21eb)

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
wlee221: Done in [c03a451](https://github.com/aws-amplify/amplify-ui/pull/2927/commits/c03a45186d872036e4955a86ebf9f43f5b21e397)!

joebuono (approved):

packages/react/__tests__/__snapshots__/exports.ts.snap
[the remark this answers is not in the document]
wlee221: IMO, Each Account Settings component is its own component, so I wouldn't consider `DeleteUser` to be a subcomponent, like how `Authenticator.SignIn` is.

For example, `ChangePassword` isn't a subcomponent, but `ChangePassword.NewPassword` is.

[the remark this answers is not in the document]
calebpollman: Let's discuss this offline

packages/react/src/components/AccountSettings/DeleteUser/types.ts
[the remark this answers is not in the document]
wlee221: Oops, forgot to comment -- done in https://github.com/aws-amplify/amplify-ui/pull/2927/commits/54d3680353a07321b15ff1a881cafcbb5fa14518.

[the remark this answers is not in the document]
wlee221: latter is correct.

packages/react/src/components/AccountSettings/DeleteUser/defaultComponents.tsx
calebpollman: Can be done in that PR, but unless we have specific functionality or styling we probably don't need a standalone component or can just assign this as:
```
const DefaultSubmitButton: AccountSettingsButton = Button
```

calebpollman (approved):

PR #2920, opened 2022-11-09
chore(cypress e2e): Update Link checker for docs site

ErikCH:
#### Description of changes
Docs builds are failing because the number of links has grown. This will add the correct number of links

#### Issue #, if available
This will fix the issue with dev docs not building.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated
- [x] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated
- [x] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2931, opened 2022-11-10
test(e2e): add e2e docs numberOfPages test to Github Workflow

0618:
#### Description of changes

**Issue**: in https://github.com/aws-amplify/amplify-ui/pull/2844, we added e2e tests to traverse all the links on each page. However, the `numberOfPages` has to be defined manually every time when adding or removing pages from docs site.

After we added several new pages to the docs site https://github.com/aws-amplify/amplify-ui/pull/2708, the Amplify Hosting Test broken with the following error message:

```
 1) Local Sitemap
                                 should have 119 links:
                                 AssertionError: expected 121 to equal 119
                                 + expected - actual
                                 -121
                                 +119
```

So we had to make follow up PR to fix it https://github.com/aws-amplify/amplify-ui/pull/2920

**Fix**: To fail quicker before waiting until the test fails in Amplify Hosting build, this PR add an e2e test to verify the `numberOfPages` at the PR check.

- ran `yarn e2e test:docs` locally.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/.github/workflows/reusable-e2e.yml
+++ b/.github/workflows/reusable-e2e.yml
@@ -400,7 +400,7 @@ jobs:
         run: yarn docs start & npx wait-on -c waitOnConfig.json -t 20000 http-get://localhost:3000
 
       - name: Run E2E tests against docs
-        run: yarn workspace e2e test:theme
+        run: yarn workspace e2e test:docs
         env:
           # Env values for testing flows
           DOMAIN: ${{ secrets.DOMAIN }}
--- a/docs/cypress/e2e/links.cy.ts
+++ b/docs/cypress/e2e/links.cy.ts
@@ -11,7 +11,7 @@ type EvtName =
   | 'RETURNING'
   | 'NO_HREF';
 let allLinks: string[] = [];
-const numberOfLinks = 121;
+const numberOfPages = 121;
 const requestedLinks: Set<string> = new Set();
 
 before(() => {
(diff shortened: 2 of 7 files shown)


wlee221 (commented):
This lgtm for immediate solution. I have few concerns though:
- maintaining `numberOfLinks` will be difficult, because number of links change a lot through docs PR. For example, this will pop up on new documentation PRs for file uploader, user profile, rna, etc.
- I'm wondering if number of links can be tested through something simpler like jest, instead of Cypress. Link number test isn't really a UI test, which is what cypress specializes in.

0618:
> This lgtm for immediate solution. I have few concerns though:
> 
> * maintaining `numberOfLinks` will be difficult, because number of links change a lot through docs PR. For example, this will pop up on new documentation PRs for file uploader, user profile, rna, etc.

We will need to update the page number, which is trivial, but should not be too difficult, especially with the check.

> * I'm wondering if number of links can be tested through something simpler like jest, instead of Cypress. Link number test isn't really a UI test, which is what cypress specializes in.

Good point. Yea, I used Cypress to take advantage of the setup we've already got.

wlee221:
> We will need to update the page number, which is trivial, but should not be too difficult, especially with the check.

Right, but this is manual and has to be repeated on each docs PR that involves links. This will require (somewhat) unrelated PR changes to the PRs. 

One example this could be frustrating is when you merge latest change from your target branch, and # of links change. 

I'm not blocking, but was wondering if we really need to check for number of links, since it is expected that # of links change.

PR #2945, opened 2022-11-11
refactor(docs): remove numberOfPages from broken link test

0618:
#### Description of changes

**Issue**: in https://github.com/aws-amplify/amplify-ui/pull/2844, we added e2e tests to traverse all the links on each page. However, the `numberOfPages` has to be defined manually every time when adding or removing pages from docs site. Otherwise, the build would fail.

**Reason**: We had to set `numberOfPages` manually because `it.each` needs the number to set the test.

**Fix**: as a trade of, we can use `Array.forEach` to loop through all pages.

**Other Changes**:
- Use `Cypress.$` to read the `sitemap` so that we can remove `readSitemapLinks` task.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/cypress/e2e/links.cy.ts
```diff
@@ -10,34 +10,43 @@ type EvtName =
...
+            .replace(`${BASE_URL}/`, '')
+            .replace('https://www.dev.ui.docs.amplify.aws/', '')
+            .replace('https://ui.docs.amplify.aws/', '')
+        );
+    });
   });
-});
 
-describe(`All pages on Sitemap`, () => {
-  it.each(numberOfLinks)('all links on page %k should work', (i: string) => {
```
0618, line 30: Here's the trade-off. `allLinks` is empty when `it.each` sets up.

0618: Also, since it's a giant test, which takes a lot of memory, the Amplify Hosting Build pipeline failed on my personal account. Tried a few times, never succeed, but it works on my local machine.

But I wonder if we can try it in our Dev since its build's been quicker and less issue than my personal account.

docs/cypress/e2e/links.cy.ts
```diff
@@ -10,34 +10,43 @@ type EvtName =
...
-    const link = allLinks[i];
-    cy.task('log', `🧪[TESTING...] page ${BASE_URL}/${link}`);
-    cy.visit(link || '/');
-    cy.get('a').each((el) => hrefWorks(el, link));
+  it('should succesfully load each url in the sitemap', () => {
+    allLinks.forEach((link, idx) => {
+      cy.task('log', `🧪[TESTING...] page #${idx} ${BASE_URL}/${link}`);
+      cy.visit(link || '/');
+      cy.get('a').each((el) => hrefWorks(el, link, allLinks, requestedLinks));
```
ioanabrooks, line 37: If I understand this right, what we are doing here is going through each page in the sitemap list, retrieving all links on each page and testing them. We have quite a bit of logic in the `hrefWorks` fn to exclude links we've already tested as being part of sitemap, which makes this a bit fragil and almost requires a test for our test. What do you think about simplifying?

I've analyzed the links on most of our pages and found a couple patterns. We have:
* external links [github, discord, aws, etc] (rule: doesn't have our baseurl)
* internal links - same as the ones in sitemap but pointing to a section (rule: contains #)
* links that are in sitemap so we don't want to re-test them

So I was thinking that instead of line 37 we can break down the links testing like so:

```
cy.get("a:not([href^='${DOCS_URL}']), a[href*='#']").each ... call simplified fn that only does cy.request, checks status, logs 
```

0618: I like this idea 👍

0618: Updated, but we still need to keep the `hrefWorks` function because we still need to check if it's in `requestedLinks`

0618: I thought about it again. We need to verify if the href in `a` tag is in sitemap, not outdated, so we can't skip them.

I'll revert the commit.

docs/cypress/e2e/links.cy.ts
```diff
@@ -10,44 +10,60 @@ type EvtName =
...
-describe(`All pages on Sitemap`, () => {
-  it.each(numberOfLinks)('all links on page %k should work', (i: string) => {
-    const link = allLinks[i];
-    cy.task('log', `🧪[TESTING...] page ${BASE_URL}/${link}`);
-    cy.visit(link || '/');
-    cy.get('a').each((el) => hrefWorks(el, link));
+  it('should succesfully load each url in the sitemap', () => {
+    allLinks.forEach((link, idx) => {
+      cy.task('log', `🧪[TESTING...] page #${idx} ${BASE_URL}/${link}`);
+      cy.visit(link || '/');
```
ioanabrooks, line 36: Can we add a query string to easily identify and filter these requests in our merics?
```
cy.visit(URL, qs: {'cypress': true}) 
```

same comment for the `cy.request` call bellow.

docs/cypress.config.ts
```diff
@@ -8,23 +8,6 @@ export default defineConfig({
...
-              siteMapContent
-            );
-            return sitemapLinks.map((link) =>
-              link
-                .replace(`${BASE_URL}/`, '')
-                .replace('https://www.dev.ui.docs.amplify.aws/', '')
-                .replace('https://ui.docs.amplify.aws/', '')
-            );
-          }
-        },
```
ioanabrooks, line 27: Cypress will record video for all tests by default, let's turn that off in config to save some memory, screenshots should be sufficient in case of test failure.

```
video: false
```

docs/cypress/e2e/links.cy.ts
ioanabrooks: I see. Can we not add a separate: 
```
cy.get("a[href^='${DOCS_URL}']).each ... 
```
following `cy.get("a:not([href^='${DOCS_URL}']), a[href*='#']").each` ?

this way we also test the links that contain '#' which now we are missing as per the TODO

docs/cypress.config.ts
ioanabrooks: Cypress also supports running tests in parallel, we should investigate if we can enable this in our CI. Ref: https://docs.cypress.io/guides/guides/parallelization

docs/cypress/e2e/links.cy.ts
0618: Good idea. Done.

docs/cypress.config.ts
0618: Good points.
`video: false` added.

[parallelization](https://docs.cypress.io/guides/guides/parallelization) requires 1) multiple machines 2) multiple tests in separated files, so it doesn't help this PR 😢

docs/cypress/e2e/links.cy.ts
0618: The major reason why we are not testing `#` is to make less request so that we can save some memory 😂

docs/cypress.config.ts
ioanabrooks: right, I should have been more clear, by `we should investigate if` I meant add a ticket for follow-up, not address in this PR.

ioanabrooks (approved):

docs/cypress/e2e/links.cy.ts
0618: I see! I separated them by `cy.get(`a[href^='/']`)` and `cy.get(`a:not([href^='/']), a[href*='#']`)` for internal and external links. It does make the `if...else` shorter now

ioanabrooks (approved):

dbanksdesign (approved):

PR #2950, opened 2022-11-11
feat(ui-react-native): Add theming for IconButton primitive

joebuono:
#### Description of changes

- Removed unused `text` property from styles
- Updated unit tests to use `@testing-library/react-native`

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


packages/react-native/src/primitives/IconButton/styles.ts
```diff
@@ -1,7 +1,12 @@
...
 
-export const styles: IconButtonStyles = StyleSheet.create({
-  pressed: { opacity: 0.6 },
-  text: { alignSelf: 'center' },
-});
+export const getThemedStyles = (theme: StrictTheme): IconButtonStyles => {
+  const { components, opacities } = theme.tokens;
+
+  return StyleSheet.create({
+    pressed: { opacity: opacities[60], ...components?.iconbutton.pressed },
```
calebpollman, line 10: Thinking we should probably also provide a `container` prop so users can apply the theme to the `Pressable` when its not pressed

ioanabrooks: +1 Also disabled and size.

calebpollman:
LGTM. Just to confirm, did we check if it impacts usage in the InApp Messaging components?

ioanabrooks (approved):

calebpollman (approved):

joebuono:
> LGTM. Just to confirm, did we check if it impacts usage in the InApp Messaging components?

Doesn't appear to impact InApp Messaging

ioanabrooks (approved):

PR #2952, opened 2022-11-11
feat(ui-react-native): Add theming for PasswordField primitive

joebuono:
#### Description of changes

Note: The `styles` were not getting applied in `PasswordField.tsx`. This PR also fixes that issue. 

Wrote unit tests, ran in iOS and Android. 

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


ioanabrooks (approved):

packages/react-native/src/primitives/PasswordField/PasswordField.tsx
```diff
@@ -24,12 +31,13 @@ export default function PasswordField({
       {...rest}
       disabled={disabled}
       secureTextEntry={obscureText}
+      style={[themedStyle.container, style]}
```
calebpollman, line 34: Is this a `container` or the field itself?

joebuono: Currently, `TextField` applies the `style` prop to the container, not the field itself. ([link](https://github.com/aws-amplify/amplify-ui/blob/rna/main/packages/react-native/src/primitives/TextField/TextField.tsx#L39)). Perhaps we should have a separate `field` style for `PasswordField`?

joebuono: Yes I think so ([link](https://github.com/aws-amplify/amplify-ui/blob/rna/main/packages/react-native/src/primitives/TextField/TextField.tsx#L45))

calebpollman: Got it. Definitely not within the scope of this PR but we may want to rethink that

calebpollman (approved):

ioanabrooks (approved):

PR #2957, opened 2022-11-11
feat(fileuploader): cancel file name edit, a11y improvements

dbanksdesign:
#### Description of changes

* Editing a file name is now cancel-able. Edits to the filename are in a local state rather than global
* Removed `isEditing` in place of the existing 'editing' fileStatus
* Added some visually hidden text for icon buttons and `ariaHidden` for icons
* Made sure input is not tab-able
* Auto-focus file name text input after hitting edit button and save after pressing enter

https://user-images.githubusercontent.com/321279/201443669-9f9fd503-1138-40c7-9e27-7296d4d10959.mp4

https://user-images.githubusercontent.com/321279/201730985-25d6bd20-5991-4be1-b22b-a1b96a80f23a.mp4

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/Storage/FileUploader/UploadButton/index.tsx
+++ b/packages/react/src/components/Storage/FileUploader/UploadButton/index.tsx
@@ -26,6 +26,7 @@ export function UploadButton({
       <VisuallyHidden>
         <input
           type="file"
+          tabIndex={-1}
           ref={hiddenInput}
           onChange={onFileChange}
           multiple={multiple}
--- a/packages/react/src/components/Storage/FileUploader/FileUploader.tsx
+++ b/packages/react/src/components/Storage/FileUploader/FileUploader.tsx
@@ -37,9 +37,6 @@ export function FileUploader({
 
   const fileStatusesRef = useRef<FileStatuses>([]);
 
-  // Tracker state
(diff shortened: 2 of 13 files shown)


packages/react/src/components/Storage/FileUploader/UploadButton/index.tsx
```diff
@@ -26,6 +26,7 @@ export function UploadButton({
       <VisuallyHidden>
         <input
           type="file"
+          tabIndex={-1}
```
0618, line 29: why do we want to "Made sure input is not tab-able"?

0618 (approved):

packages/react/src/components/Storage/FileUploader/UploadButton/index.tsx
dbanksdesign: Because the control of the input is handled by the button. We don't want keyboard users to access the input directly or screen readers to announce the input or those users might get confused with 2 focusable elements. 

Screen Shot 2022-11-01 at 2 51 44 PM

ErikCH (approved):
LGTM!

PR #2983, opened 2022-11-15
chore(ui-react-native): Restrict verify user fields to unverified contacts

ioanabrooks:
#### Description of changes

This change will restrict verify user fields to unverified contact methods.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react-core/src/Authenticator/hooks/types.ts
```diff
@@ -143,6 +143,7 @@ export type SignUpBaseProps<FieldType = {}> = {
 
 export type VerifyUserProps<FieldType = {}> = {
   skipVerification: AuthenticatorMachineContext['skipVerification'];
+  unverifiedContactMethods: AuthenticatorMachineContext['unverifiedContactMethods'];
```
calebpollman, line 146: `unverifiedContactMethods` is already passed to `VerifyUser` as its `fields` prop, so this is not required

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -39,6 +43,18 @@ export const getSanitizedRadioFields = (
...
+      unverifiedContactMethods &&
+      Object.keys(unverifiedContactMethods).findIndex((key) => key === name) < 0
+    ) {
+      logger.warn(
+        `${name} is not supported. Available values are: ${Object.keys(
+          unverifiedContactMethods
+        )}.`
+      );
+      return false;
+    }
```
calebpollman, line 56: We really just need to make sure that if a user passes a custom value to the `fields` prop of `VerifyUser` and it has a `name` that isn't one of the two keys that are allowed by `UnverifiedContactMethods` it's excluded. We don't need to add any additional props or types in this PR

ioanabrooks: Wouldn't it be a better user experience to also print out what the accepted values are though? Just saying that filed x was excluded is not helpful enough imo.

packages/react-core/src/Authenticator/hooks/types.ts
ioanabrooks: Thanks, removed!

packages/react-native/src/Authenticator/Defaults/VerifyUser/VerifyUser.tsx
```diff
@@ -19,14 +20,17 @@ const COMPONENT_NAME = 'VerifyUser';
...
+  if (type === 'Email') {
     const splitEmail = value.split('@');
     const censoredName = censorAllButFirstAndLast(splitEmail[0]);
 
     censoredVal = `${censoredName}@${splitEmail[1]}`;
-  } else if (name === 'phone') {
+  } else if (type === 'Phone Number') {
     censoredVal = censorPhoneNumber(value);
   }
   return censoredVal;
```
joebuono, line 36: Just to be clear, if the `type` !== 'Email' or 'Phone Number', the returned value is uncensored, correct?

joebuono (dismissed):
LGTM, just one question

packages/react-native/src/Authenticator/Defaults/VerifyUser/VerifyUser.tsx
ioanabrooks: The available contact methods are only email or phone [and that is unlikely to change]  so there shouldn't be another option.

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -2,9 +2,18 @@ import { Logger } from 'aws-amplify';
 import { AuthenticatorRouteComponentName } from '@aws-amplify/ui-react-core';
 
 import { RadioFieldOptions, TypedField } from '../types';
+import { ContactMethod, UnverifiedContactMethods } from '@aws-amplify/ui';
 
 const logger = new Logger('Authenticator');
 
+export const supportedContactMethods: Record<
```
calebpollman, line 9: Can this just be an array?

ioanabrooks: The reason why I went with `Record<keyof UnverifiedContactMethods,ContactMethod>` is to ensure that if in the future another unverified contact method is added, or an existing one is changed we'll get some safety here.

packages/react-native/src/Authenticator/Defaults/VerifyUser/VerifyUser.tsx
```diff
@@ -82,12 +82,12 @@ const FormFields: DefaultVerifyUserComponent['FormFields'] = ({
...
-      {fields.map(({ name, value, ...props }) => (
+      {fields.map(({ label, name, value, ...props }) => (
         <Radio
           {...props}
-          key={value}
-          value={value}
-          label={censorContactInformation(name, value)}
+          key={name}
+          value={name}
+          label={censorContactInformation(label, value)}
```
wlee221, line 90: Wouldn't `name` still make more sense to use as parameter here? Since customer might change `label` to some custom string.

ioanabrooks: Thanks for catching this! Addressed.

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
calebpollman: Totally align on the intention, but have a few callouts:
- creating an object with values that themselves do not correlate to any logic has the potential to confuse future maintainers since the existence of the values implies they have a use case, as objects are meant to store values for usage elsewhere, not type checking. Additionally, `ContactMethod` does not correlate to `UnverifiedContactMethod`, its use case is to denote the allowed types of sign up confirmation code destination
- usage of a `string` enum and utility would much better suit this use case, provide more utility in the codebase, and could be colocated with the `UnverifiedContactMethods` type (which we maintain as well)

packages/react-native/src/Authenticator/hooks/useFieldValues/__tests__/useFieldValues.spec.ts
```diff
@@ -82,6 +86,20 @@ describe('useFieldValues', () => {
     });
   });
 
+  it('logs a warning for non array fields', () => {
+    const mockFields = 'test';
+    // test js users usecase, ignore ts on purpose
+    //@ts-ignore
```
calebpollman, line 92: Rather than introduce `@ts-ignore` in test cases, cast the invalid value, example `mockFields as unknown as RadioFieldOptions[]`

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -50,6 +59,17 @@ export const getSanitizedRadioFields = (
       return false;
     }
 
+    if (
+      Object.keys(supportedContactMethods).findIndex((key) => key === name) < 0
+    ) {
+      logger.warn(
+        `'${name}' is not supported. Available values are: ${Object.keys(
```
calebpollman, line 66: Can we align this message with the rest of the messages in this file? Example:
```
`Key of ${name} is not supported, field has been ignored. Supported keys: ${...}`
```

packages/react-native/src/Authenticator/hooks/useFieldValues/__tests__/useFieldValues.spec.ts
ioanabrooks: Thanks! I also went ahead a modified another test using this pattern that slipped through.

calebpollman (approved):
🚢

wlee221 (approved):

PR #2989, opened 2022-11-15
feat(ui-react-native): Button variants

joebuono:
#### Description of changes

- Add `primary` and `secondary` Button variants
- Refactor styles for RNA subcomponent screens
- Update tests and snapshots

AuthButtons3

ButtonVariantsDemo

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated


packages/react-native/src/primitives/Button/Button.tsx
```diff
@@ -14,30 +15,55 @@ import { ButtonProps } from './types';
...
   const theme = useTheme();
   const themedStyle = getThemedStyles(theme);
 
   const pressableStyle = useCallback(
     ({ pressed }: PressableStateCallbackType): StyleProp<ViewStyle> => {
+      const containerStyle = {
+        ...themedStyle.container,
+        ...(variant === 'primary'
+          ? themedStyle.containerPrimary
+          : themedStyle.containerSecondary),
```
calebpollman, line 33: Can this just be:
```
...themedStyle[`container${capitalize(variant)}`],
```

packages/react-native/src/primitives/Tabs/styles.ts
```diff
@@ -23,7 +23,8 @@ export const getThemedStyles = (theme: StrictTheme): TabsStyles => {
       borderTopWidth: space.xxs,
       flexBasis: 0,
       flexGrow: 1,
-      padding: space.large,
+      paddingHorizontal: space.large,
+      paddingVertical: space.large,
```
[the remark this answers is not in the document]
joebuono: Because otherwise the `paddingVertical` and `paddingHorizontal` from Button override the Tab's `padding`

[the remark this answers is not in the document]
calebpollman: Hmmm, are we not applying the `style` prop in the `Button` in a way where it overrides the default style?

packages/react-native/src/primitives/Button/Button.tsx
joebuono: Yes!

packages/react-native/src/primitives/Button/styles.ts
```diff
@@ -1,13 +1,42 @@
...
+      ...components?.button.disabled,
+    },
     pressed: { opacity: opacities[60], ...components?.button.pressed },
-    text: { alignSelf: 'center', ...components?.button.text },
+    text: {
+      alignSelf: 'center',
+    },
+    textPrimary: {
+      color: colors.white,
+      fontSize: fontSizes.medium,
```
calebpollman, line 33: Should we use the same `fontSize` for both variants or have a specific value for the `secondary` variant?

joebuono: We could use a specific size. I thought that making them slightly different sizes looked right for our use case in the RNA.

calebpollman: RNA use case and default use case are separate. Although the RNA styling should be aligned with our default components, ultimately it is a consumer of the base UI and not the solo use case for the base UI

calebpollman (approved):

ioanabrooks (approved):

PR #2998, opened 2022-11-16
chore(file-uploader): Removed fileNames prop and added warning for users for required props

ErikCH:
#### Description of changes
After some discuss with @dbanksdesign we are removing the `fileNames` prop, since it is not obvious to use. Also added warning if users add the file-uploader without the correct props.

Removed test not needed
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/Storage/FileUploader/FileUploader.tsx
+++ b/packages/react/src/components/Storage/FileUploader/FileUploader.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useState, useRef, useCallback } from 'react';
 import { UploadTask, Storage } from '@aws-amplify/storage';
-import { getFileName, translate, uploadFile } from '@aws-amplify/ui';
+import { translate, uploadFile } from '@aws-amplify/ui';
 import { FileStatuses, FileUploaderProps } from './types';
 import { useFileUploader } from './hooks/useFileUploader';
 import { ComponentClassNames, Text } from '../../../primitives';
@@ -15,7 +15,6 @@ const isUploadTask = (value: unknown): value is UploadTask =>
 export function FileUploader({
   acceptedFileTypes,
   components = {},
-  fileNames,
   isPreviewerVisible,
   level,
   maxFiles,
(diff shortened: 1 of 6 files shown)


packages/react/src/components/Storage/FileUploader/FileUploader.tsx
```diff
@@ -34,6 +33,13 @@ export function FileUploader({
     Tracker = FileUploader.Tracker,
   } = components;
 
+  if (!acceptedFileTypes || !level) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'You must include the level and acceptedFileNames props to use the file uploader!'
```
ErikCH, line 39: New warning to users

packages/react/src/components/Storage/FileUploader/FileUploader.tsx
```diff
@@ -34,6 +33,13 @@ export function FileUploader({
     Tracker = FileUploader.Tracker,
   } = components;
 
+  if (!acceptedFileTypes || !level) {
+    // eslint-disable-next-line no-console
+    console.warn(
```
ioanabrooks, line 38: We have a logger util we can use for these message without having to disable linter.

```
import { Logger } from 'aws-amplify';
...
const logger = new Logger('FileUploader');
```

ioanabrooks: Bonus to add a unit test that tests the logger is called in this curcumstance

ErikCH: Sounds good! Updated in fbef484

packages/react/src/components/Storage/FileUploader/FileUploader.tsx
wlee221: I believe you have to initialize the logger first:

```
const logger = new Logger('AmplifyUI:Auth');

logger.warning('You must include the level and acceptedFileNames props to use the file uploader');
```

wlee221: I'd also:
1. put the logger above this react component so that it doesn't reinitialize on every render, and
2. consider removing exclamation mark from the logger message. Maybe we can also change it to "`FileUploader` requires level and acceptedFileNames props" so that it's immediately clear the error is about FileUploader.

wlee221 (commented):
This lgtm, just a nit about logger usage!

packages/react/src/components/Storage/FileUploader/FileUploader.tsx
ErikCH: Updated [9e9b76c](https://github.com/aws-amplify/amplify-ui/pull/2998/commits/9e9b76c84a1d805e7b995a19fc0b311da69db08a)

wlee221 (approved):

PR #3002, opened 2022-11-17
fix(rna): change `useRouteTypedFields` to a util function

wlee221:
#### Description of changes

Changes `useRouteTypedFields` to a simple util function instead of a React hook.

#### Details

`useRouteTypedFields` caused a bug with `VerifyUser` screen, because `useGetRouteTypedFields` updates and returns its value asynchronously due to its use of `useEffect`. 

This hook doesn't really need to be a hook actually, and can be turned a synchronous util function, since the value of `typedFields` is constant based on current `route` and `fields`.

`Authenticator` is already listening to `route` changes, so no additional changes need to be made regarding listening to changes.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -78,3 +89,85 @@ export const getSanitizedTextFields = (
...
+export const getTypedFields = (
+  fields: AuthenticatorLegacyField[]
+): TypedField[] => fields?.map(getTypedField);
+
+export function getRouteTypedFields({
+  fields,
+  route,
+}: Pick<UseAuthenticator, 'fields' | 'route'>): TypedField[] {
+  const isComponentRoute = isAuthenticatorComponentRouteKey(route);
+  if (isComponentRoute) {
```
calebpollman, line 164: Can we just early return if `isComponentRoute` is `false`?

calebpollman (approved):
Left a minor request

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
wlee221: Done in [9da90c2](https://github.com/aws-amplify/amplify-ui/commit/9da90c2dab4f8c5bf018c32d80b8f8dcf64b9474)

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -78,3 +89,86 @@ export const getSanitizedTextFields = (
...
+      if (isPhoneFieldType && key === 'dialCodeList') {
+        return { ...acc, dialCodes: value as string[] };
+      }
+
+      return { ...acc, [key]: value };
+    },
+    // initialize `acc` with field `name` and `type`
+    { name, type } as TypedField
+  );
+};
```
wlee221, line 149: These are just copy pasted from the original util file.

ioanabrooks (approved):

calebpollman (approved):

PR #3010, opened 2022-11-18
fix(rna): Fix memoization issue in `useAuthenticatorRoute`

wlee221:
#### Description of changes

Fixes memoization issue in `useAuthenticatorRoute`. The default selector was too restrictive, because it kept its memoization even when `route` changed. 

This PR changes the default selectors to always include "route".

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react-core/src/Authenticator/hooks/useAuthenticatorRoute/__tests__/utils.spec.ts
+++ b/packages/react-core/src/Authenticator/hooks/useAuthenticatorRoute/__tests__/utils.spec.ts
@@ -40,6 +40,7 @@ const {
   getTotpSecretCode,
   isPending,
   resendCode,
+  route,
   skipVerification,
   socialProviders,
   submitForm,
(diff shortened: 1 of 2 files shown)


calebpollman (dismissed):
LGTM 🚢

packages/react-core/src/Authenticator/hooks/useAuthenticatorRoute/__tests__/utils.spec.ts
```diff
@@ -71,17 +72,20 @@ describe('getRouteMachineSelector', () => {
   it.each([
     [
       'confirmResetPassword',
-      [...commonSelectorProps, resendCode, validationErrors],
+      [...commonSelectorProps, resendCode, validationErrors, route],
```
wlee221, line 75: I was thinking about adding `route` to `commonSelectorProps`, but that started making an assumption on order of dependencies, so I just left route at the end of the array to mimic the implementation below.

ioanabrooks: Thanks for adding a test!!

ioanabrooks (approved):

PR #3034, opened 2022-11-20
chore(ui-react-native): omit checkbox from themed Components

calebpollman:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react-native/src/theme/types.ts
+++ b/packages/react-native/src/theme/types.ts
@@ -54,6 +54,9 @@ export type StrictTokens = ReactNativeTokens<'required'>;
 // Everything optional for custom themes
 export type Tokens = ReactNativeTokens<'optional'>;
 
+// omitted from custom themes until used in another component/publicly available
+type OmittedComponents = 'checkbox' | 'divider' | 'tabs';
+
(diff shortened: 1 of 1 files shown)


packages/react-native/src/theme/types.ts
```diff
@@ -54,14 +54,17 @@ export type StrictTokens = ReactNativeTokens<'required'>;
 // Everything optional for custom themes
 export type Tokens = ReactNativeTokens<'optional'>;
 
+// omitted from custom themes until used in another component/publicly available
+type OmittedComponents = 'checkbox' | 'divider' | 'tabs';
```
ioanabrooks, line 58: Good catch! Actually, looks like icon is not used either outside of IconButton and radio not used outside of RadioGroup.

calebpollman: We use `IconButton` in the `PasswordField` right? And we use a `RadioGroup` in `VerifyUser``?

ioanabrooks (approved):

joebuono (approved):

PR #3036, opened 2022-11-21
chore(rna): refactor DefaultContainer

calebpollman:
#### Description of changes
Refactor RNA `DefaultContainer` to prevent `TextField` components from getting covered by the keyboard. This is a v1 component, will be updated to allow programmatic positioning based on current focused input in the future.

Simulator Screen Recording - iPhone 13 - 2022-11-19 at 11 41 49

#### Issue #, if available
NA

A ton of manual testing
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
+++ b/packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
@@ -19,9 +19,7 @@ export default function DefaultContainer({
   alwaysBounceVertical = ALWAYS_BOUNCE_VERTICAL,
   children,
   keyboardAvoidingViewStyle,
-
   keyboardShouldPersistTaps = KEYBOARD_SHOULD_PERSIST_TAPS,
-
   keyboardVerticalOffset = KEYBOARD_VERTICAL_OFFSET,
   scrollViewContentContainerStyle,
   style,
--- a/packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
+++ b/packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
@@ -1,42 +1,60 @@
 import React, { useMemo } from 'react';
-import { ScrollView, KeyboardAvoidingView } from 'react-native';
+import { KeyboardAvoidingView, ScrollView } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
+import { useDeviceOrientation } from '../../../hooks';
 import { useTheme } from '../../../theme';
+import { platform } from '../../../utils';
+
 import { getThemedStyles } from './styles';
 import { ContainerProps } from './types';
 
 // prevent bounce when `ScrollView` content is less than height of screnn
 const ALWAYS_BOUNCE_VERTICAL = false;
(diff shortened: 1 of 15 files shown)


packages/react-native/src/Authenticator/common/DefaultContainer/InnerContainer.tsx
```diff
@@ -1,15 +1,19 @@
...
+import { getInnerContainerStyles } from './styles';
 import { ContainerProps } from './types';
 
 export default function InnerContainer({
   children,
   ...rest
 }: ContainerProps): JSX.Element | null {
+  const theme = useTheme();
+
+  const themedStyle = getInnerContainerStyles(theme);
```
calebpollman, line 14: This should be wrapped in a `useMemo` but it needs to move to `DefaultContainer` so will handle it then

packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
```diff
@@ -1,16 +1,56 @@
...
+    () => getThemedStyles(theme, { paddingBottom, paddingTop }),
+    [paddingBottom, paddingTop, theme]
+  );
+
   return (
-    <View {...rest} style={defaultContainerstyles.container}>
-      {children}
-    </View>
+    <KeyboardAvoidingView
+      behavior="position"
```
ioanabrooks, line 38: Does this behavior work for both iOS and android platforms? In my experience had to pass in different values, for android left undefined if I remember correctly, but may be outdated.

packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
```diff
@@ -1,16 +1,56 @@
...
+        keyboardShouldPersistTaps={keyboardShouldPersistTaps}
+        contentContainerStyle={[
+          themedStyle.scrollViewContentContainer,
+          scrollViewContentContainerStyle,
+        ]}
+        style={[themedStyle.scrollView, style]}
+      >
+        {children}
+      </ScrollView>
+    </KeyboardAvoidingView>
```
ioanabrooks, line 54: Have we tested this in landscape mode?

ioanabrooks (commented):
Thanks for adding this, huge improvement! I just have 2 small questions, non-blocking.

ioanabrooks (approved):

packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
calebpollman: Been a while since i worked with this API as well and after more testing made some updates. Good catch!

packages/react-native/src/Authenticator/common/DefaultContainer/DefaultContainer.tsx
calebpollman: Thanks for catching. Had tested an earlier version that was using a `SafeAreaView` on iOS only. Made some improvements for iOS, didn't get in to Android since default React Native apps don't allow orientation updates without native code changes. Am completely open to enabling orientation changes in the RN example and following up in another PR though

ioanabrooks (approved):

joebuono (approved):

PR #3039, opened 2022-11-21
feat(rna): adding default dark mode export

dbanksdesign:
#### Description of changes

* Added a properly typed for RN default dark mode override

https://user-images.githubusercontent.com/321279/203151219-f44ef358-c30d-4d10-ab5f-dddbd5f1d755.mp4

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react-native/src/theme/types.ts
```diff
@@ -43,8 +43,8 @@ export type Components<Output> = Record<string, object> & {
   textField?: ComponentTheme<TextFieldStyles, Output>;
 };
 
-export type ColorMode = 'light' | 'dark' | 'system';
-type Override = Omit<Theme, 'overrides'> & {
+export type ColorMode = 'light' | 'dark' | 'system' | null;
```
ioanabrooks, line 46: When would this be null? we should default to 'system', not allow null imho.

calebpollman (approved):
🚢

packages/react-native/src/theme/types.ts
dbanksdesign: If they pass the return of `useColorScheme` which can be null I guess? That is why I had to add it in

dbanksdesign: https://reactnative.dev/docs/usecolorscheme

wlee221 (approved):

calebpollman (approved):

PR #3048, opened 2022-11-22
chore(ui-react-native): SetupTOTP small fixes

ioanabrooks:
#### Description of changes

* add instructions about what to do with secret key
* style fixes

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react-native/src/Authenticator/Defaults/SetupTOTP/__tests__/SetupTOTP.spec.tsx
+++ b/packages/react-native/src/Authenticator/Defaults/SetupTOTP/__tests__/SetupTOTP.spec.tsx
@@ -37,6 +37,7 @@ const {
   getConfirmingText,
   getConfirmText,
   getSetupTOTPText,
+  getSetupTOTPInstructionsText,
 } = authenticatorTextUtil;
 
 const SECRET_KEY = 'secretKey';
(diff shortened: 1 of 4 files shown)


packages/react-native/src/Authenticator/Defaults/SetupTOTP/__tests__/SetupTOTP.spec.tsx
```diff
@@ -96,13 +97,14 @@ describe('SetupTOTP', () => {
...
-    const { queryByText, rerender } = render(
+    const { getByText, queryByText, rerender } = render(
       <SetupTOTP {...props} getTotpSecretCode={mockGetTotpSecretCode} />
     );
     await waitFor(async () => {
       expect(mockGetTotpSecretCode).toHaveBeenCalledTimes(1);
+      expect(getByText(getSetupTOTPInstructionsText())).toBeDefined();
       expect(queryByText(SECRET_KEY)).toBeDefined();
-      expect(queryByText(SECRET_KEY)?.props.selectable).toBe(true);
+      expect(getByText(SECRET_KEY).props.selectable).toBe(true);
```
calebpollman, line 107: Just for my own understanding, why use both `queryByText` and `getByText`? Is it that `queryByText` can be used to check for elements that might be undefined?

calebpollman (approved):
LGTM, as long as the code is still selectable

packages/react-native/src/Authenticator/Defaults/SetupTOTP/__tests__/SetupTOTP.spec.tsx
ioanabrooks: Correct, `getByText` throws an error if the element isn't foun, while `queryByText` returns undefined. Can use either to check for element but line 107 would have been problematic as it doesn't really test anything if the element is undefined.

0618 (approved):

PR #3064, opened 2022-11-22
chore(rna): remove socialProviders props

calebpollman:
#### Description of changes
Remove `socialProviders` props from `SignIn` and `SignUp` components

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react-core/src/Authenticator/hooks/types.ts
```diff
@@ -122,7 +122,6 @@ export type SetupTOTPBaseProps<FieldType = {}> = {
 
 export type SignInBaseProps<FieldType = {}> = {
   hideSignUp?: boolean;
-  socialProviders?: AuthenticatorMachineContext['socialProviders'];
```
dbanksdesign, line 125: I assume removing this from the 'react-core' won't negatively affect the react web package? I have no idea how this works.

dbanksdesign (approved):
Just one question based on my ignorance. Feel free to ignore if it is a dumb question.

packages/react-core/src/Authenticator/hooks/types.ts
calebpollman: Valid question. React web doesn't use this type yet, but the types are flexible enough that we can declare these types directly on the subcomponents in the React package when we migrate

PR #3071, opened 2022-11-22
fix(theme): update types for ThemeProvider to accept output of createTheme

dbanksdesign:
#### Description of changes

One use-case of theming is to extend from a theme, for example the Studio generated theme. To do this, customers pass in the studio theme as the 2nd argument to the `createTheme` method. The issue is `createTheme` returns a `WebTheme` type and not a `Theme` type which is what the `ThemeProvider` expects. Loosening up the type of the `ThemeProvider` theme prop to allow either `Theme` or `WebTheme`

```javascript
const theme = createTheme({

}, studioTheme);

<ThemeProvider theme={theme}>
</ThemeProvider>
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
+++ b/packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
@@ -31,7 +31,7 @@ describe('ThemeProvider', () => {
         tokens: {
           colors: {
             font: {
-              primary: { value: 'blue' },
+              primary: { value: 'hotpink' },
             },
           },
         },
--- a/packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
+++ b/packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
@@ -1,9 +1,11 @@
 import { render, screen } from '@testing-library/react';
-import { Theme } from '@aws-amplify/ui';
+import { renderHook } from '@testing-library/react-hooks';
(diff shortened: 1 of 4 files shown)


calebpollman (dismissed):

packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
```diff
@@ -21,6 +23,31 @@ describe('ThemeProvider', () => {
...
+      studioTheme
+    );
+
+    const { result } = renderHook(() => useTheme(), {
+      wrapper: ({ children }) => (
+        <ThemeProvider theme={extendedTheme}>{children}</ThemeProvider>
+      ),
+    });
+
+    expect(result.current.tokens.colors.font.primary.value).toBe('hotpink');
```
joebuono, line 48: Why do we expect it to be `'hot pink'` instead of `'blue'`?

joebuono:
Can we add a changeset?

packages/react/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
dbanksdesign: that is my bad, this will fail. I was playing with things and messed up

joebuono (approved):

wlee221 (approved):

ioanabrooks (approved):

PR #3128, opened 2022-12-02
feat(account-settings): align classnames

wlee221:
#### Description of changes

Aligns classnames from account settings components. Naming convention taken from https://github.com/aws-amplify/amplify-ui/blob/145d0b5f596ff7c9f623898af0bb3836516c51fe/packages/react/src/components/InAppMessaging/BannerMessage/constants.ts#L2

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


calebpollman (approved):

packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -177,7 +177,11 @@ function ChangePassword({
   }
 
   return (
-    <View as="form" className="amplify-changepassword" onSubmit={handleSubmit}>
+    <View
+      as="form"
+      className="amplify-accountsettings-changepassword"
```
ioanabrooks, line 182: Would be nicer if instead of hardcoding, these class names were constants, could be used in unit tests as well, to assert they are applied as expected ;)
Ditto for the other classnames as well.

wlee221: Good point. I've added `ComponentClassName` constant. I didn't put it in ui package yet. 

I think we should refactor / consolidate connected components classnames to the ui package (especially authenticator ones are all the place), but let's keep that to a different PR.

ioanabrooks (approved):

PR #3129, opened 2022-12-02
fix(theme): making all component tokens optional

dbanksdesign:
#### Description of changes

In the large theme type refactor @calebpollman and I did, there were some misses. Some component theme types were not marking all properties as optional, which causes some TS issues. This fixes those issues and adds a test to ensure this doesn't happen again

Tests pass (after making the test which failed)

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/src/theme/__tests__/themeType.test.ts
+++ b/packages/ui/src/theme/__tests__/themeType.test.ts
@@ -58,6 +58,61 @@ describe('Amplify UI Theme', () => {
       };
       // This test doesn't test anything, but
       // if there is a TS error it will fail
+      expect(theme).toMatchInlineSnapshot(`
+        Object {
+          "name": "test",
+          "tokens": Object {
+            "components": Object {
+              "alert": Object {},
+              "authenticator": Object {},
+              "autocomplete": Object {},
+              "badge": Object {},
+              "button": Object {},
+              "card": Object {},
+              "checkbox": Object {},
+              "checkboxfield": Object {},
+              "collection": Object {},
+              "copy": Object {},
+              "divider": Object {},
+              "expander": Object {},
+              "field": Object {},
+              "fieldcontrol": Object {},
+              "fieldgroup": Object {},
+              "fieldmessages": Object {},
+              "fileuploader": Object {},
+              "flex": Object {},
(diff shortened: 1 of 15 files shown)


packages/ui/src/theme/__tests__/themeType.test.ts
```diff
@@ -0,0 +1,63 @@
+import { Theme } from '../index';
+
+describe('Amplify UI Theme', () => {
+  describe('components', () => {
+    it('should allow empty components', () => {
+      const theme: Theme = {
+        name: 'test',
+        tokens: {
+          components: {
+            alert: {},
```
dbanksdesign, line 10: Is there a better way to test a type shape like this? Also, if we add a component we will need to add it here. Grabbing all the component names and creating an object dynamically will not trigger a TS error.

packages/ui/src/theme/tokens/components/button.ts
```diff
@@ -63,7 +63,8 @@ export type ButtonTokens<Output extends OutputVariantKey> =
     | 'borderWidth'
     | 'borderStyle'
     | 'borderRadius'
-    | 'color'
+    | 'color',
+    Output
```
dbanksdesign, line 67: The 2 types of issue are this one, where we forgot to pass the `Output` generic down, and the one below where we just forgot to add a `?`

packages/ui/src/theme/__tests__/themeType.test.ts
```diff
@@ -0,0 +1,63 @@
+import { Theme } from '../index';
+
+describe('Amplify UI Theme', () => {
+  describe('components', () => {
+    it('should allow empty components', () => {
+      const theme: Theme = {
```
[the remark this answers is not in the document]
ErikCH: Not sure how we can skip this error...

ErikCH (dismissed):
Looks good, not sure how we can exclude the unused variable warning.

joebuono (approved):

ErikCH (approved):

calebpollman (approved):

joebuono (approved):

PR #3141, opened 2022-12-06
fix(account-settings): early return if `isPending` is true

wlee221:
#### Description of changes

Changes the order of early returns, so that it checks `isLoading` before it checks `!user`. This ensures that component does not prematurely log that user isn't present to console.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
+++ b/packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
@@ -71,7 +71,12 @@ function DeleteUser({
     runDeleteUser();
   }, [runDeleteUser]);
 
-  // Return null if user isn't authenticated in the first place
+  // Return null if Auth.getCurrentAuthenticatedUser is still in progress
+  if (isLoading) {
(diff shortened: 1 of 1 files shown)


packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -71,7 +71,12 @@ function DeleteUser({
...
 
-  // Return null if user isn't authenticated in the first place
+  // Return null if Auth.getCurrentAuthenticatedUser is still in progress
+  if (isLoading) {
+    return null;
+  }
+
+  // Return null if user isn't authenticated
   if (!user) {
     logger.warn('<DeleteUser /> requires user to be authenticated.');
```
wlee221, line 81: We don't want this logger to run prematurely

joebuono (approved):

ErikCH (approved):

ioanabrooks (approved):

Issue #1805, opened 2022-04-29
Bug: RadioGroupField does not use a fieldset and legend to group and label options


hbuchel:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Any

### Please describe your bug.

Currently, the RadioFieldGroup uses a `<div>` with a `<label>` to group and label its options. 

### What's the expected behaviour?

It should use a `<fieldset>` with a `<legend>` so that we can convey semantic meaning to assistive technologies. 

[Related WCAG guideline: 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/quickref/#info-and-relationships) | [Examples](https://www.w3.org/WAI/WCAG21/Techniques/html/H71.html)

### Help us reproduce the bug!

Any example on the [RadioGroupField docs page](https://ui.docs.amplify.aws/components/radiogroupfield) and when used in example app.

### Code Snippet

```typescript
// Put your code below this line.

```

### Additional information and screenshots

Just a note that I tested a "quick fix" for this by updating the RadioGroupField component to add...

- `as="fieldset"` on its [Flex component](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx#L70)
- and adding `as="legend"` on its [Label component](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx#L80)

... this correctly updated the markup, however the gap properties of the parent Flex component were not being applied in Chrome, leading to no margin between the legend and the option group. `<fieldset>` elements still seem pretty finicky when applying display: flex to them (can find similar bug reports as to [this one in Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=375693) scattered around) so that solution might require further debugging, or an alternate solution to applying the gap.

Also, note that `<legend>` must be a the first and direct descendant of its parent <fieldset>: https://html.spec.whatwg.org/multipage/form-elements.html#the-legend-element

[...]

PR #3144, opened 2022-12-06
fix(ui-react): Add fieldset and labelledby to RadioGroupField

joebuono:
#### Description of changes

Addresses the accessibility issues identified in [this PR](https://github.com/aws-amplify/amplify-ui/issues/1805). 

Due to the following [constraints in styling](https://github.com/aws-amplify/amplify-ui/issues/1805#:~:text=...%20this%20correctly%20updated,of%20its%20parent) `fieldset` and `legend` elements using `flex`, this PR takes the following approach:
- Add the `aria-labelledby` property to the outer `Flex`, and set it to the `id` of the `Label`, which is functioning as a `legend`
- Move the `role="radiogroup"` to the outer `Flex`, to reproduce the same accessibility experience of using `fieldset` and `legend` for a `radiogroup`

Current:
https://user-images.githubusercontent.com/48109584/206017482-c6c223c2-e574-4655-bd2c-a6504dc19c7b.mov

Updated:
https://user-images.githubusercontent.com/48109584/206018053-e971c443-3664-4d93-9dcd-cbfef2138767.mov

#### Issue #, if available

Fixes: https://github.com/aws-amplify/amplify-ui/issues/1805

Ran the docs locally and tested with VoiceOver on Mac

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are updated

### DIFF
--- a/packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx
+++ b/packages/react/src/primitives/RadioGroupField/RadioGroupField.tsx
@@ -7,6 +7,7 @@ import { Flex } from '../Flex';
 import { Label } from '../Label';
 import { RadioGroupContext, RadioGroupContextType } from './context';
 import { RadioGroupFieldProps, Primitive } from '../types';
+import { getTestId } from '../utils/testUtils';
 import { useStableId } from '../utils/useStableId';
 
 // Note: RadioGroupField doesn't extend the JSX.IntrinsicElements<'input'> types (instead extending 'typeof Flex')
@@ -29,6 +30,7 @@ const RadioGroupFieldPrimitive: Primitive<RadioGroupFieldProps, typeof Flex> = (
     onChange,
     name,
     size,
+    testId,
     value,
     ...rest
   },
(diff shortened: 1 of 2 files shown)


hbuchel:
@joebuono Just adding some notes from our offline conversation:
- Let's test this on iOS + voiceover (on a real device). I think that is where the main issue would lie with using the aria-labelledby technique we currently have. (see [findings here](https://adrianroselli.com/2022/07/use-legend-and-fieldset.html) also)
- Probably we can just use fieldset without adding the radiogroup role? I don't think the radiogroup role is necessary since we're using native radio elements, and the fieldset would suffice.
- Depending on the iOS testing, I would suggest going with what's recommended in that article: using a fieldset, a visually hidden legend since it doesn't behave with flex rules, an aria-hidden Text for the "visual" legend that DOES behave with flex rules, and dropping the aria-labelledby.

joebuono:
Taking a different approach to fix iOS bug

PR #3156, opened 2022-12-10
chore(ui-react-native): Refactor unit tests

ioanabrooks:
#### Description of changes

This change refactors existing React Native tests to use @testing-library instead of react-test-renderer. Also adds some assertions and tests to improve coverage.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react-native/src/InAppMessaging/components/BannerMessage/__tests__/BannerMessage.spec.tsx
+++ b/packages/react-native/src/InAppMessaging/components/BannerMessage/__tests__/BannerMessage.spec.tsx
@@ -95,9 +95,7 @@ describe('BannerMessage', () => {
       const { getByTestId } = render(<BannerMessage {...props} />);
       const testElement = getByTestId(testID);
 
-      for (const [key, value] of Object.entries(expectedProps)) {
-        expect(testElement.props[key]).toEqual(value);
-      }
+      expect(testElement.children).toContain(expectedProps.children);
     }
   );
 
@@ -116,7 +114,7 @@ describe('BannerMessage', () => {
         testProps: { onAction, title: 'secondary button' },
       },
     ],
-  ])('correctly handles a %s prop', (key, testID, { testProps }) => {
+  ])('correctly handles a %s button prop', (key, testID, { testProps }) => {
     mockUseMessageImage.mockReturnValueOnce({
       hasRenderableImage: false,
       imageDimensions: { height: undefined, width: undefined },
--- a/packages/react-native/src/InAppMessaging/components/BannerMessage/__tests__/BannerMessage.spec.tsx
+++ b/packages/react-native/src/InAppMessaging/components/BannerMessage/__tests__/BannerMessage.spec.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import TestRenderer from 'react-test-renderer';
+import { fireEvent, render } from '@testing-library/react-native';
 
 import { IN_APP_MESSAGING_TEST_ID } from '../../../constants';
 import { useMessageImage } from '../../../hooks/useMessageImage';
@@ -27,9 +27,9 @@ describe('BannerMessage', () => {
       isImageFetching: false,
     });
 
-    const renderer = TestRenderer.create(<BannerMessage {...baseProps} />);
+    const { toJSON } = render(<BannerMessage {...baseProps} />);
 
-    expect(renderer.toJSON()).toMatchSnapshot();
+    expect(toJSON()).toMatchSnapshot();
   });
 
   it('renders a message as expected with an image', () => {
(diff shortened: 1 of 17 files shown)


packages/react-native/src/InAppMessaging/components/BannerMessage/__tests__/BannerMessage.spec.tsx
```diff
@@ -83,39 +81,60 @@ describe('BannerMessage', () => {
...
+      });
+
+      const props = { ...baseProps, [key]: testProps };
+
+      const { getByTestId } = render(<BannerMessage {...props} />);
+      const testElement = getByTestId(testID);
+
+      for (const [key, value] of Object.entries(expectedProps)) {
+        expect(testElement.props[key]).toEqual(value);
+      }
```
calebpollman, line 100: Prefer explicit assertions in unit tests over iterative assertions which require the reader to spend more time understanding the unit test. Looks like the only thing we are asserting against here is `children` in `expectedProps` so should be fairly straightforward to make the assertion explicit?

packages/react-native/src/InAppMessaging/components/BannerMessage/__tests__/BannerMessage.spec.tsx
```diff
@@ -83,39 +81,60 @@ describe('BannerMessage', () => {
...
     ],
-  ])(
-    'correctly handles a %s prop',
-    (key, testID, { testProps, expectedProps }) => {
-      mockUseMessageImage.mockReturnValueOnce({
-        hasRenderableImage: false,
-        imageDimensions: { height: undefined, width: undefined },
-        isImageFetching: false,
-      });
+  ])('correctly handles a %s prop', (key, testID, { testProps }) => {
```
calebpollman, line 119: Not sure how this got past linting since it we have a rule preventing reuse of titles (`jest/no-identical-title`, probably because the title is computed), but this could be confusing for future maintainers since we are using the same title template on line 85.

packages/react-native/src/InAppMessaging/components/CarouselMessage/__tests__/CarouselMessage.spec.tsx
```diff
@@ -1,55 +1,92 @@
 import React, { ReactElement } from 'react';
-import { act, create, ReactTestRenderer } from 'react-test-renderer';
+import { render } from '@testing-library/react-native';
 
-import { Carousel } from '../../../../primitives';
+import { IN_APP_MESSAGING_TEST_ID } from '../../../constants';
 import CarouselMessage from '../CarouselMessage';
+import { defaultStyle } from '../styles';
+import { ReactTestRendererJSON } from 'react-test-renderer';
```
calebpollman, line 7: If we are updating the unit tests to no longer use `react-test-renderer` directly than i would assume we would stop using the types as well?

packages/react-native/src/InAppMessaging/components/ModalMessage/__tests__/ModalMessage.spec.tsx
```diff
@@ -100,39 +98,60 @@ describe('ModalMessage', () => {
         expectedProps: { children: 'body content' },
       },
     ],
+  ])(
+    'correctly handles a %s prop',
```
calebpollman, line 102: Same comments as what i added in _BannerMessage.spec.tsx_

packages/react-native/src/primitives/Carousel/__tests__/Carousel.spec.tsx
```diff
@@ -7,10 +7,10 @@ import {
   Text,
   ViewToken,
 } from 'react-native';
-import { act, create, ReactTestRenderer } from 'react-test-renderer';
+import { act, render } from '@testing-library/react-native';
 
 import Carousel from '../Carousel';
-import CarouselPageIndicator from '../CarouselPageIndicator';
+import { ReactTestInstance } from 'react-test-renderer';
```
calebpollman, line 13: Same comment abt not using _react-test-renderer_ types as noted earlier

packages/react-native/src/primitives/Carousel/__tests__/CarouselPageIndicator.spec.tsx
```diff
@@ -1,71 +1,73 @@
 import React from 'react';
-import { View } from 'react-native';
-import { act, create, ReactTestRenderer } from 'react-test-renderer';
+import { render } from '@testing-library/react-native';
 
 import CarouselPageIndicator from '../CarouselPageIndicator';
+import { ReactTestRendererJSON } from 'react-test-renderer';
```
calebpollman, line 5: Same comment regarding _react-test-renderer_ as previously noted

packages/react-native/src/primitives/Carousel/types.ts
```diff
@@ -7,6 +7,7 @@ export interface CarouselProps<T> {
   onClose?: () => void;
   renderItem: ListRenderItem<T>;
   style?: StyleProp<ViewStyle>;
+  testID?: string | undefined;
```
calebpollman, line 10: Shouldn't need the `undefined` if it's optional (afaik)?

calebpollman (commented):
Thanks for doing this @ioanabrooks. Left some feedback but LGTM once addressed

packages/react-native/src/InAppMessaging/components/CarouselMessage/__tests__/CarouselMessage.spec.tsx
ioanabrooks: `react-native-testing-library` is built on top of react-test-renderer thus has a dependency on it, as noted in their [readme](https://github.com/callstack/react-native-testing-library#installation) as well.

calebpollman (approved):

joebuono (approved):
Thank you for doing this refactor!

PR #3167, opened 2022-12-14
chore: upgrade dependencies for AppSec

0618:
#### Description of changes

image

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -45,8 +45,8 @@
   "dependencies": {
     "csstype": "^3.1.1",
     "lodash": "4.17.21",
-    "style-dictionary": "3.7.0",
-    "tslib": "2.4.0"
+    "style-dictionary": "^3.7.1",
+    "tslib": "^2.4.1"
   },
   "peerDependencies": {
     "aws-amplify": ">= 5.0.1",
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -45,8 +45,8 @@
   "dependencies": {
     "csstype": "^3.1.1",
     "lodash": "4.17.21",
-    "style-dictionary": "^3.7.1",
-    "tslib": "^2.4.1"
+    "style-dictionary": "3.7.1",
+    "tslib": "2.4.1"
   },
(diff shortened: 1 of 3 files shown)


packages/ui/package.json
```diff
@@ -45,8 +45,8 @@
   "dependencies": {
     "csstype": "^3.1.1",
     "lodash": "4.17.21",
-    "style-dictionary": "3.7.0",
-    "tslib": "2.4.0"
+    "style-dictionary": "^3.7.1",
```
zchenwei, line 48: I think we should pin those dependencies and upgrade manually?

0618: Why do we want to ping the version if it doesn't have any breaking changes?

zchenwei: I think we encountered this before. Some dependencies do not follow semver strictly and introduce breaking changes for minor bump. So we pin them afterwards

0618: 2 scenarios:
A. There's a minor version to fix a security issue, but we missed it.
B. There's a minor version update, but broke our code

I'd rather have scenario B unless there's some special cases that we have to add the pkg to `resolution` to ping the version.

It's weird that AppSec didn't complain about the react pkg as you linked above. And Angular is using `^2.0.0` and `^2.3.1`

reesscot: I agree with @zchenwei here because NOT pinning versions has been shown to break customers and we are already monitoring security updates via dependabot.

zchenwei: Yea, what I mean is not breaking our code since we have `yarn.lock`. It will break customer apps

wlee221: +1, we deliberately pinned the versions to prevent regressions in direct dependencies and downstream dependencies.

0618: Ok. PR updated!

wlee221 (approved):
🚢

zchenwei (approved):

Issue #3092, opened 2022-11-27
React -- setting totpIssuer / totpUser doesn't do anything


epicfaace:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

webpack

### Please describe your bug.

I set formFields as per documentation:

```
  const formFields = {
      confirmation_code: {
        totpIssuer: "issuer",
        totpUsername: "username"
      }
    }
  };
```

### What's the expected behaviour?

It should use these custom values when making the QR code.

### Help us reproduce the bug!

1. Use react / amplify and set the above form fields.
2. The bug is apparent.

### Additional information and screenshots

I wonder if this commit (https://github.com/aws-amplify/amplify-ui/commit/2e524dc9f73dc86e011614740ed6f9f400678872#diff-c585283990ae826e5a553a05d4b1abc495c11e7f1c1cfdf215f21c1dc198a36aL51) broke it. Also, please add tests to ensure this functionality doesn't regress in the future!

ErikCH:
Hi @epicfaace ! You should use `QR` not `confirmation_code`.

It would like like this.

```
const formFields = {
  setupTOTP: {
        QR: {
          totpIssuer: 'test issuer',
          totpUsername: 'amplify_qr_test_user',
        },
      },
  }

export default function App() {
  return (
    <Authenticator formFields={formFields}>
      {({ signOut }) => <button onClick={signOut}>Sign out</button>}
    </Authenticator>
  );
}
```

https://ui.docs.amplify.aws/react/connected-components/authenticator/customization#update-setup-totp-qr-issuer-and-username

Could you try that and see if that fixes your problem?

ErikCH:
You can only set the totpIssuer when you setup the TOTP.

epicfaace:
Sorry, my sample code had a typo. I was actually using `QR`, and still had the problem of it not showing up properly. Here's the exact code we're using!

[...]

PR #3183, opened 2022-12-16
fix(MFA Issuer React): MFA Issuer is not displaying

ErikCH:
#### Description of changes
The `totpIssuer` and `totpUsername` is not displaying after being updated

#### Issue #, if available
#3092 

Added test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/examples/angular/src/pages/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.component.html
+++ b/examples/angular/src/pages/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.component.html
@@ -1,4 +1,4 @@
-<amplify-authenticator>
+<amplify-authenticator [formFields]="formFields">
   <ng-template
     amplifySlot="authenticated"
     let-user="user"
--- a/packages/e2e/cypress/integration/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.steps.ts
+++ b/packages/e2e/cypress/integration/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.steps.ts
@@ -1,4 +1,7 @@
 import { Then, When } from '@badeball/cypress-cucumber-preprocessor';
+import { Decoder } from '@nuintun/qrcode';
+
+const qrcode = new Decoder();
 
 When('I enter an invalid confirmation code', () => {
   // Wait for QR code to render before entering code
@@ -9,4 +12,13 @@ When('I enter an invalid confirmation code', () => {
 
 Then('I will be redirected to the setup totp page', () => {
   cy.get('[data-amplify-qrcode]').should('be.visible');
+});
+
+Then('I check to see if QR code is correct', () => {
+  cy.get('img').then(($el) => {
+    const imagePath = $el[0]?.currentSrc; // gives URL to image
+    return qrcode.scan(imagePath).then((decode) => {
+      expect(decode.data).to.contain('otpauth://totp/My%20Web%20App'); // whatever qr is set to
+    });
+  });
 });
--- a/packages/e2e/cypress/integration/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.steps.ts
+++ b/packages/e2e/cypress/integration/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.steps.ts
@@ -16,9 +16,9 @@ Then('I will be redirected to the setup totp page', () => {
 
 Then('I check to see if QR code is correct', () => {
   cy.get('img').then(($el) => {
-    const imagePath = $el[0]?.currentSrc; // gives URL to image
+    const imagePath = $el[0]?.currentSrc;
     return qrcode.scan(imagePath).then((decode) => {
-      expect(decode.data).to.contain('otpauth://totp/My%20Web%20App'); // whatever qr is set to
+      expect(decode.data).to.contain('otpauth://totp/My%20Web%20App');
     });
   });
 });
--- a/packages/e2e/features/ui/components/authenticator/sign-in-totp-mfa.feature
+++ b/packages/e2e/features/ui/components/authenticator/sign-in-totp-mfa.feature
@@ -12,6 +12,8 @@ Feature: Sign In with TOTP MFA
     And I type my password
     And I click the "Sign in" button
     Then I will be redirected to the setup totp page
+    Then I check to see if QR code is correct
+    Then I see the "Confirm" button
 
   @angular @react @vue
   Scenario: Redirect to sign in page
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -24,6 +24,7 @@
     "@aws-amplify/ui": "^5.3.1",
     "@badeball/cypress-cucumber-preprocessor": "^12.1.0",
     "@bahmutov/cypress-esbuild-preprocessor": "^2.1.3",
+    "@nuintun/qrcode": "^3.3.0",
     "@testing-library/cypress": "^8.0.3",
     "aws-crt": "^1.10.6",
     "cypress": "^10.7.0",
--- a/packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
+++ b/packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
@@ -34,8 +34,6 @@ const mockActorReturnActorContextValues = {
 
 const getActorContextSpy = jest.spyOn(UIModule, 'getActorContext');
 
-getActorContextSpy.mockReturnValue(mockActorReturnActorContextValues);
-
 const getSortedFormFieldsSpy = jest
   .spyOn(UIModule, 'getSortedFormFields')
   .mockImplementation(() => [['name', { required: true }]]);
--- a/packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
+++ b/packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
@@ -13,9 +13,29 @@ import {
   getComparator,
   getMachineFields,
   getTotpSecretCodeCallback,
+  getQRFields,
 } from '../utils';
+import { AuthActorContext } from '@aws-amplify/ui';
 
 const setupTOTPSpy = jest.spyOn(Auth, 'setupTOTP').mockImplementation();
+
+const totpIssuer = 'testIssuer';
+const totpUsername = 'testUsername';
+const mockActorReturnActorContextValues = {
+  formFields: {
+    setupTOTP: {
+      QR: {
+        totpIssuer,
+        totpUsername,
+      },
+    },
+  },
+} as unknown as AuthActorContext;
+
+const getActorContextSpy = jest.spyOn(UIModule, 'getActorContext');
+
+getActorContextSpy.mockReturnValue(mockActorReturnActorContextValues);
+
 const getSortedFormFieldsSpy = jest
   .spyOn(UIModule, 'getSortedFormFields')
   .mockImplementation(() => [['name', { required: true }]]);
@@ -144,5 +164,24 @@ describe('getMachineFields', () => {
 
     expect(output).toHaveLength(1);
     expect(output).toStrictEqual([{}]);
+  });
+
+  describe('getQRFields', () => {
+    const state = {} as unknown as AuthMachineState;
+
+    beforeEach(() => {
+      jest.clearAllMocks();
+    });
+
+    it('returns the correct QR issuers', () => {
+      getActorContextSpy.mockReturnValue(mockActorReturnActorContextValues);
+      const QRFields = getQRFields(state);
+      expect(QRFields).toEqual({ totpIssuer, totpUsername });
+    });
+    it('returns empty object if QR field is not present', () => {
+      getActorContextSpy.mockReturnValue({});
+      const QRFields = getQRFields(state);
+      expect(QRFields).toEqual({});
+    });
   });
 });
--- a/packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
+++ b/packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
@@ -3,6 +3,7 @@ import {
   AmplifyUser,
   AuthenticatorServiceFacade,
   AuthMachineState,
+  AuthActorContext,
 } from '@aws-amplify/ui';
 
 import * as UIModule from '@aws-amplify/ui';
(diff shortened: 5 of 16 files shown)


packages/e2e/cypress/integration/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.steps.ts
```diff
@@ -1,4 +1,7 @@
 import { Then, When } from '@badeball/cypress-cucumber-preprocessor';
+import { Decoder } from '@nuintun/qrcode';
```
ErikCH, line 2: This library is a dev dependency used to get the QR code information directly from the image

packages/react-core/src/Authenticator/hooks/useAuthenticator/types.ts
```diff
@@ -38,6 +38,7 @@ export interface UseAuthenticator extends AuthenticatorServiceFacade {
   getTotpSecretCode: () => Promise<string>;
 
   /** @deprecated For internal use only */
+  QRFields: { totpIssuer?: string; totpUsername?: string };
```
ErikCH, line 41: I made this deprecated internal, because we eventually want to handle this differently.

packages/e2e/features/ui/components/authenticator/sign-in-totp-mfa.feature
```diff
@@ -12,6 +12,8 @@ Feature: Sign In with TOTP MFA
     And I type my password
     And I click the "Sign in" button
     Then I will be redirected to the setup totp page
+    Then I check to see if QR code is correct
+    Then I see the "Confirm" button
```
ErikCH, line 16: This is well more suited to be tested in a Cypress e2e test then a unit test IMO

packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
```diff
@@ -41,9 +41,8 @@ export const SetupTOTP = ({
   const [qrCode, setQrCode] = React.useState<string>();
   const [copyTextLabel, setCopyTextLabel] = React.useState<string>('COPY');
   const [secretKey, setSecretKey] = React.useState<string>('');
-
   const { totpIssuer = 'AWSCognito', totpUsername = user?.username } =
-    (fields as LegacyQRFields)?.QR ?? {};
```
ErikCH, line 46: This isn't in the fields, since it gets deleted in the `formFields.ts` file. Otherwise it creates an unusual `QR` input field on the `SetupTOTP` page. Since the behavior of `QR` doesn't match up with our other override field logic, I created a new `QRFields` from the `useAuthenticator` hook, that retrieves this information to use.

ErikCH: It's worth noting Angular and Vue do not use this logic, and garb the field data out of the xState context directly. React Native doesn't seem to allow issuer overrides, so this should not affect it.

packages/react-core/src/Authenticator/hooks/useAuthenticator/types.ts
wlee221: Should `fields` also be `@deprecated?

ErikCH: OK, I'll update it!

ErikCH: [d681c00](https://github.com/aws-amplify/amplify-ui/pull/3183/commits/d681c00fe7843e52b7b95834c9092c64c308781d)

packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
wlee221: Curious, why doesn't current code work and whether it's a regression. If so, we should avoid adding an extra return value from `useAuthenticator`, and resolve that regression instead. 

Ideally, `fields` should be the source of truth. Is `fields` from `useAuthenticator` not getting updated properly?

examples/angular/src/pages/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.component.html
```diff
@@ -1,4 +1,4 @@
-<amplify-authenticator>
+<amplify-authenticator [formFields]="formFields">
```
reesscot, line 1: Why is this necessary? Is this required for all customers using TOT{? Could it be considered a breaking change?

packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
```diff
@@ -25,7 +25,7 @@ export const SetupTOTP = ({
   className,
   variation,
 }: RouteProps): JSX.Element => {
-  const { fields, getTotpSecretCode, isPending, user } = useAuthenticator(
+  const { getTotpSecretCode, isPending, user, QRFields } = useAuthenticator(
```
reesscot, line 28: Is this a new external facing API, or is this internal only?

reesscot (commented):
Left some questions

packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
ErikCH: This is internal only.

examples/angular/src/pages/ui/components/authenticator/sign-in-totp-mfa/sign-in-totp-mfa.component.html
ErikCH: This is the examples app. I added a new e2e test that now checks the setup TOTP QR field. So I had to add this new fields in the vue/angular/react e2e tests.

packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
ErikCH: In the original implementation Vue, React and Angular all looked up the issuer information directly from the actor context in the SetupTOTP file. 

Even though the `QR` object is passed into the `formFields` prop it really has never been used as as form field, since it really isn't one. In fact it's deleted out of the form fields as seen below.

```
export const getFormFields = (
  route: FormFieldComponents,
  state: AuthMachineState
): FormFields => {
...
  delete formFields['QR'];
  return formFields;
```
When the big `react-core` refactor was done, it was assumed it was in the fields, but this was never the case. Vue and React still pulled it directly from the context.

If we keep it in the fields, then we'll need to update the `FormFields` components for each framework and ignore the `QR` field, since it doesn't exist. The `formFields` prop is used to help generate all form fields, even those that don't exist. 

I'm up for suggestions on a better way of handling this. I noticed in the refactor a `getTotpSecretCode` was added, so I assumed it was ok to add another TOTP/QR return value.

ErikCH: @calebpollman Well QR isn't a formField, it's an image, but we kept all the logic for updating and customizing components, like dialCode, etc, in `formField`. In our next breaking change we can remove it from there.

packages/e2e/package.json
```diff
@@ -24,6 +24,7 @@
     "@aws-amplify/ui": "^5.3.1",
     "@badeball/cypress-cucumber-preprocessor": "^12.1.0",
     "@bahmutov/cypress-esbuild-preprocessor": "^2.1.3",
+    "@nuintun/qrcode": "^3.3.0",
```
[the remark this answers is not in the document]
ErikCH: @calebpollman I believe a cypress test, testing the QR code generated, and looking at the image is the best way of doing this. And since it's a dev dependency, for the Cypress e2e test, the impact is minimal.

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
```diff
@@ -13,9 +13,27 @@ import {
   getComparator,
   getMachineFields,
   getTotpSecretCodeCallback,
+  getQRFields,
 } from '../utils';
+import { AuthActorContext } from '@aws-amplify/ui';
```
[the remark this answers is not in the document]
ErikCH: I'll update it.

packages/e2e/package.json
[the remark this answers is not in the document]
calebpollman: While i appreciate that you feel that it is the best way, is there any other way to test the output of the QR code without adding a 3p dep?

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
```diff
@@ -145,4 +163,23 @@ describe('getMachineFields', () => {
...
   });
+
+  describe('getQRFields', () => {
+    const state = {} as unknown as AuthMachineState;
+
+    beforeEach(() => {
+      jest.clearAllMocks();
+    });
+
+    it('returns the correct QR issuers', () => {
```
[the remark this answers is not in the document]
ErikCH: I'll change to issuer and username.

packages/react-core/src/Authenticator/hooks/useAuthenticator/useAuthenticator.tsx
```diff
@@ -43,6 +44,12 @@ export default function useAuthenticator(
   // `fields` from updating with current form state on value changes
   const serviceSnapshot = service.getSnapshot();
 
+  // legacy `QRFields` values only used for SetupTOTP page to retrieve issuer information, will be removed in future
+  const QRFields = useMemo(
+    () => getQRFields(serviceSnapshot as AuthMachineState),
+    [serviceSnapshot]
+  );
```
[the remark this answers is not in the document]
ErikCH: Unfortunately, I don't really see an easy way of keeping it in the existing fields variable. We'd have to explicitly not delete it, then have to make changes to Vue, React and Angular so the Form Fields components won't automatically try to create a new bogus form field component. On the other hand, if we keep it in `useAuthenticator` it will only affect React, and won't require anything additional.

I'm up for more suggestions if you know of an easier way. Otherwise, we could wait until the next breaking change to fix this.

packages/react-core/src/Authenticator/hooks/useAuthenticator/utils.ts
```diff
@@ -52,6 +53,16 @@ export const getComparator =
     return areSelectorDepsEqual(currentSelectorDeps, nextSelectorDeps);
   };
 
+export const getQRFields = (
+  state: AuthMachineState
+): { totpIssuer?: string; totpUsername?: string } => {
+  const fields = getActorContext(state);
```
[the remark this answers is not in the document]
ErikCH: Well technically it's in actorContext, and it's all fields.

packages/e2e/package.json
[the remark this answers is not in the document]
ErikCH: Not that I know of.

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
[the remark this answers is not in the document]
ErikCH: [0bfdc84](https://github.com/aws-amplify/amplify-ui/pull/3183/commits/0bfdc841fa6e83628660407cec2848c7b6cf1360)

packages/react-core/src/Authenticator/hooks/useAuthenticator/__tests__/utils.spec.tsx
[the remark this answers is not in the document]
ErikCH: [0bfdc84](https://github.com/aws-amplify/amplify-ui/pull/3183/commits/0bfdc841fa6e83628660407cec2848c7b6cf1360)

packages/react-core/src/Authenticator/hooks/useAuthenticator/useAuthenticator.tsx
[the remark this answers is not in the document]
calebpollman: Why can't we just put the workaround in `SetupTOTP`?

packages/e2e/package.json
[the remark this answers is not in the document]
ErikCH: Our existing QRCode library does not decode, -> https://github.com/soldair/node-qrcode/issues/129

packages/react-core/src/Authenticator/hooks/useAuthenticator/useAuthenticator.tsx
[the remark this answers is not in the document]
ErikCH: @calebpollman The problem is how we get it into the `SetupTOTP` component? It's not in the `fields` component passed from `useAuthenticator`. That's why it doesn't update the QR image for the customer.

packages/react/src/components/Authenticator/shared/FormFields.tsx
```diff
@@ -5,9 +5,10 @@ import { useAuthenticator } from '@aws-amplify/ui-react-core';
 
 export function FormFields(): JSX.Element {
   const { fields } = useAuthenticator(({ route }) => [route]);
+  const removedQRFields = fields.filter((field) => field.name !== 'QR');
```
ErikCH, line 8: Since we add in QR fields, we need to remove it back out again, so another field isn't created.

packages/react-core/src/Authenticator/hooks/useAuthenticator/useAuthenticator.tsx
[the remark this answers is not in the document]
ErikCH: Per offline discussion, I have removed the `QRFields` from `useAuthenticator`. We are now grabbing the information in `getMachineFields` instead, and then deleting it in `FormFields` component, since it's not technically a field that needs to be displayed.

packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
ErikCH: Return value from `useAuthenticator` has been removed.

packages/react/src/components/Authenticator/shared/FormFields.tsx
ErikCH: reverted

packages/react/src/components/Authenticator/SetupTOTP/SetupTOTP.tsx
ErikCH: Reverted back, after offline discussion with @calebpollman - Since it affects React native. `useAuthenticator` is now updated with new `QRFields`

packages/react-core/src/Authenticator/hooks/useAuthenticator/useAuthenticator.tsx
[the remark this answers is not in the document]
ErikCH: Reverted - added back in `QRFields` after offline discussion with @calebpollman

calebpollman (approved):
@ErikCH Thanks for sorting through this! Appreciate the work 🚢

wlee221 (approved):

PR #3185, opened 2022-12-16
chore: fixing types of primitives/shared/utils

dbanksdesign:
#### Description of changes

Starting to pay down our loose TS

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/shared/utils.ts
+++ b/packages/react/src/primitives/shared/utils.ts
@@ -34,7 +34,7 @@ export const getConsecutiveIntArray = (
   return Array.from({ length }, (_, idx) => idx + start);
 };
 
-export type Modifiers = string | number | null | boolean;
+export type Modifiers = string | number | null | undefined;
 
 /**
  * This helper function creates modifier class names that are used for our flat BEM styling
--- a/packages/react/src/primitives/shared/utils.ts
+++ b/packages/react/src/primitives/shared/utils.ts
@@ -34,7 +34,7 @@ export const getConsecutiveIntArray = (
   return Array.from({ length }, (_, idx) => idx + start);
 };
 
-export type Modifiers = string | number | null;
+export type Modifiers = string | number | null | boolean;
 
 /**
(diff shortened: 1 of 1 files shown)


packages/react/src/primitives/shared/utils.ts
```diff
@@ -34,7 +34,7 @@ export const getConsecutiveIntArray = (
   return Array.from({ length }, (_, idx) => idx + start);
 };
 
-export type Modifiers = string | number | null;
+export type Modifiers = string | number | null | boolean;
```
reesscot, line 37: Should this allow `undefined` instead, so we can allow undefined params without having to wrap each one in Boolean?
```suggestion
export type Modifiers = string | number | null | undefined;
```

reesscot: hmm, I see you already added `?` below, but not sure why we should add `boolean`.

packages/react/src/primitives/shared/utils.ts
```diff
@@ -34,7 +34,7 @@ export const getConsecutiveIntArray = (
   return Array.from({ length }, (_, idx) => idx + start);
 };
 
-export type Modifiers = string | number | null;
+export type Modifiers = string | number | null | undefined;
```
reesscot, line 37: I think you can drop my suggestion to add `undefined` since you have `modifiers?` below

packages/react/src/primitives/shared/utils.ts
```diff
@@ -45,8 +45,8 @@ export type Modifiers = string | number | null;
  */
 export const classNameModifier = (
   base: ComponentClassName,
-  modifier: Modifiers
-): string => {
+  modifier?: Modifiers
+): string | null => {
   return modifier ? `${base}--${modifier}` : null;
```
reesscot, line 50: I'm actually think it might make more sense to return empty string here, instead of `null`
```suggestion
  return modifier ? `${base}--${modifier}` : '';
```

PR #3187, opened 2022-12-17
chore: fixing types of SliderField and StepperField

dbanksdesign:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/SliderField/SliderField.tsx
+++ b/packages/react/src/primitives/SliderField/SliderField.tsx
@@ -64,7 +64,7 @@ const SliderFieldPrimitive: Primitive<SliderFieldProps, typeof Root> = (
   );
 
   const values = isControlled ? [value] : undefined;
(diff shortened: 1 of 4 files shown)


packages/react/src/primitives/SliderField/SliderField.tsx
```diff
@@ -64,7 +64,7 @@ const SliderFieldPrimitive: Primitive<SliderFieldProps, typeof Root> = (
   );
 
   const values = isControlled ? [value] : undefined;
-  const defaultValues = !isControlled ? [defaultValue] : undefined;
+  const defaultValues = !isControlled ? [defaultValue || 0] : undefined;
```
ioanabrooks, line 67: If we set `defaultValue`'s default value to 0 at line 27 above, then we won't need this change or the one at line 82, since currentValue is derived from defaultValue or value.

Issue #3082, opened 2022-11-23
React Image component is incompatible with Next.js Image


dbanksdesign:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Next.js

### Please describe your bug.

The `Image` component is not compatible with the Next.js `Image` component. The Next.js Image component expects `width` and `height` to be defined or it will throw an error. The issue is the Amplify UI Image component thinks `width` and `height` are style props and places them in a `style` prop, which Next.js doesn't like either. 

There is a workaround, you can wrap the Next.js Image with a function and pass the props:

```jsx
<Image
  as={(props) => <NextImage width={300} height={300} {...props} />}
  src=""
/>
```

But that is not great

### What's the expected behaviour?

Ideally, I should be able to do:

```jsx
<Image
  as={NextImage}
  width={300}
  height={300
  src=""
/>
```

### Help us reproduce the bug!

https://stackblitz.com/edit/nextjs-hunf2q?file=pages%2Findex.js

### Code Snippet

```typescript
// Put your code below this line.

```

dbanksdesign:
Closing this as `wontfix` after discussion with the team. Using the Amplify UI `Image` component with the `NextImage` does not really provide any value. Instead, just use the `NextImage`, Amplify UI only add some very basic default styling but is really just an `img` tag. We have now come around to the thinking that the `as` prop is not really fit for 3p components, but really should only be used to change the HTML tag that gets rendered. The 2 main use cases for `as` referencing a 3p component is `<Image as={NextImage}>` and `<Link as={RouterLink}>` both of which do not serve much of a purpose other than styling and just complicate the typings.

PR #3194, opened 2022-12-20
fix: fix Image component incompatible with Next.js Image #3082

zchenwei:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/View/View.tsx
+++ b/packages/react/src/primitives/View/View.tsx
@@ -21,6 +21,19 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
 ) => {
   const { propStyles, nonStyleProps } = useStyles(rest, style);
 
+  // No need to do extra work on style props when it is not rendered as a HTML element
+  // e.g. <Image as={NextImage} src="" width={300} height={300} />, just let all props pass to NextImage
+  const _rest =
+    typeof as === 'string'
+      ? {
+          style: propStyles,
+          ...nonStyleProps,
+        }
+      : {
+          style,
+          ...rest,
+        };
+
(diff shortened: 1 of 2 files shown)


0618 (approved):

packages/react/src/primitives/View/View.tsx
```diff
@@ -21,15 +21,27 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
...
+  // e.g. <Image as={NextImage} src="" width={300} height={300} />, just let all props pass to NextImage
+  const _rest =
+    typeof as === 'string'
+      ? {
+          style: propStyles,
+          ...nonStyleProps,
+        }
+      : {
+          style,
+          ...rest,
```
dbanksdesign, line 34: Does this mean all our style props will not work with the NextImage?

zchenwei: Discussed in dev sync

packages/react/src/primitives/View/View.tsx
```diff
@@ -21,15 +21,27 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
 ) => {
   const { propStyles, nonStyleProps } = useStyles(rest, style);
 
+  // No need to do extra work on style props when it is not rendered as a HTML element
+  // e.g. <Image as={NextImage} src="" width={300} height={300} />, just let all props pass to NextImage
+  const _rest =
```
reesscot, line 26: I think this might need a better name. How about:
```suggestion
  const stylesAndRest =
```

Issue #2488, opened 2022-08-19
PasswordResetRequiredException does not trigger code email 


adilosa:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App / Webpack 5

### What browsers are you seeing the problem on?

Firefox, Chrome

### Please describe your bug.

After importing users to a Cognito pool, the user is in the RESET_PASSWORD state [1]. When the user tries to sign in, they are kicked to the ConfirmPasswordReset screen and asked for a code, but a code email is never sent. 

This case is not currently handled in the UI.

If the user manually presses "Resend Code" the code email is immediately sent. 

[1] https://docs.aws.amazon.com/cognito/latest/developerguide/signing-up-users-in-your-app.html#signup-confirmation-verification-overview

### What's the expected behaviour?

A code email should be sent automatically.

### Help us reproduce the bug!

1. Import a user or otherwise get them in the "Reset Password" state
2. Go to the app and attempt to sign in with the username and any password
3. Arrive at the ConfirmPasswordReset screen with no code in your email inbox
4. Press "Resend Code" and see the code email received normally

example Cognito import CSV (may not match your user pool):
```
name,given_name,family_name,middle_name,nickname,preferred_username,profile,picture,website,email,email_verified,gender,birthdate,zoneinfo,locale,phone_number,phone_number_verified,address,updated_at,cognito:mfa_enabled,cognito:username
,,,,,,,,,test@example.com,True,,,,,,False,,,False,test@example.com
```

### Additional information and screenshots

Workaround:

I modified the `handleSignIn()` function to intercept the "PasswordResetRequiredException" and manually send the code using `Auth.forgotPassword()` (the same function the "Resend Code" button calls). This effectively manually presses "Resend Code" in this case without interrupting the UI flow. 

```jsx
<Authenticator
  services={{
    async handleSignIn({ username, password }) {
[...]

PR #3201, opened 2022-12-20
fix(forced password reset): Update logic for emails sent out on forced password reset

ErikCH:
#### Description of changes
When customer imports a set of users, and the users log in for the first time, they never receive an email for them to reset their password.

#### Issue #, if available
#2488 

Added e2e test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/ui/src/machines/authenticator/actions.ts
+++ b/packages/ui/src/machines/authenticator/actions.ts
@@ -1,3 +1,4 @@
+import { Auth } from 'aws-amplify';
 import { assign, stop } from 'xstate/lib/actions';
 
 import {
@@ -153,6 +154,13 @@ export const handleBlur = assign({
     };
   },
 });
+
+export const resendCode = async (context) => {
+  const { username } = context;
+  console.log('username', username);
+
+  return Auth.forgotPassword(username);
+};
 
 /**
  * This action occurs on the entry to a state where a form submit action
--- a/packages/ui/src/machines/authenticator/actions.ts
+++ b/packages/ui/src/machines/authenticator/actions.ts
@@ -157,7 +157,6 @@ export const handleBlur = assign({
 
 export const resendCode = async (context) => {
   const { username } = context;
-  console.log('username', username);
 
   return Auth.forgotPassword(username);
 };
--- a/packages/ui/src/machines/authenticator/actions.ts
+++ b/packages/ui/src/machines/authenticator/actions.ts
@@ -158,7 +158,7 @@ export const handleBlur = assign({
 export const resendCode = async (context) => {
   const { username } = context;
 
-  return Auth.forgotPassword(username);
+  return await Auth.forgotPassword(username);
 };
 
 /**
--- a/packages/ui/src/machines/authenticator/actors/resetPassword.ts
+++ b/packages/ui/src/machines/authenticator/actors/resetPassword.ts
@@ -13,6 +13,7 @@ import {
   setFieldErrors,
   setRemoteError,
   setUsername,
+  resendCode,
 } from '../actions';
 import { defaultServices } from '../defaultServices';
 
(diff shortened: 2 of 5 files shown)


packages/ui/src/machines/authenticator/actors/resetPassword.ts
```diff
@@ -100,6 +104,39 @@ export function resetPasswordActor({ services }: ResetPasswordMachineOptions) {
...
+                  always: [
+                    {
+                      target: 'pending',
+                      cond: 'shouldAutoConfirmReset',
+                    },
+                    {
+                      target: 'valid',
+                    },
+                  ],
+                },
```
ErikCH, line 122: This always runs since `confirmResetPassword` is set to `parallel` . However it will skip sending the reset password if intent is not sent.

calebpollman: Does running an extraneous action seems return modified state? cc @wlee221

ErikCH:
Converting to draft now, will need @wlee221 to look at this, before moving forward

packages/ui/src/machines/authenticator/actors/resetPassword.ts
wlee221: It will update the state long as long as `sendUpdate` is there. 

That said, I think that this does not need to be its own state since what we're adding is more of a side effect that just runs once initially. So I think just adding action can be  sufficient. 

What do you think about adding an action:

```
actions: {
  ...
  resendCode: async (context, event) => {
    if (context.intent === 'confirmPasswordReset')
      await Auth.forgotPassword(...)
    }
  }
}
```

and then calling that when we first enter `confirmResetPassword` state?

ErikCH: I wasn't sure if it was ok, to add this as an action instead, I'll try that out though!~

ErikCH: [98dc3ff](https://github.com/aws-amplify/amplify-ui/pull/3201/commits/98dc3ff23b87b70bd18fb351d9cac58cadae8b61)

packages/ui/src/machines/authenticator/actions.ts
```diff
@@ -154,6 +155,13 @@ export const handleBlur = assign({
   },
 });
 
+export const resendCode = async (context) => {
+  const { username } = context;
+  console.log('username', username);
```
wlee221, line 160:
```suggestion
```

wlee221 (commented):
Much appreciate the changes, thanks!

Just one console.log that bled in, but lgtm otherwise.

packages/ui/src/machines/authenticator/actions.ts
ErikCH: [18cbedc](https://github.com/aws-amplify/amplify-ui/pull/3201/commits/18cbedcbc8492b2e0bb871aa6cd4d915fd653919)

wlee221 (approved):

wlee221 (dismissed):

packages/ui/src/machines/authenticator/actions.ts
```diff
@@ -154,6 +155,12 @@ export const handleBlur = assign({
   },
 });
 
+export const resendCode = async (context) => {
```
calebpollman, line 158: Why does this need to be async if there is no `await`?

ErikCH: I'll add the await here!

ErikCH: [5f8bbaa](https://github.com/aws-amplify/amplify-ui/pull/3201/commits/5f8bbaabd2d2a81c3ff5ddcbd243ffbdd5db3c86)

calebpollman (approved):

wlee221 (approved):

PR #2966, opened 2022-11-13
chore(in-app-messaging): add in-app messaging documentation

calebpollman:
#### Description of changes
Add React and React Native In-App Messaging UI feature documentation

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated
- [ ] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3202, opened 2022-12-20
fix(docs): Framework chooser link for components overview

0618:
#### Description of changes

**Issue**: on the components overview page (https://ui.docs.amplify.aws/react/components), the framework chooser 1)  add a `#` to the disabled links and 2) all the links are pointing to react (unimportant since they are disabled). 

Also The wrong links breaks the tests in links.cy.ts.

image

**Fix**: remove the link from disabled frameworks

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/components/Layout/FrameworkChooser.tsx
+++ b/docs/src/components/Layout/FrameworkChooser.tsx
@@ -26,7 +26,7 @@ const FrameworkLink = ({
     : `/${framework}`;
 
   return (
-    <Link href={isDisabled ? '#' : href} passHref>
+    <Link href={href} passHref>
       <Button
         size="small"
         as="a"
--- a/docs/src/components/Layout/FrameworkChooser.tsx
+++ b/docs/src/components/Layout/FrameworkChooser.tsx
@@ -6,29 +6,6 @@ import { FRAMEWORKS, FRAMEWORK_DISPLAY_NAMES } from '@/data/frameworks';
 import metaData from '@/data/pages.preval';
 import { FrameworkLogo } from '@/components/Logo';
 
-const FrameworkButton = ({
-  isDisabled,
-  onClick,
-  classNames,
-  framework,
-}: {
-  isDisabled: boolean;
-  onClick: React.MouseEventHandler<HTMLButtonElement>;
-  classNames: string;
-  framework: string;
-}) => (
-  <Button
-    size="small"
-    as="a"
-    className={classNames}
-    onClick={onClick}
-    isDisabled={isDisabled}
-  >
-    <FrameworkLogo framework={framework} className="docs-framework-img" />
-    {FRAMEWORK_DISPLAY_NAMES[framework]}
-  </Button>
-);
-
 interface FrameworkLinkProps extends FrameworkChooserProps {
   framework: string;
   isDisabled: boolean;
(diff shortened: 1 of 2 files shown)


docs/src/components/Layout/FrameworkChooser.tsx
```diff
@@ -26,7 +26,7 @@ const FrameworkLink = ({
     : `/${framework}`;
 
   return (
-    <Link href={isDisabled ? '#' : href} passHref>
```
reesscot, line 29: Do we know why this logic was here originally?

0618: Yes, it was added 1 month ago in this PR https://github.com/aws-amplify/amplify-ui/pull/2966/files#diff-b27004b57d34aded523c1d37576d69fd905d61aca4de4c19d7f3f0e8750a1857R29

0618: It was `<Link href={isDisabled ? '' : href} passHref>` to keep an empty `href` for disabled button. Since we don't want `href` to be empty, we don't have to differeciate disabled and enabled.

calebpollman: Why did we have that conditional there to begin with?

0618: as mentioned in the description. The conditional would make all the buttons to be `react`

calebpollman: Sure, but why did we want that behavior? Not that I'm against the change or anything, but the code change reads as an update to the UX of the site and I just want to understand the actual thoughts around the initial behavior

0618: Oh, there's no UX change. 2 reasons for this small change. 1) to make the links right, even though it's disabled. 2) to make sure the link-checker (links.cy.ts) passes

reesscot: I'm still confused about how the links will remain disabled? I think this change makes them always enabled because the disabled value on button will have no effect on the a tag.

0618: > I'm still confused about how the links will remain disabled? I think this change makes them always enabled because the disabled value on button will have no effect on the a tag. Can you please test that the links remain disabled and if you click on them nothing happens?

Good point...Those looks disabled links are clickable 😂

0618: OK! I made the link conditional. If it's disabled, there's no link around the button

calebpollman: Rather than making a new component, you can just assign a `frameworkButton` variable and conditionally wrap it with the `Link`, example:
```
const FrameworkLink = ({
  framework,
  onClick,
  isDisabled,
}: FrameworkLinkProps) => {
  const { pathname, query } = useRouter();
  const isCurrent = query.platform === framework;
  const classNames = `docs-framework-link ${isCurrent ? 'current' : ''}`;
  const href = pathname.includes(platformPath)
    ? pathname.replace(platformPath, framework)
    : `/${framework}`;

  const frameworkButton = (
    <Button
      size="small"
      as="a"
      className={classNames}
      onClick={onClick}
      isDisabled={isDisabled}
    >
      <FrameworkLogo framework={framework} className="docs-framework-img" />
      {FRAMEWORK_DISPLAY_NAMES[framework]}
    </Button>
  );

  return isDisabled ? (
    frameworkButton
  ) : (
    <Link href={href} passHref>
      {frameworkButton}
    </Link>
  );
};
  ```

0618: Good point @calebpollman . updated!

calebpollman (approved):

reesscot (approved):

PR #3204, opened 2022-12-21
chore(ui-react): lint primitives

calebpollman:
#### Description of changes
Turn on linting for React primitives from `A*` to `M*`

#### Issue #, if available
NA

Ran unit tests and linted
#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/primitives/types/autocomplete.ts
```diff
@@ -144,3 +144,26 @@ export interface UseAutocompleteProps extends Partial<AutocompleteProps> {
 
   onClick?: React.MouseEventHandler<HTMLInputElement>;
 }
+
+type SetStateAction<T> = React.Dispatch<React.SetStateAction<T>>;
+
+export interface UseAutocomplete {
```
zchenwei, line 150: Maybe we should keep this stay in the same file as `useAutocomplete.tsx`?

packages/react/src/primitives/Icon/icons/IconAdd.tsx
```diff
@@ -1,13 +1,14 @@
-import classNames from 'classnames';
 import * as React from 'react';
+import classNames from 'classnames';
```
zchenwei, line 2: We have a script to generate the icons. We need to update the script as well <- https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/scripts/generateIcons.js

packages/react/src/primitives/Icon/icons/types.ts
```diff
@@ -0,0 +1,7 @@
+import { ViewProps } from '../../types';
+
+export type InternalIcon = (
+  props: ViewProps & {
+    size?: string | number;
```
zchenwei, line 5: I think we have remove `size` prop on Icon long time ago. Why is this required?

calebpollman: There is usage of the `size` prop with the internal `Icon` components throughout our codebase

packages/react/src/primitives/types/autocomplete.ts
calebpollman: I placed it with the rest of the `AutoComplete` component types for consistency. Don't feel strongly but is there a reason not include with the rest of the types?

zchenwei: No reason but looking at the `useCheckbox`, we group its type together

packages/react/src/primitives/Icon/icons/types.ts
zchenwei: I think we should clean up the use because it takes no effect.

calebpollman: Are we certain? I see it used to handle `height` and `width` [here](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/Icon/icons/IconClose.tsx#L16-L17). That being said, think that it would be fine to make changes around `size` since it's all internal, but should probably be handled in a separate PR

packages/react/src/primitives/types/autocomplete.ts
calebpollman: That's valid. I'm more apt to move `UseCheckbox` to the `Checkbox` types though

packages/react/src/primitives/Icon/icons/IconAdd.tsx
calebpollman: Removed the script :) https://github.com/aws-amplify/amplify-ui/pull/3209

packages/react/src/primitives/Icon/icons/types.ts
zchenwei: Dang! I do not know why it was handled like this. That looks weird coz we should not lean on `size` any more and let the size scale based off parent font size

packages/react/package.json
```diff
@@ -39,7 +39,7 @@
     "dev": "tsup --watch",
     "dev:build": "tsup",
     "clean": "rimraf dist node_modules",
-    "lint": "tsc --noEmit --project tsconfig.dev.json && yarn eslint ./src --ext .js,.ts,.tsx",
+    "lint": "tsc --noEmit --project tsconfig.dev.json && eslint",
```
calebpollman, line 42: Forgot to not that this change speeds up the lint command by abt `166%` (from approximately 16 seconds to 6 seconds, YMMV)

zchenwei (approved):

reesscot (approved):
LGTM

PR #3213, opened 2022-12-22
fix: update cra app babel plugin override

reesscot:
cra app no longer using `npm install`, so updating the fix for "@babel/plugin-transform-block-scoping" to yarn compatible `resolutions`

#### Description of changes
Update cra app babel plugin override since `cra` now using `yarn install` and not `npm install` like I previously thought.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/canary/apps/react/cra-ts/package.json
+++ b/canary/apps/react/cra-ts/package.json
@@ -59,5 +59,8 @@
   ],
   "resolutions": {
     "@babel/plugin-transform-block-scoping": "7.20.5"
+  },
+  "overrides": {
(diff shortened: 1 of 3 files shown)


calebpollman (approved):

joebuono (dismissed):

wlee221 (changes requested):
need both overrides and resolutions, as `@next` test against npm and `@latest` test against yarn

wlee221 (approved):

calebpollman (approved):

PR #3232, opened 2022-12-27
chore(ci): manually run codeql when `run-codeql` label is added

wlee221:
#### Description of changes

This adds logic for manually running `codeql` when `run-codeql` label is added, just like `run-tests` label.

POC'd on this PR: 
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     # We run codeql only if it's:
     #   1) pull request not from a fork (ie. internal PR), or
-    #   2) pull request from a fork (ie. external PR) that was added "run-tests" label
+    #   2) pull request from a fork (ie. external PR) that was added "run-codeql" label
     if: |
       (github.event.pull_request.head.repo.full_name == github.repository) ||
-      (github.event.action == 'labeled' && github.event.label.name == 'run-tests')
+      (github.event.action == 'labeled' && github.event.label.name == 'run-codeql')
     permissions:
       actions: read
       contents: read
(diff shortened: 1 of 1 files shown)


zchenwei (approved):
LGTM but curious why we want to run it manually on external PRs

.github/workflows/codeql.yml
```diff
@@ -5,22 +5,40 @@ on:
...
       security-events: write
+      pull-requests: write # used to remove label
 
     strategy:
       fail-fast: false
       matrix:
         language: [javascript]
 
     steps:
+      - name: Remove run-codeql label, if applicable
```
0618, line 32: curious, why do we remove the label first instead of last?

wlee221:
> LGTM but curious why we want to run it manually on external PRs

So the problem is that codeql doesn't run on bot-generated PRs, like "Version Packages" PR. We will have to run codeql manually for that.

For any other PRs, we will still run them automatically like we do now.

0618 (dismissed):

.github/workflows/codeql.yml
```diff
@@ -5,22 +5,40 @@ on:
...
 jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
+    # We run codeql only if it's:
+    #   1) pull request not from a fork (ie. internal PR), or
+    #   2) pull request from a fork (ie. external PR) that was added "run-codeql" label
+    if: |
+      (github.event.pull_request.head.repo.full_name == github.repository) ||
+      (github.event.action == 'labeled' && github.event.label.name == 'run-codeql')
```
wlee221, line 19:
```suggestion
```

Actually, these can be removed, unlike our normal `run-tests` label logic, because this codeql doesn't use any repo secrets and we can be more relaxed here.

.github/workflows/codeql.yml
wlee221: Good question, I had it run first so that labels can be re-added if someone adds a new commit before previous run adds.

0618 (approved):

zchenwei (approved):

PR #3238, opened 2022-12-29
fix(react): Manually add overflow: clip to fix chrome regression

wlee221:
#### Description of changes

Chrome 108 has a breaking change https://developer.chrome.com/blog/overflow-replaced-elements/ on how it handles image clipping. 

This specifically breaks our customers because we do `unset: all` on `amplify-image`, which also unsets `overflow` too. This needs to be set to `clip` (see above blog). 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/src/theme/css/component/image.scss
+++ b/packages/ui/src/theme/css/component/image.scss
@@ -3,4 +3,8 @@
   max-width: var(--amplify-components-image-max-width);
   object-fit: var(--amplify-components-image-object-fit);
   object-position: var(--amplify-components-image-object-position);
+
+  // need to set overflow to clip to resolve chrome 108 breaking change:
+  // https://developer.chrome.com/blog/overflow-replaced-elements/
+  overflow: clip;
 }
--- a/.changeset/smart-masks-drop.md
+++ b/.changeset/smart-masks-drop.md
@@ -1,0 +1,7 @@
+---
+"@aws-amplify/ui": patch
+---
+
+fix(react): Manually add `overflow: clip` to remediate chrome 108 breaking change.
+
+Ref: https://developer.chrome.com/blog/overflow-replaced-elements/
--- a/packages/ui/src/theme/css/styles.scss
+++ b/packages/ui/src/theme/css/styles.scss
@@ -55,6 +55,13 @@ select {
   box-sizing: border-box; /* set box-sizing after unset above */
 }
 
+// manually set amplify-image's `overflow` property to clip to
+// resolve chrome 108 breaking change with how it handles image:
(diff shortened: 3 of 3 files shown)


packages/ui/src/theme/css/component/image.scss
```diff
@@ -3,4 +3,8 @@
   max-width: var(--amplify-components-image-max-width);
   object-fit: var(--amplify-components-image-object-fit);
   object-position: var(--amplify-components-image-object-position);
+
+  // need to set overflow to clip to resolve chrome 108 breaking change:
+  // https://developer.chrome.com/blog/overflow-replaced-elements/
+  overflow: clip;
```
calebpollman, line 9: Should this be applied to all browsers or just chrome?

wlee221: Good q, i'm researching on my own right now to see if it's an acceptable change for other browsers

wlee221: Reading through https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#values:

> `clip`: [...] Similar to hidden, the content is clipped to the element's padding box. 

as opposed to default behavior where content might be rendered outside the box. I believe this is expected, so that image dimension is consistent with whatever width and height is specified.

Not a css expert though, need confirmation from primitive folks.

Milan-Shah: @zchenwei can you help in validating across other browsers?

Milan-Shah (approved):

packages/ui/src/theme/css/component/image.scss
Milan-Shah: And @0618

wlee221: Just confirmed on firefox that this setting does not affect the rendered result: 

https://user-images.githubusercontent.com/43682783/209990035-23b17e6d-147a-4372-9acb-09556ae89f42.mp4

Testing with safari now, but I'm good with the fix now.

wlee221: Just confirmed with Safari too

0618 (dismissed):

packages/ui/src/theme/css/component/image.scss
zchenwei: Validating now

zchenwei (approved):

ioanabrooks (approved):

zchenwei (approved):

PR #2875, opened 2022-11-01
fix(angular,vue): align PasswordField primitives

wlee221:
### Description of changes

This aligns Vue and Angular `PasswordField` primitives with the React counterpart.

#### General Changes
- Use `ComponentClassName` from `ui` package instead of hardcoding class names
- Unit tests to catch regressions
- Add `role: switch` to toggle buttons to match accessibility requirements in the React counterpart.

#### Angular changes
- Fix a bug where its aria-label wasn't toggled correctly

#### Vue changes
- Align prop type and their default values with React counterpart 

### Description of how you validated changes

Bunch of unit tests 😈 

### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3255, opened 2023-01-04
test(angular): Add `@testing-library/angular`

wlee221:
#### Description of changes

We were previously blocked on adding `dom-testing-library` to Angular because we were on unsupported version of Angular. Now that we've bumped to Angular 12, we can use `@testing-library/angular`.

Added snapshot test for all primitives while I was here. Other tests will follow in later PRs; purpose of this PR is to get testing library set up with some active tests running.

#### Issue #, if available

will unblock #2875 

CI

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/angular/projects/ui-angular/src/lib/primitives/button/__tests__/button.component.test.ts
```diff
@@ -0,0 +1,16 @@
...
+describe('amplify-button', () => {
+  it('renders as expected', async () => {
+    const { container } = await render(
+      '<button amplify-button>My Button</button>',
+      {
+        declarations: [ButtonComponent],
+        componentProperties: { label: 'Password', fieldId: 'mockId' },
+      }
+    );
+    expect(container).toMatchSnapshot();
```
ioanabrooks, line 14: Snapshot tests are a great start, but we should also verify that the button is rendered with the correct label
perhaps using `screen.getByText` query. Same comment for the other tests, we could add more assertions.

wlee221: Right, I 100% agree. However I wanted to scope this PR just for setting up `@testing-library/angular` (we have been blocked on this for a while). Then I will prepare another PR, across Angular and *Vue*, to establish those tests and also align any visual differences (e.g. #2875 that is WIP). 

Does that sound good to you?

ioanabrooks (approved):

reesscot (approved):

Issue #3256, opened 2023-01-04
Fix Types For Tab Children


bestickley:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Primitive components

### How is your app built?

Vite

### What browsers are you seeing the problem on?

Chrome

### Please describe your bug.

See below TS error:
![image](https://user-images.githubusercontent.com/35735118/210607232-38026f5a-02b0-4368-b352-9442e43e2f55.png)

### What's the expected behaviour?

`<Tabs>` should accept `null` as child as documented [here](https://github.com/aws-amplify/amplify-ui/pull/1117#discussion_r784169204)

### Help us reproduce the bug!

Conditionally render `<TabItem>` within `<Tab>` and see TS error.

### Code Snippet

```typescript
let RequestTabItem: ReactElement | null = null;
  if (canViewRequests) {
    RequestTabItem = (
      <TabItem
        title={<RequestsBadge count={length} loading={isFetching} />}
        onClick={() => navigate("requests")}
      />
    );
  }
  return (
    <>
      <Tabs currentIndex={tab} onChange={handleChange} marginBottom="small">
        <TabItem title="Members" onClick={() => navigate("members")} />
        
        {RequestTabItem}
      </Tabs>
```

### Additional information and screenshots

Amplify UI does correctly not log warning message when `null`. This is just a TS issue.

reesscot:
@bestickley Thanks for reporting this, we'll get it fixed!

dbanksdesign:
This should be fixed in #3270 and go out in the next release!

reesscot:
Fixed by https://github.com/aws-amplify/amplify-ui/pull/3270

dandrei:
This seems to still be an issue in the React package? `@aws-amplify/ui-react 6.1.12`:

    TS2307: Cannot find module
    @aws-amplify/ui-react/dist/types/primitives/Tabs/TabsItem
[...]

Issue #1018, opened 2021-12-18
TypeScript type errors from `@aws-amplify/ui-react`


moltar:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### Please describe your bug.

When running TS build on the project, I am getting lots of errors:

```
tsc --noEmit
```

```
node_modules/@aws-amplify/ui-react/dist/index.d.ts:29:9 - error TS2502: 'FormFields' is referenced directly or indirectly in its own type annotation.

29     var FormFields: typeof FormFields;
           ~~~~~~~~~~

node_modules/@aws-amplify/ui-react/dist/index.d.ts:224:9 - error TS2502: 'SignIn' is referenced directly or indirectly in its own type annotation.

224     var SignIn: typeof SignIn;
            ~~~~~~

node_modules/@aws-amplify/ui-react/dist/index.d.ts:225:9 - error TS2502: 'SignUp' is referenced directly or indirectly in its own type annotation.

225     var SignUp: typeof SignUp;
            ~~~~~~

node_modules/@aws-amplify/ui-react/dist/index.d.ts:5775:48 - error TS2693: 'DataStoreItemProps' only refers to a type, but is being used as a value here.

5775 declare const index_DataStoreItemProps: typeof DataStoreItemProps;
                                                    ~~~~~~~~~~~~~~~~~~

node_modules/@aws-amplify/ui-react/dist/index.d.ts:5776:54 - error TS2693: 'DataStoreCollectionProps' only refers to a type, but is being used as a value here.

5776 declare const index_DataStoreCollectionProps: typeof DataStoreCollectionProps;
                                                          ~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@aws-amplify/ui-react/dist/index.d.ts:5777:49 - error TS2693: 'DataStoreItemResult' only refers to a type, but is being used as a value here.

5777 declare const index_DataStoreItemResult: typeof DataStoreItemResult;
                                                     ~~~~~~~~~~~~~~~~~~~

[...]

PR #3264, opened 2023-01-04
chore: more TS strict fixes

dbanksdesign:
#### Description of changes

* Some minor things to destructuring props
* Making TS happy in the splitPrimitiveProps function

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Menu/MenuButton.tsx
+++ b/packages/react/src/primitives/Menu/MenuButton.tsx
@@ -23,7 +23,7 @@ export const MenuButton = React.forwardRef<
       isDisabled,
       isLoading,
       size,
-      style,
+      style = {},
       type = 'button',
       variation,
       testId,
--- a/packages/react/src/primitives/Menu/MenuButton.tsx
+++ b/packages/react/src/primitives/Menu/MenuButton.tsx
@@ -23,7 +23,7 @@ export const MenuButton = React.forwardRef<
       isDisabled,
       isLoading,
       size,
-      style = {},
+      style,
       type = 'button',
       variation,
       testId,
--- a/packages/react/src/primitives/shared/styleUtils.ts
+++ b/packages/react/src/primitives/shared/styleUtils.ts
@@ -135,7 +135,7 @@ export const convertStylePropsToStyleObj: ConvertStylePropsToStyleObj = ({
 
 export const useStyles = (
   props: ViewProps,
-  style: React.CSSProperties
+  style?: React.CSSProperties
 ): {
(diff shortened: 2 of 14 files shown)


packages/react/src/primitives/Menu/MenuButton.tsx
```diff
@@ -23,7 +23,7 @@ export const MenuButton = React.forwardRef<
       isDisabled,
       isLoading,
       size,
-      style,
+      style = {},
```
calebpollman, line 26: As `style` is used in the dep array for the memoized return value of `useStyles` (see [source](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/shared/styleUtils.ts#L164)), we should avoid setting a default parameter of an empty object here. The empty object will have a new reference on every render, which causes a re-evaluation of the aforementioned `useMemo` in `useStyles`

reesscot:
Related to https://github.com/aws-amplify/amplify-ui/issues/1018

packages/react/src/primitives/shared/styleUtils.ts
```diff
@@ -167,31 +167,47 @@ export const useStyles = (
...
-  rest: Omit<
-    PrimitiveProps,
-    keyof FlexContainerStyleProps | keyof BaseStyleProps
+  rest: Partial<
+    Omit<PrimitiveProps, keyof FlexContainerStyleProps | keyof BaseStyleProps>
   >;
 }
 
+function isStyleKey(prop: any): prop is keyof AllStyleProps {
+  if (prop in ComponentPropsToStylePropsMap) {
```
reesscot, line 176: Would just returning `prop in ComponentPropsToStylePropsMap` also work? Seems like that should already return a boolean.

packages/react/src/primitives/types/tabs.ts
```diff
@@ -60,4 +60,10 @@ export interface TabItemProps extends BaseComponentProps, BaseStyleProps {
    * Make a Tab not clickable and its content not visible to the user. Defaults to false (i.e., visible).
    */
   isDisabled?: boolean;
+
+  /**
+   * @description
+   * Accepts any number of TabItem components
+   */
+  children?: React.ReactNode;
```
reesscot, line 68: Does this fix this issue? https://github.com/aws-amplify/amplify-ui/issues/3256

wlee221 (approved):

ErikCH (approved):

PR #3268, opened 2023-01-05
chore: setup release for ui-react/v3

wlee221:
#### Description of changes

Sets up release to `@aws-amplify/ui-react@v3-latest`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


package.json
```diff
@@ -24,8 +24,8 @@
     "publish:next": "yarn changeset publish --tag next",
     "version:studio": "yarn changeset version --snapshot studio-$(git rev-parse --short=7 HEAD) && yarn angular build",
     "publish:studio": "yarn changeset publish --tag studio",
-    "version:latest": "yarn changeset version && yarn angular build",
-    "publish:latest": "yarn changeset publish"
+    "version:v3": "yarn changeset version && yarn angular build",
+    "publish:v3": "yarn changeset publish --tag v3-latest"
```
wlee221, line 28: Set the dist-tag to `v3-latest`.

Wanted to do `v3` tag first, but NPM does not allow dist tags named that can be read as semver versions.

.github/workflows/publish-latest-v3.yml
```diff
@@ -0,0 +1,85 @@
+name: Publish to latest / v3
```
reesscot, line 1: This shouldn't actually be publishing to latest though, right?
```suggestion
name: Publish to v3
```

.github/workflows/publish-latest-v3.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Checkout repo
+        uses: actions/checkout@v3
+        with:
+          # This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits
+          fetch-depth: 0
+      - name: Check if repo has unconsumed changesets
+        id: has-changesets
+        uses: ./.github/actions/has-changesets
+
+  # build-test:
```
reesscot, line 25: Will we run these tests manually?

.github/workflows/publish-latest-v3.yml
```diff
@@ -0,0 +1,85 @@
...
+  #         key: ${{ runner.os }}-canary-environments-${{ steps.cli-version.outputs.version }}-${{ hashFiles('canary/environments/**/amplify/**') }}
+  #     - name: Pull down AWS environments
+  #       if: steps.environments-cache.outputs.cache-hit != 'true'
+  #       run: yarn pull
+  #       env:
+  #         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  #         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  #       working-directory: ./canary
+  #     - name: Setup canary apps against @next
+  #       run: yarn setup:next
```
reesscot, line 52: couldn't we target the v2 tag instead?

.github/workflows/publish-latest-v3.yml
wlee221: You're right, adjusting

.github/workflows/test-prs.yml
```diff
@@ -40,7 +40,7 @@ jobs:
             const label = 'run-tests';
             github.issues.removeLabel({ owner, repo, issue_number, name: label });
   test:
-    uses: aws-amplify/amplify-ui/.github/workflows/reusable-e2e.yml@main
+    uses: ./.github/workflows/reusable-e2e.yml
```
reesscot, line 43: unrelated to this pr, but do we need to make the same change on main too?

package.json
reesscot: hmm, that makes sense

.github/workflows/publish-latest-v3.yml
wlee221: The problem is that build-test needs to happen after v3 tag release. We could either:
1. run build-test after publish, and do a quick follow up if anything goes wrong, or
2. have an `unstable-v3` tag that we can test with before the release, or
3. we run build tests manually

reesscot: Since this is a one-off, I'm ok with number 3

.github/workflows/test-prs.yml
wlee221: I think that'd need some more scrutiny, since then any branch could run their local workflow.

Let's keep this for now, let's follow up internally. (fyi, this change is safe on this branch, because this is scoped to this branch only)

reesscot (approved):

ioanabrooks (approved):

.github/workflows/publish-latest-v3.yml
0618: Just to clarify. The step would be:

1) run this publish workflow for every commit on the `ui-react/v3`
2) once the publish workflow is done, manually run the build-test

If that's right, would it be hard to remember to run the test? (I, with poor memory, would prefer doing the test automatically 🥴  )

wlee221: Oh sorry, merged before I saw the comment.

1 is a bit different, we'll have a version packages PR for v3. 

1. "Version Packages" PR is prepared.
2. Engineer does build tests
3. Engineer merges the "Version Packages" PR

PR #2768, opened 2022-10-20
ci(next-release): setup next-release build tests

wlee221:
#### Description of changes

This sets up next-release build tests.  
- Adjusted `changeCanaryPackageVersion.mjs` script to accept desired dist-tag as CLI argument
- Added `next-release` specific setup scripts to `canary/package.json`

#### Description of how you validated changes
POC running on #2770:

[POC / Publish / next-release](https://github.com/aws-amplify/amplify-ui/actions/workflows/publish-next-release.yml)

https://user-images.githubusercontent.com/43682783/197032412-5edfd2dc-c80f-444a-b61a-73832b7eed8e.mp4

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3273, opened 2023-01-05
ci(v3): enable build tests against `v3-next`

wlee221:
#### Description of changes

This is a follow up to https://github.com/aws-amplify/amplify-ui/pull/3268. While I said that we could run build tests manually, this is not true. We still need some unstable npm tag to test against before we publish.

This PR enables publishes to `v3-next` on every merge to `ui-react/v3` (just like we do for `@next`) along with build testing. Publishing to latest will require build tests as well. 

(cc @0618, I answered incorrectly to your comment).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.github/workflows/publish-next-v3.yml
```diff
@@ -155,34 +155,34 @@ jobs:
         run: yarn build
         working-directory: ./canary
 
-  log-failure-metric:
```
wlee221, line 158: Disabled logs, since these are one-off hotfixes.

canary/script/changeCanaryPackageVersion.mjs
```diff
@@ -7,6 +7,13 @@ import { globbyStream } from 'globby';
 import fs from 'fs';
```
wlee221, line 7: This change is cherry-picked from https://github.com/aws-amplify/amplify-ui/pull/2768, which enables canary to target any dist-tag. No code changes were made.

canary/package.json
```diff
@@ -7,9 +7,13 @@
     "install": "yarn react:install && yarn vue:install && yarn angular:install",
     "pull": "./environments/pull-environments.sh",
     "script:install": "yarn --cwd script",
-    "setup:next": "yarn script:install && yarn updatePackage",
+    "setup:next": "yarn script:install && yarn updatePackage:next",
+    "setup:next-release": "yarn script:install && yarn updatePackage:next-release",
```
wlee221, line 11: These changes are cherry-picked from https://github.com/aws-amplify/amplify-ui/pull/2768

0618 (approved):
So publishing to v3 will be the same process as publishing to the latest?

canary/package.json
```diff
@@ -7,9 +7,13 @@
...
+    "setup:v3-next": "yarn script:install && yarn updatePackage:v3-next",
     "test": "yarn --cwd e2e test",
-    "updatePackage": "node script/changeCanaryPackageVersion.mjs",
+    "updatePackage:next": "node script/changeCanaryPackageVersion.mjs next",
+    "updatePackage:next-release": "node script/changeCanaryPackageVersion.mjs next-release",
+    "updatePackage:v3-next": "node script/changeCanaryPackageVersion.mjs v3-next",
```
reesscot, line 16: in the future, should we just take the parameter in the script instead of having special scripts for each tag?

wlee221:
> So publishing to v3 will be the same process as publishing to the latest?

Correct.

canary/package.json
wlee221: The script does, but one nice thing about having these special scripts is that there is less room for typo then using the script directly

wlee221: That said, open to thoughts! (will be a separate PR though)

reesscot (approved):

wlee221:
I'll force merge this, since these are CI changes that won't take effect until they are merged.

Issue #2632, opened 2022-09-22
Add type support for inert attribute


hbuchel:
### On which framework/platform would you like to see this feature implemented?

React

### Which UI component is this feature-request for?

Other

### Please describe your feature-request in detail.

Make TypeScript stop complaining about the `inert` HTML attribute.

**Note:** I've found (via [related thread](https://github.com/WICG/inert/issues/58) on WICG repo) I need to set it using a conditional statement like:
```
<View inert={inertCondition ? '' : null}></ View>
``` 
instead of just 
```
<View inert></View>
```

See Stackblitz example: https://stackblitz.com/edit/react-ts-xdqru3?file=App.tsx
https://caniuse.com/?search=inert 

### We love contributors! Is this something you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

PR #3277, opened 2023-01-06
feat(primitives): support inert prop

dbanksdesign:
#### Description of changes

Adds support for the HTML `inert` attribute: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

React elements do not support the prop by default, so adding it in to make it work like any other HTML attribute. 

#### Issue #, if available

#2632

Added unit test and also tried it on the docs site:
CleanShot 2023-01-06 at 13 22 42@2x

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/View/View.tsx
+++ b/packages/react/src/primitives/View/View.tsx
@@ -15,6 +15,7 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
     ariaLabel,
     isDisabled,
     style,
+    inert,
     ...rest
   }: PrimitivePropsWithRef<ViewProps, Element>,
   ref?: React.ForwardedRef<HTMLElementType<Element>>
@@ -28,6 +29,7 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
       'data-testid': testId,
       disabled: isDisabled,
       ref,
+      inert: inert ? '' : null,
       style: propStyles,
       ...nonStyleProps,
     },
--- a/packages/react/src/primitives/types/base.ts
+++ b/packages/react/src/primitives/types/base.ts
@@ -25,7 +25,7 @@ export interface BaseComponentProps {
    * Support for the HTML `inert` attribute
    * https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert
    */
-  inert?: string | null;
+  inert?: boolean;
 }
 
 export interface AriaProps {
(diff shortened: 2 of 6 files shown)


packages/react/src/primitives/types/base.ts
```diff
@@ -19,6 +19,13 @@ export interface BaseComponentProps {
    * Used to provide a `data-testid` attribute for testing purposes
    */
   testId?: string;
+
+  /**
+   * @description
+   * Support for the HTML `inert` attribute
+   * https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert
+   */
+  inert?: string | null;
```
zchenwei, line 28: Should be boolean?

zchenwei: Then you can also `inert: inert ? '' : null,`

hbuchel: Yeah I'm wondering if the type as `string | null` might be confusing to developers who aren't familiar with its implementation (i.e. that gh issue I linked in the feature request)?

dbanksdesign: yea that makes sense, updated to boolean

packages/react/src/primitives/View/View.tsx
```diff
@@ -28,6 +29,7 @@ const ViewPrimitive = <Element extends ElementType = 'div'>(
       'data-testid': testId,
       disabled: isDisabled,
       ref,
+      inert: inert ? '' : null,
```
0618, line 32: What about this?

```suggestion
           ...(inert && { inert: "" }),
```

0618: also, if the type of `inert` is `boolean`, should it be boolean here?

`      ...(true && { inert: true }),`

zchenwei: I guess if you use boolean value here React will render `inert=true/false` but we just want `inert` on the element

zchenwei (approved):

hbuchel (approved):

packages/react/src/primitives/View/View.tsx
dbanksdesign: This is actually a React issue, we can't pass the boolean `true` here or it won't actually get rendered in the DOM. See this quick example:

https://codesandbox.io/s/crazy-wu-02mz2p?file=/src/App.js

dbanksdesign: Heather also documented this in the issue (Thanks Heather!) https://github.com/aws-amplify/amplify-ui/issues/2632

PR #3280, opened 2023-01-10
fix(primitives): remove `checked` prop from textfield

0618:
#### Description of changes

Remove `checked` prop from TextField, PasswordField, PhoneNumberField, StepperField, SliderField , SwitchField and SearchField.

Verified on local and by snapshot.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/types/textField.ts
+++ b/packages/react/src/primitives/types/textField.ts
@@ -36,4 +36,6 @@ export interface TextFieldOptions extends FieldProps, FlexContainerStyleProps {
   type?: InputProps['type'];
 }
 
-export interface TextFieldProps extends TextFieldOptions, InputProps {}
+export interface TextFieldProps
+  extends TextFieldOptions,
+    Omit<InputProps, 'checked'> {}
--- a/.changeset/sour-moose-bathe.md
+++ b/.changeset/sour-moose-bathe.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ui-react': patch
+---
+
+Remove `checked` props from TextField.
--- a/docs/src/data/test/__snapshots__/props-table.test.ts.snap
+++ b/docs/src/data/test/__snapshots__/props-table.test.ts.snap
@@ -368,13 +368,6 @@ exports[`Props Table 1`] = `
                         \\"category\\": \\"InputProps\\",
                         \\"isOptional\\": true
                     },
-                    \\"checked\\": {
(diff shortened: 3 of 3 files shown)


packages/react/src/primitives/types/textField.ts
```diff
@@ -36,4 +36,6 @@ export interface TextFieldOptions extends FieldProps, FlexContainerStyleProps {
   type?: InputProps['type'];
 }
 
-export interface TextFieldProps extends TextFieldOptions, InputProps {}
+export interface TextFieldProps
+  extends TextFieldOptions,
+    Omit<InputProps, 'checked'> {}
```
calebpollman, line 41: Not entirely sure we can remove this without considering it a breaking change :(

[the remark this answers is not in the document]
dbanksdesign: This is a good point, I think we might need to rethink the `InputProps` type, it might be doing too much. Maybe moving things out to `FieldProps` which are shared across all *Field components.

dbanksdesign: Technically yes... but maybe not in the spirit of the law? This was a bug that we were exposing this. But realistically it might NOT be a breaking change because.... the interface for an HTMLInputElement coming from React types includes the `checked` attribute and because we are extending from the element's type, this won't actually throw any TS errors. 

CleanShot 2023-01-10 at 09 47 54@2x

CleanShot 2023-01-10 at 09 48 29@2x

CleanShot 2023-01-10 at 09 45 57@2x

So maybe this change (or a similar change, see @ioanabrooks's comment) is what we want. We remove it from our types and documentation, but won't break anyone's build.

dbanksdesign (commented):
Responding to comments

packages/react/src/primitives/types/textField.ts
0618: It should be fine because `TextField` never have to use `checked` and can't use `checked`

calebpollman: Are we absolutely certain that there is no customer impact? Is there potential that a user erroneously added the `checked` prop to one of the impacted UI components and that they will see type errors when it is removed?

reesscot:
Let's hold on this change till we do our next major version bump

zchenwei:
Lets close it out since it requires more discussion on whether we should filter out HTML attributes.

PR #3291, opened 2023-01-11
chore(ui,ui-react-core): Add type guard utils

ioanabrooks:
#### Description of changes

This change adds the following type guard utils:
* `isObject`
* `isString`
* `isUndefined`

in an effort to make some progress towards removing `lodash` dependency.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react-core/src/utils/index.ts
+++ b/packages/react-core/src/utils/index.ts
@@ -1,6 +1,6 @@
 import isEmpty from 'lodash/isEmpty';
-import isObject from 'lodash/isObject';
-import isString from 'lodash/isString';
+
+import { isObject, isString } from '@aws-amplify/ui';
 
 function isEmptyArray<T>(value: T): boolean {
   return Array.isArray(value) && isEmpty(value);
@@ -11,7 +11,7 @@ export function areEmptyArrays<T>(...values: T[]): boolean {
 }
 
 function isEmptyObject<T>(value: T): boolean {
-  return isObject(value) && !Array.isArray(value) && isEmpty(value);
+  return isObject(value) && isEmpty(value);
 }
 
 export function areEmptyObjects<T>(...values: T[]): boolean {
--- a/packages/ui/src/helpers/typeGuards.ts
+++ b/packages/ui/src/helpers/typeGuards.ts
@@ -1,0 +1,26 @@
+/**
+ * Checks if `value` is an Object (excludes arrays and functions)
+ *
+ * @param value
+ * @returns
+ */
+export const isObject = (value: unknown): value is object =>
+  value != null && !Array.isArray(value) && typeof value === 'object';
+
+/**
+ * Checks if `value` is a string
+ *
+ * @param value
+ * @returns
+ */
+export const isString = (value: unknown): value is string =>
+  typeof value === 'string';
+
+/**
+ * Checks if `value` is undefined
+ *
+ * @param value
+ * @returns
+ */
+export const isUndefined = <T>(value: T | undefined): value is T =>
+  value === undefined;
--- a/.changeset/sixty-buttons-study.md
+++ b/.changeset/sixty-buttons-study.md
@@ -1,0 +1,6 @@
+---
+'@aws-amplify/ui-react-core': patch
+'@aws-amplify/ui': patch
+---
+
+chore(ui,ui-react-core): Add type guard utils
--- a/packages/react-core/src/Authenticator/hooks/useAuthenticator/utils.ts
+++ b/packages/react-core/src/Authenticator/hooks/useAuthenticator/utils.ts
@@ -7,8 +7,8 @@ import {
   getSortedFormFields,
   UnverifiedContactMethods,
   getActorContext,
+  isString,
 } from '@aws-amplify/ui';
-import isString from 'lodash/isString';
 
 import { areEmptyArrays, areEmptyObjects } from '../../../utils';
 import { AuthenticatorLegacyField, AuthenticatorLegacyFields } from '../types';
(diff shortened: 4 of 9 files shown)


packages/ui/src/helpers/typeGuards.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Checks if `value` is an Object (excludes arrays and functions)
+ *
+ * @param value
+ * @returns
```
calebpollman, line 5: Are the empty `@returns` in this file intentional 😄

packages/ui/src/helpers/typeGuards.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export const isObject = (value: unknown): value is object =>
+  value != null && !Array.isArray(value) && typeof value === 'object';
```
calebpollman, line 8: Should we align this closer to the [`lodash` implementation](https://github.com/lodash/lodash/blob/master/isObject.js) since that is what is replacing? Same question with `isString` ([`lodash` implementation](https://github.com/lodash/lodash/blob/master/isString.js))

packages/react-core/src/utils/index.ts
```diff
@@ -1,6 +1,6 @@
 import isEmpty from 'lodash/isEmpty';
```
0618, line 1: We don't do `isEmpty` this time?

0618 (approved):
LGTM

packages/ui/src/helpers/typeGuards.ts
```diff
@@ -0,0 +1,26 @@
...
+export const isString = (value: unknown): value is string =>
+  typeof value === 'string';
+
+/**
+ * Checks if `value` is undefined
+ *
+ * @param value
+ * @returns
+ */
+export const isUndefined = <T>(value: T | undefined): value is T =>
```
calebpollman, line 25: Maybe I am misunderstanding the usage of the generic, but it seems to me we would want the return type of this type predicate to return `value is undefined`?

calebpollman (commented):
Thanks for doing this! 

What do you think of moving these in to a top level _utils_ folder and putting them in to their own files? Currently the _helpers_ folder is coupled to connected component utils/helpers. IMO it would make sense to keep these in their own folder since they are more generic, and to leave _helpers_ as is to keep a separation of concerns (in particular since the connected component helpers/utils are most likely not living in the UI package long term)

packages/ui/src/helpers/typeGuards.ts
wlee221: plus one, for `isString` especially, because that one will fail if string is initialized through `new String('hi!')`.

This is not at all common practice, but IMO we might as well keep it safe with possible edge cases that lodash has considered.

ioanabrooks: The intent wasn't to replace the lodash function 1:1 but more cover the usecases we have based on current usage. But I hear you..

packages/ui/src/helpers/typeGuards.ts
zchenwei: Maybe we can use `unknown` instead? Not sure if the generic type is necessary

calebpollman: Missed on my first review but after seeing @zchenwei callout regarding the generic, not sure how well TS supports using generics with arrow functions and once we add stricter linting on for the UI package we will most likely see errors here (also breaks prettier i guess?):
image

dbanksdesign (dismissed):
LGTM! Tentative approval for @calebpollman's comment on TS strict not liking `isUndefined`

packages/ui/src/helpers/typeGuards.ts
calebpollman: Totally agree with that sentiment for future utils, but have a couple concerns with doing that here:
-  these are replacing `lodash` utils in prod code, if we miss an edge case that `lodash` has accounted for it's our customers that will end up with issues
- future maintainers may not know that these `lodash`-like utils do not cover the same cases as the actual `lodash` versions and may make assumptions based on that

packages/react-core/src/utils/index.ts
ioanabrooks: Was intending of doing `isEmpty`, `isNil` and `isEqual` as a group next.

packages/ui/src/helpers/typeGuards.ts
ioanabrooks: Gave it some thought and still feel like these should be tailored to our use cases.
*  The edge cases missing are: `isObject` excludes arrays and functions, `isString` excludes primitive object wrapper which is discouraged to use in the first place. Could you give an example of where these could be an issue in prod code? 
*  Updated jsdoc for each util fn for future maintainers, furthermore unit tests explicitly cover scenarios as fn and array not being objects for the `isObject` util, and string obj wrapper not being a string for purposes of the `isString` util.

packages/ui/src/helpers/typeGuards.ts
ioanabrooks: Good call, no need for generics, updated!

dbanksdesign (dismissed):
LGTM! I'm fine either way about the `isObject` being the exact lodash method or not.

0618 (dismissed):

packages/ui/src/helpers/typeGuards.ts
calebpollman: > The edge cases missing are: isObject excludes arrays and functions, isString excludes primitive object wrapper which is discouraged to use in the first place. Could you give an example of where these could be an issue in prod code?

Since we are a library, we can't assume to know how customers are using our code and whether a subtle change in a utility can cause issues. Additionally just because using `String()` is discouraged, that doesn't mean it won't happen.

calebpollman (approved):
🚢 🚢 🚢

0618 (approved):

PR #3294, opened 2023-01-11
chore(primitives): fix ts strict errors in checkbox, collection, expander

reesscot:
#### Description of changes
Fixes TS Strict errors in `Checkbox`, `Collection` and `Expander`.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/__tests__/__snapshots__/exports.ts.snap
+++ b/packages/react/__tests__/__snapshots__/exports.ts.snap
@@ -5041,9 +5041,6 @@ Object {
       "ariaValuetext": Object {
         "type": "string",
       },
-      "asChild": Object {
-        "type": "boolean",
-      },
       "autoCapitalize": Object {
         "type": "string",
       },
@@ -5112,9 +5109,6 @@ Object {
       },
       "dir": Object {
         "type": "string",
-      },
-      "disabled": Object {
-        "type": "boolean",
       },
       "display": Object {
         "type": "string",
--- a/packages/react/src/primitives/Checkbox/Checkbox.tsx
+++ b/packages/react/src/primitives/Checkbox/Checkbox.tsx
@@ -50,9 +50,11 @@ const CheckboxPrimitive: Primitive<CheckboxProps, 'input'> = (
   const dataId = useStableId();
   React.useEffect(() => {
     const input = document.querySelector(`[data-id="${dataId}"]`);
-    // HTMLInputElement does not have an `indeterminate` attribute
-    (input as HTMLInputElement & { indeterminate: boolean }).indeterminate =
-      isIndeterminate;
+    if (input && isIndeterminate !== undefined) {
+      // HTMLInputElement does not have an `indeterminate` attribute
+      (input as HTMLInputElement & { indeterminate: boolean }).indeterminate =
+        isIndeterminate;
+    }
   }, [dataId, isIndeterminate]);
 
   const buttonTestId = getTestId(testId, ComponentClassNames.CheckboxButton);
--- a/packages/react/src/primitives/Checkbox/Checkbox.tsx
+++ b/packages/react/src/primitives/Checkbox/Checkbox.tsx
@@ -8,11 +8,11 @@ import { Text } from '../Text';
 import { VisuallyHidden } from '../VisuallyHidden';
 import { CheckboxProps } from '../types/checkbox';
 import { Primitive } from '../types/view';
-import { getTestId } from '../utils/testUtils';
+import { getTestId } from '../utils/getTestId';
 import { useStableId } from '../utils/useStableId';
 import { useCheckbox } from './useCheckbox';
 import { ComponentClassNames } from '../shared/constants';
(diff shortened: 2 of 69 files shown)


packages/react/src/primitives/Expander/Expander.tsx
```diff
@@ -9,7 +9,7 @@ import { ExpanderProps } from '../types/expander';
 import { Primitive } from '../types/view';
 import { splitPrimitiveProps } from '../shared/styleUtils';
 
-const ExpanderPrimitive: Primitive<ExpanderProps, typeof Root> = (
+const ExpanderPrimitive: Primitive<ExpanderProps, 'div'> = (
```
reesscot, line 12: This makes the Expander html element type more explicit. Before TS was having trouble finding the type, which was causing a TS strict error on the ref param.

zchenwei (approved):

zchenwei (approved):

packages/react/__tests__/__snapshots__/exports.ts.snap
```diff
@@ -5041,9 +5041,6 @@ Object {
       "ariaValuetext": Object {
         "type": "string",
       },
-      "asChild": Object {
```
reesscot, line 5044: We don't support these props in Expander, so it should be ok to remove them

joebuono (dismissed):

packages/react/__tests__/__snapshots__/exports.ts.snap
wlee221: Did this show up in our docs somewhere?

reesscot: This never showed in our docs: https://ui.docs.amplify.aws/react/components/expander?tab=props

wlee221 (dismissed):

packages/react/src/primitives/Checkbox/Checkbox.tsx
```diff
@@ -50,9 +50,11 @@ const CheckboxPrimitive: Primitive<CheckboxProps, 'input'> = (
   const dataId = useStableId();
   React.useEffect(() => {
     const input = document.querySelector(`[data-id="${dataId}"]`);
-    // HTMLInputElement does not have an `indeterminate` attribute
-    (input as HTMLInputElement & { indeterminate: boolean }).indeterminate =
-      isIndeterminate;
+    if (input && isIndeterminate !== undefined) {
```
calebpollman, line 53: Probably not super important but just for explicitness should the second condition be `typeof isIndeterminate === 'boolean'`?

0618 (dismissed):

packages/react/src/primitives/Checkbox/useCheckbox.ts
```diff
@@ -4,8 +4,8 @@ import { isFunction } from '../shared/utils';
 import { UseCheckbox } from '../types/checkbox';
 
 export const useCheckbox = (
-  initialChecked: boolean,
-  onChangeProp: React.ChangeEventHandler
+  initialChecked: boolean | undefined,
+  onChangeProp: React.ChangeEventHandler | undefined
```
calebpollman, line 8: What do you think of making this optional instead of allowing `undefined`?

packages/react/src/primitives/Checkbox/useCheckbox.ts
```diff
@@ -4,8 +4,8 @@ import { isFunction } from '../shared/utils';
 import { UseCheckbox } from '../types/checkbox';
 
 export const useCheckbox = (
-  initialChecked: boolean,
-  onChangeProp: React.ChangeEventHandler
+  initialChecked: boolean | undefined,
```
calebpollman, line 7: Can we just keep this as `boolean` and assign a default value of `false`? _Think_ it would allow `setDataChecked` to have a type of `Dispatch<SetStateAction<boolean>>` in the `UseCheckbox` interface

calebpollman (commented):
LGTM, just left some questions

packages/react/src/primitives/Checkbox/useCheckbox.ts
reesscot: I'm good with that, will update

packages/react/src/primitives/Checkbox/Checkbox.tsx
reesscot: That does seem clearer, will update

packages/react/src/primitives/Checkbox/useCheckbox.ts
reesscot: I'd rather now change the existing behavior, in particular when it comes to the uncontrolled Checkbox. I checked current behavior and checked is initially `undefined` when it hasn't been interacted with.

zchenwei (approved):

calebpollman (approved):
🚢

PR #3311, opened 2023-01-17
chore(account-settings): fix strict ts errors for account settings components

wlee221:
#### Description of changes

Fixes `strict: true` errors in the Account Settings components.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
+++ b/packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
@@ -44,7 +44,7 @@ function ChangePassword({
   validators,
   components,
 }: ChangePasswordProps): JSX.Element | null {
-  const [errorMessage, setErrorMessage] = React.useState<string>(null);
+  const [errorMessage, setErrorMessage] = React.useState<string | null>(null);
   const [formValues, setFormValues] = React.useState<FormValues>({});
   const [validationError, setValidationError] = React.useState<ValidationError>(
     {}
@@ -158,6 +158,10 @@ function ChangePassword({
 
   const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
+    if (!user) {
+      return;
+    }
+
     const { currentPassword, newPassword } = formValues;
     if (errorMessage) {
       setErrorMessage(null);
--- a/packages/react/src/components/AccountSettings/ChangePassword/defaults.tsx
+++ b/packages/react/src/components/AccountSettings/ChangePassword/defaults.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Button, PasswordField } from '../../../primitives';
 import { ValidationErrors } from '../../shared/ValidationErrors';
 import { DefaultErrorMessage } from '../shared/Defaults';
-import { PasswordFieldComponent, SubmitButtonComponent } from '../types';
+import { PasswordFieldComponent } from '../types';
 import { ChangePasswordComponents } from './types';
 
 const DefaultPasswordField: PasswordFieldComponent = ({
@@ -20,15 +20,11 @@ const DefaultPasswordField: PasswordFieldComponent = ({
   );
 };
 
-const DefaultSubmitButton: SubmitButtonComponent = (props) => (
-  <Button {...props} />
-);
-
 const DEFAULTS: Required<ChangePasswordComponents> = {
   CurrentPasswordField: DefaultPasswordField,
   NewPasswordField: DefaultPasswordField,
   ConfirmPasswordField: DefaultPasswordField,
-  SubmitButton: DefaultSubmitButton,
+  SubmitButton: Button,
   ErrorMessage: DefaultErrorMessage,
 };
 
--- a/packages/react/src/components/AccountSettings/ChangePassword/defaults.tsx
+++ b/packages/react/src/components/AccountSettings/ChangePassword/defaults.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Button, PasswordField } from '../../../primitives';
 import { ValidationErrors } from '../../shared/ValidationErrors';
 import { DefaultErrorMessage } from '../shared/Defaults';
-import { PasswordFieldComponent } from '../types';
+import { PasswordFieldComponent, SubmitButtonComponent } from '../types';
 import { ChangePasswordComponents } from './types';
 
 const DefaultPasswordField: PasswordFieldComponent = ({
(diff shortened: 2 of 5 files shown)


packages/react/src/components/AccountSettings/ChangePassword/ChangePassword.tsx
```diff
@@ -158,6 +158,10 @@ function ChangePassword({
 
   const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
+    if (!user) {
+      return;
+    }
```
wlee221, line 163: Technically, `user` will always be defined because `ChangePassword` does not render until there's a valid user.

But TS doesn't know this because `handleSubmit` is a callback, so I'm adding this additional check here.

packages/react/src/components/AccountSettings/ChangePassword/defaults.tsx
```diff
@@ -22,7 +24,7 @@ const DEFAULTS: Required<ChangePasswordComponents> = {
   CurrentPasswordField: DefaultPasswordField,
   NewPasswordField: DefaultPasswordField,
   ConfirmPasswordField: DefaultPasswordField,
-  SubmitButton: Button,
+  SubmitButton: (props) => <Button {...props} />,
```
wlee221, line 27: This fixes error:

```
Type 'WeakValidationMap<Pick<PrimitivePropsWithRef<ButtonProps, "button">, "area" | "slot" | "title" | "id" | "className" | "testId" | "alignSelf" | "backgroundColor" | ... 344 more ... | "onTransitionEndCapture"> & RefAttributes<...>>' is not assignable to type 'WeakValidationMap<Partial<ButtonPrimitiveProps> & Required<Pick<ButtonPrimitiveProps, "isDisabled">>>'.
  Types of property 'isDisabled' are incompatible.
    Type 'Validator<boolean | null | undefined> | undefined' is not assignable to type 'Validator<boolean> | undefined'.
       Type 'Validator<boolean | null | undefined>' is not assignable to type 'Validator<boolean>'.
         Type 'boolean | null | undefined' is not assignable to type 'boolean'.
           Type 'undefined' is not assignable to type 'boolean'.
```

which was complaining that `isDisabled` is required on override component type, but is optional on `Button` itself.

packages/react/src/components/AccountSettings/ChangePassword/defaults.tsx
```diff
@@ -13,16 +13,22 @@ const DefaultPasswordField: PasswordFieldComponent = ({
...
+      {fieldValidationErrors ? (
+        <ValidationErrors errors={fieldValidationErrors} />
+      ) : null}
     </>
   );
 };
 
+const DefaultSubmitButton: SubmitButtonComponent = (props) => (
+  <Button {...props} />
+);
```
wlee221, line 25: This fixes error:

```
Type 'WeakValidationMap<Pick<PrimitivePropsWithRef<ButtonProps, "button">, "area" | "slot" | "title" | "id" | "className" | "testId" | "alignSelf" | "backgroundColor" | ... 344 more ... | "onTransitionEndCapture"> & RefAttributes<...>>' is not assignable to type 'WeakValidationMap<Partial<ButtonPrimitiveProps> & Required<Pick<ButtonPrimitiveProps, "isDisabled">>>'.
  Types of property 'isDisabled' are incompatible.
    Type 'Validator<boolean | null | undefined> | undefined' is not assignable to type 'Validator<boolean> | undefined'.
       Type 'Validator<boolean | null | undefined>' is not assignable to type 'Validator<boolean>'.
         Type 'boolean | null | undefined' is not assignable to type 'boolean'.
           Type 'undefined' is not assignable to type 'boolean'.
```

which was complaining that `isDisabled` is required on override component type but is optional on `Button` itself. I've fixed this by having props spread onto the `Button`, so that `isDisabled` is always passed.

wlee221: We use this pattern in other component (e.g. [here](https://github.com/aws-amplify/amplify-ui/blob/a5f04956c254bf3ca6e380c081590e085c9edd8a/packages/react/src/components/AccountSettings/DeleteUser/defaults.tsx#L43-L45)), and this matches what we have in docs ([link](https://ui.docs.amplify.aws/react/connected-components/account-settings/change-password#component-overrides)). But open for suggestions.

zchenwei (approved):

dbanksdesign (approved):
LGTM!

reesscot (approved):

packages/react/src/components/AccountSettings/ChangePassword/defaults.tsx
calebpollman: Totally just my opinion here, but it doesn't make a ton of sense to add a wrapper around `Button` just to appease TS, and honestly not sure why spreading the props would have an impact on optional vs required for `isDisabled`.
In both the examples, there is a prop being overridden (`variation` in the FileUploader code, `label` in the docs) so I'm not sure if we can really compare those against what we are doing here

wlee221: Good point!

> not sure why spreading the props would have an impact on optional vs required for isDisabled.

The override type comes from `CommonButtonTypes`, which [requires](https://github.com/aws-amplify/amplify-ui/blob/c7b6ae535b8a896a3fbfe8a7d6d15d136dd7fb1d/packages/react/src/components/AccountSettings/types.ts#L36) `isDisabled` for override components. On the other hand, `Button` primitive has that property optional. Now if we spread props,

```ts
const DefaultSubmitButton: SubmitButtonComponent = (
  props // this always includes `isDisabled`.
) => (<Button {...props}/>); // `isDisabled` gets spread into the primitive.
```

so this is a valid TS.

If we do not spread props,

```ts
DefaultSubmitButton: SubmitButtonComponent = Button;
```

This fails because the component types do not match; `SubmitButtonComponent` has `isDisabled` required but the other doesn't. Note that we're directly assigning a component, so props types are compared directly. 

>  it doesn't make a ton of sense to add a wrapper around Button just to appease TS

I think that default components are just a specific instance of override components. So I think it's reasonable that there exists discrepancy between override and default component (primitives in this case), and we use additional wrapper to bridge the two.

calebpollman: What do you think of just making `isDisabled` optional on the `SubmitButtonProps`? Since it is a boolean it will just evaluate to falsy if `undefined`

wlee221: That'd make sense! I'll write a commit now.

PR #3312, opened 2023-01-17
chore(ui): make theme code ts strict

dbanksdesign:

Built the UI package with TS strict turned on (there are still other errors, but none in the theme section) and ran tests

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/src/theme/utils.ts
+++ b/packages/ui/src/theme/utils.ts
@@ -34,7 +34,7 @@ function referenceValue(value: string) {
   return value;
 }
 
-export function cssValue(token: DesignToken<any>) {
+export function cssValue(token: BaseDesignToken): string | number {
   const { value } = token;
   if (isString(value)) {
     return referenceValue(value);
@@ -42,13 +42,11 @@ export function cssValue(token: DesignToken<any>) {
 
   if (isShadowToken(value) && isObject(value)) {
     return SHADOW_PROPERTIES.map((property) => {
-      // lookup property against `token` first for custom non-nested value, then lookup
-      // property against `value` for design token value, default to empty string
-      if (isShadowToken(token) && isObject(token)) {
-        return referenceValue(token[property] ?? '');
-      } else {
-        return referenceValue(value[property] ?? '');
-      }
+      return referenceValue(
+        // lookup property against `token` first for custom non-nested value, then lookup
+        // property against `value` for design token value, default to empty string
+        isShadowToken(token) ? token[property] : value[property] ?? ''
+      );
     }).join(' ');
   }
 
--- a/packages/ui/src/theme/utils.ts
+++ b/packages/ui/src/theme/utils.ts
@@ -44,8 +44,8 @@ export function cssValue(token: BaseDesignToken): string | number {
   if (isShadowToken(value)) {
     return SHADOW_PROPERTIES.map((property) => {
       return referenceValue(
-        // lookup property against `token` first for custom non-nested value, then lookup
-        // property against `value` for design token value, default to empty string
+       // lookup property against `token` first for custom non-nested value, then lookup
+       // property against `value` for design token value
         isShadowToken(token) ? token[property] : value[property]
       );
     }).join(' ');
--- a/packages/ui/src/theme/utils.ts
+++ b/packages/ui/src/theme/utils.ts
@@ -26,7 +26,8 @@ const SHADOW_PROPERTIES: ShadowPropertyKey[] = [
   'color',
 ];
 
-function referenceValue(value: string) {
+function referenceValue(value?: string) {
+  if (!value) return '';
   if (usesReference(value)) {
     const path = value.replace(/\{|\}/g, '').replace('.value', '').split('.');
     return `var(--${cssNameTransform({ path })})`;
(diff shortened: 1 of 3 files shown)


zchenwei (approved):

packages/ui/src/theme/utils.ts
```diff
@@ -34,20 +34,22 @@ function referenceValue(value: string) {
   return value;
 }
 
-export function cssValue(token: DesignToken<{ value: unknown }>) {
+export function cssValue(token: DesignToken<any>) {
```
reesscot, line 37: Don't we have some idea of the shape of the DesignToken?

packages/ui/src/theme/utils.ts
```diff
@@ -34,20 +34,22 @@ function referenceValue(value: string) {
...
-    }
+  if (isShadowToken(value) && isObject(value)) {
+    return SHADOW_PROPERTIES.map((property) => {
+      // lookup property against `token` first for custom non-nested value, then lookup
+      // property against `value` for design token value, default to empty string
+      if (isShadowToken(token) && isObject(token)) {
+        return referenceValue(token[property] ?? '');
+      } else {
+        return referenceValue(value[property] ?? '');
+      }
```
calebpollman, line 51: These two conditions look the same to me and line 47 reads the same as line 43 or am i missing something 😄

dbanksdesign: Yea I wonder if there is a cleaner way to write this code in a way that makes TS happy...

dbanksdesign: I re-did this code a bit to be a bit more readable I hope. Take a look.

calebpollman: Just throwing this out there but we should be able to remove `isObject` from the `if` blocks since we use `isObject` in `isShadowToken`. From there think we could probably just reduce the return to:
```
return referenceValue((isShadowToken(token) ? token[property] : value[property]) ?? '');
```

packages/ui/src/theme/utils.ts
dbanksdesign: Updated it to be `BaseDesignToken` because that is what we are calling `cssValue` with in `createTheme.ts`

packages/ui/src/theme/utils.ts
```diff
@@ -34,20 +34,20 @@ function referenceValue(value: string) {
...
-      return SHADOW_PROPERTIES.map((property) =>
+  if (isShadowToken(value) && isObject(value)) {
+    return SHADOW_PROPERTIES.map((property) => {
+      return referenceValue(
         // lookup property against `token` first for custom non-nested value, then lookup
         // property against `value` for design token value, default to empty string
-        referenceValue(token[property] ?? value[property] ?? '')
-      ).join(' ');
-    }
+        isShadowToken(token) ? token[property] : value[property] ?? ''
```
calebpollman, line 48: Silly question incoming: Does the nullish coalescing get applied after the ternary resolves or is it only applied to the right condition?

packages/ui/src/types/style-dictionary.d.ts
```diff
@@ -0,0 +1,21 @@
+declare module 'style-dictionary/lib/utils/deepExtend' {
+  // based on the typing specified by style-dictionary:
+  // https://github.com/amzn/style-dictionary/blob/main/lib/utils/deepExtend.js#L18-L22
+  export default function deepExtend<T extends object>(
+    objects: (object | undefined)[],
+    collision?: Function,
+    path?: string[]
+  ): T;
```
calebpollman, line 8: Nice!

packages/ui/src/theme/utils.ts
calebpollman: Looks like we were mostly thinking the same thing 💯 Left [a question](https://github.com/aws-amplify/amplify-ui/pull/3312#discussion_r1072980384) on the new code, and still thinking we can remove the `isObject` from the first `if` block but looks much better imo

packages/ui/src/theme/utils.ts
dbanksdesign: Ok I think I made it better by removing the nullish coalescing and making `referenceValue` return empty string if its input is nullish.

reesscot (approved):

zchenwei (approved):

packages/ui/src/theme/utils.ts
```diff
@@ -26,28 +26,29 @@ const SHADOW_PROPERTIES: ShadowPropertyKey[] = [
   'color',
 ];
 
-function referenceValue(value: string) {
+function referenceValue(value?: string) {
+  if (!value) return '';
```
calebpollman, line 30: ⭐

packages/ui/src/theme/utils.ts
```diff
@@ -26,28 +26,29 @@ const SHADOW_PROPERTIES: ShadowPropertyKey[] = [
...
   }
 
-  if (isObject(value)) {
-    if ('offsetX' in value) {
-      return SHADOW_PROPERTIES.map((property) =>
+  if (isShadowToken(value)) {
+    return SHADOW_PROPERTIES.map((property) => {
+      return referenceValue(
         // lookup property against `token` first for custom non-nested value, then lookup
         // property against `value` for design token value, default to empty string
```
calebpollman, line 48:
```suggestion
       // lookup property against `token` first for custom non-nested value, then lookup
       // property against `value` for design token value

calebpollman (approved):

reesscot (approved):

PR #3317, opened 2023-01-18
chore(ui-react): Add strict type check fixes for data store hooks

ioanabrooks:
#### Description of changes

Fix type errors for datastore hooks to allow enabling TS strict mode.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/hooks/actions/shared/types.ts
+++ b/packages/react/src/hooks/actions/shared/types.ts
@@ -20,7 +20,7 @@ type ModelFields<Type> = {
   [Property in keyof Type]: string | number | boolean;
 };
 
-export type DataStoreActionFields<Model> =
+export type DataStoreActionFields<Model extends PersistentModel> =
   | ModelInit<Model, PersistentModelMetaData<Model>>
   | ModelFields<ModelInit<Model, PersistentModelMetaData<Model>>>;
 
@@ -39,9 +39,9 @@ export interface UseDataStoreActionOptions<Model extends PersistentModel> {
   schema?: Schema;
 }
 
-export function isAlreadyTyped<Model>(
+export function isAlreadyTyped<Model extends PersistentModel>(
   fields: DataStoreActionFields<Model>,
-  schema: Schema
+  schema?: Schema
 ): fields is ModelInit<Model, PersistentModelMetaData<Model>> {
   return !schema;
 }
--- a/packages/react/src/hooks/actions/shared/useTypeCastFields.ts
+++ b/packages/react/src/hooks/actions/shared/useTypeCastFields.ts
@@ -1,16 +1,18 @@
 import * as React from 'react';
 import {
   ModelInit,
+  PersistentModel,
   PersistentModelMetaData,
   Schema,
 } from '@aws-amplify/datastore';
 
 import { DataStoreActionFields, isAlreadyTyped } from './types';
+import { isString } from 'lodash';
 
-interface UseTypeCastFieldsProps<Model> {
+interface UseTypeCastFieldsProps<Model extends PersistentModel> {
   fields: DataStoreActionFields<Model>;
   modelName: string;
-  schema: Schema;
+  schema?: Schema;
 }
 
 type UseTypeCastFieldsReturn<Model> =
@@ -22,7 +24,7 @@ type UseTypeCastFieldsReturn<Model> =
  * datastore based on the schema type
  * @see: See https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html
  */
-export const useTypeCastFields = <Model>({
+export const useTypeCastFields = <Model extends PersistentModel>({
   fields,
   modelName,
   schema,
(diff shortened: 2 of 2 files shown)


packages/react/src/hooks/actions/shared/useTypeCastFields.ts
```diff
@@ -1,16 +1,18 @@
 import * as React from 'react';
 import {
   ModelInit,
+  PersistentModel,
   PersistentModelMetaData,
   Schema,
 } from '@aws-amplify/datastore';
 
 import { DataStoreActionFields, isAlreadyTyped } from './types';
+import { isString } from 'lodash';
```
reesscot, line 10: For proper treeshaking, I think this is supposed to be:

```suggestion
import { isString } from 'lodash/isString';
```

packages/react/src/hooks/actions/shared/useTypeCastFields.ts
```diff
@@ -1,16 +1,18 @@
 import * as React from 'react';
 import {
   ModelInit,
+  PersistentModel,
   PersistentModelMetaData,
   Schema,
 } from '@aws-amplify/datastore';
 
 import { DataStoreActionFields, isAlreadyTyped } from './types';
+import { isString } from 'lodash/isString';
```
ioanabrooks, line 10:
```suggestion
import isString from 'lodash/isString';
```

packages/react/src/hooks/actions/shared/useTypeCastFields.ts
wlee221: I believe this should be default import iirc

wlee221 (approved):

packages/react/src/hooks/actions/shared/useTypeCastFields.ts
```diff
@@ -1,16 +1,18 @@
 import * as React from 'react';
 import {
   ModelInit,
+  PersistentModel,
   PersistentModelMetaData,
   Schema,
 } from '@aws-amplify/datastore';
 
 import { DataStoreActionFields, isAlreadyTyped } from './types';
+import isString from 'lodash/isString';
```
reesscot, line 10: nit: move import above to third party imports

reesscot (dismissed):
LGTM, thanks for fixing these!

packages/react/src/hooks/actions/shared/types.ts
```diff
@@ -39,9 +39,9 @@ export interface UseDataStoreActionOptions<Model extends PersistentModel> {
   schema?: Schema;
 }
 
-export function isAlreadyTyped<Model>(
+export function isAlreadyTyped<Model extends PersistentModel>(
   fields: DataStoreActionFields<Model>,
-  schema: Schema
+  schema?: Schema
```
calebpollman, line 44: Feel like this is a silly question, but this change now allows `isAlreadyTyped` (side note: maybe we can rename) to be called without the `schema` param, which will always return `true` correct? 

Just want to make sure that this is intentional, the change has no impact on current usage (in particular since this ultimately gets used by Studio) and is a change we definitely want. 

Last thing - know you did not write the code initially, but it might be nice to leave some comments on how this should be used as well since this function is surprisingly opaque between its name and logic

packages/react/src/hooks/actions/shared/useTypeCastFields.ts
```diff
@@ -1,16 +1,18 @@
...
 } from '@aws-amplify/datastore';
 
 import { DataStoreActionFields, isAlreadyTyped } from './types';
 
-interface UseTypeCastFieldsProps<Model> {
+interface UseTypeCastFieldsProps<Model extends PersistentModel> {
   fields: DataStoreActionFields<Model>;
   modelName: string;
-  schema: Schema;
+  schema?: Schema;
```
calebpollman, line 15: @reesscot Does loosening our types have any impact on Studio?

zchenwei (dismissed):
LGTM

packages/react/src/hooks/actions/shared/useTypeCastFields.ts
reesscot: We probably need to have this published to a tag and run the studio integ tests

dbanksdesign: We should combine this and #3382 into a branch and publish a tag to run studio integ tests.

awinberg-aws: Agreed, let's get a tag release out for testing on our end. Should be a quick yes/no outcome since only types are affected.

hein-j:
The Studio team pulled in the tagged release into their codegen integration tests and didn't find any problems.
The test cases included variants and navigation actions.
If there's something specific you'd like us to test, let us know.

awinberg-aws (approved):

hein-j (approved):

calebpollman (approved):

PR #3321, opened 2023-01-18
chore(docs): revert commit restricting tracking code to only prod

dbanksdesign:
#### Description of changes

Reverting #2391 so that we can test changes to tracking code. There is currently no way to test if changes to our tracking code works unless it goes to prod. 

It is ok that our dev site tracks usage because we can filter by URL. We were counting dev hits along with prod, but we can just change the filter:

CleanShot 2023-01-18 at 15 02 21@2x

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/utils/stage.ts
+++ b/docs/src/utils/stage.ts
@@ -1,7 +1,0 @@
-const PROD_STAGE_SITE_URL = 'https://ui.docs.amplify.aws';
-const DEV_STAGE_SITE_URL = 'dev.ui.docs.amplify.aws';
-const SITE_URL = process.env.SITE_URL || '';
-export const IS_PROD_STAGE = SITE_URL === PROD_STAGE_SITE_URL;
-// Dev Site url starts with "www" due to Algolia search, so we use `includes` to match it:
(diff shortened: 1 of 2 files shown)


reesscot (approved):

docs/src/utils/stage.ts
```diff
@@ -1,7 +0,0 @@
-const PROD_STAGE_SITE_URL = 'https://ui.docs.amplify.aws';
-const DEV_STAGE_SITE_URL = 'dev.ui.docs.amplify.aws';
```
reesscot, line 2: Are we sure this isn't used anywhere else?

zchenwei: +1

dbanksdesign: yup, did a search through the repo. And we will see the docs build test fail if it is used somewhere.

0618 (approved):
To clarify or for future reference, this PR partially reverts https://github.com/aws-amplify/amplify-ui/pull/2391, A.K.A. adds the tracking code back to non-prod environment.

(The fix for the theme switch is still there)

zchenwei (approved):

Issue #3257, opened 2023-01-04
FileUploader component isn't setting content type of svg


rozhover:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Storage (File Uploader)

### How is your app built?

Next.js

### Please describe your bug.

I have a FileUploader component in Next.js app, and i'm trying to set the uploads to public read access. It works for jpegs and pngs, but not svgs. Svg metadata is being set to "binary/octet-stream" instead of "image/svg+xml". Only when i specifically set contentType="image/svg+xml" on the FileUploader component then it works, but not sure how to set that prop per file, if multiple upload types are allowed on the component.

### What's the expected behaviour?

Expected to detect svg mime type based on the file and save in the metadata

### Help us reproduce the bug!

`<FileUploader showImages={true} level="public" accessLevel="public" storageAccessLevel="public" acceptedFileTypes={["image/*"]} onSuccess={(res) => setNewUploads(res)} />`

### Code Snippet

```typescript
// Put your code below this line.

```

ErikCH:
Hi @rozhover !

Are you saying when you download the file back from S3, that it's not detecting it's an SVG file any more? I just ran a test, and the svg file downloaded back from S3 was an S3 image .

Also I am adding a new feature in the future that you can edit the files before they upload, including changing the file name, and run your own `Storage.put`

rozhover:
Hi, thanks for the quick reply!
It does indeed work to download back, but it doesn't work to display in the browser when using an img tag. The difference i'm seeing is in the metadata, the Content-Type seems to be set as binary/octet-stream. But if I override it to set to image/svg+xml then it works to display in the image tag. 
That would be a great feature to edit files before upload. But it would be nice too if the uploader could detect svg and set the content type to image/svg+xml automatically.
Thanks for your help!

ErikCH:
Hi @rozhover !

[...]

PR #3322, opened 2023-01-19
fix(file-uploader): Add default contentType for Uploads

ErikCH:
#### Description of changes
When uploading files using the file uploader the content type is never set. Instead all content types default to `binary/octet-stream`. This causes an issue since files like svg won't have the correct contentType in S3 storage. When using `Storage.get` the file cannot be displayed in an `image` tag with the wrong content type.

To fix this now all uploads will default to the correct file type listed in the `file.type`. If none exists it will fall back to `binary/octet-stream`
 Users can still change this by updating the prop `contentType` if needed back to `binary/octet-stream` for all uploads.

#### Issue #, if available
#3257 

Added additional tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
+++ b/packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
@@ -27,6 +27,7 @@ export function uploadFile({
       progressCallback,
       errorCallback,
       completeCallback,
+      contentType: file.type || 'binary/octet-stream',
       ...rest,
     });
   } else {
@@ -34,6 +35,7 @@ export function uploadFile({
       level,
       resumable: false,
       progressCallback,
+      contentType: file.type || 'binary/octet-stream',
       ...rest,
     }).then(completeCallback, errorCallback);
   }
--- a/packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
+++ b/packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
@@ -20,6 +20,7 @@ export function uploadFile({
   errorCallback: (err: string) => void;
   completeCallback: (event) => void;
 }) {
+  const contentType = file.type || 'binary/octet-stream';
   if (isResumable === true) {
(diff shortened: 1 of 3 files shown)


packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
```diff
@@ -27,13 +27,15 @@ export function uploadFile({
       progressCallback,
       errorCallback,
       completeCallback,
+      contentType: file.type || 'binary/octet-stream',
```
ErikCH, line 30: We want all falsy values, not just null or undefined, to default to `binary/octet-stream`. Since falsy types are not valid MIME types.

This can be overridden by passing a `contentType` prop to the FileUploader if needed.

packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
```diff
@@ -27,13 +27,15 @@ export function uploadFile({
...
       ...rest,
     });
   } else {
     return Storage.put(fileName, file, {
       level,
       resumable: false,
       progressCallback,
+      contentType: file.type || 'binary/octet-stream',
```
dbanksdesign, line 38: nit: can we make a const of `file.type || 'binary/octet-stream'` so we don't have to repeat it?

dbanksdesign (approved):
just one non-blocking nit. LGTM!

packages/ui/src/helpers/storage/fileUploader/utils/uploader.ts
calebpollman: +1. Also would it make sense to use nullish coalescing here rather than the `OR` operator since the resolved value is not a boolean?

dbanksdesign (approved):
LGTM! :shipit:

reesscot (approved):

calebpollman (approved):

PR #3325, opened 2023-01-19
fix(ui): Translate 'or' text for FederatedSignIn

joebuono:
#### Description of changes

- Adds `'or'` translations for all of our i18n supported languages
- Includes `'or'` text on the `FederatedSignIn` screen for React, Vue and Angular

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/ui/src/i18n/dictionaries/authenticator/it.ts
```diff
@@ -25,9 +25,11 @@ export const itDict: AuthenticatorDictionary = {
   'Enter your username': 'Inserisci il tuo nome utente',
   'Forgot your password?': 'Password dimenticata?',
   'Hide password': 'Nascondi password',
-  'It may take a minute to arrive': "L'arrivo potrebbe richiedere qualche minuto",
+  'It may take a minute to arrive':
+    "L'arrivo potrebbe richiedere qualche minuto",
   Loading: 'Caricamento in corso',
   'New password': 'Nuova password',
+  or: 'o',
```
ioanabrooks, line 32:
```suggestion
  or: 'oppure',
```

packages/ui/src/i18n/dictionaries/authenticator/de.ts
```diff
@@ -29,6 +29,7 @@ export const deDict: AuthenticatorDictionary = {
     'Es kann eine Minute dauern, bis er ankommt',
   Loading: 'Wird geladen',
   'New password': 'Neues Passwort',
+  or: 'oder',
```
reesscot, line 32: just curious, do we know if this will work with all language sentence structures? I could imagine that some languages may not use or in the same way in the middle of the sentence... Maybe they do, I only know English and Spanish...

joebuono: Good question. This specific string is used in isolation on the FederatedSignIn screen, so I don't think we need to be concerned about sentence structure. 

AuthOr

reesscot (approved):
LGTM

wlee221 (approved):

PR #3336, opened 2023-01-23
fix(examples): ignore directories named .DS_Store in metro.config

hbuchel:
#### Description of changes

This change filters out directories named .DS_Store in the metro.config.js in our react-native example so that it doesn't try to treat it as a package.

When running `yarn react-native-example ios`, this error was encountered:

```
error Cannot find module '/Volumes/workplace/amplify-ui/packages/.DS_Store/package.json'
```
Screen Shot 2023-01-23 at 12 07 32 PM

After the change,  `yarn react-native-example ios` runs successfully.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/examples/react-native/metro.config.js
+++ b/examples/react-native/metro.config.js
@@ -31,12 +31,14 @@ function getInternalPackages(root, internalPackagesDirectory) {
 
       return readdirSync(internalDirectoryRoot, {
         withFileTypes: true,
-      }).map(({ name }) => {
-        const packagePath = path.resolve(internalDirectoryRoot, name);
-        const packageName = require(`${packagePath}/package.json`).name;
+      })
+        .filter(({ name }) => name !== '.DS_Store')
+        .map(({ name }) => {
+          const packagePath = path.resolve(internalDirectoryRoot, name);
+          const packageName = require(`${packagePath}/package.json`).name;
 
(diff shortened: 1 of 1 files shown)


examples/react-native/metro.config.js
```diff
@@ -31,12 +31,14 @@ function getInternalPackages(root, internalPackagesDirectory) {
...
-        return { packageName, packagePath };
-      });
+      })
+        .filter(({ name }) => name !== '.DS_Store')
+        .map(({ name }) => {
+          const packagePath = path.resolve(internalDirectoryRoot, name);
+          const packageName = require(`${packagePath}/package.json`).name;
+
+          return { packageName, packagePath };
+        });
```
dbanksdesign, line 41: Could we make this logic a bit more robust to check if instead there exists a package.json at the packagePath? We could do a try/catch or filtering if that file exists using fs.access or something similar?

dbanksdesign (commented):
Thanks for adding this, just one small comment to make this more robust for future use cases

examples/react-native/metro.config.js
hbuchel: Yeah, good point, let me take a look!

zchenwei: +1

PR #3340, opened 2023-01-24
fix(ui): replace tsup with tsc in dev command

calebpollman:
#### Description of changes
Remove `tsup` in preference of `tsc` for dev commands

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


package.json
```diff
@@ -81,7 +81,6 @@
     "lint-staged": ">=10",
     "prettier": "2.4.1",
     "rimraf": "^3.0.2",
-    "tsup": "^5.11.9",
```
reesscot, line 84: woohoo!!!

packages/ui/src/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './i18n';
 export * from './machines';
 export * from './theme';
 export * from './types';
+export const hehe = '';
```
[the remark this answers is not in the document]
reesscot: my favorite const ever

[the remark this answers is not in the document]
wlee221: This might necessitate a API review 😉

packages/react-core/tsconfig.dev.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    // rollup does not allow `declarationDir` to be assigned in the tsconfig
```
reesscot, line 4: What does this mean? Looks like this is in the tsconfig?

calebpollman: The dev tsconfig extends from the base tsconfig

packages/react/tsconfig.json
```diff
@@ -1,16 +1,16 @@
...
     "declaration": true,
     "esModuleInterop": true,
-    "jsx": "react-jsx",
-    "rootDir": "./src",
-    "outDir": "./dist",
     "importHelpers": true,
+    "jsx": "react",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "outDir": "dist/esm",
```
reesscot, line 10: Assuming this is ignored by rollup?

packages/ui/tsconfig.json
```diff
@@ -5,25 +5,21 @@
     }
   },
   "compilerOptions": {
-    "target": "es5",
```
reesscot, line 8: Can you confirm that the only change here was outDir? Hard to read with the alphabetizing

packages/react/tsconfig.json
calebpollman: It aligns to our `rollup` config (has the same value as `dir`) -> [source](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/rollup.config.js#L29)

packages/ui/tsconfig.json
calebpollman: That's correct

packages/react-core/tsconfig.dev.json
reesscot: ok, then maybe change the comment to say 
```
 // rollup does not allow `declarationDir` to be assigned in the base tsconfig
 ```

**Comment by reesscot** [comment]
If you can make the two changes above, I'm ready to approve

File: packages/ui/src/index.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (zchenwei) [reply packages/ui/src/index.ts:6]: I like it

**Review by zchenwei — COMMENTED** [review]
LGTM. Thanks for working on it!

File: packages/ui/tsconfig.json [hunk 6]
```diff
@@ -5,25 +5,21 @@
...
-    "declaration": true,
-    "outDir": "dist",
-    "baseUrl": ".",
-    "types": ["webpack-env", "jest"],
-    "paths": {
-      "@/*": ["src/*"]
-    },
-    "lib": ["esnext", "dom", "dom.iterable", "scripthost"]
+    "strict": false,
+    "target": "es5",
```
zchenwei, line 21: Thinking if we should update the target to align other consumer packages. Those customer facing packages target a higher ES version which makes es5 looks dumb here.

calebpollman: Agree but believe that update has the potential to be a breaking change

reesscot (approved):
LGTM

PR #3349, opened 2023-01-25
fix:(file-uploader): Swap Save and Cancel Button

ErikCH:
#### Description of changes

To adhere to best practices, we are swapping the placement of the save and cancel buttons when editing a file.

Before
Screen Shot 2023-01-25 at 10 34 32 AM

After
Screen Shot 2023-01-25 at 10 37 33 AM

We have extensive tests that test edit and save.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/.changeset/afraid-maps-speak.md
+++ b/.changeset/afraid-maps-speak.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ui-react': patch
+---
+
+fix: Swapped save and cancel buttons.
--- a/packages/react/src/components/Storage/FileUploader/UploadTracker/index.tsx
+++ b/packages/react/src/components/Storage/FileUploader/UploadTracker/index.tsx
@@ -79,21 +79,21 @@ export function UploadTracker({
           <>
             <Button
               size="small"
-              variation="primary"
-              onClick={() => {
-                onSaveEdit(tempName);
-              }}
-            >
-              Save
-            </Button>
-            <Button
-              size="small"
               onClick={() => {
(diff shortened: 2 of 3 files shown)


ioanabrooks (changes requested):
I think the mainstream is not yet settled on whether having the primary button action be last is the best practice, but if we are going to be opinionated on this we should be consistent and also change the position of the edit/remove actions [screenshot below]. Did we consider accessibility? If using keyboard only does reaching the primary action require an extra key press now?

Screen Shot 2023-01-25 at 11 26 42 AM

dbanksdesign:
This was brought up in the fit-n-finish review with the design org. TBH its not a big deal to me either way. @hbuchel do you have any thoughts on this from an accessibility point of view?

ErikCH:
> I think the mainstream is not yet settled on whether having the primary button action be last is the best practice, but if we are going to be opinionated on this we should be consistent and also change the position of the edit/remove actions [screenshot below]. Did we consider accessibility? If using keyboard only does reaching the primary action require an extra key press now?
> 
> Screen Shot 2023-01-25 at 11 26 42 AM

OK! In a separate PR I'll move the edit button so we don't have any inconsistencies. As for keyboard users they can either hit enter on the input to save, or it's two tab presses to get to save.

dbanksdesign (approved):

ioanabrooks (dismissed):

ioanabrooks (approved):
LGTM

PR #3363, opened 2023-01-27
chore(ui-react): add some unit tests to MessageLayout

calebpollman:
#### Description of changes
Add some unit tests to `MessageLayout`

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
+++ b/packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
@@ -12,11 +12,13 @@ import {
 import { CloseIconButton } from '../CloseIconButton';
 import {
   BLOCK_CLASS,
+  BODY_TEXT_TEST_ID,
   BUTTON_CLASS,
   CLOSE_BUTTON_CLASS,
   CONTENT_CLASS,
   CONTENT_TEST_ID,
   HEADER_CLASS,
+  HEADER_TEXT_TEST_ID,
   IMAGE_CONTAINER_CLASS,
   IMAGE_CONTAINER_TEST_ID,
   MESSAGE_LAYOUT_TEST_ID,
@@ -81,7 +83,7 @@ export function MessageLayout({
           >
             <Image
               alt="In-App Message Image"
-              src={image?.src}
+              src={image.src}
               style={styles.image}
             />
           </Flex>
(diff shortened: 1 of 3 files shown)


packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
```diff
@@ -118,7 +125,7 @@ export function MessageLayout({
               onClick={secondaryButton?.onAction}
               style={styles.secondaryButton}
             >
-              {secondaryButton?.title}
+              {secondaryButton.title}
```
calebpollman, line 128: The optional chaining here is unnecessary since we know that `secondaryButton` is defined from line 119 (same thing with line 140)

packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
```diff
@@ -118,7 +125,7 @@ export function MessageLayout({
               onClick={secondaryButton?.onAction}
```
ioanabrooks, line 125:
```suggestion
              onClick={secondaryButton.onAction}
```

packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
```diff
@@ -130,7 +137,7 @@ export function MessageLayout({
               onClick={primaryButton?.onAction}
```
ioanabrooks, line 137:
```suggestion
              onClick={primaryButton.onAction}
```

packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
```diff
@@ -118,7 +125,7 @@ export function MessageLayout({
               onClick={secondaryButton?.onAction}
               style={styles.secondaryButton}
             >
-              {secondaryButton?.title}
+              {secondaryButton.title}
             </Button>
           )}
           {hasPrimaryButton && (
```
ioanabrooks, line 131: Curious about why `hasPrimaryButton` is needed, couldn't this be inferred from the `primaryButton` prop? Same question for `hasSecondaryButton` and `hasRenderableImage`. Seems like we could get into a scenario where the boolean value is true but the actual button/image props are missing?  Perhaps add a comment on `UseMessageProps` for future reference, to outline why these are needed.

reesscot (approved):

wlee221 (approved):

packages/react/src/components/InAppMessaging/MessageLayout/MessageLayout.tsx
calebpollman: Believe this out of scope for this PR. While i agree that we could probably remove the button booleans,  believe the reason they were added was to increase the robustness of our code in the scenario that a customer passes the wrong type of value as `primaryButton` or `secondaryButton`. 

`hasRenderableImage` is not determined by just looking at the image object, and adding a comment abt `UseMessageProps` doesn't quite make sense to me in this file since the hook itself is not used in it, and could be missed if `useMessageProps` is modified in the future leaving an orphaned comment.

PR #3374, opened 2023-02-01
fix(workflow): artifact upload error 

0618:
#### Description of changes

- Cypress test screenshots don't upload for the following error. Fixed it by changing the workflow matrix

> Error: Artifact name is not valid: canary-cypress-error-react/next. Contains the following character:  Forward slash /

- add user email to the account-setting/delete-user example to prevent accidental deletion.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/.github/workflows/build-and-runtime-test.yml
+++ b/.github/workflows/build-and-runtime-test.yml
@@ -17,16 +17,25 @@ jobs:
       fail-fast: false
       matrix:
         # Paths for each example app we want to test build on
-        path:
-          - angular/angularcli
-          - react/cra
-          - react/cra-ts
-          - react/next
-          - react/vite
-          - react/vite3
-          - vue/vite
-          - vue/vite3
-          - vue/vuecli
+        include:
+          - framework: angular
+            build-tool: angularcli
+          - framework: react
+            build-tool: cra
+          - framework: react
+            build-tool: cra-ts
+          - framework: react
+            build-tool: next
+          - framework: react
+            build-tool: vite
+          - framework: react
+            build-tool: vite3
+          - framework: vue
+            build-tool: vite
+          - framework: vue
+            build-tool: vite3
+          - framework: vue
+            build-tool: vuecli
     steps:
       - name: Checkout Amplify UI
         uses: actions/checkout@v3
@@ -36,7 +45,7 @@ jobs:
         uses: ./.github/actions/install-with-retries
         with:
           no-lockfile: true
-          working-directory: ./canary/apps/${{ matrix.path }}
+          working-directory: ./canary/apps/${{ matrix.framework }}/${{ matrix.build-tool }}
       - name: Add Amplify CLI
         run: yarn global add @aws-amplify/cli
       - name: Get CLI versions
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-runtime-test.yml
```diff
@@ -17,16 +17,25 @@ jobs:
...
-          - react/cra-ts
-          - react/next
-          - react/vite
-          - react/vite3
-          - vue/vite
-          - vue/vite3
-          - vue/vuecli
+        include:
+          - framework: angular
+            build-tool: angularcli
```
wlee221, line 22: I do wish GitHub Actions a less wordy syntax for this, but I think this is the best we can do according to [docs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) 🥲

wlee221 (approved):
LGTM!

examples/next/pages/ui/components/account-settings/delete-user/index.page.tsx
```diff
@@ -16,8 +16,9 @@ Amplify.configure(awsExports);
 export default function App() {
   return (
     <Authenticator>
-      {({ signOut }) => (
+      {({ signOut, user }) => (
         <Card width="800px">
+          <h1>Hello {user.attributes.email}</h1>
```
calebpollman, line 21: Just curious how this will prevent accidental deletion?

.github/workflows/build-and-runtime-test.yml
reesscot: That's YML for ya

reesscot: That's YML for ya

reesscot: So this fixes the path how? I'm lost here

examples/next/pages/ui/components/account-settings/delete-user/index.page.tsx
reesscot: I don't think it does, maybe it's just for smoke testing that it's deleted?

reesscot (approved):

.github/workflows/build-and-runtime-test.yml
```diff
@@ -106,7 +115,7 @@ jobs:
         uses: actions/upload-artifact@v3
         if: ${{ failure() && steps.e2e.outcome != 'success' }}
         with:
-          name: canary-cypress-error-${{ matrix.path }}
+          name: canary-cypress-error-${{ matrix.framework }}-${{ matrix.build-tool }}
```
reesscot, line 118: Oh I see, this is where the artifact name is fixed

examples/next/pages/ui/components/account-settings/delete-user/index.page.tsx
calebpollman: If the user was deleted, then this would just break right?

0618: It prevents accidental deletion by telling the user which account they are using 😄 

I accidentally deleted because I didn't know that test account was cached in the browser 😂

.github/workflows/build-and-runtime-test.yml
0618: yep

.github/workflows/build-and-runtime-test.yml
0618: Thanks @wlee221  for the instruction here!

examples/next/pages/ui/components/account-settings/delete-user/index.page.tsx
calebpollman: But that won't actually prevent deletion right?

0618: @calebpollman right

PR #3379, opened 2023-02-01
chore(primitives): ts strict fixes for primitives

dbanksdesign:
#### Description of changes

Fixes to

* usePagination
* useSwitch
* useToggleButtonGroup
* ToggleButtonGroup

#### Issue #, if available

#1018 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Pagination/usePagination.ts
+++ b/packages/react/src/primitives/Pagination/usePagination.ts
@@ -36,8 +36,8 @@ export const usePagination = (
     }
   }, [currentPage]);
 
-  const onChange = React.useCallback((newPage: number, _: number) => {
-    setCurrentPage(newPage);
+  const onChange = React.useCallback((newPageIndex?: number) => {
+    setCurrentPage(newPageIndex ?? 0);
   }, []);
 
   return {
--- a/packages/react/src/primitives/Pagination/usePagination.ts
+++ b/packages/react/src/primitives/Pagination/usePagination.ts
@@ -37,7 +37,9 @@ export const usePagination = (
   }, [currentPage]);
 
   const onChange = React.useCallback((newPageIndex?: number) => {
-    setCurrentPage(newPageIndex ?? 0);
+    if (typeof newPageIndex === 'number') {
+      setCurrentPage(newPageIndex);
+    }
   }, []);
 
   return {
--- a/packages/react/src/primitives/SwitchField/useSwitch.ts
+++ b/packages/react/src/primitives/SwitchField/useSwitch.ts
@@ -18,9 +18,7 @@ interface UseSwitch {
 export const useSwitch = (props: UseSwitchProps): UseSwitch => {
   const { onChange, isChecked, defaultChecked, isDisabled } = props;
   const isControlled = typeof isChecked !== 'undefined';
-  const [isOn, setIsOn] = useState(
-    isControlled ? isChecked : defaultChecked ?? false
-  );
+  const [isOn, setIsOn] = useState(isControlled ? isChecked : !!defaultChecked);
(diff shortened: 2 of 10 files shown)


packages/react/src/primitives/Pagination/usePagination.ts
```diff
@@ -36,8 +36,8 @@ export const usePagination = (
     }
   }, [currentPage]);
 
-  const onChange = React.useCallback((newPage: number, _: number) => {
```
reesscot, line 39: Why was there an unused parameter here?

packages/react/src/primitives/SwitchField/useSwitch.ts
```diff
@@ -18,7 +18,9 @@ interface UseSwitch {
 export const useSwitch = (props: UseSwitchProps): UseSwitch => {
   const { onChange, isChecked, defaultChecked, isDisabled } = props;
   const isControlled = typeof isChecked !== 'undefined';
-  const [isOn, setIsOn] = useState(isControlled ? isChecked : defaultChecked);
+  const [isOn, setIsOn] = useState(
+    isControlled ? isChecked : defaultChecked ?? false
```
calebpollman, line 22: Rather than use nullish coalescing here, would it be possible to coerce `defaultChecked` to a boolean?

calebpollman (commented):
Thanks for doing this, left some questions

packages/react/src/primitives/Pagination/usePagination.ts
```diff
@@ -36,8 +36,8 @@ export const usePagination = (
...
-    setCurrentPage(newPage);
+  const onChange = React.useCallback((newPageIndex?: number) => {
```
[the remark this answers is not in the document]
dbanksdesign: Yea I might need to rethink this. It is weird which parts are potentially undefined in which places..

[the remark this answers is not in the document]
dbanksdesign: Ok redid the logic a bit. TBH it is still not great because we have a single `PaginationItem` component that renders all parts of the Pagination component and does a bit too much, and the event handlers were really weird. We should do a larger refactor, but at least it makes TS happy now :P

packages/react/src/primitives/types/pagination.ts
```diff
@@ -140,5 +140,5 @@ export interface PaginationItemProps
    * @description
    * Triggered every time the item is clicked.
    */
-  onClick?: (newPageIndex?: number, prevPageIndex?: number) => void;
+  onClick?: () => void;
```
calebpollman, line 143: Is this interface exposed to customers or internal only?

packages/react/src/primitives/Pagination/usePagination.ts
```diff
@@ -36,8 +36,8 @@ export const usePagination = (
...
+    setCurrentPage(newPageIndex ?? 0);
```
calebpollman, line 40: Maybe we could just not call `setCurrentPage` if `newPageIndex` is `undefined`? Seems like passing `0` could have unintended consequences, but totally defer to your judgement here

dbanksdesign: Oh yea good call

packages/react/src/primitives/types/pagination.ts
dbanksdesign: Internal only, PaginationItem is not exported

calebpollman (approved):
🚢

reesscot (approved):
LGTM

PR #3381, opened 2023-02-01
docs(fix): update tabs demo to be uncontrolled

zchenwei:
#### Description of changes

Docs demo should not be controlled otherwise the copied code is not workable out of the box. This confused me either if I do not look into it carefully. Also, it does not make sense to have a select to update the current index because you can switch to a disabled tab via it 🥴 

#### Issue #, if available

Kapture 2023-02-01 at 12 44 46

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/docs/src/pages/[platform]/components/tabs/demo.tsx
+++ b/docs/src/pages/[platform]/components/tabs/demo.tsx
@@ -9,7 +9,6 @@ import { demoState } from '@/utils/demoState';
 const propsToCode = (props) => {
   return (
     `<Tabs` +
-    `\n  currentIndex="${props.currentIndex}"` +
     `${props.spacing ? `\n  spacing="${props.spacing}"` : ``}` +
     `\n  justifyContent="${props.justifyContent}"` +
     `${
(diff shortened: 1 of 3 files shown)


zchenwei (commented):
Deleting code is an important skill as a SDE. What do you think Github bot?

docs/src/pages/[platform]/components/tabs/demo.tsx
```diff
@@ -9,7 +9,6 @@ import { demoState } from '@/utils/demoState';
 const propsToCode = (props) => {
   return (
     `<Tabs` +
-    `\n  currentIndex="${props.currentIndex}"` +
```
Comment (0618) [inline docs/src/pages/[platform]/components/tabs/demo.tsx:12]: I do think the dropdown is little confusing, but it's nice to demo the `currentIndex` prop. Is it a better way to demo it?

Comment (zchenwei) [reply docs/src/pages/[platform]/components/tabs/demo.tsx:12]: No we do not want to make it controlled. If you copy the code, it will not work out of the box.

0618 (approved):

PR #3382, opened 2023-02-01
chore(types): better typing of studio code

dbanksdesign:
#### Description of changes

* getOverridesFromVariants
* useNavigateAction
* Moving a test model around

#### Issue #, if available

#1018 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


reesscot:
@dbanksdesign  Action hooks use that file:
image

packages/react/src/hooks/actions/testShared.ts
```diff
@@ -1,26 +0,0 @@
-import type { ModelInit, MutableModel } from '@aws-amplify/datastore';
```
reesscot, line 1: this file is used in action hooks

packages/react/src/studio/getOverridesFromVariants.ts
```diff
@@ -30,8 +30,10 @@ export function getOverridesFromVariants<T>(
   const variantValuesFromProps: VariantValues = Object.keys(props)
     .filter((i) => variantValueKeys.includes(i) && props[i])
     .reduce((acc, key) => {
-      acc[key] = props[key];
-      return acc;
+      return {
```
reesscot, line 33: Is there a test on this change?

packages/react/src/hooks/actions/testShared.ts
dbanksdesign: ah gotcha, moved this so it is with the other test models

packages/react/src/studio/getOverridesFromVariants.ts
dbanksdesign: There are tests already for this functionality. This change is purely to make TS happy because it likes object spreading, but not arbitrary string index accessing for some reason. AFAIK the previous code and this code are equivalent.

packages/react/src/hooks/actions/testModels/todo.ts
```diff
@@ -12,6 +12,9 @@ export class Todo {
   readonly updatedAt?: string;
   constructor(init: ModelInit<Todo, TodoMetaData>) {
     this.name = init.name;
+    // this is only a test model and this is to make TS happy about
+    // the id attribute being set on the constructor
+    this.id = Date.now().toString();
```
dbanksdesign, line 17: Someone who knows DataStore / TS better than I do, is this fine? TS was complaining:

CleanShot 2023-02-02 at 09 32 03@2x

wlee221: If `id` is something that's always present through DataStore actions, you can add a `!` post-fix:

```ts
readonly id!: string
```

But someone who knows DataStore better than I do can check if this is okay 😄

dbanksdesign: Works! Thanks!

PR #3369, opened 2023-01-30
test(docs): lint MDX

0618:
#### Description of changes

**Issue**: Without Linters, it's 1) easy to introduce bug without known (e.g. forget to import variables https://github.com/aws-amplify/amplify-ui/pull/3339), 2) code issues in code blocks 3) code style inconsistent

**Fix**: The goal of this PR is to 

1. Lint all MDX fiiles under `docs/`
2. Add linter to prevent issue at build time

Lint config references:

- [eslint-mdx](https://github.com/remarkjs/remark-lint)
- [remarklint](https://github.com/remarkjs/remark-lint)

#### Description of how you validated changes

- Tested on localhost
- ToC https://docs-lint-mdx.da9impzjamkt4.amplifyapp.com/
- Ran `build` for both success and failed lint.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated
- [ ] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3389, opened 2023-02-02
refactor(docs): convert ThemeExample from mdx to tsx

0618:
#### Description of changes

This PR is a part of breaking down of the big PR https://github.com/aws-amplify/amplify-ui/pull/3369

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/src/components/ThemeExample.mdx
```diff
@@ -1,25 +0,0 @@
...
-import { Fragment } from '@/components/Fragment';
-import { ThemeAlert } from '@/components/ThemeAlert';
-import { GITHUB_REPO_FILE } from '@/data/links';
-import { DesignTokenIcon } from '@/components/Icons';
-
-### Theme
-
-<ThemeAlert />
-
-You can customize the appearance of all {props.component} components in your application with a [Theme](/theming).
```
0618, line 11: `{props.component}` would trigger linting errors, so convert the mdx to tsx

zchenwei (dismissed):

zchenwei (commented):
LGTM

docs/src/components/ThemeExample.tsx
```diff
@@ -0,0 +1,25 @@
+import { Link, View } from '@aws-amplify/ui-react';
+import { ThemeAlert } from '@/components/ThemeAlert';
+import { GITHUB_REPO_FILE } from '@/data/links';
+import { DesignTokenIcon } from '@/components/Icons';
+
+export const ThemeExample = (props) => (
+  <>
+    ### Theme
```
calebpollman, line 8: Asking out of my own ignorance around how our docs are complied but is markdown syntax supported within our `.tsx` files?

dbanksdesign: I was going to ask this as well...

calebpollman: Am curious what the linting error is? It may be a rule that doesn't make sense for us and could be turned off

0618: !! Good catch! Fixed!

0618: > Am curious what the linting error is? It may be a rule that doesn't make sense for us and could be turned off

Yea, I was thinking about that too. Then I decided to convert to tsx is because 1) this is the only .mdx file we have in the docs/components/, 2) we may convert all the .mdx files to .tsx in the future

calebpollman: Still curious what is the actual rule/error? For context, linting libraries tend to be opinionated and some times those opinions do not align what our needs are

reesscot: I think it's a rule from https://github.com/aws-amplify/amplify-ui/pull/3369

0618: > Still curious what is the actual rule/error? For context, linting libraries tend to be opinionated and some times those opinions do not align what our needs are

We would have this error

> Expected an assignment or function call and instead saw an expression.eslint[no-unused-expressions](https://eslint.org/docs/rules/no-unused-expressions)

image

If you pull https://github.com/aws-amplify/amplify-ui/pull/3390 to local, I think you should be able to reproduce it.

reesscot:
I think we need a design doc on mitigation of MDX errors before we make any of these changes.

calebpollman:
> I think we need a design doc on mitigation of MDX errors before we make any of these changes.

Agree with this comment. We may need to consider other options, linting is helpful but may not be the solution we need.

PR #3394, opened 2023-02-07
chore(canary:angular): remove codelyzer and lint

0618:
#### Description of changes

**Issue**: codelyzer depends on @angular/core@9.0.0. We need to Move away from codelyzer, so that we don't have to depend on  @angular/core@9.0.0.
**Fix**: Remove Angular canary's lint (tslint) so that we don't need codelyzer.
**TODO**: if needed, we can use eslint instead. 

Ran `ng serve`, `ng build` on local

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


ioanabrooks (commented):
I have concerns about removing `tslint` without providing an alternative. `@angular-eslint/schematics` claims to generate eslint rules based on  existing tslint config, have we looked at that?

reesscot:
@ioanabrooks While I agree that generally we would like to have linters on our code, this is a canary app. I think it's legitimate to reduce the dependencies we have to maintain on the canary apps, since they are not our production code.

reesscot (approved):

ioanabrooks (approved):
Approving with the understanding we will revisit this when we take on canary work later this quarter.

PR #3395, opened 2023-02-07
chore(file-uploader): Move edit icon in between picture and text

ErikCH:
#### Description of changes
Moved edit pencil icon

Before:
image

After:

image

#### Issue #, if available

Based on Asana and this [discussion](https://github.com/aws-amplify/amplify-ui/pull/3349#issuecomment-1404326011)

Updated snapshot

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated

### DIFF
--- a/packages/react/src/components/Storage/FileUploader/UploadTracker/index.tsx
+++ b/packages/react/src/components/Storage/FileUploader/UploadTracker/index.tsx
@@ -64,12 +64,18 @@ export function UploadTracker({
             {name}
           </Text>
         </View>
+        {showEditButton ? (
+          <Button onClick={onStartEdit} size="small" variation="link">
+            <VisuallyHidden>Edit file name {file.name}</VisuallyHidden>
+            <IconEdit aria-hidden fontSize="medium" />
+          </Button>
+        ) : null}
         <Text as="span" className={ComponentClassNames.FileUploaderFileSize}>
           {humanFileSize(size, true)}
         </Text>
       </>
     ),
-    [name, size]
+    [file.name, name, onStartEdit, showEditButton, size]
   );
 
   const Actions = useCallback(() => {
@@ -143,12 +149,6 @@ export function UploadTracker({
           <View className={ComponentClassNames.FileUploaderFileImage}>
             {icon}
           </View>
-        ) : null}
-        {showEditButton ? (
-          <Button onClick={onStartEdit} size="small" variation="link">
-            <VisuallyHidden>Edit file name {file.name}</VisuallyHidden>
-            <IconEdit aria-hidden fontSize="medium" />
-          </Button>
         ) : null}
 
         {/* Main View */}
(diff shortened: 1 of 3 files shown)


reesscot:
What if we put the edit icon to the right of text? I think that's actually what people are editing right? It's hurting my brain to have it before the text, as it feels like it's editing the preview image.

ErikCH:
> What if we put the edit icon to the right of text? I think that's actually what people are editing right? It's hurting my brain to have it before the text, as it feels like it's editing the preview image.

The problem with having it after the file name, is the file name can be very large or very short. So the pencil won't align vertically correctly when you have multiple files.

We could move it back over before the the size instead. It will still move around slightly depending on the size, but it will be a lot less noticeable. 

image

What do you think @dbanksdesign and @reesscot

reesscot:
> > What if we put the edit icon to the right of text? I think that's actually what people are editing right? It's hurting my brain to have it before the text, as it feels like it's editing the preview image.
> 
> The problem with having it after the file name, is the file name can be very large or very short. So the pencil won't align vertically correctly when you have multiple files. We could set a maximum width for the file name text, but then you might have the pencil icon really far away from the text, depending on the users viewport.
> 
> We could move it back over before the the size instead. It will still move around slightly depending on the size, but it will be a lot less noticeable. Or we could set a maximum width for file size too.
> 
> image
> 
> What do you think @dbanksdesign and @reesscot

Yeah, I like it how you're showing it in the screenshot. Right before the size, but after the text.

ErikCH:
> 

OK! Done moved

packages/react/src/components/Storage/FileUploader/UploadTracker/index.tsx
```diff
@@ -64,12 +64,18 @@ export function UploadTracker({
...
         <Text as="span" className={ComponentClassNames.FileUploaderFileSize}>
           {humanFileSize(size, true)}
         </Text>
       </>
     ),
-    [name, size]
+    [file.name, name, onStartEdit, showEditButton, size]
   );
 
   const Actions = useCallback(() => {
```
ioanabrooks, line 81: Just noticed that we are wrapping child components in `useCallback`, usually this hook is useful for handlers/functions, not components, and should be used judiciously as it could actually cause performance issues.
This comment is pertinent to the refactor work, not this PR.

ioanabrooks (approved):

packages/react/src/components/Storage/FileUploader/UploadTracker/index.tsx
ErikCH: Thanks @ioanabrooks  ! Will take a note of it

dbanksdesign (approved):
LGTM! :shipit:

reesscot (approved):

PR #3400, opened 2023-02-08
chore: ts strict fixes for ui theme

dbanksdesign:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/ui/src/theme/types/style-dictionary.d.ts
+++ b/packages/ui/src/theme/types/style-dictionary.d.ts
@@ -1,0 +1,26 @@
+declare module 'style-dictionary/lib/utils/references/usesReference.js' {
+  // based on the typing specified by style-dictionary:
+  // https://github.com/amzn/style-dictionary/blob/main/lib/utils/references/usesReference.js
+  export default function usesReference(value: string): boolean;
+}
+
+
+declare module 'style-dictionary/lib/utils/deepExtend.js' {
+  // based on the typing specified by style-dictionary:
+  // https://github.com/amzn/style-dictionary/blob/main/lib/utils/deepExtend.js#L18-L22
+  export default function deepExtend<T>(
+    objects: (object | undefined)[],
+    collision?: Function,
+    path?: string[]
+  ): T;
+}
+
+
+declare module 'style-dictionary/lib/utils/flattenProperties.js' {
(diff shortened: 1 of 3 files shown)


0618 (approved):

packages/ui/src/theme/utils.ts
```diff
@@ -65,7 +65,7 @@ export function isDesignToken(value: unknown): value is WebDesignToken {
   return isObject(value) && has(value, 'value');
 }
 
-export function isShadowToken(value: unknown): value is ShadowValue {
+export function isShadowToken(value: unknown): value is ShadowValue & object {
```
calebpollman, line 68: Fairly certain that the definition of `ShadowValue` can resolve to either an `object` of shadow props or `string`. Not fully related to this PR but maybe we should rename this to something more explicit

packages/ui/src/theme/types/style-dictionary.d.ts
```diff
@@ -0,0 +1,26 @@
+declare module 'style-dictionary/lib/utils/references/usesReference.js' {
+  // based on the typing specified by style-dictionary:
+  // https://github.com/amzn/style-dictionary/blob/main/lib/utils/references/usesReference.js
+  export default function usesReference(value: string): boolean;
```
reesscot, line 4: Is there potential for drift on these types? I'm assuming these types cannot be exported from style-dictionary, is that correct?

reesscot (approved):

wlee221 (approved):

packages/ui/src/theme/utils.ts
dbanksdesign: good point, `isShadowObject` it is

packages/ui/src/theme/types/style-dictionary.d.ts
dbanksdesign: That is correct, something if I ever get time to do I would like to rewrite the whole thing in TS and expose the correct types with better exports so we don't have to import specific files. There are a lot of projects that do this so I promise not to break any of these paths :)

reesscot (approved):

calebpollman (approved):

wlee221 (approved):

PR #3415, opened 2023-02-11
chore: ts strict authenticator

dbanksdesign:
#### Description of changes

Turned on TS strict and went to town on the Authenticator. 39 errors in src/components/Authenticator and now there are only 7 TS strict errors left! These changes should not have any outward affect, hopefully. 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Authenticator/Authenticator.tsx
+++ b/packages/react/src/components/Authenticator/Authenticator.tsx
@@ -51,7 +51,7 @@ export function AuthenticatorInternal({
   services,
   socialProviders,
   variation,
-}: AuthenticatorProps): JSX.Element {
+}: AuthenticatorProps): JSX.Element | null {
   const { route, signOut, user } = useAuthenticator(
     ({ route, signOut, user }) => [route, signOut, user]
   );
@@ -66,7 +66,7 @@ export function AuthenticatorInternal({
   });
 
   const value = React.useMemo(
-    () => ({ components: { ...defaultComponents, ...customComponents } }),
+    () => ({ ...defaultComponents, ...customComponents }),
     [customComponents]
   );
 
--- a/packages/react/src/components/Authenticator/ConfirmSignIn/ConfirmSignIn.tsx
+++ b/packages/react/src/components/Authenticator/ConfirmSignIn/ConfirmSignIn.tsx
@@ -20,14 +20,9 @@ export const ConfirmSignIn = ({
   const { isPending } = useAuthenticator((context) => [context.isPending]);
   const { handleChange, handleSubmit } = useFormHandlers();
 
-  const {
-    components: {
-      ConfirmSignIn: {
-        Header = ConfirmSignIn.Header,
-        Footer = ConfirmSignIn.Footer,
-      },
-    },
-  } = useCustomComponents();
+  const { ConfirmSignIn: components } = useCustomComponents();
+  const Header = components?.Header ?? ConfirmSignIn.Header;
+  const Footer = components?.Footer ?? ConfirmSignIn.Footer;
 
   return (
     <RouteContainer className={className} variation={variation}>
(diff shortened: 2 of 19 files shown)


packages/react/src/components/Authenticator/Authenticator.tsx
```diff
@@ -66,7 +66,7 @@ export function AuthenticatorInternal({
   });
 
   const value = React.useMemo(
-    () => ({ components: { ...defaultComponents, ...customComponents } }),
```
dbanksdesign, line 69: This `components` object felt unnecessary, we just unwrap it anyways, so I got rid of it. Also I would have liked this to do a deep merging of objects because they are nested with every part being optional, meaning although we are merging onto `defaultComponents` if `customComponents` has `SignIn: {}` it will wipe out the default `SignIn.Header` and `SignIn.Footer`. Not going to spend too much time worrying if we are going to have better component slots.

packages/react/src/components/Authenticator/ConfirmSignIn/ConfirmSignIn.tsx
```diff
@@ -20,14 +20,9 @@ export const ConfirmSignIn = ({
...
-    components: {
-      ConfirmSignIn: {
-        Header = ConfirmSignIn.Header,
-        Footer = ConfirmSignIn.Footer,
-      },
-    },
-  } = useCustomComponents();
+  const { ConfirmSignIn: components } = useCustomComponents();
+  const Header = components?.Header ?? ConfirmSignIn.Header;
+  const Footer = components?.Footer ?? ConfirmSignIn.Footer;
```
dbanksdesign, line 25: Hopefully this is readable, but we are grabbing the `ConfirmSignIn` and assigning it to `components` then do some optional chaining and nullish coalescing. Reusing this pattern for other components.

packages/react/src/components/Authenticator/ConfirmSignUp/ConfirmSignUp.tsx
```diff
@@ -96,7 +91,6 @@ const DefaultHeader = () => {
 };
 
 ConfirmSignUp.Header = DefaultHeader;
-
-ConfirmSignUp.Footer = function Footer(): JSX.Element {
+ConfirmSignUp.Footer = function Footer() {
```
dbanksdesign, line 94: Dunno how everyone feels about implicit return types, but felt right here.

packages/react/src/components/Authenticator/RouteContainer/types.ts
```diff
@@ -1,6 +1,6 @@
 export type RouteProps = {
-  className: string;
-  variation: 'default' | 'modal';
+  className?: string;
+  variation?: 'default' | 'modal';
```
dbanksdesign, line 3: These are being passed from the Authenticator and the AuthenticatorProps mark these as optional.

packages/react/src/components/Authenticator/Router/Router.tsx
```diff
@@ -11,9 +11,11 @@ import { ConfirmResetPassword, ResetPassword } from '../ResetPassword';
 import { isSignInOrSignUpRoute } from '../utils';
 import { RouterProps } from './types';
 
-type RouteComponent = (props: Omit<RouterProps, 'children'>) => JSX.Element;
+type RouteComponent = (
+  props: Omit<RouterProps, 'children'>
+) => JSX.Element | null;
 
-function RenderNothing(): JSX.Element {
+function RenderNothing() {
```
dbanksdesign, line 18: A return type here? Really?

packages/react/src/components/Authenticator/Router/types.ts
```diff
@@ -1,3 +1,3 @@
 import { RouteProps } from '../RouteContainer';
 
-export type RouterProps = { hideSignUp: boolean } & RouteProps;
+export type RouterProps = { hideSignUp?: boolean } & RouteProps;
```
dbanksdesign, line 3: Authenticator passes this and it is optional.

dbanksdesign (commented):
BEHOLD!

dbanksdesign:
Closing in favor of moving this code to the react-core code.

PR #3427, opened 2023-02-15
chore(docs): update .env.example

0618:
#### Description of changes

Add some fake values so that users can use the example value.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/.env.example
```diff
@@ -1,4 +1,4 @@
-DOCSEARCH_DOCS_APP_ID=
-DOCSEARCH_DOCS_API_KEY=
-DOCSEARCH_DOCS_INDEX_NAME=
-SITE_URL=
+DOCSEARCH_DOCS_APP_ID="FAKE_VALUE"
+DOCSEARCH_DOCS_API_KEY="FAKE_VALUE"
+DOCSEARCH_DOCS_INDEX_NAME="FAKE_VALUE"
+SITE_URL="http://localhost:3000"
```
reesscot, line 4: Can we add a note that this is used in the sitemap?

0618: It's not only for `sitemap` and it's easy to search out where `SITE_URL` is used. Also, the `#` comment in `.env` somehow stops working recently when I use the `e2e/.env` (It didn't throw good error message, so took me some time to figure out. So, I didn't add any comment for the `.env` file in this PR)

reesscot (approved):

calebpollman (approved):

PR #3433, opened 2023-02-17
fix(example/react-native): filter unused internal packages from metro package lookup

calebpollman:
#### Description of changes
Prevent metro bundler start up from breaking due to the lookup of `.DS_Store/package.json` by resolving only used internal packages

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/examples/react-native/metro.config.js
+++ b/examples/react-native/metro.config.js
@@ -8,6 +8,9 @@ const INTERNAL_DEPENDENCY_FLAG = '*';
 // only 'packages' for now, but can be extended for use with other directories
 const INTERNAL_DIRECTORY_ROOTS = ['packages'];
 
+// internal packages used by the RN example
+const INTERNAL_USED_PACKAGES = ['ui', 'react-core', 'react-native'];
+
 const EXAMPLE_APP_PACKAGE_JSON = require('./package.json');
 const EXAMPLE_APP_ROOT = __dirname;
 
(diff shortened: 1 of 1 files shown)


reesscot (approved):
Thank you for fixing this!

examples/react-native/metro.config.js
```diff
@@ -8,6 +8,9 @@ const INTERNAL_DEPENDENCY_FLAG = '*';
 // only 'packages' for now, but can be extended for use with other directories
 const INTERNAL_DIRECTORY_ROOTS = ['packages'];
 
+// internal packages used by the RN example
+const INTERNAL_USED_PACKAGES = ['ui', 'react-core', 'react-native'];
```
wlee221, line 12: Tiny random nit: what do you think about just `INTERNAL_PACKAGES`?

calebpollman: Semantically that would include `vue`, `angular` and `react` right?

wlee221: That's fair, I was thinking as "Internal Packages" as packages customers do not consume directly. I'm good to keep this name to be explicit then.

wlee221 (approved):

examples/react-native/metro.config.js
calebpollman: Oh i get it, yeah i would be more concerned if this was not inside a config file that is already using similar language

PR #3454, opened 2023-02-21
fix: TS Strict errors on toggleButton primitive 

reesscot:
#### Description of changes

This fixes TS Strict errors in the `ToggleButtonGroup` primitive. The root of the issue was that we return undefined to `onChange` from the standard `ToggleButton` primitive (see https://codesandbox.io/p/sandbox/falling-browser-f09yqb?file=%2Fsrc%2FApp.tsx&selection=%5B%7B%22endColumn%22%3A11%2C%22endLineNumber%22%3A24%2C%22startColumn%22%3A11%2C%22startLineNumber%22%3A24%7D%5D), but that wasn't properly reflected in the types. Note that undefined is only returned when someone uses `ToggleButton` inside a `ToggleButtonGroup` without setting a value, which is an incorrect use we cannot prevent.

Validated that all ToggleButton demos work correctly:
CleanShot 2023-03-03 at 11 18 59

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/primitives/ToggleButtonGroup/useToggleButtonGroup.tsx
```diff
@@ -1,7 +1,9 @@
+import { isString } from '@aws-amplify/ui';
 import * as React from 'react';
 import { isFunction } from '@aws-amplify/ui';
 
-import { ToggleButtonProps, ToggleButtonGroupProps } from '../types';
+import { isFunction } from '../shared/utils';
```
reesscot, line 5:
```suggestion
```

dbanksdesign (dismissed):

packages/react/src/primitives/types/toggleButton.ts
```diff
@@ -23,5 +23,5 @@ export interface ToggleButtonProps extends ButtonProps {
    * @description
    * Handles changes to the current value when using the ToggleButton as a controlled component
    */
-  onChange?: (value: string) => void;
+  onChange?: (value: string | undefined) => void;
```
ioanabrooks, line 26:
```suggestion
  onChange?: (value?: string) => void;
```
we shoul dbe consistent with our other components type defs; https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/types/expander.ts#L33

packages/react/src/primitives/ToggleButtonGroup/useToggleButtonGroup.tsx
```diff
@@ -42,19 +44,19 @@ export const useToggleButtonGroup = ({
...
-          value === buttonValue && !isSelectionRequired ? null : buttonValue
-        );
-      },
-      [onChange, value, isSelectionRequired]
-    );
+  const handleExclusiveChange = React.useCallback<
+    (buttonValue: string | undefined) => void
+  >(
+    (buttonValue) => {
+      if (!isFunction(onChange) || Array.isArray(value)) {
```
calebpollman, line 51: Could be wrong but adding the `Array.isArray(value)` condition reads like different behavior from what we currectly have?

packages/react/src/primitives/types/toggleButtonGroup.ts
```diff
@@ -34,5 +34,5 @@ export interface ToggleButtonGroupProps
    * @description
    * Handles changes to the current value when using the ToggleButtonGroup as a controlled component
    */
-  onChange: (value: string | string[]) => void;
+  onChange: (value: string | Array<string | undefined> | undefined) => void;
```
calebpollman, line 37: Nit: prefer shorthand `Array` syntax
```suggestion
  onChange?: (value: string | (string | undefined)[] | undefined) => void;
```

packages/react/src/primitives/types/toggleButton.ts
reesscot: General TS guidance is to not have [optional callback parameters,](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks) so we should probably update the other component type definitions.

calebpollman (approved):

PR #3460, opened 2023-02-22
feat(ui-react-core): add useAuthenticatorComponents and useAuthenticatorProps hooks

calebpollman:
#### Description of changes
Adds `useAuthenticatorComponents` and `useAuthenticatorProps` hooks, utils and types to replace `useAuthenticatorRoute`

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react-core/src/Authenticator/hooks/useAuthenticatorProps/constants.ts
```diff
@@ -0,0 +1,31 @@
+import { CommonRouteSelectorKeys, RouteSelectorKeys } from './types';
+
+const COMMON_SELECTOR_KEYS: CommonRouteSelectorKeys[] = [
```
ioanabrooks, line 3: These are all related to input and form handling, which is what the commonality is, do you think we can be more specific here and call it `FORM_SELECTOR_KEYS`? Though I'm not sold on the selector bit, why selector?

packages/react-core/src/Authenticator/hooks/useAuthenticatorProps/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+  ...rest
+}: Props) {
+  return { ...rest, handleBlur, handleChange, handleSubmit };
+}
+
+const PROPS_RESOLVERS: RoutePropsResolvers = {
+  confirmResetPassword: translateHandlerProps,
+  confirmSignIn: (props) => ({
+    ...translateHandlerProps(props),
+    challengeName: props.user.challengeName!,
```
ioanabrooks, line 49: Would love to see a note for current and future maintainers explaining the confidence behind using the non-null assertion operator.

packages/react-core/src/Authenticator/hooks/useAuthenticatorProps/useAuthenticatorProps.ts
```diff
@@ -0,0 +1,36 @@
...
+  route,
+}: UseAuthenticatorPropsParams<Route>): UseAuthenticatorRouteProps<Route> {
+  // `useAuthenticator` outputs all values from the `Authenticator` state machine
+  // with only a subset of those values required by each `route`. In order to
+  // prevemt extra calculations when interacting with `xstate` a selector param
+  // is used to only return the `props` required by the current `route`
+  const selector = useMemo(() => getRouteMachineSelector(route), [route]);
+  const selectedProps = useAuthenticator(selector);
+
+  // do not memoize, `selectedProps` does not maintain a stable reference
```
ioanabrooks, line 32: Thanks for adding these comments!

packages/react-core/src/Authenticator/hooks/useAuthenticatorComponents/__tests__/useAuthenticatorComponent.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { capitalize } from '@aws-amplify/ui';
+
+import { RenderNothing } from '../../../../components';
+import { COMPONENT_ROUTE_KEYS } from '../../constants';
+import { DEFAULTS } from '../__mock__/components';
+
+import { useAuthenticatorComponents } from '..';
+
+describe('useAuthenticatorComponents', () => {
```
ioanabrooks, line 10: From the hook implementation it seems it can take in overrides, can we add a test for that as well? Bonus covering an unhappy path scenario and verify it is handled gracefully.

calebpollman: The unhappy path test scenarios are covered by the utility that handles them. IMO do not need redundant tests but more than happy to discuss

packages/react-core/src/Authenticator/hooks/useAuthenticatorProps/constants.ts
calebpollman: `error` and `isPending` are related to state machine "states" and are not directly tied to form handling, `updateBlur` and `updateForm` should be removed once validation is moved out of the state machine entirely. The state machine itself should have no knowledge of form or input events, but there is only so much tech debt that can be handled rn.

 Additionally,  `COMMON` scales for other keys that may be added in the future, for example if we were to add `user`. 

The usage of `SELECTOR` is because these are keys that we use for the selector param when calling `useAuthenticator`

packages/react-core/src/Authenticator/hooks/useAuthenticatorProps/utils.ts
calebpollman: The issue here isn't so much the non-null assertion but is the way that the state machine and Amplify JS are typed but i don't disagree with you in terms of adding a comment.
```suggestion
    // `challengName` is always provided on `user` for the `confirmSignIn` route
    challengeName: props.user.challengeName!,
```

packages/react-core/src/Authenticator/hooks/useAuthenticatorProps/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+  ...rest
+}: Props) {
+  return { ...rest, handleBlur, handleChange, handleSubmit };
+}
+
+const PROPS_RESOLVERS: RoutePropsResolvers = {
+  confirmResetPassword: translateHandlerProps,
+  confirmSignIn: (props) => ({
+    ...translateHandlerProps(props),
+    // `challengName` is always provided on `user` for the `confirmSignIn` route
```
calebpollman, line 49:
```suggestion
    // `challengeName` is always provided on `user` for the `confirmSignIn` route
```

ioanabrooks (dismissed):
LGTM! Thanks for tackling the types on this one and for adding more tests!

ioanabrooks (approved):

calebpollman:
Closing this for the time being

PR #3461, opened 2023-02-22
chore(docs): fixing console errors and warnings

dbanksdesign:
#### Description of changes

Fixing all the warnings and errors that come up in our docs site:

* Add a mock Amplify config to appease Auth
* Fix our preloads of fonts
* Adds `data/image` to `img-src` because Geo code is being executed which tries to add a 1x1 webp image to test for webp support
* Fixes our expander use to quiet our orw

| Before | After |
| - | - |
| CleanShot 2023-02-22 at 14 43 34@2x | CleanShot 2023-02-22 at 14 42 59@2x |

Note: the 403 will go away in prod

Also the Authenticator will show a generic network error if you try to log in:

CleanShot 2023-02-22 at 14 51 14@2x

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/docs/src/pages/_document.page.tsx
+++ b/docs/src/pages/_document.page.tsx
@@ -23,7 +23,7 @@ const getCSPContent = (context: Readonly<HtmlProps>) => {
       style-src 'self' 'unsafe-inline';
       font-src 'self' data:;
       frame-src 'self' ${ANALYTICS_CSP.all.frame.join(' ')} *.youtube.com;
-      img-src 'self' blob: ${ANALYTICS_CSP.all.img.join(' ')};
+      img-src 'self' blob: data: ${ANALYTICS_CSP.all.img.join(' ')};
       connect-src 'self' *.shortbread.aws.dev ${ANALYTICS_CSP.all.connect.join(
         ' '
       )} https://*.algolia.net https://*.algolianet.com;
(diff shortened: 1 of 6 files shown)


calebpollman (dismissed):
LGTM

docs/src/pages/_document.page.tsx
```diff
@@ -41,7 +41,7 @@ const getCSPContent = (context: Readonly<HtmlProps>) => {
       ...ANALYTICS_CSP.all.frame,
       ...ANALYTICS_CSP.prod.frame,
     ].join(' ')} *.youtube.com;
-    img-src 'self' blob: ${[
+    img-src 'self' blob: data: ${[
```
reesscot, line 44: Can we lock this down to that specific data image?

dbanksdesign: Great question! Unfortunately, no. It is a schema and you can't put anything behind the `:`. In general it sounds like data schema is bad to allowlist, BUT it sounds like maybe it isn't so bad for `img-src` : https://security.stackexchange.com/questions/94993/is-including-the-data-scheme-in-your-content-security-policy-safe/95011#95011

We could root cause this and fix whatever Geo code has a side effect that is running, but that sounds like a lot of work.

ioanabrooks (dismissed):
Thank you for doing this!

reesscot (approved):
LGTM

calebpollman (approved):

PR #3464, opened 2023-02-23
FileUploader strings refactor

dbanksdesign:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
+++ b/packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
@@ -132,6 +132,7 @@ export function FileUploader({
     [setFileStatuses]
   );
 
+  const { error } = displayText;
   const errorCallback = useCallback(
     (index: number) => {
       return (err: string) => {
@@ -141,7 +142,7 @@ export function FileUploader({
           const updatedStatus = {
             ...prevStatus,
             fileState: 'error' as FileState,
-            fileErrors: displayText.error(err.toString()),
+            fileErrors: error(err.toString()),
           };
 
           prevFileStatuses[index] = updatedStatus;
@@ -152,7 +153,7 @@ export function FileUploader({
         if (typeof onError === 'function') onError(err);
       };
     },
-    [onError, setFileStatuses, displayText]
+    [onError, setFileStatuses, error]
   );
 
   const onPause = useCallback(
--- a/packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
+++ b/packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
@@ -46,25 +46,23 @@ export function FileUploader({
   const {
     dropFilesText,
     browseFilesText,
-    errorText,
-    filesUploadedText,
+    getErrorText,
+    getFilesUploadedText,
     clearButtonText,
-    remainingFilesText,
-    uploadingText,
-    maxFilesErrorText,
+    getRemainingFilesText,
+    getUploadingText,
+    getMaxFilesErrorText,
     doneButtonText,
-    pausedText,
+    getPausedText,
     pauseText,
     resumeText,
     extensionNotAllowedText,
     uploadSuccessfulText,
-    uploadButtonText,
-  } = useMemo(() => {
-    return {
-      ...defaultFileUploaderDisplayText,
-      ...overrideDisplayText,
-    };
-  }, [overrideDisplayText]);
+    getUploadButtonText,
+  } = {
+    ...defaultFileUploaderDisplayText,
+    ...overrideDisplayText,
+  };
 
   // File Previewer loading state
   const [isLoading, setLoading] = useState(false);
@@ -157,7 +155,7 @@ export function FileUploader({
           const updatedStatus = {
             ...prevStatus,
             fileState: 'error' as FileState,
-            fileErrors: errorText(err.toString()),
+            fileErrors: getErrorText(err.toString()),
           };
 
           prevFileStatuses[index] = updatedStatus;
@@ -168,7 +166,7 @@ export function FileUploader({
         if (typeof onError === 'function') onError(err);
       };
     },
-    [onError, setFileStatuses, errorText]
+    [onError, setFileStatuses, getErrorText]
   );
 
   const onPause = useCallback(
@@ -406,12 +404,12 @@ export function FileUploader({
             {uploadButtonComponent}
           </UploadDropZone>
         }
-        filesUploadedText={filesUploadedText}
+        getFilesUploadedText={getFilesUploadedText}
         clearButtonText={clearButtonText}
-        remainingFilesText={remainingFilesText}
-        uploadButtonText={uploadButtonText}
-        uploadingText={uploadingText}
-        maxFilesErrorText={maxFilesErrorText}
+        getRemainingFilesText={getRemainingFilesText}
+        getUploadButtonText={getUploadButtonText}
+        getUploadingText={getUploadingText}
+        getMaxFilesErrorText={getMaxFilesErrorText}
         doneButtonText={doneButtonText}
         fileStatuses={fileStatuses}
         isLoading={isLoading}
@@ -425,8 +423,8 @@ export function FileUploader({
         {fileStatuses?.map((status, index) => (
           <UploadTracker
             pauseText={pauseText}
-            pausedText={pausedText}
-            uploadingText={uploadingText}
+            getPausedText={getPausedText}
+            getUploadingText={getUploadingText}
             resumeText={resumeText}
             extensionNotAllowedText={extensionNotAllowedText}
             uploadSuccessfulText={uploadSuccessfulText}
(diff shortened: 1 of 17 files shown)


packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
```diff
@@ -33,6 +34,7 @@ export function FileUploader({
   accessLevel,
   variation = 'drop',
   isResumable = false,
+  displayText: _displayText,
```
calebpollman, line 37: Maybe `overrideDisplayText`?

packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
```diff
@@ -280,13 +289,13 @@ export function FileUploader({
...
             ? undefined
-            : translate('Extension not allowed'),
+            : displayText.extensionNotAllowed,
         };
 
         setFileStatuses(newFileStatuses);
       };
     },
-    [fileStatuses, setFileStatuses]
+    [fileStatuses, setFileStatuses, displayText]
```
calebpollman, line 298: Prefer the specific values used within a hook to be used in dep arrays where possible, e.g. `displayText.extensionNotAllowed`. That way if for some reason `displayText` is re-evaluated it won't trigger a recalculation here unless the actual value of  `displayText.extensionNotAllowed` updates

packages/react/src/components/Storage/FileUploader/UploadTracker/UploadTracker.tsx
```diff
@@ -56,7 +57,7 @@ export function UploadTracker({
   const showEditButton =
     fileState === FileState.INIT ||
     (fileState === FileState.ERROR &&
-      errorMessage === translate('Extension not allowed'));
+      errorMessage === displayText.extensionNotAllowed);
```
calebpollman, line 60: This has nothing to do with your PR but it's kind of weird to me that we would check against the `errorMessage` for a specific value to determine if we should render the edit button. Maybe we should have more diverse error states on `FileState`, cc @reesscot

packages/react/src/components/Storage/FileUploader/displayText.ts
```diff
@@ -0,0 +1,35 @@
...
+  dropFiles: 'Drop files here or',
+  pause: 'Pause',
+  resume: 'Resume',
+  uploadSuccessful: 'Uploaded successfully',
+  paused(percentage: number): string {
+    return `Paused: ${percentage}%`;
+  },
+};
+
+export type DisplayText = typeof defaultDisplayText;
```
calebpollman, line 35: Thinking we should probably proactively name this `FileUploaderDisplayText` to prevent type namespace conflicts as we extend this pattern in the future

calebpollman (commented):
Great start! Left minor feedback 😅

Part of me wonders if we should only have `displayText` on the `FileUploader` itself and just pass appropriately named string props to the subcomponents. This would require us to ensure that we evaluate all the string values inside `FileUploader` (could be in a util) but the pro would be that we can guarantee better logic around not passing unnecessary props used just to evaluate display values, for example the usage of `percentage` in `UploadMessage`, and reduce the amount of logic in the subcomponents themselves

packages/react/src/components/Storage/FileUploader/UploadTracker/UploadTracker.tsx
dbanksdesign: Yea I had a note of this too.

packages/react/src/components/Storage/FileUploader/displayText.ts
```diff
@@ -0,0 +1,35 @@
...
+    return `Cannot choose more than ${count} ${
+      count === 1 ? 'file' : 'files'
+    }. Remove files before updating`;
+  },
+  error(message: string): string {
+    return message;
+  },
+  doneButton: 'Done',
+  clearButton: 'Clear all',
+  extensionNotAllowed: 'Extension not allowed',
```
ioanabrooks, line 24: * The unit tests should also be updated to use `defaultDisplayText` instead of hardcoded values.
* I noticed we are not passing these strings through the translation module anymore. Is that intentional? Wondering if it could be a breaking change for customers who added their own translations.

packages/react/src/components/Storage/FileUploader/FileUploader/types.ts
```diff
@@ -13,4 +14,5 @@ export interface FileUploaderProps {
   shouldAutoProceed?: boolean;
   showImages?: boolean;
   variation?: 'drop' | 'button';
+  displayText?: Partial<DisplayText>;
```
ioanabrooks, line 17: It would be great DX if we updated the docs when making changes to the API.

dbanksdesign: I was going to handle docs in a separate PR because I want to generate docs from the source. This is on a feature branch so won't affect prod

packages/react/src/components/Storage/FileUploader/displayText.ts
dbanksdesign: Correct, and yes. When I add the docs I will add a section on using the Amplify.i18n module

packages/react/src/components/Storage/FileUploader/displayText.ts
dbanksdesign: Done

packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
```diff
@@ -368,17 +378,45 @@ export function FileUploader({
         </VisuallyHidden>
       </>
     ),
-    [isLoading, onFileChange, hasMultipleFiles, accept]
+    [isLoading, onFileChange, hasMultipleFiles, accept, displayText]
```
reesscot, line 381:
```suggestion
    [isLoading, onFileChange, hasMultipleFiles, accept, displayText.browseFiles]
```

packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
```diff
@@ -41,6 +43,13 @@ export function FileUploader({
     );
   }
 
+  const displayText = useMemo(() => {
+    return {
+      ...defaultFileUploaderDisplayText,
+      ...overrideDisplayText,
+    };
+  }, [overrideDisplayText]);
```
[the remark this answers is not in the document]
dbanksdesign: Good point, let me test that use-case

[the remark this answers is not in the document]
dbanksdesign: I did a quick example of toggling strings and it works, although the `useMemo` here is unnecessary because changing the `displayText` prop will always cause a re-render and will always run the memo function right? 

```jsx
import { Button, FileUploader } from '@aws-amplify/ui-react'; // IGNORE
import { useState } from 'react';

const es = {
  dropFilesText: 'yoink filez',
  uploadButtonText: (count) => `YOINK ${count} filez`
}

const en = {
  dropFilesText: 'yeet filez',
  uploadButtonText: (count) => `yeet ${count} filez`
}

export const DefaultFileUploaderExample = () => {
  const [text, setText] = useState(en);
  
  return (
    <>
    <Button onClick={() => {setText(en)}}>En</Button>
    <Button onClick={() => {setText(es)}}>Es</Button>
    <FileUploader
      accessLevel="public"
      displayText={text}
      provider="fast" // IGNORE
    />
    </>
  );
};

```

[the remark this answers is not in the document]
calebpollman: Changing the `displayText` prop in your example would cause the useMemo to rerun because the objects, while static, ultimately are different object instances. 

We probably don't need to memoize here imo, in particular if we update the `FileUploader` internals to evaluate all the display text in the parent (this file) we would just be passing strings which do not benefit from memoization

packages/react/src/components/Storage/FileUploader/displayText.ts
```diff
@@ -0,0 +1,35 @@
...
+    return `Upload ${count} ${count === 1 ? 'file' : 'files'}`;
+  },
+  maxFilesErrorText(count: number): string {
+    return `Cannot choose more than ${count} ${
+      count === 1 ? 'file' : 'files'
+    }. Remove files before updating`;
+  },
+  errorText(message: string): string {
+    return message;
+  },
```
calebpollman, line 21: Sorry to bring this up so late, but wondering if we should prefix the functions with `get` (`getErrorText`). Would be easier to discern which of these values are just strings and which are functions that return a string. What do you think?

dbanksdesign: Makes sense to me

dbanksdesign: Done!

calebpollman (approved):

packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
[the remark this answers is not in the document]
dbanksdesign: Makes sense, removed the useMemo here

reesscot (approved):

PR #3469, opened 2023-02-24
chore(Vue): Update Vue TS 

ErikCH:
#### Description of changes
This adds types for `useAuth` and `useAuthenticator` and removes some `any`

Reduces the number of times `useAuthenticator` reactive state needs to rerun. 

Verified all tests pass, added `useAuth` test.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/vue/src/components/authenticator.vue
```diff
@@ -107,6 +107,12 @@ unsubscribeMachine = service.subscribe((newState) => {
     });
     hasInitialized.value = true;
   }
+  const facadeValues = getServiceFacade({
+    send,
+    state: newState,
+  });
+
+  updateValues(facadeValues);
```
ErikCH, line 115: Now we will rely on on the xState subscription to update state, instead of a watch. This will reduce the number of times we update the `useAuthenticator` reactive state, and increase performance.

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,29 @@
...
 
 const service = ref() as Ref<InterpretService>;
-const useAuthenticatorValue = reactive({
-  ...facade,
-  send: '' as unknown,
-  state: '' as unknown,
-}) as any;
 
+/** @deprecated For internal use only */
 export const useAuth = (serv?: InterpretService) => {
```
ErikCH, line 10: Eventually this will be completly removed, and we will no longer directly use `state` or `send`. And rely on the facades like we do for `React`

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,29 @@
...
 
-  const { state, send } = useAuth();
+export const useAuthenticator = () => useAuthInternal;
 
-  const facadeValues = getServiceFacade({ send, state: state.value });
+export function updateValues(facadeValues: AuthenticatorServiceFacade) {
   for (const key of Object.keys(facade)) {
     //@ts-ignore
-    useAuthenticatorValue[key] = facadeValues[key];
+    useAuthInternal[key as keyof typeof useAuthInternal] =
```
ErikCH, line 26: I'm close in removing this ts-ignore, but still having issues with null.

examples/vue/src/pages/ui/components/authenticator/custom-sign-up-fields/index.vue
```diff
@@ -31,7 +31,9 @@ const services = {
   <authenticator initial-state="signUp" :services="services">
     <template v-slot:sign-up-fields>
       <authenticator-sign-up-form-fields />
-      <amplify-check-box :errorMessage="validationErrors.acknowledgement" />
+      <amplify-check-box
+        :errorMessage="validationErrors.acknowledgement as string"
```
ErikCH, line 35: Since `validationErrors.acknowledgement` is a bit of an unusual type `string or string[]` we have to typecast this as a string. Not sure what sort of version bump we should do for this. @reesscot  @calebpollman @wlee221

ErikCH: This does match Reacts implementation, it also has to cast

packages/vue/src/composables/useAuth.ts
calebpollman: Is this a public export?

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,29 @@
...
-    createValues();
-  });
-  return useAuthenticatorValue;
-};
-
-export const useAuthenticator = useInternalAuthenticator;
-
-function createValues() {
-  if (!service.value) return;
+const useAuthInternal = reactive<AuthenticatorServiceFacade>({
```
calebpollman, line 17: Is it a standard naming convention in Vue to prefix non-hooks with `use`?

packages/vue/src/composables/useAuth.ts
calebpollman: Obviously not certain on this, but this reads to me as an upstream type issue with `facade`

packages/vue/src/composables/useUtils.ts
```diff
@@ -1,33 +1,34 @@
...
-  route: '',
-  authStatus: '',
-  user: '',
-  validationErrors: [''],
-  codeDeliveryDetails: {
-    val: '',
-    DeliveryMedium: '',
-  },
-  change: (data?: AuthEventData) => null,
+  route: {} as AuthenticatorRoute,
```
calebpollman, line 15: `AuthenticatorRoute` is a string union, not sure why we would use an object here

packages/vue/src/composables/useUtils.ts
```diff
@@ -1,33 +1,34 @@
...
+  authStatus: 'configuring',
+  socialProviders: ['amazon'] as SocialProvider[],
+  initializeMachine: (data?: AuthEventData) => null,
+  user: {} as AmplifyUser,
+  totpSecretCode: null,
+  unverifiedContactMethods: { email: '', phone_number: '' },
+  validationErrors: {} as ValidationError,
+  codeDeliveryDetails: {} as CodeDeliveryDetails,
```
calebpollman, line 23: Couple things here:
- confused where the default values for `socialProviders`, `user`, `unverifiedContactMethods`, `validationErrors` and `codeDeliveryDetails` are coming from
- we shouldn't need to cast anything in this object since it is typed with `AuthenticatorServiceFacade`

calebpollman (commented):
Thanks for taking this on @ErikCH. Left some questions/feedback

packages/vue/src/composables/useAuth.ts
ErikCH: No internal, no public export.  Only `useAuthenticator` is public

packages/vue/src/composables/useAuth.ts
ErikCH: Yes. Here is an example of other composables with the `use` prefix. https://vueuse.org/functions.html

packages/vue/src/composables/useUtils.ts
ErikCH: OK, I'll remove the cast and add the route in.

packages/vue/src/composables/useUtils.ts
ErikCH: I'll remove the casts, except for `AmplifyUser` since that is a large object, with a lot of required fields.

packages/vue/src/composables/useAuth.ts
ErikCH: That's the unusual thing, facade is of type `AuthenticatorServiceFacade` and so is `useAuthInternal` so the keys exactly match. It's something to do with null being a type in the `useAuthInternal[key]`.  I'm up for suggestions. Otherwise maybe I can push this fix to a future PR.

packages/vue/src/composables/useUtils.ts
ErikCH: [b048bfb](https://github.com/aws-amplify/amplify-ui/pull/3469/commits/b048bfb115b769e358fb6b1374bbef8cfa03838d)

packages/vue/src/composables/useUtils.ts
ErikCH: [b048bfb](https://github.com/aws-amplify/amplify-ui/pull/3469/commits/b048bfb115b769e358fb6b1374bbef8cfa03838d)

examples/vue/src/pages/ui/components/authenticator/custom-sign-up-fields/index.vue
wlee221: I believe this is a technically a breaking change, because existing customer codes that don't cast to `string` will be broken.

We could see it as a bug, but since this was a code snippet that is in docs [here](https://ui.docs.amplify.aws/vue/connected-components/authenticator/customization#sign-up-fields), we should be cautious about this.

packages/vue/src/composables/useAuth.ts
calebpollman: Does it need the "For internal use only" comment then?

packages/vue/src/composables/useAuth.ts
calebpollman: Maybe I'm confused here due to my lack of understanding of Vue and `reactive`. Is `useAuthInternal` a function or an object?

packages/vue/src/composables/useAuth.ts
calebpollman: Think we should probably dig in deeper to understand this better, concerned with the casting

packages/vue/src/composables/useUtils.ts
calebpollman: Can you explain where the default values are coming from? For example, why would `unverifiedContactMethods` default to `{ email: '', phone_number: '' }`?

calebpollman: To follow up here on the casting, just because something is hard to type doesn't mean it's acceptable not to provide the necessary values. The point of TS is to add type safety to JS, and if we are casting just to appease the complier than we lose that safety

packages/vue/src/composables/useAuth.ts
ErikCH: I can remove that, but leave the deprecated command, for developers internally

packages/vue/src/composables/useAuth.ts
ErikCH: The `useAuthInternal` is a [reactive proxy object](https://vuejs.org/api/reactivity-core.html#reactive), it encapsulates state that we can share inside vue.

packages/vue/src/composables/useUtils.ts
ErikCH: Understood. The idea here is that were using dummy values so we can use it later in our testing and the keys are used to help loop over the facade and attach them to the the `useAuthenticatorInternal` reactive object.

packages/vue/src/composables/useAuth.ts
ErikCH: [dedc4ed](https://github.com/aws-amplify/amplify-ui/pull/3469/commits/dedc4ed64daa51dca51aff763f6c324d86312c6c)

packages/vue/src/composables/useAuth.ts
ErikCH: To your point, I'll change the name of `useAuthInternal` to `stateAuth` so there is no confusion that `useAuthInternal` is a composable, that are normally functions. [4f31335](https://github.com/aws-amplify/amplify-ui/pull/3469/commits/4f313353e9485f12eca6e2976e3abf6378825670)

reesscot:
Let's hold off on TS type changes in the Vue package, given we will be making changes with component slots shortly.

ErikCH:
Will wait for this TypeScript updates until some future enhancements are added

PR #3476, opened 2023-02-26
fix(ui): Add missing Dutch translations

buremer:
#### Description of changes

Addresses some missing translations in Authenticator for the Dutch language.

#### Issue #, if available

Fixes: https://github.com/aws-amplify/amplify-ui/issues/3477

Had to use this in an external project. That's how I found out that some are missing and fixed them first in that project directly, before creating a PR here.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


reesscot (approved):
Thank you for your contribution!

ioanabrooks (dismissed):

calebpollman (changes requested):
Thanks for adding these @buremer

Since some of these keys are not supported in our existing translations will need to discuss internally before we can merge

buremer:
@calebpollman No problem! Just to mention, these keys come straight from the Cognito Api call in this format and have been tested using the `I18n.putVocabularies` from `aws-amplify` package.

packages/ui/src/i18n/dictionaries/authenticator/nl.ts
```diff
@@ -3,6 +3,8 @@ import { AuthenticatorDictionary } from './types';
 export const nlDict: AuthenticatorDictionary = {
   'Account recovery requires verified contact information':
     'Accountherstel vereist geverifieerde contactgegevens',
+  'Attributes did not conform to the schema: email: The attribute is required\n':
```
calebpollman, line 6: Unfortunately we do not have control over this message and cannot include this key in our translations

packages/ui/src/i18n/dictionaries/authenticator/nl.ts
```diff
@@ -21,17 +23,30 @@ export const nlDict: AuthenticatorDictionary = {
...
   or: 'of',
   Password: 'Wachtwoord',
+  'Password did not conform with policy: Password must have numeric characters':
+    'Wachtwoord voldoet niet aan de vereisten: Wachtwoord moet nummers bevatten',
+  'Password did not conform with policy: Password must have symbol characters':
+    'Wachtwoord voldoet niet aan de vereisten: Wachtwoord moet symbolen bevatten',
+  'Password did not conform with policy: Password must have uppercase characters':
+    'Wachtwoord voldoet niet aan de vereisten: Wachtwoord moet hoofdletters bevatten',
+  'Password did not conform with policy: Password not long enough':
+    'Wachtwoord voldoet niet aan de vereisten: Wachtwoord is niet lang genoeg',
```
calebpollman, line 45: Similar to the above comment, unfortunately we do not have control over these messages and cannot include the keys in our translations

calebpollman (commented):
Thanks for your contribution and patience on getting this merged @buremer 

As noted in the PR, some of the translation keys are not controlled by Amplify UI (they come directly from [Amazon Cognito](https://aws.amazon.com/cognito/)) so we can not include them in our translations by default. If you can remove those keys we can get this merged asap. 

For your awareness, we are definitely looking in to how to better support custom error messages in the future!

buremer:
Aha! Ok thanks for clarifying @calebpollman. I wasn't sure what you meant with that it's not supported. I've removed the keys for now. If there's more I need to change, please let me know ✌️

reesscot (approved):

Issue #3465, opened 2023-02-23
Close-Icon in SearchField and Autocomplete is not aligned correctly


mrtzdev:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Primitive components

### How is your app built?

Create React App

### What browsers are you seeing the problem on?

Chrome, Firefox, Safari

### Please describe your bug.

The Close-Icon in SearchField and AutoComplete is is not aligned correctly when the size prop is used: 

```
<SearchField label="search" size="small" />
<SearchField label="search" size="large" />
```
<img width="525" alt="search-close-icon" src="https://user-images.githubusercontent.com/6218025/220965653-7780a6e0-1a19-4ffa-8d0d-85fc32857890.png">

```
<Autocomplete
      label="Autocomplete size example"
      size="small"
    /> 
<Autocomplete
    label="Autocomplete size example"
    size="large"
  />

```
<img width="598" alt="autocomplete-close-icon" src="https://user-images.githubusercontent.com/6218025/220965950-6cf724a9-cd2a-46bc-99b3-eb020337db01.png">

### What's the expected behaviour?

The icon should be centered  like this: 

<img width="457" alt="search-close-icon-right" src="https://user-images.githubusercontent.com/6218025/220966477-dbd048d7-cdb0-47d1-9fb7-1a7a607b7b79.png">

### Help us reproduce the bug!

``` 
import { SearchField, Flex } from '@aws-amplify/ui-react';

export const SizeSearchFieldExample = () => (
  <Flex direction="column">
    <SearchField label="search" size="small" />
    <SearchField label="search" />
    <SearchField label="search" size="large" />
  </Flex>
); 


### Code Snippet

```typescript
// Put your code below this line.

```

### Additional information and screenshots

The reason is the style width and height element span .amplify-icon. It uses the css var  --amplify-space-large: 1.5rem;   The IconSearch  uses only width: 1em height: 1em. 
[...]

PR #3478, opened 2023-02-27
fix(react): close icon alignment and size SearchField and Autocomplete

mrtzdev:
#### Description of changes
Aligns the close icon correctly in SearchField and Autocomplete. The size of the icons (IconClose, IconSearch) adapt to the size of the fields: small, default, large. 

**Before:** 
old-field-icons

**After:**

new-fields-icons

**Affected components:**
- SearchField
- Autocomplete
- IconClose (internal)

#### Issue #, if available
#3465 

@calebpollman  This is my first contribution to this project. If I have I have missed something, let me know what I can improve.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Field/FieldClearButton.tsx
+++ b/packages/react/src/primitives/Field/FieldClearButton.tsx
@@ -11,8 +11,8 @@ const FieldClearButtonPrimitive: Primitive<FieldClearButtonProps, 'button'> = (
   { ariaLabel = ariaLabelText, size, ...rest },
   ref
 ) => (
-  <FieldGroupIconButton ariaLabel={ariaLabel} ref={ref} {...rest}>
-    <IconClose size={size} />
+  <FieldGroupIconButton ariaLabel={ariaLabel} size={size} ref={ref} {...rest}>
+    <IconClose />
   </FieldGroupIconButton>
 );
 
--- a/packages/react/src/primitives/Icon/icons/IconClose.tsx
+++ b/packages/react/src/primitives/Icon/icons/IconClose.tsx
@@ -9,13 +9,13 @@ import { InternalIcon } from './types';
  * @internal For internal Amplify UI use only. May be removed in a future release.
  */
 export const IconClose: InternalIcon = (props) => {
-  const { className, ...rest } = props;
+  const { className, size, ...rest } = props;
 
   return (
     <View
       as="span"
-      width="1em"
-      height="1em"
+      width={size || '1em'}
+      height={size || '1em'}
       className={classNames(ComponentClassNames.Icon, className)}
       {...rest}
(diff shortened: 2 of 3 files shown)


packages/react/src/primitives/Icon/icons/IconClose.tsx
```diff
@@ -9,13 +9,13 @@ import { InternalIcon } from './types';
  * @internal For internal Amplify UI use only. May be removed in a future release.
  */
 export const IconClose: InternalIcon = (props) => {
-  const { className, size, ...rest } = props;
```
calebpollman, line 12: The `size` prop is used by other components besides the `FieldClearButton`. Is there a different approach to be taken here?

calebpollman (commented):
Thanks for opening a fix @mrtzdev. The `size` prop of `IconClose` is used by other components, so we may need to either update them or look for a different approach in fixing this. Let me know if I can help!

packages/react/src/primitives/Icon/icons/IconClose.tsx
dbanksdesign: I think we can leave it here, but the change above I believe will still work to fix the issue.

packages/react/src/primitives/Field/FieldClearButton.tsx
```diff
@@ -11,8 +11,8 @@ const FieldClearButtonPrimitive: Primitive<FieldClearButtonProps, 'button'> = (
   { ariaLabel = ariaLabelText, size, ...rest },
   ref
 ) => (
-  <FieldGroupIconButton ariaLabel={ariaLabel} ref={ref} {...rest}>
-    <IconClose size={size} />
+  <FieldGroupIconButton ariaLabel={ariaLabel} size={size} ref={ref} {...rest}>
```
dbanksdesign, line 14: Great catch! This is definitely how it should have been done.

dbanksdesign (commented):
Thank you for this! I think if you remove the edits from IconClose, but keep the others and it should work.

packages/react/src/primitives/Field/FieldClearButton.tsx
mrtzdev: Okay thanks! I remove the edits from the IconClose. If we don't add the size prop on IconClose in FealdClearButton, it should be fine.

calebpollman (approved):
LGTM. Thanks for making the changes @mrtzdev

dbanksdesign (approved):
LGTM!

simply the best

mrtzdev:
:thumbsup:  Thanks. Awesome UI library !

PR #3485, opened 2023-02-28
chore(uploader): remove unneeded `hasMultipleFiles` prop

reesscot:
#### Description of changes
This PR removes the external `hasMultipleFiles` API due to being redundant with the `maxFileCount` API. There isn't a scenario I could think of where these two props would need to be used together. `hasMultipleFiles={false}` is equivalent to `maxFileCount={1}` and `hasMultipleFiles={true}` is equivalent to `maxFileCount={2 to ♾)` or just leaving `maxFileCount` undefined.

This PR also renames `maxSize` to `maxFileSize` to be consistent with `maxFileCount` prop

* Smoke tested the example app and docs
* Ran unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


examples/next/pages/ui/components/storage/file-uploader/index.page.tsx
```diff
@@ -14,8 +14,7 @@ export default function FileUploaderEmail() {
       onSuccess={onSuccess}
       acceptedFileTypes={['image/*']}
       accessLevel="public"
-      hasMultipleFiles={true}
```
reesscot, line 17: This is actually a good example of why hasMultipleFiles is redundant. We shouldn't have to define both props here to get multi selection in the file uploader picker.

dbanksdesign (approved):

packages/react/src/components/Storage/FileUploader/hooks/useFileUploader/__tests__/useFileUploader.test.tsx
```diff
@@ -85,11 +85,11 @@ describe('useFileUploader', () => {
     expect(result.current.fileStatuses.length).toEqual(1);
   });
 
-  it('returns 1 and updates file statuses when not hasMultipleFiles and target length is 1', () => {
+  it('returns 1 and updates file statuses when not allowMultipleFiles and target length is 1', () => {
```
calebpollman, line 88: Off subject: Can we make a note to clean up these test descriptions

packages/react/src/components/Storage/FileUploader/hooks/useFileUploader/types.ts
```diff
@@ -17,3 +17,14 @@ export interface UseFileUploader extends DragActionHandlers {
...
 }
+
+export interface UseFileUploaderProps {
+  maxFileSize: number;
+  acceptedFileTypes: string[];
+  allowMultipleFiles: boolean;
+  isLoading: boolean;
+}
+
+export interface UseFileUploader {
```
calebpollman, line 28: Nit: we use hook PascalCase for the return type for a fair amount of our existing hooks. Using `UseFileUploader` to type the function breaks convention

packages/react/src/components/Storage/FileUploader/hooks/useFileUploader/useFileUploader.ts
```diff
@@ -4,17 +4,12 @@ import { checkMaxSize, returnAcceptedFiles } from '@aws-amplify/ui';
 import { Files, FileState, FileStatuses } from '../../types';
 import { UseFileUploader } from './types';
 
-export default function useFileUploader({
-  maxSize,
+export const useFileUploader: UseFileUploader = ({
```
calebpollman, line 7: Prefer named functions for larger utils, and it better aligns with existing conventions as noted in the earlier comment about usage of PascalCase and hooks names

packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
```diff
@@ -63,6 +62,10 @@ export function FileUploader({
     ...overrideDisplayText,
   };
 
+  const allowMultipleFiles =
```
calebpollman, line 65: Wondering if we can remove this entirely and just infer what we need from `maxFileCount`? Might be easier to do later once the file state lifecycle has been refactored

calebpollman (approved):
Approved, but left some questions/feedback

packages/react/src/components/Storage/FileUploader/hooks/useFileUploader/__tests__/useFileUploader.test.tsx
reesscot: Yes, will address in a subsequent pr

packages/react/src/components/Storage/FileUploader/hooks/useFileUploader/types.ts
reesscot: good catch, will update

packages/react/src/components/Storage/FileUploader/FileUploader/FileUploader.tsx
reesscot: Yeah, I can revisit when I do the file state lifecycle.

packages/react/src/components/Storage/FileUploader/hooks/useFileUploader/useFileUploader.ts
reesscot: updated

calebpollman (approved):
🚢

PR #3515, opened 2023-03-07
chore(amplify-ui): fix dev commands and clean up scripts

calebpollman:
#### Description of changes
Fix dev commands for web platforms and clean up scripts:
- use `rollup --watch` for `dev` commands
- remove default jest `coverage` generation, improving unit test speeds
- add ESM compatibility checks to `ui` and `ui-react-core`
- align script conventions
- remove _react-shim.js_ in `ui-react`
- remove `tsup` and config files

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


packages/react-native/package.json
```diff
@@ -7,13 +7,15 @@
   "license": "Apache-2.0",
   "scripts": {
     "build": "yarn dist && cp -a src/assets dist",
+    "build:dist": "tsc --project tsconfig.dist.json",
```
reesscot, line 10: For my own understanding, is this a standard react-native command?

packages/react-native/package.json
```diff
@@ -7,13 +7,15 @@
...
     "clean": "rimraf dist",
-    "dev": "yarn dist --watch",
-    "dist": "tsc --project tsconfig.dist.json",
-    "lint": "tsc --noEmit && eslint src",
+    "dev": "yarn build:dist --watch",
```
reesscot, line 12: Assuming we are doing this to avoid building the tests/mock files in watch mode?

packages/react-core/package.json
```diff
@@ -19,13 +19,15 @@
...
     "build": "rollup --config",
-    "build:esm": "tsc --project tsconfig.esm.json",
-    "dev": "yarn build:esm --watch",
-    "lint": "tsc --noEmit && eslint src",
-    "test": "jest --coverage --verbose",
-    "test:watch": "yarn test --watch"
+    "dev": "yarn build --watch",
+    "lint": "yarn typecheck && eslint src",
+    "prebuild": "rimraf dist",
+    "test": "yarn test:unit && yarn test:esm",
```
reesscot, line 26: doesn't need to be addressed here, but should we reverse these so we can pass file paths to jest?

reesscot (dismissed):
LGTM!

packages/react-core/package.json
calebpollman: Think that's what `test:unit` is for, although i personally prefer `test` to just correspond with running jest. Switching the order of commands would allow appending a path, but that feels a bit fragile to me. Alternatively we could add a CI specific test script that runs the ESM check and the unit tests to decouple the two from `test`. Open to discussion 🙂

packages/react-native/package.json
calebpollman: Close 😅 We have a second step in the build command copying the assets directory to the dist folder, which would prevent the `--watch` flag from being appended correctly if we were to just run `yarn build --watch`

packages/react-native/package.json
```diff
@@ -7,13 +7,15 @@
   "license": "Apache-2.0",
   "scripts": {
     "build": "yarn dist && cp -a src/assets dist",
```
calebpollman, line 9:
```suggestion
    "build": "yarn build:dist && cp -a src/assets dist",
```

packages/react-core/package.json
reesscot: I like the idea of putting it in a CI specific test script, since this isn't something that should need to run over and over again locally, and could primarily be caught in PRs

reesscot (approved):

packages/react-core/package.json
calebpollman: Cool cool, so thinking that we would then have something like:
```
"check:esm": "node --input-type=module --eval 'import \"@aws-amplify/ui\"'",
"test": "jest",
"test:ci": "yarn test && yarn check:esm",
"test:watch": "yarn test --watch",
```

Make sense?

wlee221: I like the change you mentioned, let's just be careful we don't break CI with this because `yarn test` for unit tests: 

https://github.com/aws-amplify/amplify-ui/blob/1bf5b330d108fb4eee28f3951fd5b0a45b4547b0/.github/workflows/reusable-e2e.yml#L181

wlee221: I think it'd make sense to do a separate PR for introducing `test:ci`, so that we don't have to worry about changing workflow in this PR

wlee221 (approved):
🚀

PR #3516, opened 2023-03-07
test(workflow): add pptr link checker

0618:
#### Description of changes

This PR is a follow-up for https://github.com/aws-amplify/amplify-ui/pull/3494.
After the AppSec approval, this PR is to enable the link checker on Github Action.

link-checker-github-action

#### Description of how you validated changes

Tested on PoC PR https://github.com/0618/amplify-ui/pull/19
Happy path: https://github.com/0618/amplify-ui/actions/runs/4359021524/jobs/7620504997
Catch bad links: https://github.com/0618/amplify-ui/actions/runs/4359240980/jobs/7620799673

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated
- [ ] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3526, opened 2023-03-08
test(workflow): add pptr link checker

0618:
Original PR https://github.com/aws-amplify/amplify-ui/pull/3516 had to be reverted https://github.com/aws-amplify/amplify-ui/pull/3523 because it report statasCode 308 from internal links as errors ([build](https://github.com/aws-amplify/amplify-ui/actions/runs/4367802459/jobs/7639863231)).

It was because if the returned status is `0`, it would keep using `308`. (Not sure why it's not an issue on forked PoC or local).

#### Description of changes

- same as https://github.com/aws-amplify/amplify-ui/pull/3516
- change `||` to `??` to catch `0`
- 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


docs/scripts/data/constants.ts
```diff
@@ -1,3 +1,3 @@
-export const PROMISE_POOL_CONCURRENCY = 10;
+export const PROMISE_POOL_CONCURRENCY = 5;
```
0618, line 1: Fix the memory leak. 

Since the Github Action Step didn't change after reducing the `PROMISE_POOL_CONCURRENCY` to 5, we don't have to test numbers bigger than 5 to reduce the testing time.

docs/scripts/util/checkLink.ts
```diff
@@ -100,7 +100,7 @@ export async function checkLink(
               ...linkData,
               statusCode,
             })
-          )?.statusCode || statusCode;
+          )?.statusCode ?? statusCode;
```
0618, line 103: This is the major change of this new PR 👿

docs/scripts/data/constants.ts
wlee221: Technically, this isn't a memory leak but is an optimization, right?

wlee221 (approved):
🚢

docs/scripts/util/checkLink.ts
reesscot: Can you please explain what this change does?

0618: Yes. 

If a link is good we would return a `statusCode: 0` ([code](https://github.com/aws-amplify/amplify-ui/blob/main/docs/scripts/util/checkLink.ts#L91)). The problems was, the redirected link got a `0`, and it was `308` before redirected, so `0 || 308` and we got `308`.

We should have been able to filter out the `308` errors when printing errors ([code](https://github.com/aws-amplify/amplify-ui/blob/main/docs/scripts/link-checker-puppeteer.ts#L20)), but since the `SITE_URL` Github Env Variable was not set to be `localhost:3000`, it was not filtered out. 

That's why we got those internal 308 errors which failed the build.

Since we've changed the Github Env Variable, the problem should be resolved.

So, it's actually not necessary to change `||` to `??`.

reesscot (approved):

wlee221 (approved):

PR #3541, opened 2023-03-13
chore: add github actions for status checks

sreeramsama:
#### Description of changes
Add github actions for status checks to use it approval based workflows.
- create-check: Creates a status check of a specified head-sha
- update-check: Updates status of an existing check

Tested against POC to run actions on pull request --> https://github.com/wlee221/amplify-ui/pull/8
[POC / Test checks](https://github.com/wlee221/amplify-ui/actions/workflows/test-checks.yml)

Tested that a workflow can update the status of a check multiple times.
For example, we tested that a check called 'POC check' is updated from 'in_progress' to 'completed'.

#### Checklist

- [x] PR description included

### DIFF
--- a/.github/actions/add-checks/action.yml
+++ b/.github/actions/add-checks/action.yml
@@ -1,0 +1,40 @@
+name: 'Add Status Checks'
+description: 'Add status check on a specified head-sha'
+inputs:
+  name:
+    description: 'Name of status check'
+    required: true
+  status:
+    description: >-
+      Status of check with possible values of 'queued', 'in_progress', 'completed'
+    required: true
+  head-sha:
+    description: 'The commit ID to add the checks to'
+    required: true
+  conclusion:
+    description: >-
+      Conclusion of status check with possible values of 'action_required',
+      'cancelled', 'failure', 'neutral', 'success', 'skipped', 'stale', 'timed_out'.
+      This is only required if status is 'completed'.
+    required: false
+runs:
+  using: 'composite'
+  steps:
+    - uses: actions/github-script@v6
+      env:
+        name: ${{ inputs.name }}
+        head_sha: ${{ inputs.head-sha }}
+        status: ${{ inputs.status }}
+        conclusion: ${{ inputs.conclusion }}
+      with:
+        script: |
+          const { name, head_sha, status, conclusion } = process.env;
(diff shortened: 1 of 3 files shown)


wlee221 (approved):
:shipit:

.github/actions/add-checks/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    - uses: actions/github-script@v6
+      env:
+        name: ${{ inputs.name }}
+        head_sha: ${{ inputs.head-sha }}
+        status: ${{ inputs.status }}
+        conclusion: ${{ inputs.conclusion }}
+      with:
+        script: |
+          const { name, head_sha, status, conclusion } = process.env;
+          const { owner, repo } = context.repo;
```
ioanabrooks, line 32: Prefer process github context values through env vars, even if seemingly within our trust boundary, it's more secure.

.github/actions/add-checks/action.yml
```diff
@@ -0,0 +1,40 @@
...
+  conclusion:
+    description: >-
+      Conclusion of status check with possible values of 'action_required',
+      'cancelled', 'failure', 'neutral', 'success', 'skipped', 'stale', 'timed_out'.
+      This is only required if status is 'completed'.
+    required: false
+runs:
+  using: 'composite'
+  steps:
+    - uses: actions/github-script@v6
```
ioanabrooks, line 23: Prefer referencing github actions through commit sha rather than version (see codeql workflow)

sreeramsama: Updated to use commit sha as in codeql workflow

.github/actions/add-checks/action.yml
wlee221: So technically, this is not prone to typical [script injection](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) attacks, because these contexts do *not* take part in script generation process. `context` is a variable that is provided outside the script generation.

That said, I'm all in for having this in environment variable for the sake of consistency, and this is still better in case `context` variable is malformed (which is a different risk from script injection, but valid). We'll update to use env var here!

ioanabrooks (approved):

wlee221 (approved):

PR #3549, opened 2023-03-15
fix(docs): Fix toc for Style Props page

ioanabrooks:
#### Description of changes

This change fixes an issue where the right side nav (table of contents) is not displayed when selecting Style Props page from the left side menu. Workaround is to refresh the page. 

The reason why this is happening is because the page is static and the [MutationObserver](https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/components/Layout/index.tsx#L50) we use relies on changes to the DOM, which occur when the browser renders a new page or updates an existing page. In the case of static pages, there are no updates to the DOM after the page is initially rendered, so the MutationObserver does not trigger.

Since the Style Props page content and examples are specific to `react`, moved the static content into a fragment, which makes it dynamic content that the observer can pick up.  Verified that the only remaining static page is the CSS variables page which does not have headings.

toc

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are updated


docs/src/components/Layout/index.tsx
```diff
@@ -35,9 +37,9 @@ export default function Page({
     const updateHeaders = debounce(() => {
       setHeadings(
         [
-          ...document
-            .querySelector('#__next')
-            .querySelectorAll('h2[id],h3[id]'),
+          ...document.querySelectorAll(
+            `${nextRootSelector} h2[id], ${nextRootSelector} h3[id]`
+          ),
```
ioanabrooks, line 42: Small optimization, use single `querySelector` instead of chaining two.

docs/src/pages/[platform]/theming/style-props/style-props.react.mdx
```diff
@@ -0,0 +1,169 @@
...
+  SizeThemeTokenExample,
+  TypographyStyles,
+  TypographyThemeTokenExample,
+  OtherStyles,
+} from './examples';
+
+These style props are available on Amplify UI primitive components and can be used to modify the corresponding styles. This allows users to quickly make style adjustments without having to go through CSS or add a theme object. Style props override any conflicting CSS style sheet or theme styling.
+
+There are three ways of passing style props to a component:
+1. Pass any valid CSS property directly (e.g., `backgroundColor="red"`)
```
Comment (ioanabrooks) [inline docs/src/pages/[platform]/theming/style-props/style-props.react.mdx:30]: Added code highlighting to backgroundColor="red"

reesscot (approved):

wlee221 (approved):

0618 (approved):

PR #3562, opened 2023-03-17
Add set-status action to set/update statuses on commit sha

sreeramsama:
#### Description of changes

We will use status API instead of checks API, because manually adding checks have shortcomings including:
- When checks are updated multiple times against the same commit, checks wouldn’t show up properly against the latest workflow.
- Status shows up as an independent update on the PR, while checks are always associated to a workflow.
- The problem is that GitHub would try to associate it even with some unrelated workflows on the PR, which is confusing.

Update: Added validation for required inputs.

POC:
PR: https://github.com/wlee221/amplify-ui/pull/11
Workflow Job: https://github.com/wlee221/amplify-ui/actions/runs/4451753099/jobs/7818767558 

Workflow to test input validation: https://github.com/wlee221/amplify-ui/actions/runs/4473455813/jobs/7860868018

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated


.github/actions/set-status/action.yml
```diff
@@ -0,0 +1,48 @@
+name: 'Set Status'
+description: 'Set status on a specified head-sha'
+inputs:
+  context:
+    description: 'String label of the status'
```
wlee221, line 5: What about just
```suggestion
    description: 'Name of the status'
```
?

wlee221 (approved):

.github/actions/set-status/action.yml
```diff
@@ -0,0 +1,48 @@
...
+        description: ${{ inputs.description }}
+        state: ${{ inputs.state }}
+        sha: ${{ inputs.sha }}
+        target_url: ${{ inputs.target-url }}
+        owner: ${{ github.event.repository.owner.login }}
+        repo: ${{ github.event.repository.name }}
+      with:
+        result-encoding: string
+        script: |
+          const { status_context, description, state, sha, target_url, owner, repo } = process.env;
```
ioanabrooks, line 39: Should we validate the required input values before calling `createCommitStatus`? (since this is a reusable action)

sreeramsama: Added validations for required inputs 
SCR-20230320-lbu

wlee221: Code nit -- what do you think about just returning early inside each if statements? That'd help us write better error messages, and help avoid nested statements.

```ts
const { status_context, description, state, sha, target_url, owner, repo } = process.env;

if (state.length == 0 || !["error", "failure", "pending", "success"].includes(state)) {
  core.setFailed('"state" input cannot be empty.');
  return;
}

if (!["error", "failure", "pending", "success"].includes(state)) {
  core.setFailed('"state" must be a string input with possible value of (...).');
  return;
}

if (context.length <= 0) {
  core.setFailed('"context" input cannot be empty.');
  return;
}

if (sha.length <= 0) {
  core.setFailed('"sha" input cannot be empty.');
  return;
}

if (!input_string.match(/^[0-9a-z]+$/)) {
  core.setFailed('"sha" must be an alphanumeric string.')
}

github.rest.repos.createCommitStatus({
  owner,
  repo,
  sha,
  state,
  context: status_context,
  description: description.length > 0 ? description : undefined,
  target_url: target_url.length > 0 ? target_url : undefined,
});
```

ioanabrooks: +1 to that and also having descriptive and helpful error messages

sreeramsama: Synced up offline with Will and we used custom messages and removed return on each.

wlee221 (approved):
Thanks for working on this! 🚢

PR #3576, opened 2023-03-21
chore: Add husky pre-push check for changeset

ioanabrooks:
#### Description of changes

This change adds a husky `pre-push` script to check for changeset before pushing changes to ui, angular, react, react-core, react-native, vue packages, specifically the `src` folder of these packages (to reduce false positives, andalso  CI jobs or dependabot wouldn't be affected). Additionally release related branches are allowlisted.

Added husky script instead of github action to fail fast instead of in CI.

Validated against my fork

* with changes to ui packages and no changeset
```
❯ git push
ERROR: These changes should result in a version bump but no changeset was found! 
Please add a changeset file by running yarn changeset.
```

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,23 +1,28 @@
 #!/bin/sh
-while read local_ref; do
-  # Check if the local branch name starts with "docs-release-pr", in which case exit successfully
-  if [[ "$local_ref" =~ ^refs/heads/docs-release-pr-.* ]]; then
-    exit 0
-  fi
-  
-  # get a list of all changed files between the local and upstream main branches
-  changed_files=$(git diff --name-only main..HEAD --no-merges)
 
-  # Check if any of the changed files are in the src folder of the angular, react, react-core, react-native, ui or vue packages
-  if echo "${changed_files}" | grep -E '^packages/(angular/projects/ui-angular|react|react-core|react-native|ui|vue)/src/'; then
-    # Check if a changeset file is present and has .md extension
-    if echo "${changed_files}" | grep -E '^\.changeset/.*\.md$'; then
-      exit 0
-    else
-      echo "ERROR: These changes should result in a version bump but no changeset was found! \nPlease add a changeset file by running yarn changeset."
-      exit 1
-    fi
+# Fail fast, exit immediately on any error
+set -e
+
+# Get the name of the current branch
+current_branch=$(git rev-parse --abbrev-ref HEAD)
+
+# Check if the local branch name starts with "docs-release-pr" or "changeset-release/main", in which case exit successfully
+if [[ "$current_branch" =~ ^(docs-release-pr-.*|changeset-release/main)$ ]]; then
+  exit 0
+fi
+
+# Get the list of all changed files between the local and upstream main branches
+changed_files=$(git diff --name-only main..HEAD --no-merges)
+
+# Check if any of the changed files are in the src folder of the angular, react, react-core, react-native, ui or vue packages
+if echo "${changed_files}" | grep -E '^packages/(angular/projects/ui-angular|react|react-core|react-native|ui|vue)/src/'; then
+  # Check if a changeset file is present and has .md extension
+  if echo "${changed_files}" | grep -E '^\.changeset/.*\.md$'; then
+    exit 0
+  else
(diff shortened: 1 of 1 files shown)


.husky/pre-push
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+while read local_ref local_sha remote_ref remote_sha; do
+  # Check if the local branch name starts with "docs-release-pr", in which case exist successfully
```
ioanabrooks, line 3:
```suggestion
  # Check if the local branch name starts with "docs-release-pr", in which case exit successfully
```

0618 (approved):
This is awesome!!! Thanks for adding the check!

.husky/pre-push
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+while read local_ref local_sha remote_ref remote_sha; do
+  # Check if the local branch name starts with "docs-release-pr", in which case exit successfully
+  if [[ "$local_ref" =~ ^refs/heads/docs-release-pr-.* ]]; then
```
wlee221, line 4: We should add `changest-release` too, because `changeset` will try to bump `CHANGELOG`, `package.json` when it creates "Version Packages" PR.

.husky/pre-push
```diff
@@ -0,0 +1,23 @@
...
+    exit 0
+  fi
+  
+  # Get the list of changed files in this commit
+  changed_files="$(git diff --name-only "${local_sha}" "${remote_sha}")"
+
+  # Check if any of the changed files are in the angular, react, react-native, ui or vue packages
+  if echo "${changed_files}" | grep -E '^packages/angular/|^packages/react/|^packages/react-core/|^packages/react-native/|^packages/ui/|^packages/vue/'; then
```
wlee221, line 12: What do you think about targeting `packages/<name>/src`? That way, it would not fail on changes to things like lint configs, tsconfigs, and jest setup that do not changesets.

This will mean there could be false positives, but this might be better than using `--no-verify` on PRs that do not need changesets.

reesscot:
Would this mean I would need separate changesets for every commit that I push up to a feature branch? That feels a little heavy handed. I'd prefer to be able to push up my "fixed stuff" commits and "squash & merge" with a changeset at the end.

.husky/pre-push
ioanabrooks: Love it! Addressed!

.husky/pre-push
ioanabrooks: Good catch, I think we should be good with the change to only target `src` folder though.

ioanabrooks:
> Would this mean I would need separate changesets for every commit that I push up to a feature branch? That feels a little heavy handed. I'd prefer to be able to push up my "fixed stuff" commits and "squash & merge" with a changeset at the end.

Excellent point, addressed by getting diff of all commits on branch.

reesscot (approved):

.husky/pre-push
wlee221: Just checked “Version Packages” PR, it changes src/version.ts so I think we need to add changeset-release/main to the ignorelist 👍

zchenwei: Yeah, it will bump `react/src/version.ts`. 🥴

ioanabrooks: Thanks for checking, added!

.husky/pre-push
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+# Fail fast, exit immediately on any error
+set -e
+
+# Get the name of the current branch
+current_branch=$(git rev-parse --abbrev-ref HEAD)
```
ioanabrooks, line 7: Decided to use git rev-parse instead of previous implementation with local_ref because I ran into some issue with local branch that hasn't been published yet and found local_ref isn't always populated, `git  rev-parse` always returns branch name regardless of whether published or not.

wlee221 (approved):
🛥️

zchenwei (approved):
LGTM

reesscot:
What do you think about allowlisting feature branches as well, since we don't need a changeset till the end for new features. Maybe just looking for a `feature` prefix?

ioanabrooks:
Closing, we don't have a patterns for marking feature branches with a prefix, this change will cause too much friction for feature branches. Fun experiment though.

PR #3615, opened 2023-03-30
chore: StorageManager unit tests

reesscot:
#### Description of changes
Adds unit tests for util functions. Also fixes a very annoying TS Strict bug. Will follow up with a bug on the JS repo.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are updated


packages/react/src/components/Storage/StorageManager/utils/uploadFile.ts
```diff
@@ -8,7 +8,7 @@ interface UploadFileProps {
   isResumable?: boolean;
   progressCallback: (progress: { loaded: number; total: number }) => void;
   errorCallback: (error: string) => void;
-  completeCallback: (event: { key: string }) => void;
+  completeCallback: (event: { key?: string }) => void;
```
reesscot, line 11: This is the correct type actually:
https://github.com/aws-amplify/amplify-js/blob/main/packages/storage/src/types/AWSS3Provider.ts#L82

ioanabrooks (approved):

packages/react/src/components/Storage/StorageManager/utils/uploadFile.ts
calebpollman: Does that mean `event` is just an empty object when there is no value for `key`? Seems like if there was no value ` key` then `event` itself would just be `undefined` be empty but 🤷  

For improved TS practices, what do you think of typing as:
```suggestionx
  completeCallback: (event: { key: string | undefined }) => void;
```

packages/react/src/components/Storage/StorageManager/utils/uploadFile.ts
```diff
@@ -29,12 +29,18 @@ export function uploadFile({
   if (isResumable === true) {
     return Storage.put(fileName, file, {
       level,
-      resumable: isResumable,
+      resumable: true,
```
calebpollman, line 32: If this is intentional (which i think it is), can we please add a comment explaining why we default to `true`? Could be confusing to come across this in the future

packages/react/src/components/Storage/StorageManager/utils/uploadFile.ts
reesscot: Makes sense to me, will update

calebpollman (approved):

PR #3617, opened 2023-03-30
chore: standardize key

dbanksdesign:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/components/Storage/StorageManager/StorageManager/types.ts
+++ b/packages/react/src/components/Storage/StorageManager/StorageManager/types.ts
@@ -52,15 +52,15 @@ export interface StorageManagerProps {
    */
   onUploadError?: (error: string) => void;
   /**
-   * Monitor upload sucess
+   * Monitor upload success
    */
   onUploadSuccess?: (event: { key: string }) => void;
   /**
    * Process file before upload
    */
   processFile?: (
-    storageFile: Required<Pick<StorageFile, 'file' | 'name'>>
-  ) => Required<Pick<StorageFile, 'file' | 'name'>>;
+    storageFile: Required<Pick<StorageFile, 'file' | 'key'>>
+  ) => Required<Pick<StorageFile, 'file' | 'key'>>;
   /**
    * Determines if thumbnails show for image files
    */
--- a/packages/react/src/components/Storage/StorageManager/hooks/useStorageManager/useStorageManager.tsx
+++ b/packages/react/src/components/Storage/StorageManager/hooks/useStorageManager/useStorageManager.tsx
@@ -39,8 +39,8 @@ export function useStorageManager(
     files: defaultFiles.map((file) => {
       return {
         ...file,
-        id: file.s3Key,
-        name: file.s3Key,
+        id: file.key,
+        key: file.key,
         status: FileStatus.UPLOADED,
       };
     }) as StorageFiles,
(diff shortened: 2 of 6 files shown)


packages/react/src/components/Storage/StorageManager/hooks/useStorageManager/useStorageManager.tsx
```diff
@@ -39,8 +39,8 @@ export function useStorageManager(
     files: defaultFiles.map((file) => {
       return {
         ...file,
-        id: file.s3Key,
-        name: file.s3Key,
+        id: file.key,
+        key: file.key,
```
calebpollman, line 43: If `id` and `key` resolve to the same value do we need both?

packages/react/src/components/Storage/StorageManager/StorageManager/types.ts
```diff
@@ -47,20 +47,24 @@ export interface StorageManagerProps {
...
    */
   onUploadSuccess?: (event: { key: string }) => void;
   /**
    * Process file before upload
    */
   processFile?: (
-    storageFile: Required<Pick<StorageFile, 'file' | 'name'>>
-  ) => Required<Pick<StorageFile, 'file' | 'name'>>;
+    storageFile: Required<Pick<StorageFile, 'file' | 'key'>>
+  ) => Required<Pick<StorageFile, 'file' | 'key'>>;
```
calebpollman, line 67: For legibility, what do you think of making a utility type from `Required<Pick<StorageFile, 'file' | 'key'>>`, example:
```
type ProcessFileParams = Required<Pick<StorageFile, 'file' | 'key'>>;

...

processFile?: (params: ProcessFileParams) => ProcessFileParams;
```

packages/react/src/components/Storage/StorageManager/StorageManager/types.ts
```diff
@@ -47,20 +47,24 @@ export interface StorageManagerProps {
...
+   * When a file is removed
+   */
+  onFileRemove?: (file: { key: string }) => void;
   /**
    * Monitor upload errors
    */
   onUploadError?: (error: string) => void;
   /**
-   * Monitor upload sucess
+   * Monitor upload success
```
calebpollman, line 59: 👀

packages/react/src/components/Storage/StorageManager/hooks/useStorageManager/useStorageManager.tsx
dbanksdesign: good question

reesscot: While they are the same value, my thinking was that I didn't want to conflate the `id` we use internally with the `key` for the file. But I'm also ok with merging them.

calebpollman: Merging would make the most sense imo, concern is that it might be confusing to understand why we have both in the future

reesscot (approved):

packages/react/src/components/Storage/StorageManager/hooks/useStorageManager/useStorageManager.tsx
reesscot: We found an edge case where the customers can update the desired key (onProcessFile), which would mutate the id if they are merged. I think we should keep them separate even if they are often the same.

calebpollman: Makes sense, can we add a comment here?

PR #3675, opened 2023-04-07
fix(field): making description show up

dbanksdesign:
#### Description of changes

Field description wasn't showing up for the internal Field component. This fixes that. 

Added tests

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are updated

### DIFF
--- a/packages/react/src/primitives/Field/Field.tsx
+++ b/packages/react/src/primitives/Field/Field.tsx
@@ -50,11 +50,12 @@ const FieldPrimitive: Primitive<FieldPrimitiveProps, typeof Flex> = (
       {...rest}
     >
       {label ? <Label visuallyHidden={labelHidden}>{label}</Label> : null}
-      {descriptiveText ? (
-        <FieldDescription labelHidden={labelHidden}>
-          {descriptiveText}
-        </FieldDescription>
-      ) : null}
+
+      <FieldDescription
+        labelHidden={labelHidden}
+        descriptiveText={descriptiveText}
+      />
+
       {children}
       {errorMessage ? (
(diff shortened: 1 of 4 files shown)


hein-j (dismissed):

packages/react/src/primitives/Field/Field.tsx
```diff
@@ -50,11 +50,12 @@ const FieldPrimitive: Primitive<FieldPrimitiveProps, typeof Flex> = (
       {...rest}
     >
       {label ? <Label visuallyHidden={labelHidden}>{label}</Label> : null}
-      {descriptiveText ? (
```
reesscot, line 53: Wait, why remove this logic? Shouldn't descriptiveText be provided to show the descriptive text?

zchenwei: +1

dbanksdesign: That logic is IN `<FieldDescription>`: https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/Field/FieldDescription.tsx#L15

reesscot: but why did that cause a bug?

dbanksdesign: This didn't cause the bug, the bug was I was passing it as a child instead of prop

reesscot (approved):

hein-j (approved):

ioanabrooks (approved):
Thank you for adding tests!

PR #3711, opened 2023-04-12
ci: refresh lockfile

wlee221:
#### Description of changes

Refreshes `yarn.lock` so that our dependencies are up to date. 

CI, and ran docs locally: 

Screenshot 2023-04-12 at 2 06 55 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


docs/package.json
```diff
@@ -83,7 +83,7 @@
     "react-markdown": "^8.0.3",
     "sitemap-urls": "^3.0.0",
     "ts-jest": "^27",
-    "ts-morph": "^15.1.0",
+    "ts-morph": "^17.0.0",
```
wlee221, line 86: We need to use ts-morph 17 because this PR bumps monorepo `typescript` to `4.9`, and only `ts-morph` 17 is compatible with `4.9`.

Reading through their [changelog](https://github.com/dsherret/ts-morph/blob/latest/packages/ts-morph/CHANGELOG.md), their only breaking change was bumping `typescript` on their end.

docs/tests/__snapshots__/props-table.test.ts.snap
```diff
@@ -90,6 +83,13 @@ exports[`Props Table 1`] = `
                         \\"category\\": \\"AlertProps\\",
                         \\"isOptional\\": true
                     },
+                    \\"heading\\": {
```
wlee221, line 86: Bumping `devDependencies` have changed the order of this prop for some reason... but the content is not changed.

docs/package.json
0618: Where do we bump the monorepo `typescript`? Is it only in the yarn.lock?

wlee221: It's from our monorepo package.json: 

https://github.com/aws-amplify/amplify-ui/blob/0da924cd5ee2884b89b0e184ab38bbd9fb57327a/package.json#L95

Since we use caret there, this gets bumped as we refresh the lockfile. This will be visible in `yarn.lock`.

reesscot (approved):

zchenwei (approved):

0618 (approved):
🚢

PR #3748, opened 2023-04-18
chore(ui-react-core-notifications): add inapp utility code

sreeramsama:
#### Description of changes
Add utility code of InAppMessaging (https://github.com/aws-amplify/amplify-ui/tree/main/packages/react-core/src/InAppMessaging)

Ran unit tests
#### Checklist

- [x] PR description included


packages/react-core-notifications/src/components/index.ts
```diff
@@ -0,0 +1 @@
+export { RenderNothing } from './RenderNothing';
```
calebpollman, line 1: Let's just import this from `@aws-amplify/ui-react-core` where it is used in in-app since `RenderNothing` has uses outside of in-app

packages/react-core-notifications/src/__tests__/__snapshots__/index.spec.ts.snap
```diff
@@ -2,6 +2,10 @@
 
 exports[`@aws-amplify/ui-react-core-notifications exports should match snapshot 1`] = `
```
calebpollman, line 3: Where is the test file that created this?

calebpollman (commented):
Good start! Some high level notes:
- rather than just duplicating the in-app code from `@aws-amplify/ui-react-core`, we should be migrating the code out of `@aws-amplify/ui-react-core`. Since this package and react-core are internal packages it is we can add `@aws-amplify/ui-react-core-notifications` as a dep in both ui-react and ui-react-native without issue. Should also make the diff much easier to read
- the in-app utility code should be in an _InAppMessaging_ directory inside _src/components_. Don't think we need them rn since we only have in-app in this package, but the _context_, _hooks_, etc. directories in _src_ should be used for storing code that is shared by multiple connected components

packages/react-core-notifications/src/__tests__/__snapshots__/index.spec.ts.snap
sreeramsama: Included in PR to add package boilerplate
https://github.com/aws-amplify/amplify-ui/pull/3747/files#diff-ec87130187bf9355dab1a52bf1950550be5768313a68bb64fa90185ea6868097

calebpollman: That makes sense 😄

sreeramsama:
> Good start! Some high level notes:
> 
> * rather than just duplicating the in-app code from `@aws-amplify/ui-react-core`, we should be migrating the code out of `@aws-amplify/ui-react-core`. Since this package and react-core are internal packages we can add `@aws-amplify/ui-react-core-notifications` as a dep in both ui-react and ui-react-native without issue. Should also make the diff much easier to read
> * the in-app utility code should be in an _InAppMessaging_ directory inside _src/components_. Don't think we need them rn since we only have in-app in this package, but the _context_, _hooks_, etc. directories in _src_ should be used for storing code that is shared by multiple connected components

Moved InAppMessaging code out of `react-core` and into `react-core-notifications/src/InAppMessaging/`

calebpollman (approved):
🚢

packages/react-core/src/index.ts
```diff
@@ -20,30 +20,7 @@ export {
...
-  MessageLayout,
-  MessagePayloadStyle,
-  MessageTextAlign,
-  ModalMessageCommonProps,
-  ModalMessageComponent,
-  OnMessageAction,
-  useInAppMessaging,
-  useMessage,
-} from './InAppMessaging';
+export { RenderNothing } from './components';
```
thaddmt, line 23: Not blocking but does this even need to be exported? I can probably go either way but seems like the component doesn't really do much.

thaddmt (approved):
LGTM other than small question

PR #3762, opened 2023-04-21
chore(liveness): add script to prevent liveness from being automatically published to latest

thaddmt:
#### Description of changes
- Change sets `react-liveness` package to `private: true` by default so it is not accidentally published
- Script sets field `private: false` in ui-react-liveness and is called during next publish lifecycle so next tags are produced for liveness
- Follow up pr for releasing liveness separately from our main packages - https://github.com/aws-amplify/amplify-ui/pull/3763

#### Checklist

- [ ] PR description included


scripts/setLivenessPackagePrivate.js
```diff
@@ -0,0 +1,18 @@
...
+
+// Read the package.json file
+const packageJson = require(`../packages/react-liveness/package.json`);
+
+// Set the "private" field to true
+packageJson.private = true;
+
+// Write the updated package.json file
+fs.writeFileSync(
+  './packages/react-liveness/package.json',
```
thaddmt, line 16: `writeFileSync` and `require` have different paths because require is relative and writeFileSync seems to be run from the directory the script is fun from. Not sure if this will work the same in gh action but I tested locally

.github/workflows/reusable-tagged-publish.yml
```diff
@@ -36,6 +36,8 @@ jobs:
           skip-cypress-binary: true # publishing doesn't need cypress
       - name: Build packages
         run: yarn build
+      - name: Set liveness package to be public
```
0618, line 39: Why do we set liveness package to public every time when publishing?

I thought we publish liveness separately?

thaddmt: This is only when doing a tagged publish (like we do for @next) so that we ensure that it is published on each tagged release. We don't want to publish liveness on @latest because we are not confident in our automated testing and the smoke testing is too laborious for each release

wlee221 (approved):

0618 (approved):

PR #3764, opened 2023-04-21
chore(docs): add liveness api reference for react web

thaddmt:
#### Description of changes
- Adds an api ref doc to liveness docs

Screenshot 2023-04-24 at 6 02 20 PM

#### Checklist

- [ ] PR description included

### DIFF
--- a/docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx
+++ b/docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx
@@ -1,5 +1,6 @@
 import { Tabs, TabItem, Table, TableHead, TableRow, TableCell, TableBody, View } from '@aws-amplify/ui-react';
-import { ResponsiveTable, ResponsiveTableCell } from '@/components/ResponsiveTable';
+import ReactPropsTable from '@/components/propsTable/ReactPropsTable';
+import { FACE_LIVENESS_DETECTOR_PROPS, FACE_LIVENESS_DETECTOR_COMPONENTS, FACE_LIVENESS_DETECTOR_CONFIG } from './react-props';
 
 ## Full API Reference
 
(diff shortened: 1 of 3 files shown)


docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx
```diff
@@ -0,0 +1,74 @@
+import { Tabs, TabItem, Table, TableHead, TableRow, TableCell, TableBody, View } from '@aws-amplify/ui-react';
+import { ResponsiveTable, ResponsiveTableCell } from '@/components/ResponsiveTable';
+
+## Full API Reference
+
+Below is the full list of props that can be used with the `FaceLivenessDetector` component. You can also reference the [source code here](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react-liveness/src/components/FaceLivenessDetector/service/types/liveness.ts#L4).
+
+<ResponsiveTable highlightOnHover>
```
Comment (0618) [inline docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx:8]: Would it be better if we don't hard code it? Import the data instead?

docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx
```diff
@@ -0,0 +1,74 @@
...
+    <TableRow>
+      <ResponsiveTableCell label="Name">`disableInstructionScreen`</ResponsiveTableCell>
+      <ResponsiveTableCell label="Description">Optional parameter for the disabling the Start/Get Ready Screen, default: false.</ResponsiveTableCell>
+      <ResponsiveTableCell label="Type">`boolean`</ResponsiveTableCell>
+      <ResponsiveTableCell label="Type">`false`</ResponsiveTableCell>
+    </TableRow>
+    <TableRow>
+      <ResponsiveTableCell label="Name">`config`</ResponsiveTableCell>
+      <ResponsiveTableCell label="Description">Optional parameter for advanced options for the component.</ResponsiveTableCell>
+      <ResponsiveTableCell label="Type">`FaceLivenessDetectorConfig`</ResponsiveTableCell>
```
Comment (ayush987goyal) [inline docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx:57]: We need API ref for this config as well (maybe just below this table itself?)

docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx
Comment (ayush987goyal) [reply docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx:8]: +1 
Perhaps a json structure that can be iterated over so that new additions are no-op from updates here.

Comment (thaddmt) [reply docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx:10]: I looked into iterating the types and didn't really see to do this here that would be a no-op solution. Importing the types still requires building an object with the type and iterating over the keys. But we would still need to manually add a description and the proper human readable typing. What do you all think?

I.e

```
interface ExampleInterface {
  name: string;
  age: number;
  email: string;
}

const example: ExampleInterface = {
  name: "John Doe",
  age: 30,
  email: "john.doe@example.com",
};

for (const key of Object.keys(example)) {
  // Create the react component with description and type here based on the key
}

```

Comment (ayush987goyal) [reply docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx:10]: Do you think we can use something like [TypeDoc](https://github.com/TypeStrong/typedoc) to extract structured information as JSON and then consume? Then the API definition would be the single source of truth and rest we handle via build steps.

https://atanasstoyanov.medium.com/extracting-structured-type-information-2e7480ca9f83

This could be a bigger work scope and hence will not block the change for the same. But we should explore it as it would be helpful for all the components.

ayush987goyal (dismissed):

docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx
Comment (thaddmt) [reply docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx:10]: yeah i think that makes sense! can look into it as a followup but as you said probably bigger than the scope of this pr

docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx
```diff
@@ -0,0 +1,146 @@
+import { Tabs, TabItem, Table, TableHead, TableRow, TableCell, TableBody, View } from '@aws-amplify/ui-react';
+import { ResponsiveTable, ResponsiveTableCell } from '@/components/ResponsiveTable';
+
+## Full API Reference
+
+### FaceLivenessDetectorProps
+
+Below is the full list of props that can be used with the `FaceLivenessDetector` component. You can also reference the [source code here](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react-liveness/src/components/FaceLivenessDetector/service/types/liveness.ts#L4).
+
+<ResponsiveTable highlightOnHover>
```
Comment (calebpollman) [inline docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx:10]: Semi-related to the above. Think you can just the [`ReactPropsTable`](https://github.com/aws-amplify/amplify-ui/blob/main/docs/src/components/propsTable/ReactPropsTable.tsx) component to avoid hard coding

Comment (thaddmt) [reply docs/src/pages/[platform]/connected-components/liveness/full-api.react.mdx]: Switched to using react props table! thanks!

calebpollman (approved):

reesscot (approved):

PR #3771, opened 2023-04-25
chore(liveness): add strict typing for react-liveness

thaddmt:
#### Description of changes
- Added types for blazefaceFaceDetection.ts
- Added types for liveness utils file

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/blazefaceFaceDetection.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/blazefaceFaceDetection.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import * as tf from '@tensorflow/tfjs-core';
 import * as blazeface from '@tensorflow-models/blazeface';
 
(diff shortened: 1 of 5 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/utils/blazefaceFaceDetection.ts
```diff
@@ -67,33 +66,33 @@ export class BlazeFaceFaceDetection extends FaceDetection {
...
+        const { topLeft, bottomRight, probability, landmarks } = prediction;
+
+        const [right, top] = topLeft as [number, number]; // right, top because the prediction is flipped
+        const [left, bottom] = bottomRight as [number, number]; // left, bottom because the prediction is flipped
+        const width = Math.abs(right - left);
+        const height = Math.abs(bottom - top);
+        const rightEye = (landmarks as number[][])[0];
+        const leftEye = (landmarks as number[][])[1];
+        const nose = (landmarks as number[][])[2];
+        const mouth = (landmarks as number[][])[3];
```
calebpollman, line 81: Obviously it would be ideal if the `NormalizedFace` interface would allow us to not have to cast these values, but  get that we cannot because of the type unions in the interface. Since we are not using type predicates here what do you think of adding some meaningfully named utility types for `[number, number]` and `number[][]`?

thaddmt: Yeah I think that makes sense. I think that the types that would make sense here are probably something like:

`type Coordinate = [number, number]`

and `landmarks` being something like `Coordinate[]`

what do you think?

calebpollman: Makes sense to me!

thaddmt:
Docs failures seem unrelated to this change

sreeramsama (dismissed):

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/blazefaceFaceDetection.ts
```diff
@@ -67,33 +66,33 @@ export class BlazeFaceFaceDetection extends FaceDetection {
...
+        const nose = (landmarks as Coordinate[])[2];
+        const mouth = (landmarks as Coordinate[])[3];
+
+        return {
+          top,
+          left,
+          width,
+          height,
+          timestampMs,
+          probability: probability as number,
```
calebpollman, line 89: Maybe I'm misreading, but this line was previously:
```ts
        probability: probability[0] as number,
```

Was the index look up unneccesary?

thaddmt: Good catch, this is supposed to be `probability[0]` though I think this value is not currently used

reesscot (approved):

Issue #3767, opened 2023-04-24
Bug (Authenticator): Missing support for SELECT_MFA_TYPE challenge


PritamDutt:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

### Please describe your bug.

With `Cognito User` MFA configuration set to `Choose preferred delivery method` Authenticator Crashes

> Ref: User Setting in Cognito
![image](https://user-images.githubusercontent.com/31215724/233923619-5560d545-b3fa-40ca-ac32-4e9f5a5c1167.png)

> Ref: Error on Chrome Console
![image](https://user-images.githubusercontent.com/31215724/233925129-7bca3971-5048-454e-86a5-e6aa73edbfe4.png)

> Ref: UI Screen when crash happens
![image](https://user-images.githubusercontent.com/31215724/233924357-13dcb5cb-ed9a-41cc-8794-7458ed621412.png)

### What's the expected behaviour?

User should be presented with available MFA options and pick his/her choice

### Help us reproduce the bug!

- Install amplify 
- Create project using React-UI
- Create User without providing phone
- Configure TOTP MFA
- Update user attributes to include phone, and verify it
- Change MFA Configuration to `Choose preferred delivery method` on Cognito Console
- Try authenticating user

### Code Snippet

```typescript
// Put your code below this line.

```

0618:
Hi @PritamDutt If the user didn't provide a Phone number, the SMS MFA methods should be disabled as the following screenshot.

![image](https://user-images.githubusercontent.com/11983489/234138049-90c3af79-7d06-49d5-9678-7891e502dc3e.png)


Could you provide more details on how to reproduce the error?

0618:
Thanks for the detailed step. I just reproduced the error.

It might be a bug, but needs more investigation.

PritamDutt:
Hi @reesscot,

[...]

PR #3787, opened 2023-04-27
fix(auth): add SELECT_MFA_TYPE error

0618:
#### Description of changes

**Issue**: If the admin force user to set up MFA without specify a type (SMS or TOTP), user would not able to login
**Reason**: 'SELECT_MFA_TYPE' is not supported by Amplify UI yet.
**Fix**: 
1. Add an error to the console for now to help users understand the situation
2. Add a `shouldSelectMfaType` to unblock/reject the sign-in process

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/3767

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/ui/src/machines/authenticator/actors/signIn.ts
+++ b/packages/ui/src/machines/authenticator/actors/signIn.ts
@@ -105,6 +105,11 @@ export function signInActor({ services }: SignInMachineOptions) {
                     target: '#signInActor.setupTOTP',
                   },
                   {
+                    cond: 'shouldSelectMfaType',
+                    actions: ['setUser', 'setChallengeName'],
+                    target: 'rejected',
+                  },
+                  {
                     cond: 'shouldConfirmSignIn',
                     actions: ['setUser', 'setChallengeName'],
                     target: '#signInActor.confirmSignIn',
@@ -548,6 +553,16 @@ export function signInActor({ services }: SignInMachineOptions) {
         },
         shouldSetupTOTP: (_, event): boolean => {
           return isExpectedChallengeName(getChallengeName(event), 'MFA_SETUP');
+        },
+        shouldSelectMfaType: (_, event): boolean => {
+          console.error(
+            `ERROR: 'SELECT_MFA_TYPE' is not supported by Amplify UI yet.
+            Please use only one MFA type or set up the MFA selector using this document: https://docs.amplify.aws/lib/auth/mfa/q/platform/js/#advanced-use-cases/`
+          );
+          return isExpectedChallengeName(
+            getChallengeName(event),
+            'SELECT_MFA_TYPE'
+          );
         },
       },
       services: {


packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -104,6 +104,11 @@ export function signInActor({ services }: SignInMachineOptions) {
                     actions: ['setUser', 'setChallengeName'],
                     target: '#signInActor.setupTOTP',
                   },
+                  {
+                    cond: 'shouldSelectMfaType',
+                    actions: ['setUser', 'setChallengeName'],
+                    target: 'rejected',
+                  },
```
calebpollman, line 111: I get a bit lost in the state machine, can you explain what is happening here?

0618: Sure. `cond` to check the user `shouldSelectMfaType` to be true, then `rejected`. 

The actions are copied from `shouldSetupTOTP` above because we'll only need to change `rejected` to something like `#signInActor.setupMfaType` once we add the feature.

calebpollman: Couple follow up questions?
- in what scenarios is the condition reached?
- Are the actions what should happen for this condition of changing `rejected` in the future?

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -549,6 +554,16 @@ export function signInActor({ services }: SignInMachineOptions) {
...
           return isExpectedChallengeName(getChallengeName(event), 'MFA_SETUP');
         },
+        shouldSelectMfaType: (_, event): boolean => {
+          console.error(
+            `ERROR: 'SELECT_MFA_TYPE' is not supported by Amplify UI yet.
+            Please use only one MFA type or set up the MFA selector using this document: https://docs.amplify.aws/lib/auth/mfa/q/platform/js/#advanced-use-cases/`
+          );
+          return isExpectedChallengeName(
+            getChallengeName(event),
+            'SELECT_MFA_TYPE'
```
calebpollman, line 564: IF we don't support `SELECT_MFA_TYPE` this will always return `false` right?

packages/ui/src/machines/authenticator/actors/signIn.ts
0618: > in what scenarios is the condition reached?

The scenario is in the description or https://github.com/aws-amplify/amplify-ui/issues/3767

The Admin add options from Amplify Console to allow user to choose MFA type, so Amplify UI would get a `SELECT_MFA_TYPE`.

> Are the actions what should happen for this condition or for ease of changing rejected in the future?

the actions are for ease of changing in the future.

reesscot: I don't think we should be factoring in the future changes, let's keep our changes focused on the fix

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -549,6 +554,16 @@ export function signInActor({ services }: SignInMachineOptions) {
         shouldSetupTOTP: (_, event): boolean => {
           return isExpectedChallengeName(getChallengeName(event), 'MFA_SETUP');
         },
+        shouldSelectMfaType: (_, event): boolean => {
+          console.error(
+            `ERROR: 'SELECT_MFA_TYPE' is not supported by Amplify UI yet.
+            Please use only one MFA type or set up the MFA selector using this document: https://docs.amplify.aws/lib/auth/mfa/q/platform/js/#advanced-use-cases/`
+          );
```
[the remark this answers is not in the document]
0618: > Should this be console.error or a just console.warn?

Should be `console.error` because it doesn't the expected result and requires the user to take actions.

> No need for a template literal here, can just be wrapped in double quotes

1. It's easier to read to cut the long line into multiple lines
2. don't have to add `\` or `\n`

> It's unclear to me what following the documentation linked here will provide the user in terms of not coming across this issue. Can you explain it to me?

I'm not sure I understand the question 😂

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -549,6 +554,19 @@ export function signInActor({ services }: SignInMachineOptions) {
         shouldSetupTOTP: (_, event): boolean => {
           return isExpectedChallengeName(getChallengeName(event), 'MFA_SETUP');
         },
+        shouldSelectMfaType: (_, event): boolean => {
+          const isSelectedMfaType = isExpectedChallengeName(
```
reesscot, line 558: This is confusing, why wouldn't this be false if if it's an unexpected challenge name?

reesscot:
Just looking at the screencast, I don't think we want to allow them to be in the "authenticated state" if they have this selected, since we won't have actually done the multifactor auth piece? I think we need to look at where this change needs to happen in the state machine, seems like it needs to block logic and return the user to the signin page with an  a generic error message shown to the user.

packages/ui/src/machines/authenticator/actors/signIn.ts
wlee221: FYI, this condition can be reached without admin doing this directly on AWS console: https://github.com/aws-amplify/amplify-js/issues/4646#issuecomment-1462908573

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -104,6 +104,11 @@ export function signInActor({ services }: SignInMachineOptions) {
                     actions: ['setUser', 'setChallengeName'],
                     target: '#signInActor.setupTOTP',
                   },
+                  {
+                    cond: 'shouldSelectMfaType',
+                    actions: ['setUser', 'setChallengeName'],
```
wlee221, line 109: We shouldn't `setUser` here (or arguably even `challengeName`), since we're intentionally not progressing further if we come across `SELECT_MFA_TYPE`.

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -549,6 +554,19 @@ export function signInActor({ services }: SignInMachineOptions) {
         shouldSetupTOTP: (_, event): boolean => {
           return isExpectedChallengeName(getChallengeName(event), 'MFA_SETUP');
         },
+        shouldSelectMfaType: (_, event): boolean => {
```
wlee221, line 557: `isSelectMFATypeChallenge`, or `isSelectMFAChallenge` would make more sense. 

We would use `shouldSelectMFAType` if we had a mfa selector screen to redirect users to, but we do not yet.

packages/ui/src/machines/authenticator/actors/signIn.ts
wlee221, line 558: nit: `isSelectMFATypeChallenge` or `isSelectMFAChallenge` would make more sense. `selected` sounds like end user already selected something.

0618:
> Just looking at the screencast, I don't think we want to allow them to be in the "authenticated state" if they have this selected, since we won't have actually done the multifactor auth piece? I think we need to look at where this change needs to happen in the state machine, seems like it needs to block logic and return the user to the signin page with an a generic error message shown to the user.
> 
> Can we change this to a Draft PR since it's not ready yet?

That makes sense. I update to keep the user on the sign-in page and show a remote error. Please see the new screencast in the description

packages/ui/src/machines/authenticator/actors/signIn.ts
0618: Done!

packages/ui/src/machines/authenticator/actors/signIn.ts
0618: Right, it would be false if it's an unexpected challenge name(`SELECT_MFA_TYPE`). a.k.a. It would be false if it's not `SELECT_MFA_TYPE`

packages/ui/src/machines/authenticator/actors/signIn.ts
wlee221: I think `cond` name still needs to be updated 👍

0618: Ok. Done!

packages/ui/src/machines/authenticator/__tests__/signIn.test.ts
```diff
@@ -0,0 +1,13 @@
...
+
+describe('createAuthenticatorMachine', () => {
+  it('should create AuthenticatorMachine and run actor when sign-in', () => {
+    const authenticatorMachine = createAuthenticatorMachine();
+    const expectedValue = { signIn: 'runActor' };
+    const actualState = authenticatorMachine.transition('signIn', {
+      type: 'SUBMIT',
+    });
+
+    expect(actualState.matches(expectedValue)).toBeTruthy();
```
calebpollman, line 11: Prefer checking for actual values over truthiness

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -549,6 +556,20 @@ export function signInActor({ services }: SignInMachineOptions) {
...
         },
+        /* START ISSUE 3767 - Temporary fix. Will change to handle 'SELECT_MFA_TYPE' */
+        isSelectMFATypeChallenge: (_, event): boolean => {
+          const isSelectMfaTypeChallenge = isExpectedChallengeName(
+            getChallengeName(event),
+            'SELECT_MFA_TYPE'
+          );
+          if (isSelectMfaTypeChallenge) {
+            console.error(
+              "ERROR: 'SELECT_MFA_TYPE' is not supported by Amplify UI yet. \n Please use only one MFA type or set up the MFA selector using this documen: https://docs.amplify.aws/lib/auth/mfa/q/platform/js/#advanced-use-cases/"
```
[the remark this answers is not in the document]
0618: In the error message, I'm trying to communicate
1) 'SELECT_MFA_TYPE' is not supported by Amplify UI, but users can implement it on their own using the API by following the link
2) 'SELECT_MFA_TYPE' will be supported by Amplify UI sometime in the future because we've got the feature request issue

Any thought?

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -104,6 +104,13 @@ export function signInActor({ services }: SignInMachineOptions) {
                     actions: ['setUser', 'setChallengeName'],
                     target: '#signInActor.setupTOTP',
                   },
+                  /* START ISSUE 3767 - Temporary fix. Will change to handle 'SELECT_MFA_TYPE' */
+                  {
+                    cond: 'isSelectMFATypeChallenge',
+                    actions: 'setRemoteError',
```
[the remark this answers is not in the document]
0618: Yes, shows the red warning as in the screencast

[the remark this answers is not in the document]
calebpollman: Is that what we should be doing? The end user has no control over this situation

packages/ui/src/machines/authenticator/__tests__/signIn.test.ts
0618: I know! I wish Xstate could do that. Or maybe I missed something, but that's in Xstate's [guide](https://xstate.js.org/docs/guides/testing.html#mocking-effects)

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -549,6 +556,20 @@ export function signInActor({ services }: SignInMachineOptions) {
         shouldSetupTOTP: (_, event): boolean => {
           return isExpectedChallengeName(getChallengeName(event), 'MFA_SETUP');
         },
+        /* START ISSUE 3767 - Temporary fix. Will change to handle 'SELECT_MFA_TYPE' */
+        isSelectMFATypeChallenge: (_, event): boolean => {
+          const isSelectMfaTypeChallenge = isExpectedChallengeName(
```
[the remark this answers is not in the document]
0618: I prefer to change it back to `shouldSelectMFAType` to make it consistent with `shouldSetupTOTP` etc. . Since @wlee221 commented on this earlier, what's your thought?

packages/ui/src/machines/authenticator/actors/signIn.ts
0618: We don't know until this line

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -104,6 +104,13 @@ export function signInActor({ services }: SignInMachineOptions) {
                     actions: ['setUser', 'setChallengeName'],
                     target: '#signInActor.setupTOTP',
                   },
+                  /* START ISSUE 3767 - Temporary fix. Will change to handle 'SELECT_MFA_TYPE' */
```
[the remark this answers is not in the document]
0618: I see your point that we can add less comment to make it look cleaner, but wrapping it up would make it easier to be removed by script (though we don't have the script yet), and easier and mistake-proofing for the engineer to identify them in the future

packages/ui/src/machines/authenticator/actors/signIn.ts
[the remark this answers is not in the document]
wlee221: I don't think we should use the word `should`, since that says Authenticator "should" do something, which we do not. I'm okay with giving slight modifications to the second variable name to have them different, or come up with some better name that starts with `should` (e.g. `shouldWarnSelectMFAChallenge`, but shorter)...

Happy to sync offline too.

packages/ui/src/machines/authenticator/__tests__/signIn.test.ts
0618: Removed the test 0d9d07f19f1ddd26804daa5e6865a4845fb58bd3

packages/ui/src/machines/authenticator/actions.ts
```diff
@@ -97,7 +97,10 @@ export const setFieldErrors = assign({
 
 export const setRemoteError = assign({
   remoteError: (_, event: AuthEvent) => {
-    if (event.data.name === 'NoUserPoolError') {
+    if (
+      event.data.name === 'NoUserPoolError' /* START ISSUE 3767 */ ||
+      event.data.challengeName === 'SELECT_MFA_TYPE' /* END ISSUE 3767 */
```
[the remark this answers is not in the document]
0618: Done 0d9d07f19f1ddd26804daa5e6865a4845fb58bd3

packages/ui/src/machines/authenticator/actors/signIn.ts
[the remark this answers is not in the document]
0618: Done 0d9d07f19f1ddd26804daa5e6865a4845fb58bd3

packages/ui/src/machines/authenticator/actors/signIn.ts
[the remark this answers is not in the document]
0618: Done 0d9d07f19f1ddd26804daa5e6865a4845fb58bd3

packages/ui/src/machines/authenticator/__tests__/signIn.test.ts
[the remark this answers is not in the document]
0618: Removed 0d9d07f19f1ddd26804daa5e6865a4845fb58bd3

ioanabrooks (approved):

reesscot:
Let's chat on the approach as a team before we merge this.

0618:
Close the PR and make a UI docs change instead because we'd like to support SELECT_MFA_TYPE instead of adding this temporary error message.

Issue #3683, opened 2023-04-10
StorageManager: Add x-amz-meta header


brian-visikon:
### On which framework/platform would you like to see this feature implemented?

React

### Which UI component is this feature-request for?

Storage (File Uploader)

### Please describe your feature-request in detail.

I would like to add metadata to the files I upload using the [StorageManager](https://ui.docs.amplify.aws/react/connected-components/storage/storagemanager) component, but it seems I cannot add custom metadata to the upload.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata

### Please describe a solution you'd like.

Here is example of I would like to be able do:

```
<StorageManager
          processFile={({ file, key }) => {
            return { 
               file,  key, headers: [{ 'x-amz-meta-id': uuidv4()}] 
            };
            onUploadSuccess={({ key, headers }) => {
               setDone(true);
               console.log('Uploaded', key, headers);
            }}
          }}
        />
```

I obviously could add the ID to the filename in the `processFile()` callback, but this is not elegant and it quickly becomes problematic. Especially once I start passing more than just an ID. There could be contextual info also, such as a reference to the enabled feature flags.

### We love contributors! Is this something you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

reesscot:
@brian-visikon You are correct that currently it is not possible to pass metadata, but I can see how this would be useful and it is supported by the underlying Amplify JS `Storage.put`. Are there any other options you would find useful from the "other options" section on this page? https://docs.amplify.aws/lib/storage/upload/q/platform/js/#encrypted-uploads

brian-visikon:
Adding `x-amz-meta`  headers is the only thing I am missing.

dbanksdesign:
@brian-visikon we merged a change that would allow this. It is currently published in the `@next` tag and will be fully released tomorrow. Here is an example:

https://github.com/aws-amplify/amplify-ui/blob/main/examples/next/pages/ui/components/storage/storage-manager/metadata/index.page.tsx

dbanksdesign:
This was released in v[1.2.0](https://www.npmjs.com/package/@aws-amplify/ui-react-storage/v/1.2.0)!

PR #3788, opened 2023-04-27
feat(storage-manager): add metadata

dbanksdesign:
#### Description of changes

Adding `metadata` prop to StorageManager and to the `processFile` function so users can add metadata per-session and per-file. 

#### Issue #, if available

#3683

Added an example

#### Checklist

- [x] PR description included

### DIFF
--- a/docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
+++ b/docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
@@ -64,7 +64,7 @@ export const STORAGE_MANAGER = [
     name: `processFile?`,
     description:
       'Called immediately before uploading a file to allow you to edit the key or the file itself.',
-    type: `(props: {key: string, file: Blob, metadata?: Record<string, string>}) => {key: string, file: Blob, metadata?: Record<string, string>};`,
+    type: `(props: {key: string, file: Blob}) => {key: string, file: Blob} & Record<string, string>;`,
   },
   {
     name: `path?`,
@@ -80,11 +80,6 @@ export const STORAGE_MANAGER = [
     name: `displayText?`,
     description: 'Text to override in the component.',
     type: 'Partial<StorageManagerDisplayText>',
-  },
-  {
-    name: `metadata?`,
-    description: '',
-    type: `Record<string, string>`,
   },
   {
     name: `components?.Container?`,
--- a/docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
+++ b/docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
@@ -64,7 +64,7 @@ export const STORAGE_MANAGER = [
     name: `processFile?`,
     description:
       'Called immediately before uploading a file to allow you to edit the key or the file itself.',
-    type: `(file: {key: string, file: Blob}) => {key: string, file: Blob};`,
+    type: `(props: {key: string, file: Blob, metadata?: Record<string, string>}) => {key: string, file: Blob, metadata?: Record<string, string>};`,
   },
   {
     name: `path?`,
@@ -80,6 +80,11 @@ export const STORAGE_MANAGER = [
     name: `displayText?`,
     description: 'Text to override in the component.',
     type: 'Partial<StorageManagerDisplayText>',
+  },
+  {
+    name: `metadata?`,
+    description: '',
+    type: `Record<string, string>`,
   },
   {
     name: `components?.Container?`,
--- a/docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
+++ b/docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
@@ -64,7 +64,7 @@ export const STORAGE_MANAGER = [
     name: `processFile?`,
     description:
       'Called immediately before uploading a file to allow you to edit the key or the file itself.',
-    type: `(props: {key: string, file: Blob}) => {key: string, file: Blob} & Record<string, string>;`,
+    type: `(params: {key: string, file: Blob}) => {key: string, file: Blob} & Record<string, string>;`,
   },
   {
     name: `path?`,
--- a/examples/next/pages/ui/components/storage/storage-manager/metadata/index.page.tsx
+++ b/examples/next/pages/ui/components/storage/storage-manager/metadata/index.page.tsx
@@ -8,16 +8,11 @@ import '@aws-amplify/ui-react/styles.css';
 import awsExports from './aws-exports';
 Amplify.configure(awsExports);
 
-const processFile: StorageManagerProps['processFile'] = ({
-  file,
-  key,
-  metadata,
-}) => {
+const processFile: StorageManagerProps['processFile'] = ({ file, key }) => {
   return {
     file,
     key,
     metadata: {
-      ...metadata,
       id: key,
     },
   };
@@ -30,9 +25,6 @@ export function StorageManagerExample() {
       accessLevel="private"
       maxFileCount={3}
       showThumbnails={true}
-      metadata={{
-        test: 'bar',
-      }}
       processFile={processFile}
     />
   );
--- a/examples/next/pages/ui/components/storage/storage-manager/metadata/index.page.tsx
+++ b/examples/next/pages/ui/components/storage/storage-manager/metadata/index.page.tsx
@@ -1,0 +1,40 @@
+import { Amplify } from 'aws-amplify';
+import { withAuthenticator } from '@aws-amplify/ui-react';
+import {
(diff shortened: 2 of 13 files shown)


docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
```diff
@@ -81,6 +81,11 @@ export const STORAGE_MANAGER = [
     description: 'Text to override in the component.',
     type: 'Partial<StorageManagerDisplayText>',
   },
+  {
+    name: `metadata?`,
+    description: '',
```
Comment (reesscot) [inline docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts:86]: Add a description?

examples/next/pages/ui/components/storage/storage-manager/metadata/index.page.tsx
```diff
@@ -0,0 +1,40 @@
...
+      accessLevel="private"
+      maxFileCount={3}
+      showThumbnails={true}
+      metadata={{
+        test: 'bar',
+      }}
+      processFile={processFile}
+    />
+  );
+}
```
calebpollman, line 39: Based on this example, it's unclear to me why we need to add the `metadata` prop, since a user could just do this?
```tsx
const metadata = { test: 'bar' }

const processFile: StorageManagerProps['processFile'] = ({
  file,
  key,
}) => {
  return {
    file,
    key,
    metadata: {
      ...metadata,
      id: key,
    },
  };
};

export function StorageManagerExample() {
  return (
    <StorageManager
      accessLevel="private"
      maxFileCount={3}
      showThumbnails={true}
      processFile={processFile}
    />
  );
}
```

packages/react-storage/src/components/StorageManager/types.ts
```diff
@@ -32,6 +32,12 @@ export type StorageFiles = StorageFile[];
 
 export type DefaultFile = Pick<StorageFile, 'key'>;
 
+type ProcessFileProps = Required<Pick<StorageFile, 'file' | 'key'>> & {
```
calebpollman, line 35: Prefer `Props` for component props, `Params` for function params

packages/react-storage/src/components/StorageManager/utils/uploadFile.ts
```diff
@@ -23,6 +24,7 @@ export function uploadFile({
   completeCallback,
   isResumable = false,
   provider,
+  metadata,
```
calebpollman, line 27: If we do keep `metadata` we don't need to destructure it since the it will be part of the `rest` props

examples/next/pages/ui/components/storage/storage-manager/metadata/index.page.tsx
dbanksdesign: Yea this example is kinda bad. I was adding `metadata` on the StorageManager so users could add non-file-specific metadata without having to use `processFile`. If you are already using `processFile` there would be no need to use both at the same time.

calebpollman: IMO we should really just offer one prop since having two ways to do this can cause confusion and potentially lead the user to use both erroneously. 

Since we already support `processFile` what do you think of only updating the `ProcessFile` interface to return `metadata`?

packages/ui/src/theme/tokens/components/fileUploader.ts
```diff
@@ -12,7 +12,8 @@ type BaseDropZoneTokens<OutputType> = DesignTokenProperties<
 
 export interface FileUploaderTokens<OutputType extends OutputVariantKey> {
   dropzone?: DesignTokenProperties<
-    'gap' | 'paddingBlock' | 'paddingInline' | 'textAlign'
+    'gap' | 'paddingBlock' | 'paddingInline' | 'textAlign',
+    OutputType
```
dbanksdesign, line 16: Note: this is a bug because with the 2nd arg all these properties become required

packages/react-storage/src/components/StorageManager/types.ts
```diff
@@ -32,6 +32,12 @@ export type StorageFiles = StorageFile[];
 
 export type DefaultFile = Pick<StorageFile, 'key'>;
 
+type ProcessFileParams = Required<Pick<StorageFile, 'file' | 'key'>>;
+
+type ProcessFile = (
+  props: ProcessFileParams
```
calebpollman, line 38:
```suggestion
  params: ProcessFileParams
```

docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts
```diff
@@ -64,7 +64,7 @@ export const STORAGE_MANAGER = [
     name: `processFile?`,
     description:
       'Called immediately before uploading a file to allow you to edit the key or the file itself.',
-    type: `(file: {key: string, file: Blob}) => {key: string, file: Blob};`,
+    type: `(props: {key: string, file: Blob}) => {key: string, file: Blob} & Record<string, string>;`,
```
Comment (calebpollman) [inline docs/src/pages/[platform]/connected-components/storage/storagemanager/props.ts:67]: type: `(params: {key: string, file: Blob}) => {key: string, file: Blob} & Record<string, string>;`,

packages/react-storage/src/components/StorageManager/utils/uploadFile.ts
calebpollman:
```suggestion
```

packages/react-storage/src/components/StorageManager/utils/uploadFile.ts
```diff
@@ -42,6 +44,7 @@ export function uploadFile({
        * https://github.com/aws-amplify/amplify-js/blob/main/packages/storage/src/types/AWSS3Provider.ts#L59
        */
       provider: provider as `AWSS3`,
+      metadata,
```
calebpollman, line 47:
```suggestion
```

packages/react-storage/src/components/StorageManager/utils/uploadFile.ts
```diff
@@ -51,6 +54,7 @@ export function uploadFile({
       progressCallback,
       contentType,
       provider: provider as `AWSS3`,
+      metadata,
```
calebpollman, line 57:
```suggestion
```

calebpollman (commented):
LGTM! Left some suggestions, lmk what you think 😅

packages/react-storage/src/components/StorageManager/types.ts
dbanksdesign: Man I am sloppy

calebpollman (approved):
🚢

packages/ui/src/theme/tokens/components/fileUploader.ts
reesscot: Did you already fix this in StorageManager, or did you mean to fix this in FileUploader?

packages/react-storage/src/components/StorageManager/types.ts
```diff
@@ -32,6 +32,12 @@ export type StorageFiles = StorageFile[];
 
 export type DefaultFile = Pick<StorageFile, 'key'>;
 
+type ProcessFileParams = Required<Pick<StorageFile, 'file' | 'key'>>;
+
+type ProcessFile = (
```
reesscot, line 37: I don't know the answer, but will this be exported since the interface below is exported?

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
```diff
@@ -74,17 +74,21 @@ export function useUploadFiles({
...
       if (file) {
-        const processedFile =
-          typeof processFile === 'function'
-            ? processFile({ file, key })
-            : { file, key };
+        const {
+          file: processedFile,
+          key: processedKey,
+          ...processedRest
+        } = typeof processFile === 'function'
```
reesscot, line 81: I think we actually have a isFunction util type, don't we?

packages/ui/src/theme/tokens/components/fileUploader.ts
dbanksdesign: StorageManger basically just extends this type so it is fixed in both :)

reesscot: So do we need to rename this type? When we delete the FileUploader tokens in next release, won't StorageManager break?

reesscot (approved):

packages/ui/src/theme/tokens/components/fileUploader.ts
dbanksdesign: When we do that breaking change we will inline that type into the storagemanager type

PR #3798, opened 2023-05-01
feat(storage-manager): let processFile be async or sync

dbanksdesign:
#### Description of changes

Allowing the `processFile` prop to return a Promise or immediately returning. This allows for doing things like reading the file before upload. 

#### Checklist

- [x] PR description included

### DIFF
--- a/.changeset/hungry-owls-admire.md
+++ b/.changeset/hungry-owls-admire.md
@@ -1,6 +1,5 @@
 ---
 "@aws-amplify/ui-react-storage": minor
-"@aws-amplify/ui-react": minor
 ---
 
 feat(storage-manager): make processFile async. This allows for reading the file contents and performing async validations or mutations like creating a hash of the file contents. 
--- a/.changeset/hungry-owls-admire.md
+++ b/.changeset/hungry-owls-admire.md
@@ -1,0 +1,34 @@
+---
+"@aws-amplify/ui-react-storage": minor
+"@aws-amplify/ui-react": minor
+---
+
+feat(storage-manager): make processFile async. This allows for reading the file contents and performing async validations or mutations like creating a hash of the file contents. 
+
+```jsx
+const processFile = async ({ file }) => {
+  const fileExtension = file.name.split('.').pop();
+
+  return file
+    .arrayBuffer()
+    .then((filebuffer) => window.crypto.subtle.digest('SHA-1', filebuffer))
+    .then((hashBuffer) => {
+      const hashArray = Array.from(new Uint8Array(hashBuffer));
+      const hashHex = hashArray
+        .map((a) => a.toString(16).padStart(2, '0'))
+        .join('');
+      return { file, key: `${hashHex}.${fileExtension}` };
+    });
+};
+
+export const StorageManagerHashExample = () => {
+  return (
+    <StorageManager
+      acceptedFileTypes={['image/*']}
+      accessLevel="public"
+      maxFileCount={1}
+      processFile={processFile}
+    />
+  );
+};
+```
--- a/packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/resolveFile.ts
+++ b/packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/resolveFile.ts
@@ -1,10 +1,9 @@
 import { isFunction } from '@aws-amplify/ui';
-import { StorageManagerProps } from '../../types';
+import { ProcessFileParams, StorageManagerProps } from '../../types';
 
-interface ProcessFileParams extends Pick<StorageManagerProps, 'processFile'> {
-  file: File;
-  key: string;
-}
+interface ResolveFileParams
+  extends Pick<StorageManagerProps, 'processFile'>,
+    ProcessFileParams {}
 
 /**
  * Utility function that takes the processFile prop, along with a file a key
@@ -15,7 +14,7 @@ export const resolveFile = ({
   processFile,
   file,
   key,
-}: ProcessFileParams): Promise<Pick<ProcessFileParams, 'file' | 'key'>> => {
+}: ResolveFileParams): Promise<ProcessFileParams> => {
   return new Promise((resolve, reject) => {
     const result = isFunction(processFile)
       ? processFile({ file, key })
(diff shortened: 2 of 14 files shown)


packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
```diff
@@ -75,42 +75,75 @@ export function useUploadFiles({
       };
 
       if (file) {
-        const {
-          file: processedFile,
-          key: processedKey,
-          ...processedRest
-        } = isFunction(processFile)
+        const processedFile = isFunction(processFile)
```
bombguy, line 78: If the `processFile` is an async function. Will this wait for the resolution?

bombguy: Ah I found it below in the else statement. All Good

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/resolveFile.ts
```diff
@@ -0,0 +1,31 @@
...
+  return new Promise((resolve, reject) => {
+    const result = isFunction(processFile)
+      ? processFile({ file, key })
+      : { file, key };
+    if (result instanceof Promise) {
+      result.then(resolve).catch(reject);
+    } else {
+      resolve({
+        ...result,
+      });
```
calebpollman, line 28: Unless I'm missing something, don't think we need to spread `result` here

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/resolveFile.ts
```diff
@@ -0,0 +1,31 @@
+import { isFunction } from '@aws-amplify/ui';
+import { StorageManagerProps } from '../../types';
+
+interface ProcessFileParams extends Pick<StorageManagerProps, 'processFile'> {
```
calebpollman, line 4: Can we use the `ProcessFileParams` defined [here](https://github.com/aws-amplify/amplify-ui/pull/3798/files#diff-41fc325b42b2c3ea38b8bb65e8f03aba29456c1a904f8a8e7a87b9329730047aR35)?

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
```diff
@@ -75,43 +75,38 @@ export function useUploadFiles({
...
-          setUploadingFile({ id });
-        }
+        resolveFile({ processFile, file, key }).then(
+          ({ file, key, ...rest }) => {
+            onUploadStart?.({ key });
+            if (isResumable) {
+              const uploadTask = uploadFile({
+                ...rest,
+                file,
+                key: path + key,
```
calebpollman, line 85: Is there any concern with the value of `key` being passed here since the user may have defined a specific value for it in `processFile`?

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
```diff
@@ -75,43 +75,38 @@ export function useUploadFiles({
...
-            level: accessLevel,
-            completeCallback: onComplete,
-            progressCallback: onProgress,
-            errorCallback: onError,
-            provider,
-          });
-          setUploadingFile({ id });
-        }
+        resolveFile({ processFile, file, key }).then(
+          ({ file, key, ...rest }) => {
```
calebpollman, line 79: Don't think we need to destructure `file` any more 😅

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
```diff
@@ -75,43 +75,38 @@ export function useUploadFiles({
...
+                key: path + key,
+                isResumable: false,
+                level: accessLevel,
+                completeCallback: onComplete,
+                progressCallback: onProgress,
+                errorCallback: onError,
+                provider,
+              });
+              setUploadingFile({ id });
+            }
```
calebpollman, line 107: Think this could be simplified to:
```suggestion
            const uploadTask = uploadFile({
              ...rest,
              file,
              key: path + key,
              isResumable,
              level: accessLevel,
              completeCallback: onComplete,
              progressCallback: onProgress,
              errorCallback: onError,
              provider,
            }) as unknown as UploadTask;
            setUploadingFile({ id, uploadTask: isResumable ? uploadTask : undefined });
```

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/resolveFile.ts
dbanksdesign: No you are right

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
dbanksdesign: As in, concern with prepending the key with `path` ? That is what we had before

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
calebpollman: But since we are just directly passing the value of it below it wouldn't it just be included with `rest`?

packages/react-storage/src/components/StorageManager/hooks/useUploadFiles/useUploadFiles.ts
calebpollman: Yeah, maybe out of scope of this PR but just curious

.changeset/hungry-owls-admire.md
```diff
@@ -0,0 +1,34 @@
+---
+"@aws-amplify/ui-react-storage": minor
+"@aws-amplify/ui-react": minor
```
calebpollman, line 3: Sorry to catch this late, do we need to bump `ui-react`?

calebpollman (dismissed):

.changeset/hungry-owls-admire.md
dbanksdesign: Nope

reesscot (approved):

PR #3811, opened 2023-05-03
chore: update private package names

wlee221:
#### Note

This PR is expected to fail checks, because it changes workflows. Please see POC badge below for change verification.

#### Description of changes

Updates `private` packages in our repo to use scoped namespace.

I've changed non workspace packages too in this PR. They don't have to be technically changed because they don't get installed, but I thought it'd be better to be consistent and have a naming convention set up for current and future packages.

CI, and ran monorepo workplace commands.

https://github.com/aws-amplify/amplify-ui/actions/runs/4888370381

#### Checklist

- [x] PR description included


package.json
```diff
@@ -39,7 +39,6 @@
   "workspaces": {
     "packages": [
       "docs",
-      "angular-docs",
```
wlee221, line 42: This doesn't exist anymore.

reesscot (dismissed):

docs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "name": "docs",
+  "name": "@aws-amplify/ui-docs",
```
0618, line 3: Do we need to change all the packages' names? For example `docs` is not consumed anywhere.

I thought only changing those under `packages/` is sufficient, isn't it?

wlee221: Fair point. It's not consumed anywhere, but it's also part of `yarn workspaces` and gets installed as part of workspaces.

wlee221: In general though, I think it'd be better to make this change for all packages (see PR description), so that we have a consistent naming standards throughout the monorepo. For example, if someone is creating a new private package, they wouldn't need to worry about how this package is consumed for possible security implications.

reesscot: I think it's preferable to scope them all to be completely safe

0618 (dismissed):

thaddmt (dismissed):

wlee221:
Found that this requires more workflow changes, that might require a force merge. Will reopen it when ready.

examples/angular/package.json
```diff
@@ -11,7 +12,6 @@
     "lint": "ng lint",
     "e2e": "ng e2e"
   },
-  "private": true,
```
0618, line 14: no longer private?

calebpollman (dismissed):

.changeset/config.json
```diff
@@ -10,13 +10,13 @@
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
   "ignore": [
-    "amplify-ui-angular-mono",
```
0618, line 13: This mind also need to be changed? [└── package.json # name: 'amplify-ui-angular-mono`, private: true](https://github.com/aws-amplify/amplify-ui/blob/b9d19c49621e5cfd0da06d5cff5ea42c092abed0/packages/angular/CONTRIBUTING.md#L23-L24)

examples/angular/package.json
wlee221: Yep, I moved it up because this `private` was too hard to find 😅

zchenwei (approved):
LGTM

examples/angular/package.json
0618: 🤦‍♀️  sorry, not sure why I missed the green one

0618 (approved):

sreeramsama (approved):

.changeset/config.json
wlee221: Thanks! I removed these references in https://github.com/aws-amplify/amplify-ui/pull/3811/commits/a18b61963d4a1c3565cbd6af8699add66db0bb53.

0618 (approved):
Thanks! LGTM!

zchenwei (approved):
🚢

reesscot (approved):

sreeramsama (approved):

PR #3812, opened 2023-05-03
chore(e2e): add injected video feed for liveness tests

thaddmt:
#### Description of changes
- Added an injected video through chrome settings to replace the webcam
- the faceRecording.y4m file is ~30 MB, this was the smallest I could make the file, originally a 10 second video was 100 MB
- All tests runs will now successfully reach the face check and will reach timeout because the face recording does not attempt to match the oval, this does successfully test our face recognition features through blazeface

foo

#### Checklist

- [ ] PR description included


0618 (approved):
NICE!

thaddmt:
It doesn't work on electron :(

reesscot (approved):

packages/e2e/package.json
```diff
@@ -6,14 +6,14 @@
...
-    "test:geo": "cypress run --spec 'features/ui/components/geo/*.feature'",
-    "test:liveness": "cypress run --spec 'features/ui/components/liveness/*.feature'",
-    "test:storage": "cypress run --spec 'features/ui/components/storage/**/*.feature'",
-    "test:theme": "cypress run --spec 'features/ui/theme/*.feature'",
-    "test:guides": "cypress run --spec 'features/ui/guides/*.feature'",
+    "test:examples": "cypress run --spec 'features/ui/hooks/*.feature','features/ui/components/**/*.feature' --browser chrome",
+    "test:authenticator": "cypress run --spec 'features/ui/components/authenticator/*.feature' --browser chrome",
+    "test:hooks": "cypress run --spec 'features/ui/hooks/*.feature' --browser chrome",
+    "test:geo": "cypress run --spec 'features/ui/components/geo/*.feature' --browser chrome",
+    "test:liveness": "cypress run --spec 'features/ui/components/liveness/*.feature' --browser chrome",
```
calebpollman, line 13: Definitely not a blocker for this PR but wondering if we should scope the `--browser` flag to just Liveness. Think ideally we should be able to test other browsers for the rest of the connected components

thaddmt: Spoke with @reesscot @ioanabrooks and @0618 on this and we opted to switch to chrome for now since the default we use all around is  electron which is not indicative of our customer experience. Agree though we should probably be testing on more browswers in the future.

Also in this case during the e2e test run we only run `test:examples` command, these other ones are more for convenience but changed them to just match

calebpollman (approved):

packages/e2e/cypress.config.ts
```diff
@@ -21,6 +21,15 @@ export default defineConfig({
 
       Object.assign(config.env, process.env);
 
+      // This is a chrome launch option which enables fake videos
+      on('before:browser:launch', (_browser, launchOptions) => {
```
wlee221, line 25: This is cool 😮

wlee221 (approved):

PR #3816, opened 2023-05-04
fix(accessibility): add role=status to Button loading text span

andrewwaltos:
#### Description of changes

Loading text is being announced by screen reader when using Mac VoiceOver, but not when using JAWS or NVDA. I moved the span wrapping the spinner and loading text to be rendered regardless of the button's loading state. The spinner and the loading text still only render based on isLoading. Also added role="status" to that span to trigger announcing the text change via screen readers.

#### Checklist

- [x] PR description included

### DIFF
--- a/.changeset/wild-singers-beg.md
+++ b/.changeset/wild-singers-beg.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ui-react': patch
+---
+
+fix(accessibility): add role=status to Button loading text span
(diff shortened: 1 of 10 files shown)


calebpollman:
👋 @andrewwaltos! Thanks for opening this, we will consider the proposed change and update here

hbuchel:
Hi @andrewwaltos, thanks for opening this PR which addresses this gap. We've had some discussion around refactoring out this loading functionality from our Button primitive so that it's more readily customizable by users. Ideally, we'd like users to be able to manage their own roles and ARIA attributes, especially for things that should be announced.

For now, we're going to close this PR but I have opened an issue to track (https://github.com/aws-amplify/amplify-ui/issues/4343) and added it to our roadmap. Thank you, again, for this contribution!

PR #3830, opened 2023-05-05
chore(liveness): add ts-strict to xstate machine

thaddmt:
#### Description of changes
- Attempted to add ts-strict to the liveness xstate machine by adding non-null assertions to places that use the xstate context
- Issue: the initial xstate context we set has many of the context fields as undefined and those fields are hydrated throughout the life cycle of xstate. This requires us to set the type of the context to allow undefined values but later when that context is used in xstate actions it expects those values could be undefined thus all the ts errors.
- Attempted solutions
  - I tried using [type states](https://xstate.js.org/docs/guides/typescript.html#typestates) but this seems to work better when the actions are not re-used and are defined in line with each state object ( we currently reference each action by their string name in the state in a separate actionsobject). We could move all the actions to in-line with the state object but sorta seems messy to do this way, though maybe it's the safest TS wise.
  - Tried using their experimental feature [typegen](https://xstate.js.org/docs/guides/typescript.html#typegen) but also did not have much luck here either as the generated types did not seem to apply to the actions object as well
  - The only other option is to do a non-null assertion or type guard in each action for TS to know that the various context values are defined

#### Checklist

- [ ] PR description included


packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
```diff
@@ -432,7 +412,7 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
       }),
       updateFailedAttempts: assign({
         failedAttempts: (context) => {
-          return context.failedAttempts + 1;
+          return context.failedAttempts! + 1;
```
thaddmt, line 415: Instead of using the non-null assertion operator here we could instead typecast `context` like so. What does everyone think?

```
context as any as HydratedLivenessContext
```

wlee221 (approved):
LGTM, I'm aligned with you that this is as close as we can get without needing to massively rethink how we use xstate.

PR #3862, opened 2023-05-12
feat(liveness): remove 3 second countdown from liveness check

thaddmt:
#### Description of changes
- Per UX request, removing the countdown from liveness check
- Aligned with PM and UX on the new experience
  - User will go straight from `move back` to `connecting` to matching the oval
  - For users with longer `connecting` steps we will show an error if they move their face closer during the countdown
 
foobar
timeout

#### Checklist

- [ ] PR description included


packages/react-liveness/package.json
```diff
@@ -60,7 +60,6 @@
     "@xstate/react": "3.0.0",
     "classnames": "2.3.1",
     "nanoid": "3.1.31",
-    "react-countdown-circle-timer": "^2.5.4",
```
reesscot, line 63: nice!

reesscot (approved):

reesscot:
Should the error message still refer to the countdown if it's missing? 
image

thaddmt:
> Should the error message still refer to the countdown if it's missing? image

yep! just waiting on new text from the ux

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/__tests__/index.test.ts
```diff
@@ -163,7 +163,7 @@ describe('Liveness Machine', () => {
     jest.advanceTimersToNextTimer(); // waitForSessionInformation
     await flushPromises(); // detectFaceDistanceBeforeRecording
     jest.advanceTimersToNextTimer(); // checkFaceDistanceBeforeRecording
-    service.send({ type: 'START_RECORDING' });
+    // await flushPromises();
```
calebpollman, line 166: Assuming this is unintentional? 😅

thaddmt: whoops haha, this is when i was testing stuff :P will fix

calebpollman (dismissed):

reesscot (approved):

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
```diff
@@ -739,6 +752,7 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
         context.componentProps!.onError?.(error);
       },
       callErrorCallback: async (context, event) => {
+        console.log({ event });
```
wlee221, line 755:
```suggestion
```

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
```diff
@@ -198,46 +202,46 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
...
+        initial: 'waitForSessionInfo',
         states: {
-          detectFaceDistanceDuringCountdown: {
+          waitForSessionInfo: {
+            after: {
+              0: {
+                target: '#livenessMachine.recording',
+                cond: 'hasServerSessionInfo',
+              },
+              10: { target: 'detectFaceDistanceDuringLoading' },
```
wlee221, line 225: Just for my curiosity, why do we have to do these two tasks async? (0 vs 10). Is it possible to combine them?

thaddmt: I guess this is the pattern I originally saw done in this machine for essentially doing conditionals. If the first condition does not pass then we do the second one 10 ms later. If there is a better way of doing it happy to change it :)

packages/react-liveness/package.json
zchenwei, line 63: I do not see a `yarn.lock` update?

thaddmt: oh let me regenerate it! good catch

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
wlee221: If you're only interested in doing one or the other, you can do this:

```ts
waitForSessionInfo: {
  always: [
    {
      target: '#livenessMachine.recording',
      cond: 'hasServerSessionInfo',
    },
    { target: 'detectFaceDistanceDuringLoading' },
  ],
},
```
and essentially work as `if` `else` statement. This is more [idiomatic](https://xstate.js.org/docs/guides/guards.html#multiple-guards) to handle conditional transitions 🙏

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
```diff
@@ -198,46 +201,46 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
...
+          checkFaceDistanceDuringLoading: {
             after: {
               0: {
                 target: 'failure',
                 cond: 'hasNotEnoughFaceDistanceBeforeRecording',
               },
-              200: {
-                target: 'detectFaceDistanceDuringCountdown',
+              100: {
+                target: 'waitForSessionInfo',
```
wlee221, line 243: Can we replace this with the above recommendation too?

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
thaddmt: ah thanks will change!

zchenwei (approved):
🚢

wlee221 (approved):

gustavovalverde:
Just an FYI as we have users using this functionality, and it's very confusing having the `Avoid moving closer during countdown and ensure only one face is in front of camera.` as there is no countdown, and there are no more faces in front of the camera.

reesscot:
@gustavovalverde that's a good callout, can you please open a github issue for that?

thaddmt:
@gustavovalverde that string should be updated in the 2.x version to be `Avoid moving closer when connecting.` and no longer has references to countdown. On the 1.x version it should still be the same though

PR #3872, opened 2023-05-12
React/expose files ref storage manager

kwwendt:
#### Description of changes

- Storage Manager component now has a `filesRef` prop that can be used to clear files currently managed in the components state.
- Following the existing code path for updates to the files state, I added a `clearFilesAction`, a case in the reducer to handle the action, and associated interface changes.
- In the `StorageManager` component, I expose the `clearFiles` function using the `useImperitiveHandle` React hook which will allow components that have passed a ref to the `StorageManager` component to clear the files.

I created my own example using the existing framework in the repository. I added a storage component and a button to clear form values (including the Storage Manager files). I created a `files` ref using the `useRef` React component. I passed this to the `StorageManager` component and was able to call `files.current.clearValues()` when the reset button was clicked. This worked as expected.

#### Checklist

- [X] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageManager/StorageManager.tsx
+++ b/packages/react-storage/src/components/StorageManager/StorageManager.tsx
@@ -38,6 +38,7 @@ function StorageManager({
   components,
   provider,
   path,
+  filesRef,
 }: StorageManagerProps): JSX.Element {
   if (!acceptedFileTypes || !accessLevel || !maxFileCount) {
     logger.warn(
@@ -75,6 +76,7 @@ function StorageManager({
 
   const {
     addFiles,
+    clearFiles,
     files,
     removeUpload,
     setUploadingFile,
@@ -83,6 +85,14 @@ function StorageManager({
     setUploadSuccess,
     setUploadResumed,
   } = useStorageManager(defaultFiles);
+
+  React.useImperativeHandle(filesRef, () => {
+    return {
+      clearFiles: () => {
+        clearFiles();
+      },
+    };
+  });
 
   const dropZoneProps = useDropZone({
     onChange: (event: React.DragEvent<HTMLDivElement>) => {
--- a/.changeset/cyan-teachers-obey.md
+++ b/.changeset/cyan-teachers-obey.md
@@ -1,0 +1,18 @@
+---
+'@aws-amplify/ui-react-storage': minor
+---
+
+What changed:
+
+- Added exposed filesRef for the Storage Manager component.
+- This allows for clearing of the files list from a separate component where the Storage Manager component is being used.
+
+Why was the change made:
+
+- There was no easy way to clear the list of files without unmounting the component.
+- This capability is important when using Storage Manager inside a form. After submit, clear all entries including uploaded files.
+
+How should a customer update their code:
+
+- No changes are required by the customer since the added prop is optional in the Storage Manager component.
+- If customers want to take advantage of this change, they can create a files ref and include it in the Storage Manager component, then made calls to clearFiles from the ref.
--- a/packages/react-storage/src/components/StorageManager/hooks/useStorageManager/actions.ts
+++ b/packages/react-storage/src/components/StorageManager/hooks/useStorageManager/actions.ts
@@ -18,6 +18,12 @@ export const addFilesAction = ({
   };
 };
 
+export const clearFilesAction = (): Action => {
+  return {
+    type: StorageManagerActionTypes.CLEAR_FILES,
+  };
+};
+
 export const setUploadingFileAction = ({
   id,
   uploadTask,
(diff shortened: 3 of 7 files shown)


packages/react-storage/src/components/StorageManager/StorageManager.tsx
```diff
@@ -84,6 +86,14 @@ function StorageManager({
     setUploadResumed,
   } = useStorageManager(defaultFiles);
 
+  React.useImperativeHandle(filesRef, () => {
+    return {
+      clearFiles: () => {
+        clearFiles();
+      },
+    };
+  });
```
calebpollman, line 95: Nit: can be simplified to:
```
React.useImperativeHandle(ref, () => ({ clearFiles }));
```

calebpollman (commented):
Thanks for opening this @kwwendt! We are excited about the contribution and would like to work with you to get it merged and released. 

The approach makes sense but wanted to discuss whether the following update can be made:

Rather than adding a `React.Ref` prop, can we look in to wrapping `StorageManager` with `React.forwardRef`? Although It will be a breaking change, usage of `React.forwardRef` is idiomatic when creating ref APIs with `useImperativeHandle`, and allows future flexibility of the forwarded ref (for example, if we wanted to attach the forwarded ref to a DOM node).

This will require a bit more work on your end as the return type of `React.ForwardRef` does not allow for adding additional properties to its output, so Typescript will complain about [these lines](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react-storage/src/components/StorageManager/StorageManager.tsx#L255-L259). The type issue  can be mitigated by calling `Object.assign` with the `StorageManager` as the first parameter and an object of the subcomponents as the second parameter rather than the adding the properties individually. 

Below is some example code with comments that hopefully illustrates the request a bit further:
```
// currently `StorageManager`, `StorageManagerBase` is just an example name
function StorageManagerBase(
  {
    // ...destructured props here
  }: StorageManagerProps,
  //  prefer `StorageManagerHandle` as the interface name for the fowarded ref, replaces `FilesRefType` (still would just contain `clearFiles`)
  ref: React.ForwardedRef<StorageManagerHandle>
) {
  // ...do StorageManager stuff
};

const StorageManager = Object.assign(
  React.forwardRef<StorageManagerHandle, StorageManagerProps>(
    StorageManagerBase
  ),
  {
    Container,
    DropZone,
    FileList,
    FileListHeader,
    FilePicker,
  }
);
```

Please let us know if that makes sense and is agreeable on your end, if so we will work with you to get this added in to our next major release of _@aws-amplify/ui-react-storage_.  Appreciate the contribution! 🎉

packages/react-storage/src/components/StorageManager/StorageManager.tsx
kwwendt: Ack. Will fix in next commit with additional changes. Thanks!!

kwwendt:
Thanks @calebpollman !! I appreciate the review and the feedback. 

Those recommendations make sense and I can see how that would allow for more extensibility in the future.

I'll make those changes and corresponding doc changes and update the PRs.

Thanks again!

calebpollman (commented):
@kwwendt Implementation looks great! Couple things:
- can you add some unit tests using the new API
- can you update the branch this PR is pointing at from `main` to `next-release/main`

kwwendt:
Added test cases and updated branch to `next-release/main`.

Should I also update the docs PR to point to `next-release/main` or `docs-release`? Or leave it be?

Thanks again for the review and feedback :)

calebpollman:
> Added test cases and updated branch to `next-release/main`.

🎉

> Should I also update the docs PR to point to `next-release/main` or `docs-release`? Or leave it be?

Yes please, good catch 👀 

> Thanks again for the review and feedback :)

Of course, thank you for your contribution!

packages/react-storage/src/components/StorageManager/hooks/useStorageManager/__tests__/useStorageManager.test.ts
```diff
@@ -41,6 +41,23 @@ describe('useUploadFiles', () => {
...
+      result.current.addFiles({
+        files: [
+          new File(['test1'], 'test1.txt', { type: 'text/plain' }),
+          new File(['test2'], 'test2.txt', { type: 'text/plain' }),
+        ],
+        getFileErrorMessage: () => '',
+      })
+    );
+
+    expect(result.current.files.length).toBe(4);
```
calebpollman, line 56: Does the initial value of `result.current.files.length` start at `2`?

calebpollman (approved):
LGTM 🚢

packages/react-storage/src/components/StorageManager/hooks/useStorageManager/__tests__/useStorageManager.test.ts
kwwendt: The `defaultFiles` contains 2 files which I passed in to the `useStorageManager` hook, plus the addition of 2 more files.

reesscot (approved):

PR #3873, opened 2023-05-12
chore: import maplibre styles

sreeramsama:
#### Description of changes
Importing missing maplibre styles in `ui-react-geo` package.
We need the same ones in ui-react-geo as we have in ui-react: https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/styles.css

- Tested MapView and LocationSearch components in sample app with these imports.
#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-geo/src/styles.css
+++ b/packages/react-geo/src/styles.css
@@ -1,1 +1,4 @@
 @import '../../ui/dist/styles.css';
+@import '@maplibre/maplibre-gl-geocoder/dist/maplibre-gl-geocoder.css';
+@import 'maplibre-gl-js-amplify/dist/public/amplify-geocoder.css';
+@import 'maplibre-gl/dist/maplibre-gl.css';
--- a/packages/react-geo/src/styles.css
+++ b/packages/react-geo/src/styles.css
@@ -1,4 +1,4 @@
-@import '../../ui/dist/styles.css';
 @import '@maplibre/maplibre-gl-geocoder/dist/maplibre-gl-geocoder.css';
-@import 'maplibre-gl-js-amplify/dist/public/amplify-geocoder.css';
 @import 'maplibre-gl/dist/maplibre-gl.css';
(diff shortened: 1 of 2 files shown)


calebpollman (approved):

packages/react-geo/src/styles.css
```diff
@@ -1 +1,4 @@
 @import '../../ui/dist/styles.css';
+@import '@maplibre/maplibre-gl-geocoder/dist/maplibre-gl-geocoder.css';
+@import 'maplibre-gl-js-amplify/dist/public/amplify-geocoder.css';
+@import 'maplibre-gl/dist/maplibre-gl.css';
```
wlee221, line 4: Can we order the imports so that it's:

1. 3p deps (`maplibre-gl`, `@maplibre`)
2. Internal dependencies (`maplibre-gl-js-amplify`)
3. Local dependencies (`../../ui/dist/...`)

wlee221 (approved):

zchenwei (approved):

packages/react-geo/src/styles.css
```diff
@@ -1 +1,4 @@
+@import '@maplibre/maplibre-gl-geocoder/dist/maplibre-gl-geocoder.css';
+@import 'maplibre-gl/dist/maplibre-gl.css';
+@import 'maplibre-gl-js-amplify/dist/public/amplify-geocoder.css';
```
ioanabrooks, line 3: We should write some unit tests to prevent regressions and make sure these styles are applied as expected. Re-ordering is nice but now they're not in the order they were in the react pkg, have we thoroughly verified it works as before?

calebpollman: +1

sreeramsama: Yes. We have a meeting today to discuss how to proceed with this. Will update PR based on that.

sreeramsama: I have verified that the new order of css imports works the same.

PR #3874, opened 2023-05-13
chore: update e2e workflow to run react native tests on every commit to main

zchenwei:
#### Description of changes

It is reasonable to skip react native e2e tests for unrelated changes in a PR but we need to run the tests against every commit to main because the unrelated changes such as examples, environments, workspace commands, etc could be related somehow. Always running the tests on main can help us catch the issues early before the release.

I kicked off a PoC workflow where e2e tests for native platform were triggered for irrelevant changes.

PoC workflow - https://github.com/aws-amplify/amplify-ui/actions/runs/5005906216

#### Checklist

- [x] PR description included


.github/workflows/reusable-e2e.yml
```diff
@@ -340,7 +347,7 @@ jobs:
 
   e2e-ios:
     needs: check-changed-packages
-    if: ${{ needs.check-changed-packages.outputs.has-changed-packages == 'true' }}
+    if: ${{ always() && (inputs.skip-changed-packages-check || needs.check-changed-packages.outputs.has-changed-packages == 'true') }}
```
zchenwei, line 350: `always` make sure the job always execute even though its dependent jobs get skipped

ioanabrooks: This is confusing. But more importantly, it seems like `always` makes the job not cancellable, check this [thread](https://github.com/orgs/community/discussions/26303)

zchenwei: Another way is we skip `check-changed-packages` at step level without using `always()`.

zchenwei: In scenario where `check-changed-packages` was skipped, `e2e-ios` and `e2e-android` would be skipped alongside.

zchenwei: Yeah this is tricky. May need to use `!cancel() && !failure()` or Skip at step level can get rid of this spaghetti conditional

.github/workflows/reusable-e2e.yml
```diff
@@ -327,7 +334,7 @@ jobs:
 
   e2e-ios:
     needs: check-changed-packages
-    if: ${{ needs.check-changed-packages.outputs.has-changed-packages == 'true' }}
+    if: ${{ inputs.skip-changed-packages-check || needs.check-changed-packages.outputs.has-changed-packages == 'true' }}
```
0618, line 337: why don't we just remove the condition if we "always run e2e tests for native platform on main"?

0618: Oh, NVM, I forgot we use the the re-usable-e2e at other places

0618 (approved):

.github/workflows/reusable-e2e.yml
zchenwei: We still want to skip on PRs but leaving it running async on main so that we can catch ❌ in case some changes seems unrelated but in reality they are related

calebpollman, line 337: Not going to pretend i understand how the syntax works here but is it possible to align the syntax of the two conditions?

zchenwei: We cannot align the two. The latter - `needs.check-changed-packages.outputs.has-changed-packages`  is the result of the check that gets skipped.

wlee221: Think we can put 
```suggestion
    if: ${{ inputs.skip-changed-packages-check == 'true' || needs.check-changed-packages.outputs.has-changed-packages == 'true' }}
```

zchenwei: Is this valid? Thinking it make sense that output is a string. But let me kick off a run to test it out

wlee221: oh wait, yeah this might not work as expected because you have the input type as `input` `type` as a `boolean`.

We've used `string` because of this issue https://github.com/actions/runner/issues/1483 but looks like it's fixed since then.

wlee221: Up to thoughts, but what do you think about enforcing the input type to be a `string` for the sake of consistency? We can look at the new `boolean` inputs and change our workflows holistically.

zchenwei: That's a good history reference. No wonder our conditionals try to evaluate a string

zchenwei: I am good to make it consistent for now and revisit the workflows as a whole

zchenwei: Updated @wlee221 @calebpollman

wlee221 (approved):
LGTM, as long as POC passes

zchenwei (commented):
PoC run - https://github.com/aws-amplify/amplify-ui/actions/runs/5005906216

wlee221 (approved):

calebpollman (approved):

ioanabrooks (approved):

0618 (approved):

PR #3908, opened 2023-05-17
chore(vue): refactor vue composable

wlee221:
#### Description of changes

- move state machine initialization logic to the composable instead of doing it in the Authenticator.
- remove side effects from composable codes.
- fix race condition between `useAuth` and `useAuthenticator` in `authenticator.vue`.
- fix incorrect usage of `createSharedComposable` and use `createSharedComposable` at the entry point level so that its state can be shared.

#### Verification of changes

Existing CIs. Manually checked:
- rendering of our example app, and confirmed that subscription only gets called once
- refreshing with and without Authenticator, and 
- toggling Authenticator rendering on and off.

https://github.com/aws-amplify/amplify-ui/assets/43682783/31d6351f-3a8b-4407-91b0-acc349b22265

#### Checklist

- [x] PR description included


packages/vue/src/components/authenticator.vue
```diff
@@ -77,17 +64,13 @@ const emit = defineEmits([
   'verifyUserSubmit',
   'confirmVerifyUserSubmit',
 ]);
-const machine = createAuthenticatorMachine();
 
-const service = useInterpret(machine);
-let unsubscribeHub: () => void;
```
wlee221, line 83: Move machine initialization and hub listening logic to the composable.

packages/vue/src/components/authenticator.vue
```diff
@@ -77,17 +64,13 @@ const emit = defineEmits([
...
 let unsubscribeMachine: () => void;
 
-const { state, send } = useActor(service);
-useAuth(service);
-
```
wlee221, line 88: Previously, this file used `useActor`, `useAuth`, and `useAuthenticator` in one file. This actually had a race condition based on which one finishes updating the facade value.

For example, `state.matches('signUp')` might execute faster than `useAuthenticator` finishes computing the facade value. This could lead to UI rendering early before contexts are ready.

packages/vue/src/components/authenticator.vue
```diff
@@ -77,17 +64,13 @@ const emit = defineEmits([
...
 const hasInitialized = ref(false);
 
+const { service, send, state } = useAuth();
```
wlee221, line 72: I dumbed down the `useAuth`, and `state` and `send` are no longer used for rendering logic. We would ideally just remove `useAuth` in this file too, but we still rely it in some places below publicly, and I wanted to minimize big changes to this file.

packages/vue/src/components/authenticator.vue
```diff
@@ -109,21 +92,22 @@ unsubscribeMachine = service.subscribe((newState) => {
   }
 }).unsubscribe;
 
+const { route, signOut, toSignIn, toSignUp, user } = toRefs(
+  // `useAuthenticator` is casted for temporary type safety on this file.
+  useAuthenticator() as AuthenticatorServiceFacade
```
wlee221, line 97: I added temporary type safety to this file so that we can trust at least the type safety of the changes I make below.

packages/vue/src/components/authenticator.vue
```diff
@@ -224,41 +208,19 @@ const onConfirmVerifyUserSubmitI = (e: Event) => {
...
 };
 
-// watchers
-
-/**
- * Update service facade when context updates
- */
-
-const user: Ref<AmplifyUser | null> = ref(null);
-const signOut = ref();
```
wlee221, line 234: Now inferred directly to `useAuthenticator`, which resolves the race condition between getting `user` and evaluating `state.matches('authenticated').

packages/vue/src/components/authenticator.vue
```diff
@@ -522,10 +481,10 @@ const hasRouteComponent = computed(() => {
...
-    v-if="state?.matches('authenticated')"
-    :user="user"
-    :state="state"
-    :signOut="signOut"
-    :send="send"
+    v-if="route === 'authenticated'"
+    :user="(user as any)"
+    :state="(state as any)"
+    :signOut="(signOut as any)"
+    :send="(send as any)"
```
wlee221, line 488: Added for backwards compatibility and to avoid breaking changes.

From my experience my vscode could not infer the type through slots, but wanted to be sure.

packages/vue/src/components/confirm-reset-password.vue
```diff
@@ -15,8 +14,7 @@ import BaseFormFields from './primitives/base-form-fields.vue';
 
 const { state, send } = useAuth();
 
-const useAuthShared = createSharedComposable(useAuthenticator);
```
wlee221, line 18: This is actually a wrong usage of `createSharedComposable`. 

`createSharedComposable` creates a closure with shared state value when it is called. Previously, each auth subcomponents called its own `createdSharedComposable`, which creates its own closure, and ended up not sharing states.

https://vueuse.org/shared/createSharedComposable/

packages/vue/src/components/sign-up.vue
```diff
@@ -12,9 +11,8 @@ import AuthenticatorSignUpFormFields from './authenticator-sign-up-form-fields.v
 
 import { useAuthenticator } from '../composables/useAuth';
 
-const useAuthShared = createSharedComposable(useAuthenticator);
-const facadeValues = useAuthShared();
-const props = useAuthShared();
+const facadeValues = useAuthenticator();
+const props = useAuthenticator();
```
wlee221, line 15: Didn't want to go too far into refactoring (this needs changing the templates too), so didn't change code outside using the `sharedAuthenticator`.

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,55 @@
...
+import { UseAuth } from '../types';
 import { facade } from './useUtils';
-import { InterpretService } from '@/components';
-
-const service = ref() as Ref<InterpretService>;
-const useAuthenticatorValue = reactive({
-  ...facade,
-  send: '' as unknown,
-  state: '' as unknown,
-}) as any;
```
wlee221, line 12: These top level globals are gone by using `createSharedComposable` below.

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,55 @@
...
-  createValues();
-  watchEffect(() => {
-    createValues();
-  });
-  return useAuthenticatorValue;
-};
 
-export const useAuthenticator = useInternalAuthenticator;
+export const useAuth = createSharedComposable((): UseAuth => {
+  const machine = createAuthenticatorMachine();
```
wlee221, line 18: Unlike React hooks, this gets only run once because composable gets called in `setup` [[docs](https://vuejs.org/guide/reusability/composables.html#conventions-and-best-practices)]. 

Only the reactive part gets updated independently.

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,55 @@
...
-function createValues() {
-  if (!service.value) return;
+  onScopeDispose(() => {
+    unsubscribeHub();
+  });
+
+  return { service: service, send, state };
+});
 
+export const useAuthenticator = createSharedComposable(() => {
```
wlee221, line 32: Using `createSharedComposable` is *not* a breaking change, because we've been relying on the top level global `useAuthenticator` ref value to share the return value.

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,55 @@
...
-  useAuthenticatorValue.send = send;
-  useAuthenticatorValue.state = state;
-}
+  const useAuthenticatorValue = reactive({
+    ...facade,
+    send,
+    state,
+  }) as any;
+
+  watchEffect(() => {
```
wlee221, line 41: By putting this in a shared composable, this actually gets run significantly less since composable only maintains one watcher.

This reduced down the initial rendering of `Authenticator` on `sign-up-with-email` from 15 to 4.

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,55 @@
...
-export const useAuthenticator = useInternalAuthenticator;
+export const useAuth = createSharedComposable((): UseAuth => {
+  const machine = createAuthenticatorMachine();
+  const service: AuthInterpreter = interpret(machine).start();
+
+  const { state, send } = useActor(service);
+
+  const unsubscribeHub = listenToAuthHub(service, defaultAuthHubHandler);
 
-function createValues() {
```
wlee221, line 31: Moved `createdValues` to inside `useAuthenticator` so that it isn't a side-effect anymore.

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,55 @@
...
-
-export const useAuth = (serv?: InterpretService) => {
-  if (serv) {
-    service.value = serv;
-  }
-  return useActor(service.value);
-};
-
-const useInternalAuthenticator = () => {
-  createValues();
```
wlee221, line 22: This initial `createValues` isn't actually necessary because `watchEffect` runs immediately prior to component rendering: https://vuejs.org/api/reactivity-core.html#watcheffect

wlee221:
I'll add unit tests for `useAuth` and `useAuthenticator`.

packages/vue/jest.config.cjs
```diff
@@ -9,7 +9,7 @@ module.exports = {
   ],
   coverageThreshold: {
     global: {
-      branches: 28,
+      branches: 22,
```
wlee221, line 12: I temporarily reduced the threshold since this PR touched Auth subcomponents as part of the refactor. 

I added unit tests to the actual code changes (`useAuth` and `useAuthenticator`) with 100% coverage, so I *think* this is acceptable for the time being as these subcomponents don't even have tests in the first place. 

```
  useAuth.ts                                       |     100 |      100 |     100 |     100 |    
```

packages/vue/src/composables/__tests__/useAuth.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  it('calls unsubscribe when component unmounts', async () => {
+    const wrapper = mount(TestComponent);
+    wrapper.unmount();
+
+    expect(listenToAuthHubSpy).toBeCalledTimes(1);
+    expect(unsubscribeHubSpy).toBeCalledTimes(1);
+  });
+
+  it('unsubscribes to hub only once even if multiple instances are mounted', () => {
```
wlee221, line 96:
```suggestion
  it('unsubscribes only once even if multiple instances are mounted', () => {
```

packages/vue/jest.config.cjs
reesscot: Fine with me, it's naturally going to add more branches due to code extraction. Despite the numbers, we probably actually have more coverage since you tested the extracted code.

reesscot (approved):

packages/vue/jest.config.cjs
ioanabrooks: Would feel better if any of the other values were increased (fn/lines/stmts)

wlee221: That's very fair. Currently I could increase `functions` by 1 and `statements`, but that probably isn't convincing.

Let me add unit tests for `Authenticator.vue`, since that is the primary consumer of the composables. I'll update you when I get it in.

wlee221: Added some tests for `authenticator.vue`. However, it's still difficult to increase the scope with current scope of refactor because of the rest of code has:
1. a lot of unused/unsupported public API ([example](https://github.com/aws-amplify/amplify-ui/blob/main/packages/vue/src/components/authenticator.vue#L147-L225)) and
1. a lot of usage relying on explicit `authState` look up ([example](https://github.com/aws-amplify/amplify-ui/blob/main/packages/vue/src/components/sign-in.vue#L120)).

Talked offline with @ioanabrooks, we will make follow up PRs to increase the coverage.

packages/vue/src/components/authenticator.vue
```diff
@@ -224,41 +208,19 @@ const onConfirmVerifyUserSubmitI = (e: Event) => {
...
 
 const hasRouteComponent = computed(() => {
-  return !(
-    state.value.matches('authenticated') ||
-    state.value.matches('idle') ||
-    state.value.matches('setup') ||
-    state.value.matches('signOut') ||
-    actorState.value?.matches('autoSignIn')
-  );
+  const routesWithoutComponents: AuthenticatorRoute[] = [
```
calebpollman, line 216: Nittish: would prefer evaluation against `routesWithComponents` (and removing the bang on line 223). Idea being that if a new non-component route were added `hasRouteComponent ` would still evaluate as expected for the routes that render

packages/vue/src/components/authenticator.vue
calebpollman: Can we add a comment? Know that there is a good chance this code might just disappear during a refactor, but in case it still exists in the future 😅

packages/vue/src/components/confirm-reset-password.vue
calebpollman: Thanks for digging in here!

packages/vue/src/components/reset-password.vue
```diff
@@ -8,14 +8,12 @@ import {
...
-import { createSharedComposable } from '@vueuse/core';
 
 const attrs = useAttrs();
 const emit = defineEmits(['resetPasswordSubmit', 'backToSignInClicked']);
 
-const useAuthShared = createSharedComposable(useAuthenticator);
-const { send, submitForm } = useAuthShared();
-const { error, isPending } = toRefs(useAuthShared());
+const { send, submitForm } = useAuthenticator();
+const { error, isPending } = toRefs(useAuthenticator());
```
calebpollman, line 16: Just curious, is the reason that we the separate calls to `useAuthenticator` here because `send` and `submitForm` maintain stable values whereas `error` and `isPending` are reactive?

packages/vue/src/components/sign-up.vue
calebpollman: Makes sense, although doing a quick (and uninformed) glance through the file it does read as though we would be able to use either just `facade` or `props`

packages/vue/src/composables/useAuth.ts
```diff
@@ -1,43 +1,55 @@
...
+
+  return { service, send, state };
+});
 
+export const useAuthenticator = createSharedComposable(() => {
   const { state, send } = useAuth();
 
-  const facadeValues = getServiceFacade({ send, state: state.value });
-  for (const key of Object.keys(facade)) {
-    //@ts-ignore
```
calebpollman, line 38: ❤️

calebpollman (dismissed):
🚢

packages/vue/src/components/reset-password.vue
wlee221: Yep, correct. We can reduce it down to one calls, but I'll put that in the subcomponents refactor!

packages/vue/src/components/authenticator.vue
wlee221: Done in 9fec28ccd842ead311f17a2a9de0fe6ea78b9e1f 🙏

packages/vue/src/components/authenticator.vue
wlee221: Agreed, done in 9fec28ccd842ead311f17a2a9de0fe6ea78b9e1f (and sorted in another commit)

ioanabrooks (approved):
LGTM!

zchenwei (commented):
Not going to pretend I understand the logic. But have a random thought, maybe it is worth to create share test utils to use across frameworks going forward like a few mock services are sharable?

zchenwei:
> > But have a random thought, maybe it is worth to create share test utils between frameworks like a few mock services are sharable?
> 
> 100%! Maybe we could have something like the shared config package? That would be great, I admit I copy pasted a lot of them 😅

Yeah, that is a good idea. That can help us reach test parity

PR #3923, opened 2023-05-19
fix: fix translations for enter your email

reesscot:
Please enter the commit message for your changes. Lines starting

#### Description of changes

Our AuthenticatorDictionary TS type didn't match actual defaultTexts constant for ENTER_EMAIL which has "Email" capitalized. This lead to missing translations due to case sensitivity. Will follow up with other translation mismatches, but this only addresses "Enter your Email".

#### Checklist

- [ ] PR description included


packages/ui/src/i18n/dictionaries/authenticator/types.ts
```diff
@@ -17,7 +17,7 @@ type AuthenticatorText =
   | 'Dismiss alert'
   | 'Email'
   | 'Enter your code'
-  | 'Enter your email'
+  | 'Enter your Email'
```
reesscot, line 20: This type was wrong and enforces the wrong translation key actually used by the Authenticator code:
https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/i18n/dictionaries/authenticator/defaultTexts.ts#L30

packages/ui/src/i18n/dictionaries/authenticator/en.ts
```diff
@@ -24,7 +24,7 @@ export const enDict: AuthenticatorDictionary = {
   'Enter your Birthdate': 'Enter your Birthdate',
   'Enter your code': 'Enter your code',
   'Enter your Confirmation Code': 'Enter your Confirmation Code',
-  'Enter your email': 'Enter your email',
+  'Enter your Email': 'Enter your Email',
```
ioanabrooks, line 27: Have you run integration tests (including RN) on this change?

reesscot: Shouldn't they be running below? I think changes to ui package should probably trigger the RN tests?

ioanabrooks: The job run on the initial commit got cancelled in favor of the merge commit which looks like doesn't have changes to relevant packages so RN e2es got skipped.

ioanabrooks (approved):

ioanabrooks (approved):

packages/ui/src/i18n/dictionaries/authenticator/fr.ts
```diff
@@ -20,7 +20,7 @@ export const frDict: AuthenticatorDictionary = {
   'Dismiss alert': `Supprimer l'alerte`,
   Email: 'Email',
   'Enter your code': 'Saisissez cotre code de confirmation',
-  'Enter your email': 'Saisissez votre adresse email',
+  'Enter your Email': 'Saisissez votre adresse email',
```
zchenwei, line 23:
```suggestion
  'Enter your Email': 'Saisissez votre adresse Email',
```

wlee221 (approved):

zchenwei (approved):
LGTM. But yeah, looks like the RN e2e is skipped because a following commit is unrelated. I think it should be fine since the update does not affect en UI and the tests always run against what's actually get rendered on the screen. Plus we will always run the tests on main.

reesscot:
> LGTM. But yeah, looks like the RN e2e is skipped because a following commit is unrelated. I think it should be fine since the update does not affect en UI and the e2e tests always run against what actually gets rendered on the screen. Plus we will always run the tests on main.

The RN e2e test previously ran before the changset commit and passed! 🙌

PR #3986, opened 2023-05-24
chore(e2e): re-enable InApp Messagging RN e2e tests

ioanabrooks:
#### Description of changes

This change:
* re-enables InApp Messaging e2e tests with increased timeout for banner to show (this increased overall Android test job duration to over an hour, optimizations needed to be put in place before merging, details below).
* removes Flipper debugging library for Android as well to improve build time (by ~17%).
* adds fail fast config for cucumber, to have the tests bail at first error, this helps in situations where tests are failing due to metro or simulator/emulator issues, instead of waiting for all tests to fail before re-trying the job.
  * this could be annoying for actual issues, where we would see them one by one but helps our velocity when releasing. 
* speed up android tests by using `replaceText` instead of `typeText` for input fields (reduced from ~50m to ~35m, see below for job links)
* for Android start 2 emulators and run tests in parallel (further reduced to ~27m)
* now that tests are configured to fail fast (fail at first failure), added  setting to retry a failing test at least once, this will save time for flaky tests.

Gaps in testing:
* test Analytic events, currently blocked bc Pinpoint endpoint cannot be mocked (open [ticket](https://github.com/aws-amplify/amplify-js/issues/11371) with js)

Successful runs:
* https://github.com/aws-amplify/amplify-ui/actions/runs/5072926330/jobs/9111620239
  * before removing flipper, build time: 6m 41s

* https://github.com/aws-amplify/amplify-ui/actions/runs/5082164610/jobs/9131905325
  * after removing flipper, build time: 5m 11s

* https://github.com/aws-amplify/amplify-ui/actions/runs/5092544954/jobs/9154360836
* https://github.com/aws-amplify/amplify-ui/actions/runs/5095435634/jobs/9160535386
* https://github.com/aws-amplify/amplify-ui/actions/runs/5107132192/jobs/9181406540
  * 35% Android test run reduction by using replaceText instead of typeText (50min -> 34min)
* https://github.com/aws-amplify/amplify-ui/actions/runs/5112860754/jobs/9191707728
  * with 2 emulators running only about 5min faster, with the downside that it increases chances for errors during emulator startup (since there are 2 of them)
* https://github.com/aws-amplify/amplify-ui/actions/runs/5115358565/jobs/9196734992
  * with 2 emulators, running tests in headless mode, reduced to 27m, less errors experienced during startup
  * note: have tried running tests in parallel before, with macos-12 machines, machines crashed about 40% of the time 

#### Checklist

- [ ] PR description included


examples/react-native/android/app/build.gradle
```diff
@@ -265,7 +265,7 @@ android {
 }
 
 dependencies {
-    androidTestImplementation('com.wix:detox:+') { transitive = true }
+    androidTestImplementation('com.wix:detox:+')
```
ioanabrooks, line 268: As per Detox [docs](https://wix.github.io/Detox/docs/introduction/project-setup#41-patching-build-scripts)

.github/workflows/reusable-e2e.yml
```diff
@@ -510,7 +510,6 @@ jobs:
         run: yarn react-core build
 
       - name: Build react-core-notifications package
-        if: steps.restore-package-cache.outputs.cache-hit != 'true'
```
ioanabrooks, line 513: There is no cache to draw from, since RN e2e tests run on macos machines, previous steps run on ubuntu

.github/workflows/reusable-e2e.yml
```diff
@@ -574,9 +575,7 @@ jobs:
 
       - name: Run E2E tests against react-native example
         run: |
-          # all network requests going from the device on this port will be forwarded to the virtual machine where the mock server is running
-          $ANDROID_HOME/platform-tools/adb reverse tcp:9091 tcp:9091
```
ioanabrooks, line 578: Removed in favor of setting this via Detox config, which Detox 20 now supports.

.github/workflows/reusable-e2e.yml
```diff
@@ -547,24 +546,35 @@ jobs:
 
       - name: Install emulator
         run: |
-          $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --install 'build-tools;33.0.2' platform-tools
+          echo "y" | $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --licenses
+          $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --install 'build-tools;33.0.2' platform-tools 'platforms;android-27'
           echo "y" | $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --install "system-images;android-27;default;x86_64"
           echo "no" | $ANDROID_HOME/cmdline-tools/latest/bin/avdmanager create avd --force --name Pixel_5_API_27 --device "pixel_5" --abi x86_64 --package "system-images;android-27;default;x86_64"
-          printf 'hw.cpu.ncore=2\n' >> ~/.android/avd/Pixel_5_API_27.avd/config.ini
```
ioanabrooks, line 553: Removing this, as logs print `Running on a system with less than 6 logical cores. Setting number of virtual cores to 1`

packages/e2e/README.md
```diff
@@ -33,8 +33,10 @@ To run existing tests on `react-native` framework:
 1. Run `yarn e2e detox:build:android`
 1. Run `yarn e2e detox:e2e:android`
 
-To forward network requests on port 9091 to the mock server run:
-`adb reverse tcp:9091 tcp:9091`
+Troubleshooting
+
+If the emulator fails to connect to Metro server, run:
+`adb reverse tcp:8081 tcp:8081`
```
calebpollman, line 39: Can we update this comment to explicitly state this is an Android only command? Also this comment might be more useful in the example app README

packages/e2e/detox/integration/common/shared.ts
```diff
@@ -1,6 +1,19 @@
...
 import { by, element, expect } from 'detox';
 
+export const typeInInputField = async (
+  inputFieldMatcher: Detox.NativeMatcher,
+  text: string
+) => {
+  const inputField = element(inputFieldMatcher);
+  await device.disableSynchronization();
+  // replaceText is much faster than typeText
+  // downside: doesn't trigger text input callbacks (such as validations) so we need to tap return key
```
calebpollman, line 11: To confirm, we will not be able to validate on `blur` events?

packages/e2e/README.md
ioanabrooks: This is under the Android section of the readme, should be clear it's Android specific..

packages/e2e/detox/integration/common/shared.ts
ioanabrooks: No, this explains why we need to add the command to tap the return key 2 lines below. Taping return triggers the callbacks. Open to rephrasing :)

calebpollman: Still a bit confused, will `change` events still trigger?

ioanabrooks: Yes, they will trigger after text in the field is replaced. So not character by character. We have e2es that test validation errors and are passing. did you have specific concerns?

packages/e2e/README.md
ioanabrooks: Think I edited my prev comment instead of adding new one but: this is an issue for app built by Detox, not otherwise.

calebpollman: Realizing that i should add some context here, this is a command that i sometimes need to run for the example app when developing against Android, so was thinking that it might make sense there because its usefulness is not scoped to just the E2E tests

packages/e2e/detox/integration/common/shared.ts
calebpollman: Yes, want to ensure that this update will not be problematic for validation events in the future. IIRC this is only need for Android?

calebpollman (approved):
LGTM, thanks @ioanabrooks

packages/e2e/README.md
ioanabrooks: Synced offline, will add this note in the examples readme as follow-up.

packages/e2e/detox/integration/common/shared.ts
ioanabrooks: Synced offline, will follow up with single validation test that exercises typing to be covered.

reesscot (approved):

PR #3996, opened 2023-05-25
chore: remove geo and notifications code from ui-react

sreeramsama:
#### Description of changes
- Remove geo and notifications code from `ui-react`
- Remove unused imports in `ui-react` (like maplibre related dependencies)
- Remove geo related styles from `ui-react`
- Update e2e and unit test github workflows to only build `react-core-notifications` for related packages like `react-notifications` and `react-native`.

- unit and e2e tests

#### Checklist

- [ ] PR description included


.github/workflows/reusable-e2e.yml
```diff
@@ -154,10 +154,6 @@ jobs:
         if: ${{ matrix.package == 'react' }}
         run: yarn react-core build
 
-      - name: Build react-core-notifications package
```
reesscot, line 157: Can you explain why this is no longer needed?

.github/workflows/reusable-unit.yml
```diff
@@ -91,10 +91,7 @@ jobs:
 
       - name: Build react-core-notifications package
         if: |
-          matrix.package == 'react' ||
           matrix.package == 'react-native' ||
-          matrix.package == 'react-storage' ||
-          matrix.package == 'react-liveness' ||
```
reesscot, line 97: Was this ever needed? Doesn't seem related

packages/react/package.json
```diff
@@ -59,13 +58,8 @@
     "classnames": "2.3.1",
     "deepmerge": "4.2.2",
     "lodash": "4.17.21",
-    "mapbox-gl": "1.13.1",
```
reesscot, line 62: 💯

packages/react/package.json
```diff
@@ -119,12 +111,6 @@
       "import": "{ Authenticator }",
       "limit": "120 kB"
     },
-    {
```
reesscot, line 122: We need to update our canaries to run the size check on the new packages. Have we done that?

packages/react/src/styles.css
```diff
@@ -1,4 +1 @@
 @import '../../ui/dist/styles.css';
-@import '@maplibre/maplibre-gl-geocoder/dist/maplibre-gl-geocoder.css';
```
reesscot, line 2: 💯

.github/workflows/reusable-e2e.yml
sreeramsama: We do not need to build `react-core-notifications` for `ui-react` anymore as we moved inappmessaging code out and do not have any code depending on `react-core-notifications` now.

Earlier, inappmessaging code in `ui-react` depended on `react-core-notifications`.

.github/workflows/reusable-unit.yml
sreeramsama: It is kind of a dependency chain.
Inappmessaging code in `ui-react` depended on `react-core-notifications`.
`react-storage` depends on `ui-react`, which depended on `react-core-notifications`.
`react-liveness` depends on `ui-react`, which depended on `react-core-notifications`.

We can remove `react`, `react-storage` and `react-liveness` from this condition now as inappmessaging code is being moved out of `react`, so `react` and other `react-*` packages do not need `react-core-notifications`.

We have inappmessaging code in `react-notifications` and `react-native`, so we want to build `react-core-notifications` for those packages.

reesscot: OK, that makes more sense, thanks for clarifying!

reesscot (approved):
LGTM, thanks for the explanations!

.github/workflows/reusable-e2e.yml
sreeramsama: Added this back as we need to build `react-core-notifications` for example app to use.

ioanabrooks (approved):

packages/react/package.json
zchenwei: We finally drop those dependencies!

packages/react/package.json
sreeramsama: Will have a PR to `main` for that

PR #4006, opened 2023-05-30
fix: update link

0618:
#### Description of changes

Update link caught by link-checker https://github.com/aws-amplify/amplify-ui/actions/runs/5124356237/jobs/9216447982?pr=3986

#### Checklist

- [ ] PR description included

### DIFF
--- a/docs/src/pages/[platform]/getting-started/internationalization/internationalization.react.mdx
+++ b/docs/src/pages/[platform]/getting-started/internationalization/internationalization.react.mdx
@@ -3,7 +3,7 @@ import { ThemeProviderDirectionExample } from './examples';
 
 ## What's included?
 
-Amplify UI components follow Web Standards with all components with the aim of making internationalization straightforward. We use [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Margins_borders_padding) for margins, borders, and padding. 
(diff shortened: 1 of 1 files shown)


calebpollman (commented):
I'm not having issues with the old URL but am seeing a redirect when i navigate to it. Does the link checker fail on redirects?

sreeramsama (approved):

0618:
> I'm not having issues with the old URL but am seeing a redirect when i navigate to it. Does the link checker fail on redirects?

Yes, we fail on those redirected links so that we can keep our links up-to-date

calebpollman:
> > I'm not having issues with the old URL but am seeing a redirect when i navigate to it. Does the link checker fail on redirects?
> 
> Yes, we fail on those redirected links so that we can keep our links up-to-date

But if the redirect works we should be good? Not sure if we really need to block our pipeline if there is ultimately a working link still

0618:
> But if the redirect works we should be good? Not sure if we really need to block our pipeline if there is ultimately a working link still

Redirected links are usually outdated links. We fail on those because we'd like to use fresh links. 

If it's to tedious to update 1 or 2 out-dated links, we can accumulate to, say, 5 redirected links before failing.

What do you think?

calebpollman:
> > But if the redirect works we should be good? Not sure if we really need to block our pipeline if there is ultimately a working link still
> 
> Redirected links are usually outdated links. We fail on those because we'd like to use fresh links.
> 
> If it's to tedious to update 1 or 2 out-dated links, we can accumulate to, say, 5 redirected links before failing.
> 
> What do you think?

MSTM

PR #3494, opened 2023-03-02
refactor(docs): link checker to use puppeteer on local

0618:
#### Description of changes

This PR is a refactor of the [Cypress Link Checker](https://github.com/aws-amplify/amplify-ui/blob/main/docs/cypress/e2e/links.cy.ts)

**Major refactor**:
- Tool: Cypress -> Puppeteer
- Run time: Amplify Hosting build -> PR test and other Github e2e tests.
- Logic: visit each page and test each link -> visit each page and save all the links, then test all the links

Note: This PR doesn't implement Github Action change, so it only allows us to test on local. It's still useful because:
- It takes about 10 minutes, which is quicker than Cypress test (~1 hour).
- It makes the PR smaller.

**TODO**:
- [x] compare the time without logging https://github.com/0618/amplify-ui/actions/runs/4326617474/jobs/7554353865
- [x] exit after running the whole test
- [ ] take care of link like https://docs.amplify.aws/lib/auth/getarted/q/platform/js
- [ ] Add Github Action ~~after AppSec Review~~
- [ ] Remove Cypress link checker https://github.com/aws-amplify/amplify-ui/pull/2844

Link-Checker

#### Description of how you validated changes

Tested on local and fork: 

- Happy pass https://github.com/0618/amplify-ui/actions/runs/4306642283/jobs/7510827771
- Found 404 https://github.com/0618/amplify-ui/actions/runs/4289772659/jobs/7473206462
- Found 302 https://github.com/0618/amplify-ui/actions/runs/4309083193/jobs/7516136062 
- Check all the errors and log them at the end https://github.com/0618/amplify-ui/actions/runs/4328714956/jobs/7558728678 (Github Actions doesn't log in order, but it logs the error at the end on local)
- Remove excessive logs https://github.com/0618/amplify-ui/actions/runs/4339875958/jobs/7577965655 
- Found errors with PromisePool https://github.com/0618/amplify-ui/actions/runs/4340073150/jobs/7578331679

<details>
<summary>Comparision of Cypress and Puppeteer Number of Links Tested</summary>

* Cypress: docs-cypress-2023-02-28T08:22:34-08:00.log
* Puppeteer 1: https://github.com/0618/amplify-ui/actions/runs/4289183901/jobs/7472039228
* Puppeteer 2: https://github.com/0618/amplify-ui/actions/runs/4306642283/jobs/7510827771

Page # | Path | Cypress | Puppeteer 1 |   | Puppeteer 2
-- | -- | -- | -- | -- | --
[...]

PR #4022, opened 2023-06-02
test(link-checker): not fail on redirected links

0618:
#### Description of changes

- Current link-checker https://github.com/aws-amplify/amplify-ui/pull/3494 catches all errors other than `200`. 
Amplify UI is ok to keep the redirected links as long as they work, so in this PR, we make the link-checker fail only on errors >`400`. 

- Also fix a bug: external 308 links should not be redirected. The link-checker should only add framework name in front of internal 308 links to test if the links is valid.
Now, if there is an external 308 link (say, https://reactjs.org/docs/higher-order-components.html on react-native page), we will have prepend a framework (it will be like "https://reactjs.org/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/react-native/docs/higher-order-components.html"). With the fix in this PR, we'll stop modifying external links.

PoC PR https://github.com/aws-amplify/amplify-ui/pull/4023

- Happy test without error https://github.com/aws-amplify/amplify-ui/actions/runs/5160735355/jobs/9297064838
- Test with redirected link https://github.com/aws-amplify/amplify-ui/actions/runs/5182389705/jobs/9339122804?pr=4023

#### Checklist

- [ ] PR description included


docs/scripts/link-checker-puppeteer.ts
```diff
@@ -51,7 +51,10 @@ function reportResult(links: LinkInfo[]) {
         }
       }
     );
-    throw new Error(`${errorLinks.length} broken links found`);
+    const failedLinks = errorLinks.filter((link) => link.statusCode >= 400);
+    if (failedLinks.length) {
+      throw new Error(`${errorLinks.length} broken links found`);
```
thaddmt, line 56: Nit: Should this be failedLinks.length?

thaddmt (dismissed):

docs/scripts/util/checkLink.ts
```diff
@@ -46,12 +49,12 @@ async function returnStatus({
      * e.g. 308 from "/guides" passes (because it's redirected to "react/guides"),
      *  but 308 from "connected-components/authenticator/headless" fails and returns a 404 ((because it's redirected to "react/connected-components/authenticator/advanced").
      */
-    if (statusCode === 308) {
-      const hostNameRegex = RegExp(`http(s)?:\/\/[^/]*`, 'i'); // matches everything between http(s)?: to "/", which is the hostname. e.g., "https://github.com/".
-      const platform = pageUrl.replace(hostNameRegex, '').split('/')[1];
-      const newHref = `${
-        href.match(hostNameRegex)[0]
-      }/${platform}${href.replace(hostNameRegex, '')}`;
+    if (statusCode === 308 && isInternalLink(href)) {
```
0618, line 52: In the comment on line 46, it says "check if it's an internal redirection...", but there was not logic for that, which caused the bug mentioned in the PR description.

So, add `isInternalLink` here

docs/scripts/util/checkLink.ts
```diff
@@ -18,6 +18,9 @@ export type LinkInfo = {
 };
 const requestedUrl: Set<string> = new Set();
 
+const hostName = 'http://localhost:3000';
```
reesscot, line 21: Won't this fail when the docs are running in production mode on a different port?

0618: That's a good point, which was the reason why it used a regex to find the hostname.

However, link-checker only test against localhost in the reusable-e2e https://github.com/aws-amplify/amplify-ui/blob/00129b743a7c117e0c9a3bf7cb834203c1c855e5/.github/workflows/reusable-e2e.yml#L687, so 'http://localhost:3000' is good enough for our current use case.

Do you think we should handle more use cases?

reesscot: But surely we aren't running docs in dev mode in e2e tests? Do we have some config that's making it pick port 3000?

0618: BTW, the target site's hostName is set in the environment variable, which is 'http://localhost:3000' in Github

https://github.com/aws-amplify/amplify-ui/blob/1bf5b330d108fb4eee28f3951fd5b0a45b4547b0/docs/src/data/sitePaths.ts#L8

If we turn it on in Prod, we'll need to change `SITE_URL` in Prod env.

docs/scripts/link-checker-puppeteer.ts
0618: Done 2d1d1956afbeabd1ae3f3cf4f4872521c1da839e

docs/scripts/util/checkLink.ts
0618: If we want to test against Prod, I think we can have `Build and RunTime Test` to do that. So that'll be like

- PR test: localhost
- Build and RunTIme test: Prod "https://ui.docs.amplify.aws/"

Does that make sense?

reesscot: I was wrong about this, just tried locally and next.js prod build runs on port 3000, like you have.

reesscot (approved):

thaddmt (approved):
LGTM

PR #4026, opened 2023-06-05
fix(liveness): fix flickering progress bar bug

thaddmt:
#### Description of changes
- Updated progress bar to continue showing when face is matched
- Flickering occurred because when the face match state was `MATCHED` we would hide the progress bar

foobar

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
@@ -69,7 +69,7 @@ export const Hint: React.FC<HintProps> = ({ hintDisplayText }) => {
     [FaceMatchState.TOO_MANY]: hintDisplayText.hintTooManyFacesText,
     [FaceMatchState.TOO_CLOSE]: hintDisplayText.hintTooCloseText,
     [FaceMatchState.TOO_FAR]: hintDisplayText.hintTooFarText,
-    [FaceMatchState.MATCHED]: undefined,
+    [FaceMatchState.MATCHED]: hintDisplayText.hintHoldFaceForFreshnessText,
   };
 
   const IlluminationStateStringMap: Record<IlluminationState, string> = {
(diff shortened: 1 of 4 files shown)


0618 (approved):
So we no longer want to hide the bar once the face is matched?

thaddmt:
> So we no longer want to hide the bar once the face is matched?

that's correct, I think it may have been a bug that it was hidden before since that was causing the flickering progress bar

packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
```diff
@@ -153,7 +153,7 @@ export const Hint: React.FC<HintProps> = ({ hintDisplayText }) => {
       // the TOO_CLOSE text, but for FACE_IDENTIFED, CANT_IDENTIFY, TOO_MANY
       // we are defaulting to the TOO_FAR text (for now). For MATCHED state,
       // we don't want to show any toasts.
-      return faceMatchState !== FaceMatchState.MATCHED ? (
+      return (
```
wlee221, line 156: Are we removing this condition because hint flickered too?

packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
```diff
@@ -69,7 +69,7 @@ export const Hint: React.FC<HintProps> = ({ hintDisplayText }) => {
     [FaceMatchState.TOO_MANY]: hintDisplayText.hintTooManyFacesText,
     [FaceMatchState.TOO_CLOSE]: hintDisplayText.hintTooCloseText,
     [FaceMatchState.TOO_FAR]: hintDisplayText.hintTooFarText,
-    [FaceMatchState.MATCHED]: undefined,
+    [FaceMatchState.MATCHED]: hintDisplayText.hintHoldFaceForFreshnessText,
```
wlee221, line 72: Why has this changed?

packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
thaddmt: Yep, the hint text is the text like `move back`, `move forward`, or `hold still` that shows above the progress bar. The progress bar and the hint text were both flickering.

packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
thaddmt: Previously when in the `MATCHED` state we did not show any hint text or progress bar which was the source of the flickering because you need to stay in the matched state for a period of time. If you were to move out of the `MATCHED`state the progress bar would re-appear with the hint text and thus the source of the flickering when moving between states

wlee221 (approved):

reesscot (approved):

PR #4054, opened 2023-06-07
StorageImage Component

### LINKED ISSUE #3428: AmplifyS3Image - Not available in latest react library

React

Other

I was looking to use _**AmplifyS3Image**_ component as per this below documentation but looks like this component is not available in the latest React package. Is there a plan to include this in near future? Is there a way I can still include and use it in my project? May be using legacy package?

https://docs.amplify.aws/ui/storage/s3-image/q/framework/react/ 

Error on `npm start`

`export 'AmplifyS3Image' (imported as 'AmplifyS3Image') was not found in '@aws-amplify/ui-react' (possible exports: AccountSettings, Alert, AmplifyProvider, Authenticator, Autocomplete, Badge, Button, ButtonGroup, Card, CheckboxField, Collection, ComponentClassNames, ComponentClassObject, ComponentPropsToStylePropsMap, ComponentPropsToStylePropsMapKeys, Divider, Expander, ExpanderItem, FieldGroupIcon, FieldGroupIconButton, FileUploader, Flex, Geocoder, Grid, Heading, HighlightMatch, Icon, Image, InAppMessageDisplay, InAppMessagingProvider, Link, Loader, LocationSearch, MapView, Menu, MenuButton, MenuItem, Pagination, PasswordField, PhoneNumberField, Placeholder, Radio, RadioGroupField, Rating, ScrollView, SearchField, SelectField, SliderField, StepperField, SwitchField, TabItem, Table, TableBody, TableCell, TableFoot, TableHead, TableRow, Tabs, Text, TextAreaField, TextField, ThemeProvider, ToggleButton, ToggleButtonGroup, View, VisuallyHidden, components, createTheme, defaultDarkModeOverride, defaultTheme, primitives, translations, useAmplify, useAuthenticator, useBreakpointValue, useInAppMessaging, usePagination, useTheme, withAuthenticator, withInAppMessaging)`

_No response_

- [ ] ⚠️ This feature might incur a breaking change.

zchenwei:
#### Description of changes

The PR adds a new StorageImage that allows users to load the images managed by Amplify Storage. Doc updates and e2e tests will be included in separate PRs.

*Example:*
```jsx
import { StorageImage } from '@aws-amplify/ui-react-storage';

<StorageImage alt="StorageImage" imgKey="image.jpg" accessLevel="public" />
```

#### Issue #, if available

Closes #3428 

#### Checklist

- [x] PR description included


packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,29 @@
...
+  onStorageError,
+  ...rest
+}: StorageImageProps): JSX.Element => {
+  const { url, error } = useStorageURL(
+    imgKey,
+    { level: accessLevel, identityId },
+    fallbackSrc
+  );
+
+  if (error && onStorageError) {
```
calebpollman, line 22: Can we use isFunction here rather than checking for truthiness of `onStorageError`?

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,29 @@
...
+    onStorageError(error);
+  }
+
+  return (
+    <Image className={ComponentClassNames.StorageImage} src={url} {...rest} />
```
calebpollman, line 27: Prefer rest props ahead of explicit props with intentional overrides, for example:
```
export const StorageImage = ({
  imgKey,
  accessLevel,
  identityId,
  fallbackSrc,
  onStorageError,
  className = ComponentClassNames.StorageImage,
  src,
  ...rest
}: StorageImageProps): JSX.Element => {
  // ...stuff
  return <Image {...rest} classname={className} src={src ? src : url} />
}
```

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const StorageImage = ({
+  imgKey,
+  accessLevel,
+  identityId,
+  fallbackSrc,
+  onStorageError,
+  ...rest
+}: StorageImageProps): JSX.Element => {
+  const { url, error } = useStorageURL(
```
calebpollman, line 16: Can we prevent this hook from running if the `src` prop is provided?

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useStorageURL } from '@aws-amplify/ui-react/internal';
+
+import type { StorageImageProps } from './types';
+
+export const StorageImage = ({
+  imgKey,
+  accessLevel,
+  identityId,
+  fallbackSrc,
+  onStorageError,
```
calebpollman, line 13: Why `onStorageError` instead of `onError`?

packages/react-storage/src/components/StorageImage/_tests_/StorageImage.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+        onStorageError={onStorageError}
+      />
+    );
+
+    const img = await screen.findByRole('img');
+    expect(onStorageError).not.toHaveBeenCalled();
+    expect(img).toHaveAttribute('src', imgURL);
+  });
+
+  it('should set image src attribute to fallbackSrc and invoke onStorageError when Storage get is rejected', async () => {
```
calebpollman, line 58: Future readers may not have immediate context in to "get" being a method of `Storage`:
```suggestion

  it('should set image src attribute to fallbackSrc and invoke onStorageError when Storage.get is rejected', async () => {
```

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Image, ComponentClassNames } from '@aws-amplify/ui-react';
+import { useStorageURL } from '@aws-amplify/ui-react/internal';
+
+import type { StorageImageProps } from './types';
+
+export const StorageImage = ({
+  imgKey,
+  accessLevel,
+  identityId,
+  fallbackSrc,
```
calebpollman, line 12: What is then expected behavior  if `imgKey`, `fallbackSrc` and `src` are all provided?

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -14,31 +14,35 @@ export interface UseStorageURLResult {
...
+  fallbackURL?: string
 ): UseStorageURLResult & { fetch: () => () => void } => {
   const [result, setResult] = React.useState<UseStorageURLResult>({
     isLoading: true,
   });
 
   // Used to prevent an infinite loop on useEffect, because `options`
   // will have a different reference on every render
-  const serializedOptions = JSON.stringify(options);
+  const optionsRef = React.useRef(options);
```
calebpollman, line 26: What if `options` changes?

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -14,31 +14,35 @@ export interface UseStorageURLResult {
...
 
+    const options = optionsRef.current;
     const promise = Storage.get(key, options);
 
     // Attempt to fetch storage object url
     promise
       .then((url) => setResult({ url, isLoading: false }))
-      .catch((error: Error) => setResult({ error, isLoading: false }));
+      .catch((error: Error) =>
+        setResult({ url: fallbackURL, error, isLoading: false })
```
calebpollman, line 38: Do we need to verify the validity of `fallbackUrl`?

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -14,31 +14,35 @@ export interface UseStorageURLResult {
...
+      );
 
     // Cancel current promise on unmount
     return () => Storage.cancel(promise);
   };
 
-  React.useEffect(fetch, [key, options, serializedOptions]);
+  React.useEffect(fetch, [key, fallbackURL]);
```
calebpollman, line 45: The usage of `key` and `fallbackUrl` as dependencies of this effect is opaque, can you please add a comment on why `fetch` should be rerun if they change? Actually, should `fetch` rerun if `fallbackUrl` changes?

packages/react/src/primitives/types/image.ts
```diff
@@ -48,6 +54,12 @@ export interface ImageOptions extends ImageStyleProps {
    */
   srcSet?: React.ImgHTMLAttributes<HTMLImageElement>['srcSet'];
 
+  /**
+   * @description
+   * Image will be loaded lazily if set to true.
+   */
+  isLazy?: boolean;
```
calebpollman, line 61: Not sure if this will scale if we are to support "eager" in the future since having `isEager` will conflict `isLazy` if both are `true`

calebpollman (commented):
Great start @zchenwei!

packages/react-storage/src/components/StorageImage/StorageImage.tsx
dbanksdesign: I believe onStorageError is if the Storage.get command fails somehow, and onError is the HTML onError event for an image element?

packages/react/src/hooks/useStorageURL.ts
thaddmt, line 26: Just for my own understanding why is options different on each render?

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -14,31 +14,35 @@ export interface UseStorageURLResult {
...
 
   const fetch = () => {
     setResult({ isLoading: true });
 
+    const options = optionsRef.current;
     const promise = Storage.get(key, options);
 
     // Attempt to fetch storage object url
     promise
       .then((url) => setResult({ url, isLoading: false }))
```
thaddmt, line 36: Have we tested what happens if the key is not an image?

packages/react/src/primitives/types/image.ts
```diff
@@ -48,6 +54,12 @@ export interface ImageOptions extends ImageStyleProps {
    */
   srcSet?: React.ImgHTMLAttributes<HTMLImageElement>['srcSet'];
 
+  /**
+   * @description
+   * Image will be loaded lazily if set to true.
```
thaddmt, line 59: Should we note what the default value will be here?

packages/react-storage/src/components/StorageImage/StorageImage.tsx
calebpollman: Got it. Is `onStorageError` too generic since it refers to `Storage` directly? Not sure if there will be additions to this component's  interactions with `Storage` in the future, but if so might be kind of having `onStorageError` and adding a more specific namespaces for other `Storage` error handlers. Maybe `onStorageGetError` or `onGetError`?

packages/react-storage/src/components/StorageImage/StorageImage.tsx
zchenwei: Think we decided to omit the `src` prop on `StorageImage` and instead provide  a `fallbackSrc` prop which makes more sense semantically and functionally in case the Storage.get call gets rejected. People should go right ahead using the `Image` primitive with `src` prop.

packages/react/src/hooks/useStorageURL.ts
zchenwei: It is programmatically possible. In reality, is there a use case that users would update the configs dynamically with the same object? 😅

zchenwei: If a user call `useStorageURL` and construct the option in line, then a new object will be created on each render and thus causes an infinite loop in `useEffect`

```jsx
useStorageURL(
    imgKey,
    { xxx }
  )
```

packages/react/src/hooks/useStorageURL.ts
zchenwei: An URL will always be generated from the API, but if it is not meant for an image, the `img` tag will not load it and trigger `onError`

packages/react/src/hooks/useStorageURL.ts
calebpollman: > In reality, is there a use case that users would update the configs dynamically with the same object?

If the user passes an updated `options` then the output is going to be incorrect since the new `options` will just be ignored. Also this is a breaking change from the current implementation

packages/react/src/primitives/types/image.ts
zchenwei: Think we do not need an extra API for `eager` coz these 2 are mutual exclusive and modifies the same `loading` attribute, which by default is `eager` if `undefined`

packages/react/src/hooks/useStorageURL.ts
zchenwei: Think that will require a new HTTP request,. Since the `img` tag will try to make the request anyway and triggers `onError` if that's invalid, think should be sufficient.

packages/react/src/hooks/useStorageURL.ts
zchenwei: Sure, will add a comment. I was thinking initialize url to `fallbackURL` so that we do not need to put it into `useEffect` but then realized this was wrong because that way `img` will try to load the fallback in the first place 😅 .

packages/react/src/hooks/useStorageURL.ts
zchenwei: We never use the serialized one and it causes infinite loop at the beginning 😅  <- https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/hooks/useStorageURL.ts#L25-L41

I can make the change to serialize the object but still think the use case is rare. It feels strange that you would update the options programmatically in an app without changing the key to another object. Like in [`useDataStore` hook](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/hooks/useDataStore.tsx#L50), we set the dependency array to `[]` though some other configs may change

packages/react-storage/src/components/StorageImage/StorageImage.tsx
zchenwei: I am open to any idea about the naming here. `onStorageGetError` sounds good to me.

packages/react/src/primitives/types/image.ts
zchenwei: Makes sense

packages/react/src/primitives/types/image.ts
calebpollman: What if additional values are allowed for `loading` in the future?

packages/react/src/hooks/useStorageURL.ts
calebpollman: Can we add a unit test ensuring this behavior?

packages/react/src/hooks/useStorageURL.ts
calebpollman: Still unclear on why a change to `fallbackUrl` should trigger a new fetch?

packages/react/src/hooks/useStorageURL.ts
zchenwei: Agree to make it be part of the dependency. But if we do not serialize it, how could you avoid infinite loop?

packages/react/src/primitives/types/image.ts
zchenwei: What about just remove this prop then since we already expose the HTML attributes? Users can set the `loading` attribute anyway if necessary

packages/react/src/hooks/useStorageURL.ts
zchenwei: Think we can move it out of the `useEffect` by:
```js
...
if (result.error) {
   result.url = fallbackURL;
}
...
```

packages/react/src/hooks/useStorageURL.ts
zchenwei: Sure

packages/react-storage/src/components/StorageImage/StorageImage.tsx
reesscot: +1 for `onStorageGetError` since it's more specific.

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+
+import { isFunction } from '@aws-amplify/ui';
+import { Image, ComponentClassNames } from '@aws-amplify/ui-react';
+import { useStorageURL } from '@aws-amplify/ui-react/internal';
+
+import type { StorageImageProps } from './types';
+
+export const StorageImage = ({
+  imgKey,
```
reesscot, line 10: nit: alphabetize props

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+
+import { isFunction } from '@aws-amplify/ui';
+import { Image, ComponentClassNames } from '@aws-amplify/ui-react';
+import { useStorageURL } from '@aws-amplify/ui-react/internal';
+
+import type { StorageImageProps } from './types';
```
reesscot, line 7: 💯

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,30 @@
...
+    { level: accessLevel, identityId },
+    fallbackSrc
+  );
+
+  if (error && isFunction(onStorageError)) {
+    onStorageError(error);
+  }
+
+  return (
+    <Image {...rest} className={ComponentClassNames.StorageImage} src={url} />
```
reesscot, line 28: Should we allow passing a custom classname?

zchenwei: Yeah, thats a good callout

packages/react/src/primitives/types/image.ts
reesscot: Agreed that we just drop this prop since `loading` is already supported natively.

packages/react/src/hooks/useStorageURL.ts
reesscot: I don't think we need to test this behavior as it's standard browser behavior and we can't control whether customer provide a valid fallbackUrl.

packages/react/src/primitives/types/image.ts
zchenwei: Will drop it

wlee221 (approved):

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import type { StorageImageProps } from './types';
+
+export const StorageImage = ({
+  accessLevel,
+  className,
+  fallbackSrc,
+  identityId,
+  imgKey,
+  onStorageGetError,
```
wlee221, line 16: Have we considered a simpler name like `onError`?

zchenwei: `onError` is the native built-in handler which will trigger when the image failed to load. We need a separate handler to deal with API error. See here - https://github.com/aws-amplify/amplify-ui/pull/4054#discussion_r1231355583

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -0,0 +1,36 @@
...
+}: StorageImageProps): JSX.Element => {
+  const { url, error } = useStorageURL(
+    imgKey,
+    { level: accessLevel, identityId },
+    fallbackSrc
+  );
+
+  if (error && isFunction(onStorageGetError)) {
+    onStorageGetError(error);
+  }
```
calebpollman, line 27: This should be in an effect right?

packages/react-storage/src/components/StorageImage/_tests_/StorageImage.test.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Storage } from 'aws-amplify';
+import { ComponentClassNames } from '@aws-amplify/ui-react';
+
+import { StorageImage } from '../StorageImage';
+
+describe('StorageImage:', () => {
```
calebpollman, line 9: Nit: remove colon
```suggestion
describe('StorageImage', () => {
```

packages/react-storage/src/components/StorageImage/StorageImage.tsx
calebpollman: Actually now that i think abt it, shouldn't the callback just be passed to `useStorageUrl`?

packages/react-storage/src/components/StorageImage/_tests_/StorageImage.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+        onStorageGetError={onStorageError}
+      />
+    );
+
+    const img = await screen.findByRole('img');
+    expect(onStorageError).not.toHaveBeenCalled();
+    expect(img).toHaveAttribute('src', imgURL);
+  });
+
+  it('should set image src attribute to fallbackSrc and invoke onStorageError when Storage.get is rejected', async () => {
```
calebpollman, line 59:
```suggestion
  it('should set image src attribute to fallbackSrc and invoke onGetStorageError when Storage.get is rejected', async () =>
```

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -14,19 +14,21 @@ export interface UseStorageURLResult {
...
 
   // Used to prevent an infinite loop on useEffect, because `options`
   // will have a different reference on every render
-  const serializedOptions = JSON.stringify(options);
+  const serializedOptions = JSON.stringify(options ?? {});
 
   const fetch = () => {
     setResult({ isLoading: true });
 
+    const options = JSON.parse(serializedOptions) as S3ProviderGetConfig;
```
calebpollman, line 31: Please do not reuse namespaces already declared in the outer scope of a function. It increases the cognitive load for readers attempting to grok what is happening when two variables share a name

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -38,7 +40,12 @@ export const useStorageURL = (
     return () => Storage.cancel(promise);
   };
 
-  React.useEffect(fetch, [key, options, serializedOptions]);
+  React.useEffect(fetch, [key, serializedOptions]);
+
+  // Set the url to fallbackURL if error happens
+  if (result.error) {
+    result.url = fallbackURL;
+  }
```
calebpollman, line 48: This should be done inside the catch block of `promise` to follow React conventions

packages/react/src/hooks/useStorageURL.ts
calebpollman: The consuming component can just memoize `options` before passing it to `useStorageUrl`

zchenwei: That's true but `useStorageURL` could be used somewhere else as a hook. Think it should handle the case internally instead of leaving the issue to its consumers?

packages/react/src/hooks/useStorageURL.ts
zchenwei: 😄 See our discussion before, that will make `fallbackURL` a dependency in the effect <- https://github.com/aws-amplify/amplify-ui/pull/4054#discussion_r1231366005
But yeah, I do not have a strong opinion on this.

packages/react/src/hooks/useStorageURL.ts
zchenwei: Then we can do whatever we want

packages/react-storage/src/components/StorageImage/StorageImage.tsx
zchenwei: Yeah, for sure. Lets remove the error from return and do it internally

packages/react-storage/src/components/StorageImage/_tests_/StorageImage.test.tsx
zchenwei: 💯

packages/react/src/hooks/useStorageURL.ts
```diff
@@ -14,31 +15,32 @@ export interface UseStorageURLResult {
...
-    setResult({ isLoading: true });
-
-    const promise = Storage.get(key, options);
-
-    // Attempt to fetch storage object url
-    promise
-      .then((url) => setResult({ url, isLoading: false }))
-      .catch((error: Error) => setResult({ error, isLoading: false }));
+  options?: S3ProviderGetConfig,
+  errorConfig?: UseStorageURLErrorConfig
```
calebpollman, line 19: This can be done in a follow up PR but imo would be a cleaner API if we extracted the key/values from `errorConfig` and updated the params `useStorageURL` to be an object:
```ts
type UseStorageURLParams {
  key: string;
  fallbackURL?: string;
  onStorageGetError?: (error: Error) => void;
  options?: S3ProviderGetConfig;
}
```

calebpollman (approved):
🚢

packages/react/src/hooks/useStorageURL.ts
zchenwei: Yeah, for sure. This is a miss on me.

zchenwei: Addressed in #4253

PR #4057, opened 2023-06-08
fix vue and add modifyJSON

0618:
#### Description of changes

- import _aws-exports.js_ to Vue Vite mega-apps
- add modifyJSON to modify JSON config.

**Why adding modify JSON?**
1. It's more like a user behavior -- to modify the exist file, instead of copy-paste from a template file 
2. It's more flexible and don't have to update the template manually -- the default JSON config may be updated in a new release

Note: There will be a follow-up PR to implement modifyJSON for other mega-apps

PoC PR: https://github.com/aws-amplify/amplify-ui/pull/4058

#### Checklist

- [ ] PR description included


.github/scripts/modifyJSON.js
0618, line 1: esbuild-register is not available in .github/scripts/mega-app-copy-files.sh, so use js instead of ts.

.github/scripts/mega-app-copy-files.sh
```diff
@@ -82,6 +82,10 @@ fi
 if [[ "$FRAMEWORK" == 'vue' ]]; then
     echo "cp templates/components/vue/App.vue mega-apps/${MEGA_APP_NAME}/src/App.vue"
     cp templates/components/vue/App.vue mega-apps/${MEGA_APP_NAME}/src/App.vue
+    echo "cp $AWS_EXPORTS_FILE mega-apps/${MEGA_APP_NAME}/src/aws-exports.js"
+    cp $AWS_EXPORTS_FILE mega-apps/${MEGA_APP_NAME}/src/aws-exports.js
+    echo "node ../.github/scripts/modifyJSON.js mega-apps/${MEGA_APP_NAME}/tsconfig.json add \"compilerOptions.allowJs true\""
+    node ../.github/scripts/modifyJSON.js mega-apps/${MEGA_APP_NAME}/tsconfig.json add "compilerOptions.allowJs true"
```
ioanabrooks, line 88: Can't we just have a different tsconfig.json per framework (i.e. `mega-apps/${MEGA_APP_NAME}/${FRAMEWORK}/tsconfig.json`)? Having a script that traverses the json seems unnecessary and am concerned with maintaining it.

0618: "have a different tsconfig.json per framework" is our current solution, which is problematic as mentioned in the PR description.

Also, it doesn't traverse the json file, it convert the json file to a Javascript object.

I think adding or removing a key from the JSON file is more like a user behavior.

If we could have modified the file earlier, we might avoid several failures because the `create-xxx` library changed their default config, which made our template outdated.

ioanabrooks: Got it, super valid. Concerned with maintainability of complex-ish scripts with no testing. Let's add more input validation and console.logs. I'll add comments below.

.github/scripts/modifyJSON.js
```diff
@@ -0,0 +1,62 @@
...
+// Parse the option parameter
+const [path, value] = option.split(' ');
+
+// Modify JSON based on the action
+if (action === 'add') {
+  // Get the property value to add
+  const propertyValue = JSON.parse(value);
+  // Traverse the path and add the property value
+  let current = json;
+  const keys = path.split('.');
```
ioanabrooks, line 31: What happens if I pass in a path that is not valid, not in the file already, or just typo?

.github/scripts/modifyJSON.js
```diff
@@ -0,0 +1,62 @@
...
+  current[keys[keys.length - 1]] = propertyValue;
+} else if (action === 'delete') {
+  // Traverse the path and delete the property
+  let current = json;
+  const keys = path.split('.');
+
+  for (let i = 0; i < keys.length - 1; i++) {
+    const key = keys[i];
+    if (!current[key]) {
+      console.log('Error: Invalid path');
```
ioanabrooks, line 46: Can we make this message more explicit and actionable? Like $path doesn't exist in the file, maybe also print out the contents of the  file to help identify the issue?

.github/scripts/modifyJSON.js
```diff
@@ -0,0 +1,62 @@
...
+} else {
+  console.log('Error: Invalid action');
+  process.exit(1);
+}
+
+// Write the modified JSON back to the file
+const modifiedData = JSON.stringify(json, null, 2);
+fs.writeFileSync(file, modifiedData, 'utf8');
+
+console.log(`JSON file '${file}' modified successfully.`);
```
ioanabrooks, line 62: Maybe we can also print out the file for easier debugging.

build-system-tests/templates/components/vue/App.vue
```diff
@@ -9,7 +9,6 @@
 
 <script lang="ts">
 import { defineComponent } from 'vue';
-// @ts-ignore
```
ioanabrooks, line 12: Do we even need HelloWorld component?

.github/scripts/modifyJSON.js
0618: If it's not there already, we'll add the key even if it's a typo, we'd add it 😂

build-system-tests/templates/components/vue/App.vue
0618: Right. We don't need the HelloWorld component at all. I didn't remove it because I didn't want to add an extra step which is not relevant to our test.

.github/scripts/mega-app-copy-files.sh
wlee221: Have you considered using 3p packages like `json`? e.g.

```bash
# might need to strip the comment first
json -I -f package.json -e "this.compilerOptions.allowJs=true"
```

0618: That's a great idea. I didn't consider using a 3p package since it's a pretty simple script, but yea, `json` looks like a solid option. I'll update the PR using that.

0618: Done 4c3eae26d20047ca7bc24bfa626b1802aab1f0ec

.github/scripts/mega-app-copy-files.sh
```diff
@@ -82,6 +82,12 @@ fi
 if [[ "$FRAMEWORK" == 'vue' ]]; then
     echo "cp templates/components/vue/App.vue mega-apps/${MEGA_APP_NAME}/src/App.vue"
     cp templates/components/vue/App.vue mega-apps/${MEGA_APP_NAME}/src/App.vue
+    echo "cp $AWS_EXPORTS_FILE mega-apps/${MEGA_APP_NAME}/src/aws-exports.js"
+    cp $AWS_EXPORTS_FILE mega-apps/${MEGA_APP_NAME}/src/aws-exports.js
+
+    # Add `allowJs: true` to tsconfig for aws-exports.js
+    echo "npx json -I -f mega-apps/${MEGA_APP_NAME}/tsconfig.json -e \"this.compilerOptions.allowJs=true\""
+    npx json -I -f mega-apps/${MEGA_APP_NAME}/tsconfig.json -e "this.compilerOptions.allowJs=true"
```
wlee221, line 90:
```suggestion
    npx json@11.0.0 -I -f mega-apps/${MEGA_APP_NAME}/tsconfig.json -e "this.compilerOptions.allowJs=true"
```

just for extra safety 🙏

wlee221 (approved):
Just one comment, but looks good!

ioanabrooks (approved):

.github/scripts/mega-app-copy-files.sh
0618: Good point. I locked the package in package.json b2155aeea115b972356432279c28ee46d78bc416

PR #4062, opened 2023-06-08
chore(liveness): increase ws connection timeout to 10s, default was 2s

thaddmt:
#### Description of changes
- Original WS connection timeout was 2s
- On slower connections we would see the connection timing out and retrying 3 additional times
- Tested with Slow 3G throttle profile on chrome which adds 2s of latency to each request

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
@@ -10,6 +10,7 @@ import {
   RekognitionStreamingClientConfig,
   StartFaceLivenessSessionCommand,
 } from '@aws-sdk/client-rekognitionstreaming';
+import { WebSocketFetchHandler } from '@aws-sdk/middleware-websocket';
 import { VideoRecorder } from './videoRecorder';
 
 export interface StartLivenessStreamInput {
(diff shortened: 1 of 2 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
```diff
@@ -117,6 +118,7 @@ export class LivenessStreamProvider extends AmazonAIInterpretPredictionsProvider
       credentials,
       region: this.region,
       customUserAgent: getAmplifyUserAgent(),
+      requestHandler: new WebSocketFetchHandler({ connectionTimeout: 10_000 }),
```
thaddmt, line 121: What should the value of timeout here be? Currently set to 10s but this is probably too high

zchenwei: 10 kinda makes sense though base on the number you mentioned in the description, plus high timeout does not cause side effects for users on fast network but it allows more wait time for users on slower network?

thaddmt: that's correct, mostly affects slower network users

ayush987goyal (approved):

zchenwei (approved):

reesscot (approved):
LGTM

PR #4063, opened 2023-06-08
fix(build system test): Angular mega app setup

0618:
#### Description of changes

Several changes for Angular:

- re-install for ng as a workaround for https://github.com/aws-amplify/amplify-js/issues/11455
- remove angular from dependencies since it's deprecated ([ref](https://www.npmjs.com/package/angular))
- remove `--minimal` because users don't use it for prod
- not install dependencies for lib app

PoC PR: https://github.com/aws-amplify/amplify-ui/pull/4064/files

#### Checklist

- [ ] PR description included


.github/scripts/mega-app-install-build.sh
```diff
@@ -4,11 +4,17 @@ echo "cd build-system-tests/mega-apps/${MEGA_APP_NAME}"
...
 if [ "$FRAMEWORK" == 'react' ]; then
-    echo "\$DEPENDENCIES='\$DEPENDENCIES react-dom@$FRAMEWORK_VERSION @aws-amplify/ui-react-storage @aws-amplify/ui-react-geo @aws-amplify/ui-react-notifications'"
+    # add react-dom
+    echo "DEPENDENCIES='\$DEPENDENCIES react-dom@$FRAMEWORK_VERSION @aws-amplify/ui-react-storage @aws-amplify/ui-react-geo @aws-amplify/ui-react-notifications'"
     DEPENDENCIES="$DEPENDENCIES react-dom@$FRAMEWORK_VERSION @aws-amplify/ui-react-storage @aws-amplify/ui-react-geo @aws-amplify/ui-react-notifications"
-    echo "Dependencies to be installed: $DEPENDENCIES"
+elif [ "$FRAMEWORK" == 'angular' ]; then
+    # remove angular
+    echo "DEPENDENCIES="@aws-amplify/ui-$FRAMEWORK aws-amplify""
+    DEPENDENCIES="@aws-amplify/ui-$FRAMEWORK aws-amplify"
```
thaddmt, line 13: Think i'm a little confused by this comment, is this pointing out the lack of the `angular` library here?

0618: Because Angular is depracated as mentioned in the description https://www.npmjs.com/package/angular

I can make it more clear in the comment

0618: Updated comment 7995bdf05e8df8483e05fb59fdba2b2af756c732

.github/workflows/build-system-test.yml
```diff
@@ -70,29 +70,15 @@ jobs:
 
           - framework: angular
             framework-version: latest
-            build-tool: angular-app
+            build-tool: angular-cli
```
wlee221, line 73: ✨

.github/scripts/mega-app-install-build.sh
wlee221: I think the bash files are becoming too tightly coupled with the workflow, which makes it hard to use and maintain since we need to make workarounds like this `if` statement. I propose we make the workflow dumber and make the bash file smarter to avoid it.

I would propose something like this:

```bash
# can use JS/TS to ease the implementation
./mega-app.setup.sh --framework="react" --framework-version="16" --build-tool="cra" --pkg-manager="yarn" --language="ts"
```

(we used a similar approach in [`environments/pull-environments.sh`](https://github.com/aws-amplify/amplify-ui/blob/build-system-test/main/environments/pull-environments.sh)).

I believe this is more maintainable and easier to use locally:
- the flags are more explicit
  - this lets script to have more control and lets the developer know what each input does.
- less workarounds 
  - this removes the need to override `DEPENDENCIES` since the bash will do all the work.
- easier for internal developers to use this script on its own
  - with previous setup, we needed to replicate the GitHub Actions workflow steps, through three different files and env vars. 
  - This approach makes the script more standalone (and we can create aliases like we did [here](https://github.com/aws-amplify/amplify-ui/blob/build-system-test/main/environments/auth/package.json#L6-L7))

I realize this is a big change request, happy to talk offline 🙏

0618: Great minds. @reesscot made a similar suggestions a few days ago that we can use the script to auto-create many mega-apps on our local machines. I've cut a ticket for that.

Your suggestion is very inspirational. Thanks for the implementation details!

Since this PR is to fix the current Angular mega-apps, I'd like to keep it small and specific. I'd also like to see all the mega-apps pass the test, which can be used as a baseline.

I'll have a separate PR to refactor the workflow 😁

wlee221 (approved):
LGTM!

.github/scripts/mega-app-install-build.sh
wlee221: Makes sense, thanks!

ioanabrooks (approved):

PR #4095, opened 2023-06-13
chore(liveness): add liveness version user agent to rekognition streaming api call

thaddmt:
#### Description of changes
- Updated custom user agent to include liveness package version
- New User agent - `aws-sdk-js/3.348.0 ua/2.0 os/macOS#10.15.7 lang/js md/browser#Chrome_114.0.0.0 api/rekognitionstreaming#3.348.0 aws-amplify/5.1.7-js-ui-react-liveness-1.0.2`
- Old user agent - `aws-sdk-js/3.309.0 os/Android/13 lang/js md/browser/Chrome_113.0.5672.132 api/rekognitionstreaming/3.309.0 aws-amplify/5.2.1_js`
- Seems like aws sdk converts spaces to dashes in custom user agent

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
@@ -1,7 +1,4 @@
-import {
-  Credentials as AmplifyCredentials,
-  getAmplifyUserAgent,
-} from '@aws-amplify/core';
+import { Credentials as AmplifyCredentials } from '@aws-amplify/core';
 import { AmazonAIInterpretPredictionsProvider } from '@aws-amplify/predictions';
 import {
   ClientSessionInformationEvent,
@@ -118,7 +115,7 @@ export class LivenessStreamProvider extends AmazonAIInterpretPredictionsProvider
     const clientconfig: RekognitionStreamingClientConfig = {
       credentials,
       region: this.region,
-      customUserAgent: `${getAmplifyUserAgent()} ${getLivenessUserAgent()}`,
+      customUserAgent: `${getLivenessUserAgent()}`,
       requestHandler: new WebSocketFetchHandler({ connectionTimeout: 10_000 }),
     };
 
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
@@ -12,6 +12,7 @@ import {
 } from '@aws-sdk/client-rekognitionstreaming';
 import { WebSocketFetchHandler } from '@aws-sdk/middleware-websocket';
 import { VideoRecorder } from './videoRecorder';
+import { getLivenessUserAgent } from '../../utils/platform';
 
 export interface StartLivenessStreamInput {
   sessionId: string;
@@ -117,7 +118,7 @@ export class LivenessStreamProvider extends AmazonAIInterpretPredictionsProvider
     const clientconfig: RekognitionStreamingClientConfig = {
       credentials,
       region: this.region,
-      customUserAgent: getAmplifyUserAgent(),
+      customUserAgent: `${getAmplifyUserAgent()} ${getLivenessUserAgent()}`,
       requestHandler: new WebSocketFetchHandler({ connectionTimeout: 10_000 }),
     };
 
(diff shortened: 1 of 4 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/utils/platform.ts
```diff
@@ -0,0 +1,7 @@
+import { VERSION } from '../../../version';
+
+const BASE_USER_AGENT = `ui-react-liveness ${VERSION}`;
```
thaddmt, line 3: Decided not to include `aws-amplify` or extra `/` because this is already appended after the amplify js version

Did not include `@` symbols because they get converted to spaces

packages/react-liveness/src/components/FaceLivenessDetector/utils/__tests__/platform.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    expect(customUserAgent).toContain(VERSION);
+  });
+
+  it('should return a user agent string with custom content', () => {
+    const customUserAgent = getLivenessUserAgent('specialEnding');
+
+    expect(customUserAgent).toContain('ui-react-liveness');
+    expect(customUserAgent).toContain(VERSION);
+    expect(customUserAgent).toContain('specialEnding');
+  });
```
wlee221, line 18: Do we plan to use custom content, and if not, should we not worry about it for now?

packages/react-liveness/src/components/FaceLivenessDetector/utils/platform.ts
wlee221: Can we do `@ui-react-liveness/${VERSION}` like we did for Auth? Assuming this works, this will sort nicely on our metrics.

https://github.com/aws-amplify/amplify-ui/blob/c85b2eeac7b42b3cd5fbd765f0f27e4a0bec1b8b/packages/ui/src/helpers/authenticator/utils.ts#L21

packages/react-liveness/src/components/FaceLivenessDetector/utils/__tests__/platform.test.ts
thaddmt: I'm open to removing it, I mostly kept it in to follow the same pattern as the JS team but can't think of a good reason for it either

wlee221: Yep, let's remove it then 😈

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
```diff
@@ -117,7 +118,7 @@ export class LivenessStreamProvider extends AmazonAIInterpretPredictionsProvider
     const clientconfig: RekognitionStreamingClientConfig = {
       credentials,
       region: this.region,
-      customUserAgent: getAmplifyUserAgent(),
+      customUserAgent: `${getAmplifyUserAgent()} ${getLivenessUserAgent()}`,
```
calebpollman, line 121: Could be way off base, but should we exclude the `customUserAgent` from the `clientConfig` if the credentials are non-cognito/custom? Just trying to confirm whether we need to consider anything related to custom credential providers here

thaddmt: Don't think you're off base, I was actually considering whether we need the Amplify user agent at all? We only use the aws-amplify library for getting credentials in this case which is probably not very useful as debug information.

I can see multiple options here:
1. We only use the liveness ui package user agent string
2. We only include the Amplify JS user agent string if we use the default amplify credentials, if they pass the credential override we don't include it

What do you think?

packages/react-liveness/src/components/FaceLivenessDetector/utils/platform.ts
thaddmt: yep we can! my main call out is just that @ will become - when appended to the query string

wlee221: Yep, let's just use `/` then 🙏  `ui-react-liveness/1.0.2`

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
thaddmt: Ended up including both for now

calebpollman (approved):

reesscot (approved):

wlee221 (approved):

zchenwei (approved):

Issue #4090, opened 2023-06-13
Bug(Docs): Value input for Ratings demo breaks if you hit backspace


hbuchel:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

Other

### Which UI component?

Other

### How is your app built?

n/a

### Please describe your bug.

On the [Ratings doc page](https://ui.docs.amplify.aws/react/components/rating) in the demo, if you hit backspace/delete on the value input, the value turns to `NaN` which breaks the demo and you can no longer type in the input; you need to reload the page to reset it. Hitting backspace on the maxValue input causing the page to crash.

<img width="400" alt="Screenshot of Ratings doc page which shows the Demo section, the value input reads NaN and the component preview does not show up" src="https://github.com/aws-amplify/amplify-ui/assets/376920/57a4d281-6d50-4da6-bfe1-0689f692dfcf">

### What's the expected behaviour?

You should be able to clear out the value input and enter a new value.

### Help us reproduce the bug!

Tab or click into the value input and hit backspace or try to delete the number.

### Code Snippet

```typescript
// Put your code below this line.

```

PR #4097, opened 2023-06-13
fix(docs): ratings demo values inputs

hbuchel:
#### Description of changes

Updates the Ratings demo inputs for value and maxValue to use number type input and removes the parseInt on the onChange. 

Unrelated, there is a small rendering bug in the Ratings element itself which causes a height discrepancy when you use a value less than 1 that I will open a issue for (verified here https://codesandbox.io/p/sandbox/ratings-height-bug-68rgd5?file=%2Fsrc%2FApp.tsx%3A8%2C40 and you can see it in the demo if you change the value to .5 there is a small shift)

**Before**
Hit backspace in the value and maxValue inputs to see current behavior on [Ratings demo](https://ui.docs.amplify.aws/react/components/rating)

**After**

https://github.com/aws-amplify/amplify-ui/assets/376920/84473d6f-10a2-40d2-b1ac-be348faa7684

Fixes https://github.com/aws-amplify/amplify-ui/issues/4090

#### Checklist

- [ ] PR description included

### DIFF
--- a/docs/src/pages/[platform]/components/rating/RatingPropControls.tsx
+++ b/docs/src/pages/[platform]/components/rating/RatingPropControls.tsx
@@ -55,7 +55,9 @@ export const RatingPropControls: RatingPropControlsInterface = ({
         value={value}
         placeholder="Set Value"
         type="number"
-        onChange={(event) => handleValueChange(event.target.value)}
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+          handleValueChange(event.target.value)
+        }
       />
 
       <TextField
@@ -63,7 +65,9 @@ export const RatingPropControls: RatingPropControlsInterface = ({
         value={maxValue}
         placeholder="Set Max Value"
         type="number"
-        onChange={(event) => handleMaxValueChange(event.target.value)}
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+          handleMaxValueChange(event.target.value)
+        }
       />
 
       <SelectField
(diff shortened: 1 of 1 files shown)


docs/src/pages/[platform]/components/rating/RatingPropControls.tsx
```diff
@@ -37,15 +37,17 @@ export const RatingPropControls: RatingPropControlsInterface = ({
       <TextField
         label="value"
         value={value}
+        type="number"
         placeholder="Set Value"
-        onChange={(event) => setValue(parseInt(event.target.value))}
+        onChange={(event) => setValue(event.target.value)}
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/rating/RatingPropControls.tsx:42]: this doesn't cause a type error?

Comment (hbuchel) [reply docs/src/pages/[platform]/components/rating/RatingPropControls.tsx:42]: That's what I though, though for some reason it doesn't throw an error. 
Screenshot 2023-06-13 at 4 10 38 PM
I don't think we'd want parseInt though either because it restricts using decimal values like 2.5

Comment (hbuchel) [reply docs/src/pages/[platform]/components/rating/RatingPropControls.tsx]: @calebpollman @reesscot updated (and updated preview video in description)

reesscot (dismissed):

docs/src/pages/[platform]/components/rating/RatingPropControls.tsx
```diff
@@ -32,20 +32,38 @@ export const RatingPropControls: RatingPropControlsInterface = ({
   setFillColor,
   setEmptyColor,
 }) => {
+  function handleValueChange(value: string) {
+    if (value === '') {
+      setValue(0);
+    } else if (parseFloat(value)) {
+      setValue(parseFloat(value));
+    }
```
Comment (calebpollman) [inline docs/src/pages/[platform]/components/rating/RatingPropControls.tsx:40]: Do we need an `else` here? Same question for ln 48

docs/src/pages/[platform]/components/rating/RatingPropControls.tsx
Comment (hbuchel) [reply docs/src/pages/[platform]/components/rating/RatingPropControls.tsx]: > Looking at the screenshot above, it doesn't look `event` is typed at all, which may be why there was no type error (implicit `any`)

Oh, you're right, thanks. Updating the type for event would have shown the type error then. I added types for event for the controls here that had implicit any type.

docs/src/pages/[platform]/components/rating/RatingPropControls.tsx
Comment (hbuchel) [reply docs/src/pages/[platform]/components/rating/RatingPropControls.tsx:40]: I think for these demo controls it's fine without; we don't want to update the value if you start typing letters.

calebpollman (approved):
🚢

reesscot (approved):

Issue #3458, opened 2023-02-22
[FR] Authenticator required fields handling


ioanabrooks:
### On which framework/platform would you like to see this feature implemented?

React, React Native, Angular, Vue

### Which UI component is this feature-request for?

Authenticator

### Please describe your feature-request in detail.

Improve developer and end-user experience with Authenticator required fields across web and React Native platforms.

For React Native add visual indicators of required fields and don't allow forms to submit with empty values, offloading checks to the sdk, and sending unnecessary calls over the network.

For web allow customers to override the HTML5 behavior for required input fields.

### We love contributors! Is this something you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

calebpollman:
Related: https://github.com/aws-amplify/amplify-ui/issues/3452

reesscot:
We should also update the required field handling in Angular and Vue as well
https://github.com/aws-amplify/amplify-ui/issues/3928

PR #4107, opened 2023-06-14
chore(ui-react-native): Add RN Authenticator fields validations

ioanabrooks:
#### Description of changes

This change adds validations for email and required fields for RN Authenticator only.
 
* Added unit test for email validation util and required fields validation.
* Added e2e test for required field and validation check.
* Added coverage report to `react-native` package.

```
---------------------------------------------------|---------|----------|---------|---------|-------------------
File                                               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
---------------------------------------------------|---------|----------|---------|---------|-------------------
All files                                          |    99.3 |    96.09 |   98.98 |   99.29 |      
```

#### Issue #, if available
https://github.com/aws-amplify/amplify-ui/issues/3458

Ran tests locally, e2e run in [CI](https://github.com/aws-amplify/amplify-ui/actions/runs/5295406091)

https://github.com/aws-amplify/amplify-ui/assets/68251134/9ad9784c-23d1-4581-9bb4-c3511da4a368

#### Checklist

- [x] PR description included


packages/react-core/src/Authenticator/hooks/types.ts
```diff
@@ -75,14 +75,14 @@ export type CommonRouteProps = {
...
   handleSubmit: UseAuthenticator['submitForm'];
+  validationErrors?: UseAuthenticator['validationErrors'];
 };
 
 /**
  * Base Route component props
  */
 export type ConfirmResetPasswordBaseProps<FieldType = {}> = {
   resendCode: UseAuthenticator['resendCode'];
-  validationErrors?: UseAuthenticator['validationErrors'];
```
ioanabrooks, line 85: Added to common props as opposed to each individual Base Route component props (same comment for below components).

packages/react-native/src/Authenticator/hooks/useFieldValues/types.ts
```diff
@@ -26,10 +28,13 @@ export interface UseFieldValuesParams<FieldType extends TypedField> {
    * machine "SUBMIT"" event handler, validates `field` value against machine validation rules
    */
   handleSubmit: MachineEventHandlers['handleSubmit'];
+
+  validationErrors?: AuthenticatorMachineContext['validationErrors'];
 }
 
 export interface UseFieldValues<FieldType extends TypedField> {
   fields: FieldType[]; // return either radio or text
+  formValidationErrors?: ValidationError;
```
calebpollman, line 37: What do you think if naming this `fieldValidationErrors`? Thinking it better aligns with the hook name, and also since these are validation errors at the field level

packages/ui/src/helpers/authenticator/__tests__/utils.test.ts
```diff
@@ -74,3 +79,16 @@ describe('configureComponent', () => {
     );
   });
 });
+
+describe('isValidEmail', () => {
```
calebpollman, line 83: Since `isValidEmail` accepts `undefined` as its parameter can we add a test case for it?

packages/react-core/src/Authenticator/hooks/types.ts
```diff
@@ -75,6 +75,7 @@ export type CommonRouteProps = {
   handleBlur: UseAuthenticator['updateBlur'];
   handleChange: UseAuthenticator['updateForm'];
   handleSubmit: UseAuthenticator['submitForm'];
+  validationErrors?: UseAuthenticator['validationErrors'];
```
calebpollman, line 78: Not sure if this the correct path forward. Why not use the `ValidationProps` interface?

calebpollman (commented):
Looks good from a technical perspective, left some feedback around semantics/types. Thanks for doing this @ioanabrooks!

packages/react-core/src/Authenticator/hooks/types.ts
ioanabrooks: Wasn't available when I was working on it :)  Will adjust.

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
```diff
@@ -37,6 +52,22 @@ export default function useFieldValues<FieldType extends TypedField>({
     return getSanitizedTextFields(fields, componentName);
   }, [componentName, fields, isRadioFieldComponent]);
 
+  const runValidation = useCallback(
```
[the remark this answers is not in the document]
ioanabrooks: While it is true that `runValidation` isn't passed as a prop and only used within the scope of this function (`useFieldValues`), since we are calling it on blur events, I thought it was worth memoizing.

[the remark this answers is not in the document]
calebpollman: What is the benefit if it is not preventing rerenders?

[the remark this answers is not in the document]
ioanabrooks: My thought is that by memoizing `runValidation`, we ensure the same function instance will be reused , since it's dependency array is empty. Do you see any downsides?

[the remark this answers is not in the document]
calebpollman: More that it's unclear to me what benefits memoizing provides versus declaring `runValidation` outside the scope of `useFieldValues`? For example why not something like this:
```tsx
const runValidations = (field: TextFieldOptionsType, value: string | undefined): ValidationError => {
  const fieldErrors = [];
  if (field.required && !value) {
    fieldErrors.push(getRequiredFieldText());
  }
  if (field.type == 'email') {
    fieldErrors.push(isValidEmail(value) ? '' : getInvalidEmailText());
  }
  return { [field.name]: fieldErrors };
}

function useFieldValues(params) {
  // ...field stuff
  
  const onBlur: TextFieldOnBlur = (event) => {
    // call `onBlur` passed as text `field` option
    field.onBlur?.(event);
    
    // call machine blur handler
    handleBlur({ name, value: values[name] });
    
    setfieldValidationErrors(runValidation(field, values[name]));
  };
}
```

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
```diff
@@ -12,14 +22,19 @@ import {
...
 export default function useFieldValues<FieldType extends TypedField>({
   componentName,
   fields = [],
   handleBlur,
   handleChange,
   handleSubmit,
+  validationErrors,
 }: UseFieldValuesParams<FieldType>): UseFieldValues<FieldType> {
   const [values, setValues] = useState<Record<string, string>>({});
+  const [fieldValidationErrors, setfieldValidationErrors] =
```
calebpollman, line 36: Typo: `setfieldValidationErrors` -> `setFieldValidationErrors`

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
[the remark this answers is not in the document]
ioanabrooks: I've played with this and it's rerendering the validation error on every blur when the error is the same. Will tinker a bit, think we need to only rerender when there are new errors [i.e. the validation errors have changed].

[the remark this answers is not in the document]
ioanabrooks: Refactored, moved out of the scope of the hook.

hbuchel:
Question about setting context for assistive tech: How are we associating the field error messages with the input? Do we need to pass it as an accessibilityHint on the [TextInput](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react-native/src/primitives/TextField/TextField.tsx#L58)? Not sure if that's the correct approach; I think that's what they're doing here https://formidable.com/open-source/react-native-ama/components/TextInput/#haserror I also this [issue](https://github.com/facebook/react-native/issues/30848) is still marked open on the RN repo :(

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
```diff
@@ -112,5 +132,10 @@ export default function useFieldValues<FieldType extends TypedField>({
     handleSubmit?.(submitValue);
   };
 
-  return { fields: fieldsWithHandlers, disableFormSubmit, handleFormSubmit };
+  return {
+    fields: fieldsWithHandlers,
+    disableFormSubmit,
+    fieldValidationErrors: { ...fieldValidationErrors, ...validationErrors },
```
calebpollman, line 138: Definitely may be missing something here, but curious if `validationErrors` still needs to be merged with `fieldValidationErrors` since (if I am reading correctly} `runValidations` has already concatenated the values contained in `validationErrors` [here](https://github.com/aws-amplify/amplify-ui/pull/4107/files#diff-52e68fc0024cea567f183a7a72a8813c90a961a5ee667f03abc479015cf676baR211)?

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -176,3 +184,38 @@ export function getRouteTypedFields({
...
+    }
+  }
+
+  // add state machine validation errors, if any
+  if (stateValidations && stateValidations[field.name]) {
+    const stateValidation = stateValidations[field.name];
+    if (isString(stateValidation)) {
+      fieldErrors.push(stateValidation);
+    } else {
+      fieldErrors = fieldErrors.concat(stateValidation);
```
calebpollman, line 216: Rather than assigning `fieldErrors` as a mutable variable, prefer returning `fieldErrors.concat(stateValidation)` here

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -176,3 +184,38 @@ export function getRouteTypedFields({
...
+  }
+  if (field.type === 'email') {
+    if (!isValidEmail(value)) {
+      fieldErrors.push(getInvalidEmailText());
+    }
+  }
+
+  // add state machine validation errors, if any
+  if (stateValidations && stateValidations[field.name]) {
+    const stateValidation = stateValidations[field.name];
```
calebpollman, line 212: Think this would be easier to read if `stateValidation` was assigned as a variable outside the `if` statement:
```
const stateFieldValidation = stateFieldValidations?.[field.name];
if (stateFieldValidation) {
  // ...do stuff
}
```

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
```diff
@@ -176,3 +184,38 @@ export function getRouteTypedFields({
 
   return isVerifyUserRoute ? radioFields : getTypedFields(fields);
 }
+
+/**
+ *
+ * @param field
+ * @param value
+ * @param stateValidations
+ * @returns
```
calebpollman, line 193: Who knows? 😅

hbuchel (dismissed):

packages/ui/src/helpers/authenticator/__tests__/utils.test.ts
```diff
@@ -74,3 +79,24 @@ describe('configureComponent', () => {
...
+  it('should return true for a valid email address', () => {
+    expect(isValidEmail('test@example.com')).toBe(true);
+    expect(isValidEmail('TEST@EXAMPLE.COM')).toBe(true);
+  });
+
+  it('should return false for an invalid email address', () => {
```
calebpollman, line 89: Nittish: Rather than running multiple tests that assert the condition within a single block the `jest` idiomatic `it.each` can be used for descriptive failures as each test runs independently, as well as improved readability of log output:
```
it.each([
  { desc: 'no domain name', email: 'test@' },
  { desc: 'an invalid domain', email: 'test@.' },
  // ...other test scenarios
])('should return false for an email address with %s', ({ email }) => {
  expect(email).toBe(false);
});

File: packages/ui/src/helpers/authenticator/utils.ts [hunk 12]
```diff
@@ -180,3 +180,12 @@ export function trimValues<T extends Record<string, string>>(
     {} as T
   );
 }
+
+export const isValidEmail = (value: string | undefined) => {
+  if (!value) return false;
+
+  // source: HTML5 spec https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+  const emailRegex =
```
calebpollman, line 188: Should this be a constant?

calebpollman (commented):
Looking good, excited to see this completed. Let me know your thoughts on the feedback

packages/react-native/jest.config.js
```diff
@@ -1,10 +1,13 @@
 module.exports = {
   preset: 'react-native',
   modulePathIgnorePatterns: ['<rootDir>/dist/'],
+  collectCoverage: true,
   collectCoverageFrom: [
     '<rootDir>/src/**/*.{js,jsx,ts,tsx}',
     '!<rootDir>/src/**/*{c,C}onstants.ts',
   ],
+  // ignore coverage for top level "export" and style files
+  coveragePathIgnorePatterns: ['<rootDir>/src/(index|styles).(ts|tsx)'],
```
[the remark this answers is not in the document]
ioanabrooks: The only nice thing about this is that it doesn't even show up on the coverage report. I.e. with this change:
```
All files                                          |   99.19 |    96.15 |   98.98 |   99.19 |                   
 src                                               |     100 |      100 |     100 |     100 |                   
  version.ts                                       |     100 |      100 |     100 |     100 |    
```
Without
```
All files                                          |   99.19 |    96.15 |   98.98 |   99.19 |                   
 src                                               |     100 |      100 |     100 |     100 |                   
  index.ts                                         |       0 |        0 |       0 |       0 |                   
  version.ts                                       |     100 |      100 |     100 |     100 |      
```
don't feel strongly tho.

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
ioanabrooks: This covers the case where there is no blur or change event but there are state machine errors. We don't want to lose them.

packages/react-native/src/Authenticator/hooks/useFieldValues/utils.ts
ioanabrooks: 😅

packages/ui/src/helpers/authenticator/__tests__/utils.test.ts
ioanabrooks: I think the test is simple enough that it doesn't need "descriptive failures", it reads easier for me than `it.each` but ultimately don't feel super strong about it, could have a follow up to refactor other tests where `it.each` could be used. One observation I have is that currently this test runs in 1ms with all assertions, having a separate test for each assertion means each run in 1ms (I tested this). While it is negligible, I think we should pay attention to run times as well, as it can add up if this (`it.each`) is the pattern we want to adopt going forward.

packages/ui/src/helpers/authenticator/utils.ts
ioanabrooks: Yes, and it is a constant.

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
```diff
@@ -53,11 +59,18 @@ export default function useFieldValues<FieldType extends TypedField>({
...
       // call `onBlur` passed as text `field` option
       field.onBlur?.(event);
 
       // call machine blur handler
       handleBlur({ name, value: values[name] });
+
+      setFieldValidationErrors({
+        ...fieldValidationErrors,
+        [name]: runFieldValidation(field, values[name], validationErrors),
+      });
```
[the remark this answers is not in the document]
ioanabrooks: Will address as follow-up since touches other parts not related to validation.

packages/ui/src/helpers/authenticator/utils.ts
calebpollman: I mean a `CONSTANT` 😅 , since there is no need to declare this for each call

packages/ui/src/helpers/authenticator/__tests__/utils.test.ts
calebpollman: We already use `it.each` throughout the codebase. Adding 1 ms per test is fine if we are able to locate failing test case without having to comment out each test individually to find the failure (hypothetical situation). That being said this was a nit, so feel free to ignore it

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
calebpollman: Maybe we should move the logic related to the machine errors out of `runValidations` and create a util to be used here so we can have one source of truth in how state machine validations are merged with the local state validations?

ioanabrooks: I was hesitant to make extra changes to this bc we're removing it with the upcoming auth work, what do you think, is it a blocker?

packages/ui/src/helpers/authenticator/utils.ts
ioanabrooks: oh gotcha now, makes sense!

packages/ui/src/helpers/authenticator/__tests__/utils.test.ts
ioanabrooks: I'll add an asana task!

packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
calebpollman: My concern is mostly just that if for some reason the merging related code needs to get updated, it might be difficult to understand what's happening here without prior context. That being said if anything does come up it'll most likely be addressed by you or i so okay with it as is

calebpollman (approved):
Not sure what's going on with CI here but approved 🚢

reesscot (approved):

PR #4122, opened 2023-06-16
chore(liveness): improve face match percentage performance

thaddmt:
#### Description of changes
- Previously in a single 7 second session the face match percentage bar would re-render up to 800 times
- Reduced re-renders by rounding face match percentage values up and memoizing the component
- Now renders are down ~80 renders in a 7 second session
- To further reduce renders we can perform fewer face detection calls to Blazeface, but having a higher number of calls helps increase accuracy of sessions as we get more granularity into face match position

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -66,7 +66,7 @@ const centeredLoader = (
 /**
  * For now we want to memoize the HOC for MatchIndicator because to optimize renders
  * The LivenessCameraModule still needs to be optimized for re-renders and at that time
- * we shoudl be able to remove this memoization
+ * we should be able to remove this memoization
  */
 const MemoizedMatchIndicator = React.memo(MatchIndicator);
 
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -22,8 +22,8 @@ import {
   Hint,
   RecordingIcon,
   Overlay,
-  MatchIndicator,
   selectErrorState,
+  MemoizedMatchIndicator,
 } from '../shared';
 import { LivenessClassNames } from '../types/classNames';
 import {
@@ -251,7 +251,9 @@ export const LivenessCameraModule = (
             {isRecording &&
             !isFlashingFreshness &&
             showMatchIndicatorStates.includes(faceMatchState!) ? (
-              <MatchIndicator percentage={faceMatchPercentage!} />
+              <MemoizedMatchIndicator
+                percentage={Math.ceil(faceMatchPercentage!)}
+              />
             ) : null}
 
             {isNotRecording && (
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -246,7 +246,9 @@ export const LivenessCameraModule = (
             {/* 
               We only want to show the MatchIndicator when we're recording
               and when the face is in either the too far state, or the 
-              initial face identified state
+              initial face identified state. Using the MemoizedMatchIndicator here
+              so that even when this component re-renders the indicator is only
+              re-rendered if the percentage prop changes.
             */}
             {isRecording &&
             !isFlashingFreshness &&
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -63,6 +63,13 @@ const centeredLoader = (
   />
 );
 
+/**
+ * For now we want to memoize the HOC for MatchIndicator because to optimize renders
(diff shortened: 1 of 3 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/shared/MatchIndicator.tsx
```diff
@@ -38,3 +38,5 @@ export const MatchIndicator: React.FC<MatchIndicatorProps> = ({
     </div>
   );
 };
+
+export const MemoizedMatchIndicator = React.memo(MatchIndicator);
```
ayush987goyal, line 42: Nice! Should we do the same for the feedback component?

zchenwei (approved):
LGTM

packages/react-liveness/src/components/FaceLivenessDetector/shared/MatchIndicator.tsx
calebpollman, line 42: Could be completely off base here, but it's unclear to me how wrapping `MatchIndicator` with `React.memo` is improving performance since we are only passing `percentage` to it from `LivenessCameraModule`. Thinking the usage of `Math.ceil` [here](https://github.com/aws-amplify/amplify-ui/pull/4122/files#diff-5d58d3114c36274fae912867705fd03fd3a44874ea478e657d41ce9224c520dbR255) might be enough since it reduces the amount of re-renders by only passing whole numbers to `MatchIndicator`?

thaddmt: This was just focused on the face match percentage for now, will address that separately

thaddmt: I was thinking the same thing but despite that, whenever the parent component re-rendered it would also cause this component to re-render. Adding the `React.memo` to it made it so that this child component would not re-render even though the parent had. We have a separate task for tracking performance of the parent component in this case

calebpollman: Makes sense. Maybe can add a comment here with a concise explanation for future maintainers?

calebpollman (dismissed):

hbuchel (dismissed):

packages/react-liveness/src/components/FaceLivenessDetector/shared/MatchIndicator.tsx
wlee221: Just curious, why not directly use `React.useMemo` inside `LivenessCameraModule.tsx`?

thaddmt: no particular reason, just decided to do it this way since the change was based around this package. Do we prefer doing it that way instead?

calebpollman: FWIW think it makes sense to memo where it is needed, ideally the memory used for the memoization frees up once `LiveCameraModule` unmounts, and also removes the mental overhead of whether to use `MemoizedMatchIndicator` or `MatchIndicator` for future devs (know it's unlikely `MatchIndicator` will ever be reused but just saying)

thaddmt: Sure, i can update it to just be useMemo in place!

wlee221: Think in general we prefer `React.useMemo`:
- we usually use `React.useMemo`: [search](https://github.com/search?q=repo%3Aaws-amplify%2Famplify-ui+React.memo&type=code)
- By using `useMemo`, the parent component has more control over how it's memoized
- Other components that uses the same component aren't affected

That said `React.memo` works perfectly fine in this case because we are just memoizing the whole component, but I think it'll be nicer if we stick with `useMemo` for the reasons above.

thaddmt: Ended up switching to use `React.useMemo` but realized this slightly increased the number of renders because the memoized component will forcibly re-render when the LivenessCameraModule component is re-rendered (something else we plan to optimize separately).

Because `react.memo` was used at the HOC level it will reduce re-renders regardless if it's parent component re-renders.

Anyways what do you guys think?

calebpollman: Started wondering abt that actually since the HOC is applied outside the scope of `LivenessCameraModule`. IMO wrapping with `React.memo` is fine, would prefer it happens in _LivenessCameraModule.tsx_ with an explanatory comment. Ideally we should be able to isolate the rendering issues in `LivenessCameraModule` by breaking it in to smaller components and remove the `memo` in the future

thaddmt: Added the react.memo and a comment explainer

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -63,6 +63,13 @@ const centeredLoader = (
   />
 );
 
+/**
+ * For now we want to memoize the HOC for MatchIndicator because to optimize renders
+ * The LivenessCameraModule still needs to be optimized for re-renders and at that time
+ * we shoudl be able to remove this memoization
```
calebpollman, line 69:
```suggestion
 * we should be able to remove this memoization
 ```

calebpollman (dismissed):

calebpollman (approved):

wlee221 (approved):

Issue #4123, opened 2023-06-16
Bug (FaceLivenessDetector): When closing the stream connection an error about websocket closed or is closing always appears


thaddmt:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Next.js

### What browsers are you seeing the problem on?

Chrome

### Please describe your bug.

When the websocket connection for a liveness session ends an error message usually pops up saying `WebSocket is already in CLOSING or CLOSED state.` which is confusing to customers looking at the logs as they believe something has gone wrong.

<img width="1018" alt="Screenshot 2023-06-16 at 4 15 32 PM" src="https://github.com/aws-amplify/amplify-ui/assets/68032955/01baf8b5-773b-4b82-a575-39faec26b91a">

### What's the expected behaviour?

We should not show this error unless something has gone wrong.

### Help us reproduce the bug!

Attempt a liveness session

### Code Snippet

```typescript
// Put your code below this line.
<FaceLivenessDetectorCore
              sessionId={createLivenessSessionApiData.sessionId}
              region={'us-east-1'}
              onUserCancel={onUserCancel}
              onAnalysisComplete={getLivenessDetectionHandler}
              disableInstructionScreen={disableStartScreen}
              onError={onError}
              config={{ credentialProvider }}
/>
```

h4j4x:
Any solution?
I have the same error, "onError" callback is giving: SERVER_ERROR: Request has invalid parameters.

thaddmt:
@h4j4x I don't believe that's related to this issue here since this issue is specifically about the websocket message showing up after a successful session.

Can you open up a separate ticket with your issue or ask in discord so we can get more information? Off the top of my head I would make sure that the `region` value and the `sessionId` value you pass to the component are all created in the same region and account.

thaddmt:
https://github.com/aws-amplify/amplify-ui/pull/4150

PR #4138, opened 2023-06-20
fix(liveness): remove websocket error showing up after connection closes

thaddmt:
#### Description of changes
- Fixes issue - https://github.com/aws-amplify/amplify-ui/issues/4123
- Previously a websocket connection is already closed error would show up after successful liveness checks
- This was happening because when closing the input stream the aws sdk client would try to send another websocket event, but if we have already sent the `stopVideo` event then the server will have already closed the connection. Hence we no longer need to close the input stream ourselves since the server will have already closed the connection

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
@@ -1176,10 +1176,7 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
         await livenessStreamProvider!.dispatchStopVideoEvent();
       },
       async getLiveness(context) {
-        const { livenessStreamProvider } = context;
         const { onAnalysisComplete } = context.componentProps!;
-
-        livenessStreamProvider!.endStream();
 
         // Get liveness result
         await onAnalysisComplete();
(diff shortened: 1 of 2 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
```diff
@@ -1176,11 +1176,8 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
         await livenessStreamProvider!.dispatchStopVideoEvent();
       },
       async getLiveness(context) {
-        const { livenessStreamProvider } = context;
         const { onAnalysisComplete } = context.componentProps!;
 
-        livenessStreamProvider!.endStream();
```
calebpollman, line 1182: Just for my own understanding, was this in place in to handle a scenario where the stream was still active from a previous attempt?

thaddmt: This was just in place so that we could ensure that the websocket connection was closed and that our input stream was closed.

But considering the websocket connection gets closed by the server we don't need to call this as at this point we have already stopped recording video.

calebpollman (approved):

wlee221 (approved):

PR #4189, opened 2023-06-30
ci: use relative path to reference workflows

wlee221:
#### Description of changes

This PR changes **internal** workflows to reference workflows using relative paths instead of absolute paths. **External** workflows that are used to test external contributions are untouched.

#### Problem

Currently we run all reusable workflows against main, not the actual branch the PR is on. This makes it very difficult to test workflow changes before we merge the PR to main, and need to rely on ephemeral "POC" workflows. This slows down development, especially in high sev situations.

#### Solution

We will use relative path instead of absolute path to reference internal workflows. This will resolve the DX issues, because now internal developers can test workflow changes before merging into `main`. This will also much improve the feature branch development as well.

#### Security Considerations

This is only for **internal** PRs, and external PRs that run on forks are untouched. External contributors cannot create / modify / push to internal branches anyway, so this change is safe. And even if external contributors were to somehow run these workflows, GitHub does not give forks any write permissions or access to secrets anyway for `push` and `pull_request` events.

#### Checklist

- [x] PR description included


zchenwei (approved):
LGTM

ioanabrooks (approved):

PR #4207, opened 2023-07-07
fix(e2e): Update mocks for storage tests

ioanabrooks:
#### Description of changes

This change updates the mocks for storage to conform to S3 specification (https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html#API_UploadPart_ResponseElements) - i.e. mock etag header.

The next version of amplify js introduces XML payload validation, which means once it is released, tests will fail.
The issue was surfaced while running tests against custom js client in https://github.com/aws-amplify/amplify-ui/pull/4182/files

Tested change against above PR to be sure, tests pass: https://github.com/aws-amplify/amplify-ui/actions/runs/5482812674/jobs/9988700067

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/e2e/cypress/integration/common/shared.ts
+++ b/packages/e2e/cypress/integration/common/shared.ts
@@ -90,12 +90,11 @@ Given(
       throw error;
     }
 
-    cy.intercept(routeMatcher, (req) => {
-      req.reply((res) => {
-        res.headers[headerName] = headerValue; // Add header to the response
-        res.send({ fixture }); // Respond with fixture data
-      });
-    }).as('route');
+    cy.intercept(routeMatcher, {
+      headers: {
+        [headerName]: headerValue,
+      },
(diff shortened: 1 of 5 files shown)


thaddmt (approved):

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -79,6 +79,25 @@ Given(
   }
 );
 
+Given(
+  'I intercept {string} with fixture {string} and add header {string} with value {string}',
```
calebpollman, line 83: Unless I'm reading wrong think we can probably hard code the second and third positional parameters here since all call sites ate passing the same values. IMO unlikely that the function will be reused in the future, and can add the parameters if/when needed

dbanksdesign (approved):

PR #4227, opened 2023-07-11
fix(react-native): border widths in the theme don't throw runtime errors

dbanksdesign:
#### Description of changes

* Fixed a typing issue where the theme would only accept strings as borderWidths
* Made sure borderWidths and space values turn into numbers even without 'rem' or 'px'

#### Checklist

- [x] PR description included

### DIFF
--- a/.changeset/cold-grapes-collect.md
+++ b/.changeset/cold-grapes-collect.md
@@ -1,0 +1,32 @@
+---
+"@aws-amplify/ui-react-native": patch
+"@aws-amplify/ui-react": patch
+"@aws-amplify/ui": patch
+---
+
+fix(react-native): border widths, spacing, font sizes, opacities in the theme don't throw runtime errors.
+
+These are all valid in a theme now:
+
+```typescript
+const theme: Theme = {
+  tokens: {
+    borderWidths: {
+      small: '4',
+      medium: '1rem',
+      large: 6,
+    },
+    opacities: {
+      '10': '0.2',
+    },
+    space: {
+      small: 4,
+      medium: '6',
+      large: '{space.small.value}',
+    },
+    fontSizes: {
+      small: '1rem',
+    },
+  },
+}
+```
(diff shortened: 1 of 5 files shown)


packages/react-native/src/theme/createTheme.ts
```diff
@@ -51,6 +51,9 @@ const setupToken = ({
     if (value.includes('px')) {
       return parseInt(value, 10);
     }
+    if (path[0] === 'borderWidths' || path[0] === 'space') {
+      return parseFloat(value);
+    }
```
calebpollman, line 56: To make the logic in `setupToken` a bit easier to reason about in the future, what do you think abt combining the conditional block here the block immediately below and the block on line 43 and abstracting a util for checking whether the value should be passed to `parseFloat` ? e.g. 
```
// contrived name
const shouldParseFloatValue = (pathKey: string) => pathKey === 'space' || pathKey === 'borderWidths' || pathKey === 'opacities';

const setupToken = (
  // ...setupToken params
) {
  // ...other setupToken logic

  if (shouldParseFloat(path[0])) {
    if (value.includes('rem')) {
      return Math.floor(parseFloat(value) * 16 * spaceModifier);
    }
  
    return parseFloat(value);
  }
}

File: .changeset/cold-grapes-collect.md [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+
+These are all valid in a theme now:
+
+```typescript
+const theme: Theme = {
+  tokens: {
+    borderWidths: {
+      small: '4',
+      medium: '1rem',
+      large: 6,
```
hbuchel, line 17: Is this something specific in RN where this works without a unit? Or do we just not want it to break completely if people try to do it

dbanksdesign: Specific to RN because when we use the theme we use it in a RN StyleSheet which is typed rather than outputting to a CSS string which is untyped. Most numeric values (space, borderWidth, fontSize, etc.) are expected to be a number rather than a string. In CSS its all just strings and numbers can be cast to strings but not the other way around.

calebpollman (approved):

Issue #4228, opened 2023-07-11
Should I need Node.js Polyfills to use  @aws-amplify/ui-react-liveness or bug?


HiroshiOHSUGA:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Liveness

### How is your app built?

Webpack5

### What browsers are you seeing the problem on?

Chrome

### Which region are you seeing the problem in?

not used yet.

### Please describe your bug.

Importing @aws-amplify/ui-react-liveness causes 'ReferenceError: process is not defined' on browser.

It looks this line.
https://github.com/aws-amplify/amplify-ui/blob/15f57d73b6d202bafbd39815aaf31bf3a36e5f09/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts#L31

Is this bug ?
or

* Does FaceLivenessDetector support only Node.js runtime ?
* Does FaceLivenessDetector require some polyfill about process.env for browser runtime?
* If `process.env` was required, What value should I set to NEXT_PUBLIC_STREAMING_API_URL ?

### What's the expected behaviour?

* Importing @aws-amplify/ui-react-liveness without runtime error on browser runtime.
* Showing some UI by FaceLivenessDetector

### Help us reproduce the bug!

Only importing package and run on browser.

### Code Snippet

```typescript
// main.tsx
// only import
import { FaceLivenessDetector } from "@aws-amplify/ui-react-liveness";

console.log(FaceLivenessDetector); // just avoiding tree shaking.
```

### Console log output

```Uncaught ReferenceError: process is not defined
    at ../node_modules/@aws-amplify/ui-react-liveness/dist/esm/components/FaceLivenessDetector/service/utils/streamProvider.mjs (streamProvider.mjs:1:528)
    at __webpack_require__ (bootstrap:19:1)
[...]

PR #4231, opened 2023-07-11
fix(liveness): fix process not defined issues by removing endpoint overrides 

thaddmt:
#### Description of changes
- bug fix for - https://github.com/aws-amplify/amplify-ui/issues/4228

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
@@ -28,7 +28,7 @@ export interface Credentials {
   sessionToken: string;
 }
 
-const ENDPOINT = process.env.NEXT_PUBLIC_STREAMING_API_URL;
+const ENDPOINT = process && process.env.NEXT_PUBLIC_STREAMING_API_URL;
 export const TIME_SLICE = 1000;
 
 function isBlob(obj: unknown): obj is Blob {
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
@@ -28,7 +28,6 @@ export interface Credentials {
   sessionToken: string;
 }
 
-const ENDPOINT = process && process.env.NEXT_PUBLIC_STREAMING_API_URL;
 export const TIME_SLICE = 1000;
 
 function isBlob(obj: unknown): obj is Blob {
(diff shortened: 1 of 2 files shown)


zchenwei (approved):
LGTM. Think people are dropping `&&` in favor of `??` according to the linting rule

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
```diff
@@ -28,7 +28,7 @@ export interface Credentials {
   sessionToken: string;
 }
 
-const ENDPOINT = process.env.NEXT_PUBLIC_STREAMING_API_URL;
+const ENDPOINT = process && process.env.NEXT_PUBLIC_STREAMING_API_URL;
```
calebpollman, line 31: `NEXT_PUBLIC_STREAMING_API_URL` is NextJS specific right? Obviously this has been in place for some time but would the usage of `NEXT_PUBLIC_STREAMING_API_URL` here  cause issues for non-NextJS consumers?

thaddmt: Yeah that could be the case for sure. Further context here is this was just left in the code to provide a backdoor way for changing the streaming endpoint to be the liveness gamma endpoint.

We made this bug fix because a customer discovered with plain webpack 5 and no polyfills they would run into runtime errors due to process not being defined.

reesscot: I don't think it's specific to Next.js, it's just named that way because of how we were using it in a Next.js app.  We could change it to something like `PUBLIC_STREAMING_API_URL`?

calebpollman: Tbh a bit unclear why we need this at all if `process.env.NEXT_PUBLIC_STREAMING_API_URL` resolving to `undefined` doesn't cause failures

thaddmt: @calebpollman Ah it's because when `process` is undefined which is the case where you do not have a polyfill then there are errors about process being undefined and we are trying to access the object `env` on it.

Ideally this value should not affect customers at all and should only be used for debugging. 

@reesscot I thought for next web env variables they need to be prefixed with `NEXT_`? - https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

thaddmt:
Decided offline to remove this code for overriding endpoints completely for now, will move it to a branch if we need it

reesscot (approved):

calebpollman (approved):
🚢

wlee221 (approved):

PR #4236, opened 2023-07-12
chore(ui): add color design token for field control error state

hbuchel:
#### Description of changes

Adds a `color` design token to fieldControl and updates the ShowPasswordButton error state css to use this token for it's color property. We only had border-color defined for error state previously, but controls like the ShowPasswordButton have a color defined.

**Before**
Screenshot 2023-07-12 at 2 35 15 PM

**After**
Screenshot 2023-07-12 at 2 23 31 PM

#### Checklist

- [ ] PR description included


.changeset/cold-drinks-tickle.md
```diff
@@ -0,0 +1,7 @@
+---
+"@aws-amplify/ui": patch
```
dbanksdesign, line 2: Do we need to bump ui-react too? And actually ui-angular and ui-view so they can get these CSS changes?

dbanksdesign (approved):
Just one question on changesets, but LGTM

.changeset/cold-drinks-tickle.md
hbuchel: If we bump ui, it will make angular/vue/react to have an "Updated dependency" entry in the changelog (like this screenshot from an old version packages pr), so I think we are good with just bumping ui and don't need to explicitly add the others.

Screenshot 2023-07-13 at 10 04 55 AM

dbanksdesign (approved):

calebpollman (approved):

PR #4253, opened 2023-07-17
test(react-storage): add e2e tests for StorageImage

zchenwei:
#### Description of changes

- Update the params `useStorageURL` to be an object as a follow-up from https://github.com/aws-amplify/amplify-ui/pull/4054#discussion_r1265600264
```ts
type UseStorageURLParams {
  key: string;
  options?: S3ProviderGetConfig;
  fallbackURL?: string;
  onStorageGetError?: (error: Error) => void;
}
```
- Add e2e tests for StorageImage

#### Checklist

- [x] PR description included


reesscot:
Why are so many seemingly unrelated commits showing on this PR?

zchenwei:
> Why are so many seemingly unrelated commits showing on this PR?

I created this branch long time ago based on the dev branch because the component was not merged into main back then. I can clean up those messages when squashing 😅

examples/next/pages/ui/components/storage/storage-image/private-access-level/index.page.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <StorageImage
+        alt="This is a private image."
+        imgKey="private-e2e.jpeg"
+        accessLevel="private"
+        onLoad={onLoad}
+      />
+      {isLoaded ? (
+        <Text>The private image is loaded.</Text>
+      ) : (
+        <Text>The private image is loading.</Text>
```
reesscot, line 35: nit: use a Loader component instead?

zchenwei: Need some text identifier. <- https://github.com/aws-amplify/amplify-ui/pull/4253/files#diff-01339a8afdb72420eb88feb0f899baeee864c2ed881ff787ab26823e5eeb40c2R11-R12

reesscot (approved):
LGTM

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -15,23 +15,17 @@ export const StorageImage = ({
...
-  const options = React.useMemo(
-    () => ({
-      accessLevel,
-      identityId,
-    }),
-    [accessLevel, identityId]
-  );
-
-  const errorConfig = React.useMemo(
+  const useStorageURLParams = React.useMemo(
```
calebpollman, line 18: Nit: would only memoize the `options` and pass the other values inline to `useStorageURL` to scope the memoization value to the `options` related values

calebpollman (commented):
Just left a bit of feedback, lmk your thoughts!

packages/e2e/features/ui/components/storage/storage-image/load-private-image-from-s3.feature
```diff
@@ -0,0 +1,15 @@
...
+  Background:
+    Given I'm running the example "ui/components/storage/storage-image/private-access-level"
+
+  @react
+  Scenario: I successfully load a private image
+    When I type my "email" with status "CONFIRMED"
+    And I type my password
+    And I click the "Sign in" button
+    Then I see "The private image is loading."
+    Then I see "The private image is loaded."
```
ioanabrooks, line 12: If I understand correctly, this only tests that the onLoad callback executed correctly, not that the image actually is displayed. Can we add a test to check for the image? Retrieve by test ID ideally or even by alt (I saw it defined)

```
cy.get('[alt="This is a private image."]')
.should('be.visible')
```
Same comment for protected/public scope.

packages/react-storage/src/components/StorageImage/StorageImage.tsx
zchenwei: Would also need to memoize the `onStorageGetError` with `useCallback`?

packages/e2e/features/ui/components/storage/storage-image/load-private-image-from-s3.feature
zchenwei: Think a broken image will also be rendered on page with all the props?

examples/next/pages/ui/components/storage/storage-image/private-access-level/index.page.tsx
zchenwei: Think that should work. Good idea

packages/react-storage/src/components/StorageImage/StorageImage.tsx
calebpollman: No since `onStorageGetError` is just passed directly to `useStorageURL` nor is `onStorageGetError` extended by within `StorageImage` (by say,  wrapping within another callback) there is no need for it to be memoized as there is no update to its reference within `StorageImage`. 

Given the scenario of a user defining an `onStorageGetError` callback that has not been memoized before being passed as a prop to `StorageImage` (`onStorageGetError` does not have a stable reference):

- including `onStorageGetError` in the `useStorageURLParams` with `options` would have the unintended consequence of causing `useStorageURLParams` to be re-evaluated on each render

 - wrapping in own `useCallback` would result on re-evaluation on each render

Additionally, since the `useEffect` in `useStorageGetURL` only calls `Storage.get` if `key` is updated the reference of `onStorageGetError` changing will not cause an additional call to `Storage.get`.

hbuchel (approved):

ioanabrooks (approved):

calebpollman (approved):

reesscot:
Can we point this to a feature branch so we can merge the whole feature together?

zchenwei:
> Can we point this to a feature branch so we can merge the whole feature together?

Going to set up one today

PR #4266, opened 2023-07-20
Add colorTheme to Button

hbuchel:
#### Description of changes

Adds colorTheme prop to Buttons to allow for more customization options.

**Todos**
- ~Fix prettier breaking long variable names~ (fixed)
- ~Remove test class names~
- Will track actual doc updates in this separate docs PR https://github.com/aws-amplify/amplify-ui/pull/4270 after this is merged to ButtonRefresh/main 

**Usage**
```
<Button colorTheme="success">Button text</Button>
```
[Preview](https://buttonrefresh-docstest.dvmvffzts1tcu.amplifyapp.com/react/components/button) (not final docs, just for testing)

#### --Tour of changes-- ####
**react/Button.tsx**
Use the new colorTheme prop added to Button to output a modifier class. 

**react/Button.ts**
Creates a new ButtonColorTheme type for the colorTheme prop to use.

**ui/colors.ts**
Adds another step (5) to our overlay ramp because 10 looked too harsh for a subtle hover effect.

**ui/button.ts**
Adds the necessary new design tokens to use for our variation + colorTheme combinations
[Here is a utility on the test docs](https://buttonrefresh-docstest.dvmvffzts1tcu.amplifyapp.com/react/components/button#themeviewer) to view the design tokens across state, color theme, and variations. 
```
primary {
   // base state tokens for primary (essentially our branded color theme)
   _disabled: {},
   _loading: {},
   _hover: {},
   _focus: {},
   _active: {},
   info: {
      // base state tokens for primary--info
     _hover: {},
     _focus: {},
     _active: {},   
   },
   error: {},
   success: {},
   warning: {}, 
}
```

**ui/button.scss**
Added new internal css custom properties defined at the base of `.amplify-button {}` to use for each variation + colorTheme combination. The initial values match those of our default Button. In each variation + colorTheme combination we will modify the values of those css custom properties to use their relevant design token. We use a similar method for loading/disabled properties.
```
.amplify-button {
  --amplify-internal-button-background-color: transparent;
  --amplify-internal-button-border-color: var(--amplify-components-button-border-color);
  --amplify-internal-button-color: var(--amplify-components-button-color);
  --amplify-internal-button-focus-box-shadow: var(--amplify-components-button-focus-box-shadow);
  --amplify-internal-button-border-width: var(--amplify-components-button-border-width);
  // rest
 } 
```
Each variation's style definitions will follow this pattern (using sass `&` to target parent selector as we have previously):
```
&--primary {
  // base primary styles (i.e. no color theme is used, uses existing branded styles)
  &--info {
    // base info styles
    &:hover{}, &:focus{}, &:active{}
  }
  &--error {
    // base error styles
    &:hover{}, &:focus{}, &:active{}
  }
  &--warning {
    // base warning styles
    &:hover{}, &:focus{}, &:active{}
  }
  &--success {
    // base success styles
    &:hover{}, &:focus{}, &:active{}
  }
  &--overlay { 
    // base overlay styles
    &:hover{}, &:focus{}, &:active{}
  }
}
```
Within each "base" and state style block, we will update the internal css custom properties that were defined above.
```
// example with primary + warning:
&--primary {
   &--warning {
      --amplify-internal-button-background-color: var(--amplify-components-button-primary-warning-background-color);
      --amplify-internal-button-border-color: var(--amplify-components-button-primary-warning-border-color);
      --amplify-internal-button-color: var(--amplify-components-button-primary-warning-color);
      &:focus {
        --amplify-internal-button-background-color: var(--amplify-components-button-primary-warning-focus-background-color);
        --amplify-internal-button-border-color: var(--amplify-components-button-primary-warning-focus-border-color);
        --amplify-internal-button-color: var(--amplify-components-button-primary-warning-focus-color);
        --amplify-internal-button-focus-box-shadow: var(--amplify-components-button-primary-warning-focus-box-shadow);
      }
  }
} 
```
Buttons with a colorTheme will have modifier classes that combine the variation with the colorTheme, so:
```
<Button variation="primary" colorTheme="success">button</Button>
```
Will produce `amplify-button--primary--success`:
```
<button class="
  amplify-button 
  amplify-field-group__control 
  amplify-button--primary 
  amplify-button--primary--success" 
  data-fullwidth="false" data-variation="primary" type="button">button</button>
```  

#### Checklist

- [x] PR description included


packages/ui/src/theme/css/component/button.scss
hbuchel, line 1: The `/* prettier-ignore */` in this file is a temporary fix for CSS variables compiling with extra spaces (which breaks the css) and should be removed as part of this PR.

packages/ui/src/theme/tokens/components/button.ts
```diff
@@ -133,8 +160,160 @@ export const button: Required<ButtonTokens<'default'>> = {
     backgroundColor: { value: 'transparent' },
     borderColor: { value: '{colors.border.tertiary.value}' },
   },
-
   // variations
+  outlined: {
```
hbuchel, line 164: Even though our default Button variation doesn't have a name, we need something to construct the custom properties pattern for the default variation that doesn't clash with the existing warning variation. 

If we did: 
```
button: { 
  info{}, 
  warning{}, 
  // ..etc 
}
```
instead, where the default button variation's color themes sit at the root, we would end up with tokens for the warning color theme that clashes with the tokens for the existing warning variation `--amplify-components-button-warning-background-color` so that is why they are nested inside of outlined here, even though there is no explicit 'outlined' variation.

packages/ui/src/theme/tokens/components/button.ts
```diff
@@ -196,7 +520,7 @@ export const button: Required<ButtonTokens<'default'>> = {
   link: {
     backgroundColor: { value: 'transparent' },
     borderColor: { value: 'transparent' },
-    borderWidth: { value: '{space.zero.value}' },
+    borderWidth: { value: '{borderWidths.small.value}' },
```
hbuchel, line 523: Updated this to have a non-zero border width so that this button variation can work in Window's high contrast mode.

dbanksdesign (commented):
Just a few minor things, looks good otherwise!

packages/ui/src/theme/css/component/button.scss
```diff
@@ -23,346 +31,669 @@
...
   }
+
+  &--outlined {
+    &--info {
+      --amplify-internal-button-background-color: var(--amplify-components-button-outlined-info-background-color);
+      --amplify-internal-button-border-color: var(--amplify-components-button-outlined-info-border-color);
+      --amplify-internal-button-color: var(--amplify-components-button-outlined-info-color);
+
+      &:hover,
+      &.amplify-button--test-hover {
```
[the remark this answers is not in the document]
hbuchel: I actually was not intending to keep those and was just using them to test the states easier.

packages/ui/src/theme/tokens/components/button.ts
```diff
@@ -133,8 +160,160 @@ export const button: Required<ButtonTokens<'default'>> = {
...
+        borderColor: { value: '{colors.blue.60.value}' },
+        backgroundColor: { value: '{colors.blue.10.value}' },
+        color: { value: '{colors.blue.100.value}' },
+      },
+      _focus: {
+        borderColor: { value: '{colors.blue.100.value}' },
+        backgroundColor: { value: '{colors.blue.10.value}' },
+        color: { value: '{colors.blue.100.value}' },
+        // Should we make field control box shadows for info, warning, success?
+        boxShadow: {
```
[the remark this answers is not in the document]
hbuchel: Oh good catch. Fixed in the latest commit [here](https://github.com/aws-amplify/amplify-ui/pull/4266/commits/cdd6ca7de4363efefa825c4c27e6f960d4b1b950) Also deployed to preview docs site but it might still be building depending on when you see this :)

packages/ui/src/theme/css/component/button.scss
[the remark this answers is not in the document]
dbanksdesign: Ah gotcha, makes sense

dbanksdesign (approved):
LGTM!

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -24,14 +24,34 @@ const ButtonPrimitive: Primitive<ButtonProps, 'button'> = (
...
     ...rest
   },
   ref
 ) => {
+  // These variations support colorThemes. 'undefined' accounts for our
+  // 'default' variation which is not named. This is done so we do not need to
+  // include a breaking change of defining a default value for variation.
+  const supportsColorThemes = ['link', 'primary', undefined].includes(
+    variation
+  );
```
ioanabrooks, line 37: Nit: not a fan of "magic values", would prefer having a const defined outside the component, i.e.

```
const supportedVariations = ['link', 'primary', undefined];
```

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -24,14 +24,34 @@ const ButtonPrimitive: Primitive<ButtonProps, 'button'> = (
...
+
+  // Use the variation to construct our color theme modifier classes; in the
+  // case of our 'default' variation, use 'outlined'
+  const variationColorThemeModifier = variation ? variation : 'outlined';
```
ioanabrooks, line 41:
```suggestion
  const variationColorThemeModifier = variation ?? 'outlined';
```

ioanabrooks (commented):
Overall lgtm, just a couple nits

packages/react/src/primitives/Button/Button.tsx
hbuchel: Thanks, updated in latest!

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -24,14 +28,30 @@ const ButtonPrimitive: Primitive<ButtonProps, 'button'> = (
     size,
     type = 'button',
     variation,
+    colorTheme,
```
calebpollman, line 31: Nit: alphabetize

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -24,14 +28,30 @@ const ButtonPrimitive: Primitive<ButtonProps, 'button'> = (
...
     ...rest
   },
   ref
 ) => {
+  // Test if variation supports color themes.
+  const supportsColorThemes = supportedVariations.includes(variation);
+
+  // Use the variation to construct our color theme modifier classes; in the
+  // case of our 'default' variation, use 'outlined'
+  const variationColorThemeModifier = variation ?? 'outlined';
```
calebpollman, line 41: Should this depend on the resolved value of  `supportsColorThemes`?

packages/react/src/primitives/Button/Button.tsx
```diff
@@ -24,14 +28,30 @@ const ButtonPrimitive: Primitive<ButtonProps, 'button'> = (
...
     ComponentClassNames.FieldGroupControl,
     classNameModifier(ComponentClassNames.Button, variation),
+    // Check if variation supports colorThemes before applying
+    // colorTheme modifying class.
+    supportsColorThemes &&
+      colorTheme &&
+      classNameModifier(
+        ComponentClassNames.Button,
+        `${variationColorThemeModifier}--${colorTheme}`
+      ),
```
calebpollman, line 54: Does `classNames` filter out `false` parameters?

packages/react/src/primitives/Button/__tests__/Button.test.tsx
```diff
@@ -46,6 +46,239 @@ describe('Button test suite', () => {
     expect(menu.classList).toContain(`${ComponentClassNames['Button']}--menu`);
   });
 
+  it('should render color themes for default variation', async () => {
```
calebpollman, line 49: To isolate test behavior can we break these out in to individual tests using `it.each`? Example:
```
const DEFAULT_SUPPORTED_COLOR_THEMES= ['info', 'success']


it.each(DEFAULT_SUPPORTED_COLOR_THEMES)('should render the %s color theme for the default variation', async (colorTheme) => {
  const testId = `default${capitalize(colorTheme)}ColorTheme`
  const button = await screen.findByTestId(testId);
  expect(button.classList).toContain(`amplify-button--outlined--${colorTheme}`);
})
```

Same question for 103, 173, and 245.

packages/react/src/primitives/types/button.ts
```diff
@@ -11,6 +11,13 @@ export type ButtonVariations =
   | 'warning'
   | 'destructive';
 
+export type ButtonColorThemes =
```
calebpollman, line 14: Prefer singular form for string unions

packages/react/src/primitives/Button/Button.tsx
hbuchel: If I'm understanding your question, no, `variationColorThemeModifier` is used to build out our modifier class. In the case of our default button, where there is no variation defined, it uses the name 'outlined' to build out the class. In every other case, it uses the variation name. Then, on line 49~ish, we use `supportsColorTheme` (and if `colorTheme` exists) to determine if that modifier class should be applied. Does that make sense?

packages/react/src/primitives/Button/__tests__/Button.test.tsx
hbuchel: @calebpollman Updated this in latest if you can please take a look

hbuchel: @calebpollman Oops I missed the line 245 part, do you think we should have tests for each colorTheme for the unsupported variations?

packages/react/src/primitives/Button/__tests__/Button.test.tsx
```diff
@@ -3,9 +3,17 @@ import { render, screen } from '@testing-library/react';
...
 import { ComponentClassNames } from '../../shared';
 
 describe('Button test suite', () => {
+  const SUPPORTED_COLOR_THEMES = [
+    'info',
+    'success',
+    'warning',
+    'error',
+    'overlay',
+  ];
```
calebpollman, line 16: Nit: declare constant at global scope

```suggestion
const SUPPORTED_COLOR_THEMES = [
  'info',
  'success',
  'warning',
  'error',
  'overlay',
];

describe('Button test suite', () => {
```

packages/react/src/primitives/types/button.ts
```diff
@@ -11,6 +11,13 @@ export type ButtonVariations =
   | 'warning'
   | 'destructive';
 
+export type ButtonColorTheme =
+  | 'info'
+  | 'error'
+  | 'warning'
```
calebpollman, line 17: This might just be a me problem, but having a hard time understanding why there is both a `warning` variant and a `warning` color scheme? Is this just to prevent a breaking change for `variation`?

packages/react/src/primitives/Button/__tests__/Button.test.tsx
```diff
@@ -46,6 +54,94 @@ describe('Button test suite', () => {
...
+  it.each(SUPPORTED_COLOR_THEMES)(
+    'should render the %s color theme for the primary variation',
+    async (colorTheme) => {
+      const testId = `primary-${colorTheme}-ColorTheme`;
+      render(
+        <Button
+          variation="primary"
+          testId={testId}
+          colorTheme={colorTheme as ButtonColorTheme}
+        ></Button>
```
calebpollman, line 83: NIt: prefer self inclosing tags for JSX
```suggestion
        />
```

packages/react/src/primitives/Button/__tests__/Button.test.tsx
```diff
@@ -46,6 +54,94 @@ describe('Button test suite', () => {
...
   });
 
+  it.each(SUPPORTED_COLOR_THEMES)(
+    'should render the %s color theme for the default variation',
+    async (colorTheme) => {
+      const testId = `default-${colorTheme}-ColorTheme`;
+      render(
+        <Button
+          testId={testId}
+          colorTheme={colorTheme as ButtonColorTheme}
```
calebpollman, line 64: _Think_ we can avoid casting here by annotating `SUPPORTED_COLOR_THEMES` with `ButtonColorTheme[]`

```
const SUPPORTED_COLOR_THEMES: ButtonColorTheme = [];
```

packages/react/src/primitives/Button/Button.tsx
calebpollman: Okay got it now. Since we don't reuse `variationColorThemeModifier` anywhere, maybe we just inline the logic on line 53 since it is only relevant if `supportsColorTheme` is truthy? Alternatively, think it is possible to abstract the relevant pieces here into a variable that resolves to either a `string` or `undefined` to pass to `classNameModifier`:
```
const colorThemeModifier = 
  supportedVariations.includes(variation) && colorTheme
  ? `${variation ?? 'outlined'}--${colorTheme}`
   : undefined;

const componentClasses = classNames(
  ComponentClassNames.Button,
  classNameModifier(ComponentClassNames.Button, variation),
  classNameModifier(ComponentClassNames.Button, colorThemeModifier),
  classNameModifier(ComponentClassNames.Button, size),
  // ...more modifiers
),
```

packages/react/src/primitives/Button/__tests__/Button.test.tsx
calebpollman: Forgive me if I'm totally off here, but those variations are to be removed in the future correct?

packages/react/src/primitives/types/button.ts
hbuchel: This is one of the things that spurred this refresh; there was a `warning` and `destructive` variation added for Button for Account Settings. They are both red (which sort of deviates from our pattern of using orange for warning like in Alert/Message). Additionally, having variations that are really just color modifications from an existing variation is a bit limiting for adding future color modifications (like what do we call a variation that is a green button that has a solid background color vs a green button that has an outline?). With this change, we have to worry less about naming and have more options for buttons where a color scheme is useful (i.e. buttons that might have an action associated with them like "approve" or "delete") or where they can aid with better contrast (i.e. our Alert/Message primitive has a colored background which can clash with a branded color theme like our teal buttons have). 

This isn't in this PR, but ideally, we should deprecate the `warning` and `destructive` variations and customers would use:
```
<Button variation="primary" colorTheme="error" />
```
to achieve the current "destructive" style. And they would use:
```
<Button colorTheme="error" />
```
to achieve the existing "warning" style (which is really another red danger button).

Let me know if that helps clarify!

packages/react/src/primitives/Button/__tests__/Button.test.tsx
hbuchel: No worries, that's right! We should deprecate warning and destructive variations.

The `menu` variation is a little weird, it's used specifically for our Menu primitive and we don't _reallllly_ advertise it as an option (like we don't include it in our Button demo on the docs, its buried down further on the page, and we don't support it in Figma). This is one of our Radix primitives, which we've considered refactoring to drop the dependency. I think for now it's safe to not have it support colorTheme and maybe in the future we can revisit depending on where we net out with the Menu primitive.

calebpollman: Cool, since those tests aren't going to be around for long not too concerned

packages/react/src/primitives/types/button.ts
calebpollman: IMO `error` means something has already gone wrong (which might mean it should be scoped to "message" related components), `warning` should warn the user, and `destructive` should indicate an action that is well, _destructive_ hehe. Totally onboard with removing them as variations but do think they have semantic meaning that `error` does not convey. Kind of wonder if we should include `destructive` as a `colorTheme` and consider removing `error`?

packages/react/src/primitives/Button/__tests__/Button.test.tsx
hbuchel: Yes, that works. Updated.

packages/react/src/primitives/types/button.ts
hbuchel: I actually considered using color namespaces from our theme (red, green, blue, purple, etc) as names for the colorThemes and foregoing any association with semantics/action but during the doc review we settled on this (using just warning, success, info, error) as a more limited start. Perhaps it could be expanded in the future to support any namespace from our color palette, and these end up just being special "named" color themes that inherit those base not-special color themes

Generally, I don't want to actually tie too much semantic meaning to a name of a color scheme; but these are so common across design systems that I think it's ok. Especially if we're already associating these names with colors elsewhere in our own design system and reinforcing that pattern. Happy to be convinced though if we want to chat offline with Danny!

packages/react/src/primitives/Button/Button.tsx
hbuchel: I like that! Updated in the latest

packages/react/src/primitives/Button/Button.tsx
hbuchel: I think the above change^ addresses this (pulling out `classNameModifier` as a flag) and using 
```
classNameModifier(ComponentClassNames.Button, colorThemeModifier),
``` 
here instead

PR #4288, opened 2023-07-27
Breadcrumbs

dbanksdesign:
#### Description of changes

```jsx
  <Breadcrumbs items={[
    {
      href: '/',
      label: 'Home'
    },
    {
      href: '/category',
      label: 'Category'
    },
    {
      href: '/category/type',
      label: 'Type'
    },
  ]} />
```

```jsx
  <Breadcrumbs.Container>
    {breadcrumbs.map(({ href, label }, idx) => {
      const isCurrent = breadcrumbs.length - 1 === idx;
      return (
        <Breadcrumbs.Item>
          <Breadcrumbs.Link isCurrent={isCurrent} href={href}>{label}</Breadcrumbs.Link>
          {isCurrent ? null : <Breadcrumbs.Separator>/</Breadcrumbs.Separator>}
        </Breadcrumbs.Item>
      )}
    )}
  </Breadcrumbs.Container>
```
https://breadcrumbs.d32sljejpnv3v5.amplifyapp.com/

* Breadcrumbs should be able to accept any children to allow for other elements like a dropdown
* Breadcrumbs should work with routing libraries
    * https://reactrouter.com/en/main/components/link
    * https://nextjs.org/docs/pages/api-reference/components/link

Inspiration
* https://chakra-ui.com/docs/components/breadcrumb/usage
* https://mui.com/material-ui/react-breadcrumbs/
* https://react-spectrum.adobe.com/react-aria/useBreadcrumbs.html
* https://ant.design/components/breadcrumb

#### Checklist

- [x] PR description included


docs/scripts/generate-props-tables-data.ts
```diff
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import path from 'path';
 import { globbyStream } from 'globby';
-import { getAllTypesData, getCatalog } from './util';
+import { getAllTypesData, getCatalog, componentsWithChildren } from './util';
```
dbanksdesign, line 4: Moved this const to util so we can use it in other scripts

docs/scripts/util/getCatalog.ts
```diff
@@ -48,6 +49,30 @@ function getCatalog() {
   const catalog: Catalog = {} as Catalog;
   for (const [componentName, [node]] of source.getExportedDeclarations()) {
     let properties: Properties = {};
+    const children = componentsWithChildren[componentName];
```
dbanksdesign, line 52: This code adds nested components into the catalog. New primitives need to be added to the catalog.d.ts and componentsWithChildren

docs/src/components/Layout/Breadcrumbs.tsx
```diff
@@ -0,0 +1,38 @@
+import Link from 'next/link';
```
dbanksdesign, line 1: Using the Breadcrumbs primitive in our docs site to show breadcrumbs!

docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsThemeExample.tsx
```diff
@@ -0,0 +1,39 @@
+import { Breadcrumbs, ThemeProvider, createTheme } from '@aws-amplify/ui-react';
+
+const theme = createTheme({
```
Comment (dbanksdesign) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsThemeExample.tsx:3]: Even though this function call is not necessary, it does make this example TS and JS friendly!

docs/src/pages/[platform]/components/breadcrumbs/react.mdx
```diff
@@ -0,0 +1,150 @@
...
+  BreadcrumbsStyleExample,
+  CustomSeparatorIconExample,
+  BreadcrumbsArrayExample,
+} from './examples'
+
+import { ComponentStyleDisplay } from '@/components/ComponentStyleDisplay'
+import { Example, ExampleCode } from '@/components/Example'
+import ThemeExample from '@/components/ThemeExample.mdx'
+
+
```
Comment (dbanksdesign) [inline docs/src/pages/[platform]/components/breadcrumbs/react.mdx:15]: Didn't include a Demo since there is only 1 top level prop, `separator`, and most of the customization comes from the composition which can't really be done in our demo component

docs/src/styles/docs/main.scss
```diff
@@ -14,7 +14,9 @@
 }
 
 .docs-content-body {
-  padding: var(--docs-horizontal-padding);
+  padding-block-start: var(--amplify-space-medium);
```
dbanksdesign, line 17: This just makes the body of docs pages not start so far down

packages/react/src/primitives/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const BreadcrumbsPrimitive: Primitive<BreadcrumbProps, 'nav'> = (
+  { className, children, separator = '/', ...rest },
+  ref
+) => {
+  const componentClasses = classNames(
+    ComponentClassNames.Breadcrumbs,
+    className
+  );
+  const value = React.useMemo(() => {
```
dbanksdesign, line 24: eslint was telling me I need to memoize this or it'll cause a re-render every time. Seems silly though?

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+} from '../types';
+import { ComponentClassNames } from '../shared/constants';
+import { Link } from '../Link';
+import { BreadcrumbsContext } from './BreadcrumbsContext';
+
+const BreadcrumbLinkPrimitive: Primitive<BreadcrumbLinkProps, 'span'> = (
+  { className, children, href, ...rest },
+  ref
+) => {
+  const { isCurrent, isDisabled } = React.useContext(BreadcrumbsContext);
```
dbanksdesign, line 19: These come from the context so customers only need to put `isCurrent` on the Breadcrumbs.Item and not both that and Breadcrumbs.Link

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const value = React.useMemo(() => {
+    return { isCurrent, isDisabled };
+  }, [isCurrent, isDisabled]);
+
+  return (
+    <View as="li" className={componentClasses} ref={ref} {...rest}>
+      <BreadcrumbsContext.Provider value={value}>
+        {children}
+      </BreadcrumbsContext.Provider>
+      {isCurrent ? null : (
```
dbanksdesign, line 43: Only show the separator if its not the last item

dbanksdesign (commented):
Walkthrough

docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsWithDropdownExample.tsx
Comment (hbuchel) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsWithDropdownExample.tsx:1]: FYI, I opened a PR https://github.com/aws-amplify/amplify-ui/pull/4301 for this weird behavior (issue with Select primitive): 
Screenshot 2023-07-31 at 1 03 49 PM

packages/ui/src/theme/css/component/breadcrumbs.scss
```diff
@@ -0,0 +1,37 @@
+// No styles on .amplify-breadcrumbs
+
+.amplify-breadcrumbs__list {
```
hbuchel, line 3: Probably people will want to do "something" on narrow layouts with their breadcrumbs, but do you think we should have a default experience of wrapping this list (at the minimum a flex-wrap: wrap, but with a token for theming) so it can collapse on narrow viewports? Currently it will cause a horizontal page scroll.

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,60 @@
...
+      'disabled',
+      isDisabled
+    ),
+    className
+  );
+
+  const linkProps = {
+    'aria-current': isCurrent ? 'page' : undefined,
+    'aria-disabled': isDisabled ? 'true' : undefined,
+    as: isCurrent ? 'span' : 'a',
```
hbuchel, line 38: This results in kind of weird HTML, a span with an href:
```
<span class="amplify-link amplify-breadcrumbs__link amplify-breadcrumbs__link--current" href="/react/components/breadcrumbs" aria-current="page">Breadcrumbs</span>
```
What do you think about outputting a `<Text />` element instead or conditionally adding the href?

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,60 @@
...
+      ComponentClassNames.BreadcrumbsLink,
+      'disabled',
+      isDisabled
+    ),
+    className
+  );
+
+  const linkProps = {
+    'aria-current': isCurrent ? 'page' : undefined,
+    'aria-disabled': isDisabled ? 'true' : undefined,
```
hbuchel, line 37: I left a previous comment about this before the change to output the current element as text, but do we have a use case intended for passing `aria-disabled` to the link? If not, I'd remove it all together. If we need it for something weird we _might consider_ this: https://www.scottohara.me/blog/2021/05/28/disabled-links.html

packages/react/src/primitives/Breadcrumbs/Breadcrumbs.tsx
reesscot: I think that's because you're wrapping `separator` in an object, which will be a new object every time. If `value` is a new object on every render, then it will cause renders to all component consuming the context. If you don't want to do this, you could just pass separator straight to value since that appears to be the only thing in your context value.

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+import { classNameModifierByFlag } from '../shared/utils';
```
reesscot, line 4: nit: alphabetize imports

packages/react/src/primitives/types/breadcrumbs.ts
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Sizes } from './base';
+import { ElementType, PrimitiveProps, BaseViewProps } from './view';
+import { BaseLinkProps } from './link';
+
+/** @deprecated For internal use only */
+export interface BaseBreadcrumbProps extends BaseViewProps {
```
calebpollman, line 7: Out of scope for this PR, but what if we stop exporting all our primitive types so we no longer need to add deprecation messages for new types?

packages/react/src/primitives/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
...
+    };
+  }, [separator]);
+
+  return (
+    <View
+      as="nav"
+      aria-label="Breadcrumb"
+      className={componentClasses}
+      ref={ref}
+      {...rest}
```
calebpollman, line 36: Is it intentional that `rest.as` and `rest.aria-label` override the specified values above?

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { ComponentClassNames } from '../shared/constants';
+import { Link } from '../Link';
+import { Text } from '../Text';
+import { BreadcrumbsContext } from './BreadcrumbsContext';
+
+const BreadcrumbLinkPrimitive: Primitive<BreadcrumbsLinkProps, 'a'> = (
+  { className, children, href, ...rest },
+  ref
+) => {
+  const { isCurrent } = React.useContext(BreadcrumbsContext);
```
calebpollman, line 20: With `isCurrent` resolving to the same value for each ` BreadcrumbLinkPrimitive`, it's a bit unclear to me how the conditional logic works if more than one `BreadcrumbLinkPrimitive` is passed as the `children` to `Breadcrumbs`?

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+  );
+
+  if (isCurrent) {
+    return (
+      <Text
+        as="span"
+        aria-current="page"
+        className={componentClasses}
+        ref={ref}
+        {...rest}
```
calebpollman, line 38: If the types we allow for `BreadcrumbLinkPrimitive` are for an `anchor` element, any concerns with passing them to a `Text` element?

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+      isCurrent
+    ),
+    className
+  );
+
+  if (isCurrent) {
+    return (
+      <Text
+        as="span"
+        aria-current="page"
```
calebpollman, line 35: Just confirming that we want these overridden by `rest`

packages/react/src/primitives/types/breadcrumbs.ts
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * @description
+   * The size property will affect the font size of the breadcrumbs.
+   */
+  size?: Sizes;
+  separator?: React.ReactNode;
+  children?: React.ReactNode;
```
calebpollman, line 14: Do we need to explicitly add `children`?

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+        {children}
+      </BreadcrumbsContext.Provider>
+      {isCurrent ? null : (
+        <View
+          as="span"
+          aria-hidden="true"
+          className={ComponentClassNames.BreadcrumbsSeparator}
+        >
+          {separator}
+        </View>
```
calebpollman, line 45: What if the customer wants to include the `separator` as part of the `children` they pass to `BreadcrumbItem` (meaning no need for a standalone `separator`)?

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+    ),
+    className
+  );
+  const value = React.useMemo(() => {
+    return { isCurrent };
+  }, [isCurrent]);
+
+  return (
+    <View as="li" className={componentClasses} ref={ref} {...rest}>
+      <BreadcrumbsContext.Provider value={value}>
```
calebpollman, line 35: Asked abt `isCurrent` in another comment, but maybe that would be resolved by having a context that is specific to  `BreadcrumbItem`?

packages/react/src/primitives/Breadcrumbs/BreadcrumbsContext.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+
+interface BreadcrumbsContextInterface {
```
calebpollman, line 3: Prefer appending `Type` over `Interface` for Context types namespaces (existing convention)

packages/react/src/primitives/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
...
+    >
+      <View as="ol" className={ComponentClassNames.BreadcrumbsList}>
```
calebpollman, line 38: What do you think of combining with the parent `View` to have a single wrapping `View`?

dbanksdesign: Semantically breadcrumbs should be within a `<nav>` and an `<ol>`: https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/

packages/react/src/primitives/types/breadcrumbs.ts
dbanksdesign: I like that, it feels weird to use these deprecated comments

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
dbanksdesign: They could do that if they wanted if they pass null to the separator prop and include whatever separator they want in BreadcrumbItem children

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
dbanksdesign: Ah good catch

dbanksdesign: Actually, maybe?

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
calebpollman: But the `View` here would still be rendered correct?

packages/react/src/primitives/Breadcrumbs/BreadcrumbsContext.tsx
```diff
@@ -0,0 +1,12 @@
...
+  separator?: React.ReactNode;
+  // We pass down isCurrent from
+  // Breadcrumbs.Item to Breadcrumbs.Link
+  // because we need that information on both elements
+  isCurrent?: boolean;
+}
+
+export const BreadcrumbsContext =
```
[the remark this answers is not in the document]
dbanksdesign: Done

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
dbanksdesign: Ah good point, we could conditionally render this IF there is a separator

docs/src/data/links.tsx
```diff
@@ -461,6 +462,13 @@ export const navigationComponents: ComponentNavItem[] = [
     platforms: ['react'],
     icon: MdTab,
   },
+  {
+    href: '/components/breadcrumbs',
+    label: 'Breadcrumbs',
+    body: `The breadcrumbs component is a navigation list used to visualize the information architecture of a website and the users current place in the hierarchy.`,
```
thaddmt, line 468:
```suggestion
    body: `The breadcrumbs component is a navigation list used to visualize the information architecture of a website and the user's current place in the hierarchy.`,
```

packages/react/src/primitives/Breadcrumbs/BreadcrumbLink.tsx
dbanksdesign: If you have 2 BreadcrumbLinks inside 1 BreadcrumbItem, they would both get the same `isCurrent`. Although, hopefully you wouldn't do that

packages/react/src/primitives/Breadcrumbs/BreadcrumbItem.tsx
thaddmt: Just a question here since these comments made me think of it, but if someone wanted to make breadcrumbs look like `/ path / to / file /` they would probably have to do this with `separator=null` and manually prepend the `/` on each item. Then on the current item they would have to add the `/` both before and after?

calebpollman: The default `/` is added in the parent `Breadcrumbs`: https://github.com/aws-amplify/amplify-ui/pull/4288/files#diff-38d21fc1975039b31cc31890ae807f45a11e592c8f78104b19a5d5aeb11643a6R17

dbanksdesign: Updated it so there is a `Breadcrumbs.Separator` component, which is the default separator, but could also be used in composition to do `/ path / to / file /` :  https://github.com/aws-amplify/amplify-ui/pull/4288/files#diff-dd142c60c767c65e50f6ed3e579ccd3974c6ce921aa7c73aee8465aec2f3a35b

docs/src/pages/[platform]/components/breadcrumbs/examples/DefaultBreadcrumbsExample.tsx
```diff
@@ -0,0 +1,17 @@
+import { Breadcrumbs } from '@aws-amplify/ui-react';
+
+export default function DefaultBreadcrumbsExample() {
+  return (
+    <Breadcrumbs>
+      <Breadcrumbs.Item>
+        <Breadcrumbs.Link href="/react/components">Home</Breadcrumbs.Link>
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/DefaultBreadcrumbsExample.tsx:7]:
```suggestion
        <Breadcrumbs.Link href="/react">Home</Breadcrumbs.Link>
```

docs/src/pages/[platform]/components/breadcrumbs/examples/NextBreadcrumbsExample.tsx
```diff
@@ -0,0 +1,38 @@
+import Link from 'next/link';
+import { Breadcrumbs } from '@aws-amplify/ui-react';
+import { useRouter } from 'next/router';
+
+export default function NextBreadcrumbsExample() {
+  const { asPath } = useRouter();
+
+  const nestedRoutes = asPath
+    .split('#')[0]
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/NextBreadcrumbsExample.tsx:9]: Why not use https://developer.mozilla.org/en-US/docs/Web/API/URL to get [pathname](https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname)?

docs/src/pages/[platform]/components/breadcrumbs/examples/NoSeparatorExample.tsx
```diff
@@ -0,0 +1,17 @@
+import { Breadcrumbs } from '@aws-amplify/ui-react';
+
+export default function NoSeparatorExample() {
+  return (
+    <Breadcrumbs separator={null}>
+      <Breadcrumbs.Item>
+        <Breadcrumbs.Link href="/react/components">Home</Breadcrumbs.Link>
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/NoSeparatorExample.tsx:7]:
```suggestion
        <Breadcrumbs.Link href="/">Home</Breadcrumbs.Link>
```

docs/src/pages/[platform]/components/breadcrumbs/examples/CustomSeparatorExample.tsx
```diff
@@ -0,0 +1,19 @@
+import { Breadcrumbs } from '@aws-amplify/ui-react';
+
+export default function CustomSeparatorExample() {
+  return (
+    <Breadcrumbs
+      separator={<Breadcrumbs.Separator>{`|`}</Breadcrumbs.Separator>}
+    >
+      <Breadcrumbs.Item>
+        <Breadcrumbs.Link href="/react/components">Home</Breadcrumbs.Link>
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/CustomSeparatorExample.tsx:9]:
```suggestion
        <Breadcrumbs.Link href="/">Home</Breadcrumbs.Link>
```

docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsThemeExample.tsx
```diff
@@ -0,0 +1,41 @@
...
+    },
+  },
+});
+
+export default function BreadcrumbsThemeExample() {
+  return (
+    <ThemeProvider theme={theme}>
+      <Breadcrumbs>
+        <Breadcrumbs.Item>
+          <Breadcrumbs.Link href="/react/components">Home</Breadcrumbs.Link>
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsThemeExample.tsx:28]:
```suggestion
          <Breadcrumbs.Link href="/">Home</Breadcrumbs.Link>
```

docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsThemeExample.tsx
```diff
@@ -0,0 +1,44 @@
...
+  },
+});
+
+export default function BreadcrumbsThemeExample() {
+  return (
+    <ThemeProvider theme={theme}>
+      <Breadcrumbs
+        items={[
+          {
+            href: '/react/components',
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/BreadcrumbsThemeExample.tsx:29]:
```suggestion
            href: '/',
```

docs/src/pages/[platform]/components/breadcrumbs/examples/CustomSeparatorIconExample.tsx
```diff
@@ -0,0 +1,24 @@
+import { Breadcrumbs } from '@aws-amplify/ui-react';
+import { MdChevronRight } from 'react-icons/md';
+
+export default function CustomSeparatorIconExample() {
+  return (
+    <Breadcrumbs
+      items={[
+        {
+          href: '/react/components',
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/breadcrumbs/examples/CustomSeparatorIconExample.tsx:9]:
```suggestion
          href: '/',
```

packages/react/src/primitives/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,63 @@
...
+  Primitive,
+  ForwardRefPrimitive,
+} from '../types';
+
+import { BreadcrumbItem } from './BreadcrumbItem';
+import { BreadcrumbLink } from './BreadcrumbLink';
+import { BreadcrumbSeparator } from './BreadcrumbSeparator';
+import { BreadcrumbContainer } from './BreadcrumbContainer';
+
+const DefaultSeparator = <BreadcrumbSeparator>{'/'}</BreadcrumbSeparator>;
```
calebpollman, line 15: What do you think of using `"/"` as the default value of `children` in `BreadcrumbsSeparator` so in composable mode the default `Breadcrumbs.Separator` renders the slash?

packages/react/src/primitives/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { BreadcrumbContainer } from './BreadcrumbContainer';
+
+const BreadcrumbsPrimitive: Primitive<BreadcrumbsProps, 'nav'> = (
+  { className, items, separator = BreadcrumbSeparator, ...rest },
+  ref
+) => {
+  const ariaLabel = rest['aria-label'] ?? 'Breadcrumb';
+
+  return (
+    <BreadcrumbContainer aria-label={ariaLabel} className={className} ref={ref}>
```
calebpollman, line 22: Might be intentional and if so feel free to disregard but should `rest` be passed to the `BreadcrumbsContainer`?
```suggestion
    <BreadcrumbContainer
      {...rest}
      aria-label={ariaLabel}
      className={className}
      ref={ref}
    >
```

calebpollman (approved):
🚢

reesscot (approved):
Shipit!

Issue #4249, opened 2023-07-14
[React Liveness Component] `Cannot read properties of undefined (reading 'height')` in `useMediaStreaminVideo`


bradjones1:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Liveness

### How is your app built?

Expo web (Webpack 5)

### What browsers are you seeing the problem on?

Chrome

### Please describe your bug.

It appears that `videoConstraints` can be undefined in the constructor at

https://github.com/aws-amplify/amplify-ui/blob/78fdfd6c8268c56204f905402162ad8cb40a0c8e/packages/react-liveness/src/components/FaceLivenessDetector/hooks/useMediaStreamInVideo.ts#L14

Sentry log:

https://not-vanilla-llc.sentry.io/share/issue/beb24914031d41f0ab4ffbb5bec9e5db/

### What's the expected behaviour?

No exception thrown.

### Help us reproduce the bug!

This is from my production error logging and I have not been able to reproduce myself.

calebpollman:
Hi @bradjones1! Reading how your app is built it sounds like you may be using React Native Web, can you confirm?

bradjones1:
@calebpollman Confirmed. The app also runs on React Native, however since there aren't RN bindings yet for Liveness I redirect those users over to a set of pages on the RN Web experience to complete the liveness check. Those who start on the RN Web UI stay there, obviously.

calebpollman:
Thanks for confirming @bradjones1. @thaddmt Do you have any ideas here?

reesscot:
@thaddmt Let's add a check for `videoConstraints` before accessing `height`

thaddmt:
This should be addressed in the latest release

PR #4314, opened 2023-08-02
fix: ensure that media track constraints is not undefined

thaddmt:
#### Description of changes
- Addresses - https://github.com/aws-amplify/amplify-ui/issues/4249
- Initialized video track constraints and checked for the value before using it

#### Checklist

- [ ] PR description included


packages/react-liveness/src/components/FaceLivenessDetector/hooks/useMediaStreamInVideo.ts
```diff
@@ -11,6 +11,10 @@ export function useMediaStreamInVideo(
   stream: MediaStream,
   videoConstraints: MediaTrackConstraints
 ): UseMediaStreamInVideo {
+  if (!videoConstraints) {
```
reesscot, line 14: Is it even possible or `videoConstraints` to be `undefined` now? https://github.com/aws-amplify/amplify-ui/pull/4314/files#diff-49f4d37e97030928199c54c5f8be21e23f8cb9565e90c1cc0845e6c3ad7a7fe9R71

thaddmt: It shouldn't be possible but I still added this just in case, the only reason this wasn't getting caught before was we were using `videoConstraints!` at caller side. Happy to remove but was just covering bases

reesscot: Where are we using `videoConstraints!`?

thaddmt: It's used when getting the video media recorder object - https://github.com/aws-amplify/amplify-ui/blob/fix-undefined-media/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts#L878

useMediaStreamInVideo is used to get information about the current media track and pass that to the video html component in the front end here https://github.com/aws-amplify/amplify-ui/blob/fix-undefined-media/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx#L103

reesscot (approved):

packages/react-liveness/src/components/FaceLivenessDetector/StartLiveness/helpers.ts
```diff
@@ -1,4 +1,4 @@
-export function getVideoConstraints(): MediaTrackConstraints | null {
+export function getVideoConstraints(): MediaTrackConstraints {
```
calebpollman, line 1: Why does this need to be a function if the object it returns is static?

thaddmt: Hm this is a leftover from before when we used to return different video constraints for mobile vs desktop. Should be safe to change it but wasn't sure if we should in this PR. What do you think?

calebpollman: It would definitely be a bigger lift, but it does seem like we could simplify `videoConstraints` usage be making it a constant and referencing it directly where needed rather than passing around as a param between functions. Does that make sense?

reesscot: I'm good with changing it here

thaddmt: Yep can make the change

calebpollman (approved):
🚢

dbanksdesign (approved):
LGTM!

ioanabrooks (approved):

PR #4325, opened 2023-08-04
fix(react): update 4.x version to use ^ version of aws-amplify

thaddmt:
#### Description of changes
- Patching 4.x version to use ^ for `aws-amplify` peer dependency
- Version packages PR which released `ui-react-@4.6.4` - https://github.com/aws-amplify/amplify-ui/pull/3973
- Actual commit - https://github.com/aws-amplify/amplify-ui/commit/52ece83b6b54c8ec24bad33c3ceed8b74e70822a

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -69,7 +69,7 @@
     "tslib": "2.4.1"
   },
   "peerDependencies": {
-    "aws-amplify": ">= 5.0.1",
+    "aws-amplify": "^5.0.1",
     "react": ">= 16.14.0",
     "react-dom": ">= 16.14.0"
   },
--- a/.changeset/giant-bugs-rush.md
+++ b/.changeset/giant-bugs-rush.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
+---
+
+fix(react): update 4.x version to use ^ version of aws-amplify


canary/package.json
```diff
@@ -8,9 +8,11 @@
     "pull": "./environments/pull-environments.sh",
     "script:install": "yarn --cwd script",
     "setup:next": "yarn script:install && yarn updatePackage:next",
+    "setup:reactv4": "yarn script:install && yarn updatePackage:reactv4",
     "setup:next-release": "yarn script:install && yarn updatePackage:next-release",
     "test": "yarn --cwd e2e test",
     "updatePackage:next": "node script/changeCanaryPackageVersion.mjs next",
+    "updatePackage:reactv4": "node script/changeCanaryPackageVersion.mjs react-v4-next",
```
calebpollman, line 15: Do we need this if the _canaries_ directory is gone?

packages/react/package.json
```diff
@@ -69,7 +69,7 @@
     "tslib": "2.4.1"
   },
   "peerDependencies": {
-    "aws-amplify": ">= 5.0.1",
+    "aws-amplify": "^5.0.1",
```
calebpollman, line 72: Just for my own knowledge, is there a reason we aren't doing all the platform packages?

calebpollman (commented):
LGTM, just left a couple questions

canary/package.json
thaddmt: I only should have removed `canary/apps/react/vite3`, but either way this code won't make it into our main branch :). It will only get pushed to `hotfix` and we will not be cherrypicking this to main because it wouldnt make sense to

packages/react/package.json
thaddmt: This patch is only targetting react v4 because the commit is the last known good commit for v4. We also only have the changeset for v4 for this reason. For other platform packages we want to upgrade we will need to find a separate commit and make a separate hotfix.

reesscot (approved):

packages/react/package.json
calebpollman: Got it ✅

canary/package.json
calebpollman: No worries, should have looked a bit closer. Thanks!

calebpollman (approved):

PR #4328, opened 2023-08-07
fix(e2e): switch liveness e2e tests to use y4m video file

thaddmt:
#### Description of changes
- Guidance from cypress is to switch to using y4m files - https://github.com/cypress-io/cypress/issues/27424#issuecomment-1667644297

#### Checklist

- [ ] PR description included


ioanabrooks (dismissed):

packages/e2e/cypress.config.ts
```diff
@@ -24,7 +24,7 @@ export default defineConfig({
       // This is a chrome launch option which enables fake videos
       on('before:browser:launch', (_browser, launchOptions) => {
         launchOptions.args.push(
-          '--use-file-for-fake-video-capture=cypress/fixtures/faceRecording.mjpeg'
```
calebpollman, line 27: Is there an associated _mjpeg_ file we should delete?

thaddmt: We could delete it, but for some reason I felt it might be useful to keep it in case we ever need to switch to it. 
We luckily happened to have the y4m file on my desktop but would just be annoying to have to create it again. What do you think?

calebpollman (approved):

ioanabrooks (approved):

PR #4359, opened 2023-08-17
[DRAFT/WIP]: DropZone component

dbanksdesign:
#### Description of changes

https://dropzone.d32sljejpnv3v5.amplifyapp.com/

#### Checklist

- [ ] PR description included


packages/react/src/primitives/DropZone/__tests__/useDropZone.test.ts
```diff
@@ -1,26 +1,40 @@
 import * as React from 'react';
 import { act, renderHook } from '@testing-library/react-hooks';
-
```
dbanksdesign, line 3: Moving these tests from storage package to react package and updating them

packages/react/src/primitives/DropZone/useDropZone.ts
```diff
@@ -0,0 +1,116 @@
+import { useState } from 'react';
+import { UseDropZoneProps, UseDropZoneReturn } from './types';
+import { isFunction } from '@aws-amplify/ui';
+
+type DragFile =
```
dbanksdesign, line 5: The objects the drag event get are different than the drop event 🤷

dbanksdesign (commented):
behold

.changeset/itchy-crabs-return.md
```diff
@@ -0,0 +1,28 @@
+---
+"@aws-amplify/ui-react-storage": minor
+"@aws-amplify/ui-react": minor
+---
```
hbuchel, line 4: Did Changeset bot not detect the ui package changes? These changes should cause a bump?

packages/react/src/primitives/DropZone/types.ts
hbuchel, line 1: The types not living in src/primitives/types is what is breaking the props table tab. We'll need to fix how we generate that data if we want to colocate the types with the primitive. Is that something we want to do? Is there a benefit to moving them? I know we do a kind of weird @deprecate tag currently to mark types internally. Is this to avoid that?

dbanksdesign: Yea this was to avoid that and also colocate the types with the component, but maybe I'll just put it where the other types are and do a cleanup all at once.

packages/react/src/primitives/DropZone/useDropZone.ts
```diff
@@ -0,0 +1,116 @@
...
+  return { accepted, rejected };
+}
+
+export function useDropZone({
+  onDropComplete,
+  acceptedFileTypes = [],
+}: UseDropZoneProps): UseDropZoneReturn {
+  const [isDragActive, setIsDragActive] = useState(false);
+  const [isDragAccept, setIsDragAccept] = useState(false);
+  const [isDragReject, setIsDragReject] = useState(false);
```
dbanksdesign, line 54: One question I have is should this just be a single enum state variable like `dragState` which has type `'default' | 'accept' | 'reject'` @hbuchel @calebpollman @reesscot

calebpollman: Definitely think we should go with a string union since the booleans conflict with eachother

docs/src/pages/[platform]/components/dropzone/react.mdx
Comment (hbuchel) [inline docs/src/pages/[platform]/components/dropzone/react.mdx:1]: Question: Do you think we can show an example of using this with an input[type="file"] or button like we do with StorageManager? Maybe as part of an Accessibility section (like how to combine this primitive with others to make a keyboard accessible + dropzone solution)? Not sure if it's out of scope for the docs for this piece, maybe that's better suited for a guide?

hbuchel (commented):
Couple more questions:

packages/react-storage/src/components/StorageManager/StorageManager.tsx
```diff
@@ -93,21 +94,24 @@ function StorageManagerBase(
   React.useImperativeHandle(ref, () => ({ clearFiles }));
 
   const dropZoneProps = useDropZone({
```
calebpollman, line 96: Can we just pass `onDropComplete` and `acceptedFileTypes` directly to the `DropZone`?

dbanksdesign: We can, but wouldn't that be a breaking change for the StorageManager DropZone component slot?

docs/src/pages/[platform]/components/dropzone/react.mdx
Comment (dbanksdesign) [reply docs/src/pages/[platform]/components/dropzone/react.mdx:1]: Done and done

reesscot (approved):
LGTM!

hbuchel (approved):
LGTM!

packages/react/src/primitives/DropZone/types.ts
```diff
@@ -0,0 +1,42 @@
...
+
+interface DragEvents {
+  onDragStart: (event: React.DragEvent<HTMLDivElement>) => void;
+  onDragEnter: (event: React.DragEvent<HTMLDivElement>) => void;
+  onDragLeave: (event: React.DragEvent<HTMLDivElement>) => void;
+  onDragOver: (event: React.DragEvent<HTMLDivElement>) => void;
+  onDrop: (event: React.DragEvent<HTMLDivElement>) => void;
+}
+
+export type DragState = 'accept' | 'reject' | 'none';
```
calebpollman, line 33: What do you think of `inactive` in place of `none`?

packages/react-storage/src/components/StorageManager/StorageManager.tsx
calebpollman: Potentially, but in the same way that changing the underlying DZ could also be considered a breaking change

reesscot (approved):

ioanabrooks (dismissed):

reesscot (approved):
LGTM

ioanabrooks (approved):

PR #4362, opened 2023-08-18
fix(docs): plain link styles

hbuchel:
#### Description of changes

Our styles for links on the docs site are overriding our updated Button primitive styles which broke our home page "Get started building" Button hover style.

Updated our link override styles on docs site to only target links with no classes (so basically links that are generated in markdown or plain <a></a> links used in documentation)

**After**

CleanShot 2023-08-18 at 11 43 16

#### Checklist

- [ ] PR description included

### DIFF
--- a/docs/src/styles/docs/base.scss
+++ b/docs/src/styles/docs/base.scss
@@ -79,7 +79,7 @@ body {
   line-height: 1.5;
 }
 
-a {
+a:not([class]) {
   color: var(--amplify-components-link-color);
 
   &:hover,
--- a/docs/src/styles/docs/base.scss
+++ b/docs/src/styles/docs/base.scss
@@ -79,6 +79,9 @@ body {
   line-height: 1.5;
 }
 
+/* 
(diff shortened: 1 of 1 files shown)


reesscot (dismissed):

docs/src/styles/docs/base.scss
```diff
@@ -79,7 +79,7 @@ body {
   line-height: 1.5;
 }
 
-a {
+a:not([class]) {
```
calebpollman, line 82: Do you think there is any value in adding a comment explaining what we are doing with the `not` pseudo selector here? Thinking it might be helpful since we are applying to all `a` elements

calebpollman (dismissed):
Just left a non-blocking question

docs/src/styles/docs/base.scss
hbuchel: Yeah, I agree. Added a comment in latest commit

calebpollman (approved):

reesscot (approved):
LGTM

PR #4371, opened 2023-08-23
chore(liveness): Custom websocket handler

thaddmt:
#### Description of changes
- Added a custom websocket fetch handler which can send custom closure codes received from the front end
- Hacked into the `VideoEvent` handler and added an object of `{code: number}` to send the closure code to the websocket handler
  - The websocket handler only can receives known events that are defined already. The only two pre-defined events are the `VideoEvent` and the `ClientSessionInformationEvent`
  - Other option would be for rekognition to add a new `ClosureEvent` which we can then listen for in the websocket fetch handler
- Ignored the file in tests/coverage since it is almost entirely copied from - https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-websocket/src/websocket-fetch-handler.ts

#### Checklist

- [ ] PR description included


packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
```diff
@@ -0,0 +1,267 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
```
thaddmt, line 1: This file is mostly copied from - https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-websocket/src/websocket-fetch-handler.ts#L176

The file was not typed so added a bunch of typescript linting ignores

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+        for await (const inputChunk of data) {
+          const decodedString = this.utf8decoder.decode(inputChunk);
+          if (decodedString.includes('closeCode')) {
+            const match = decodedString.match(/"closeCode":([0-9]*)/);
+            if (match) {
+              const closeCode = match[1];
+              socket.close(parseInt(closeCode));
+            }
+            continue;
+          }
```
thaddmt, line 247: This is the only new section of the file. It makes it so that if we find an event that contains a `closeCode` and a number, we send that number as the closure code for the websocket stream.

the events are received as Uint8Arrays that need to be decoded into strings

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
```diff
@@ -0,0 +1,267 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+/* eslint-disable @typescript-eslint/require-await */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-return */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/**
+ * @jest-environment node
+ */
```
calebpollman, line 9: Why do we need to declare `@jest-environment` for this file?

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
calebpollman: Get why this was done but it kind of negates the effort we have gone in to regarding type safety and linting in the rest of the React packages 😢 If we absolutely need to do this think there should be an explanation above so future maintainers understand this was an exception and should not be used as precedent

packages/react-liveness/jest.setup.ts
```diff
@@ -7,3 +8,8 @@ import '@testing-library/jest-dom';
 if (typeof window.URL.createObjectURL === 'undefined') {
   window.URL.createObjectURL = jest.fn();
 }
+
+Object.defineProperty(window, 'TextDecoder', {
+  writable: true,
+  value: TextDecoder,
+});
```
calebpollman, line 15: Do we need this for every test file? If not prefer declaration in the test file needed for unit test eff, _jest.setup.ts_ runs individually for each test file

calebpollman (commented):
Don't have concerns with the implementation itself, just left some questions around jest and linting

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
```diff
@@ -758,7 +759,22 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
         if (freshnessColorEl) {
           freshnessColorEl.style.display = 'none';
         }
-        await context.livenessStreamProvider?.endStream();
+
+        let closureCode = 4000;
```
ioanabrooks, line 763: Adding closureCode to `WS_CLOSURE_CODE` enum feels cleaner to me. Perhaps `WS_CLOSURE_CODE.DEFAULT`

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
ioanabrooks: +1 And can we have a followup task to work through the linting errors and remove the ignores?

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+        }
+      } catch (err) {
+        // We don't throw the error here because the send()'s returned
+        // would already be settled by the time sending chunk throws error.
+        // Instead, the notify the output stream to throw if there's
+        // exceptions
+        streamError = err as Error | undefined;
+      } finally {
+        // WS status code: https://tools.ietf.org/html/rfc6455#section-7.4
+        socket.close(1000);
```
ioanabrooks, line 259: Is it worth putting this hardcoded value in an enum along with other closure codes we might use?

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
thaddmt: Created a follow up task and added a comment

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
thaddmt: Accidentaly left this in here from testing, removed

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts
thaddmt: I was just trying to keep this as similar as possible to the original file so I didn't change it. I think with websockets it's well known that 1000 is the success code i guess?

ioanabrooks (dismissed):

packages/react-liveness/jest.config.js
```diff
@@ -6,6 +6,8 @@ module.exports = {
     '!<rootDir>/**/index.(ts|tsx)',
     // do not collect from top level version and styles files
     '!<rootDir>/src/(styles|version).(ts|tsx)',
+    // do not collect coverage from filed copies from aws-sdk
+    '!<rootDir>/src/components/FaceLivenessDetector/service/utils/CustomWebSocketFetchHandler.ts',
```
calebpollman, line 10: Thought abt this a bit and have to ask why we would choose to ignore collecting coverage on code we have copied from elsewhere. Hard not to see it as a liability, what do you think @thaddmt

reesscot: +1, are there tests we can copy over from the aws sdk to cover this code?

thaddmt: yep i believe you are right, we should be tracking this as we will own this code going forward

calebpollman (dismissed):
🚢  Thanks for adding the tests!

calebpollman (approved):

hbuchel (approved):

PR #4414, opened 2023-09-05
added new isMultiple and selectSize props

lavr001:
#### Description of changes

Added new props `isMultiple` and `selectSize` to the `SelectField` component

https://select-field-add-multiple-attr.dvmvffzts1tcu.amplifyapp.com/react/components/selectfield

#### Checklist

- [x] PR description included


docs/src/pages/[platform]/components/selectfield/react.mdx
```diff
@@ -175,6 +177,42 @@ A disabled field will not be focusable or mutable, and it will not be submitted
   </ExampleCode>
 </Example>
 
+### Multiple state
+
+The `multiple` attribute specifies that multiple options can be selected at once. If the `multiple` attribute is present, the default value for the `size` attribute is 4. Use the `isMultiple` prop to set the state for the SelectField.
```
Comment (hbuchel) [inline docs/src/pages/[platform]/components/selectfield/react.mdx:182]:
```suggestion
Use the `isMultiple` prop to specify that multiple options can be selected at once.
```
I think we can simplify this down.

docs/src/pages/[platform]/components/selectfield/react.mdx
```diff
@@ -175,6 +177,42 @@ A disabled field will not be focusable or mutable, and it will not be submitted
   </ExampleCode>
 </Example>
 
+### Multiple state
```
Comment (hbuchel) [inline docs/src/pages/[platform]/components/selectfield/react.mdx:180]:
```suggestion
### Multiple selections
```
nit: Maybe a different heading? Since we're not really doing anything with state.

packages/ui/src/theme/css/component/select.scss
```diff
@@ -61,4 +63,15 @@
       --amplify-components-select-large-padding-inline-end
     );
   }
+
+  &[multiple],
+  &[size] {
+    overflow: auto;
+    padding: 0.25rem 1rem;
```
hbuchel, line 70: Lets add tokens for these padding values. And the option padding below (we can sync on that if needed).

packages/ui/src/theme/css/component/select.scss
```diff
@@ -61,4 +63,15 @@
...
+
+    option {
+      color: var(--amplify-components-select-option-color);
```
hbuchel, line 73: Reminder for us to double check this color property to make sure it works in Firefox + windows/chrome since that has been fussy before.

hbuchel (commented):
Looking good! Let's make sure to fix the stray package.json and yarn.lock changes that snuck in. Some early feedback:

reesscot:
I don't think the `yarn.lock` file changes should be in here. Can you remove it and merge from main?

docs/src/pages/[platform]/components/selectfield/react.mdx
```diff
@@ -175,6 +177,41 @@ A disabled field will not be focusable or mutable, and it will not be submitted
...
+```jsx file=./examples/SelectFieldMultipleStateExample.tsx
+
+```
+
+  </ExampleCode>
+</Example>
+
+### Select size
+
+The `size` attribute sets the number of visible options in a drop-down list. Use the `selectSize` prop to set the `size` attribute value. If the value of the `selectSize` prop is greater than 1, but lower than the total number of options in the list, the browser will add a scroll bar to indicate that there are more options to view.
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/selectfield/react.mdx:200]: This is a bit confusing to me. Am I supposed to use `size` or `selectSize` to control the number of visible options in a drop-down list? I think what you are trying to say is that the HTML size attribute controls the number of visible options in a drop-down list, but it might be clearer not to talk about that at all and only mention the `selectSize` prop, since that's what users will be using for this.

package.json
```diff
@@ -95,6 +95,7 @@
   "devDependencies": {
     "@changesets/changelog-github": "0.4.1",
     "@changesets/cli": "^2.26.2",
+    "eslint-plugin-index": "^1.1.5",
```
ioanabrooks, line 98: What is the reason for introducing this dependency?

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -65,15 +69,17 @@ const SelectPrimitive: Primitive<SelectProps, 'select'> = (
...
-      <Flex
-        className={classNames(
-          ComponentClassNames.SelectIconWrapper,
-          classNameModifier(ComponentClassNames.SelectIconWrapper, size)
-        )}
-        color={iconColor}
-      >
-        {icon ?? icons?.expand ?? <IconExpandMore />}
-      </Flex>
+      {isMultiple || selectSize! > 1 ? null : (
```
ioanabrooks, line 72: Could we provide a default value for selectSize instead of relaxing the not null check? What happens here if selectSize is not defined?

calebpollman: +1. Using a default of `0` would make the most sense IMO

docs/src/pages/[platform]/components/selectfield/react.mdx
Comment (lavr001) [reply docs/src/pages/[platform]/components/selectfield/react.mdx]: Agree, removed `size` reference, so only `selectSize` is used. May be for a future improvements to avoid this confusion we can swap functionality between those props, so our custom prop size would have a native html attr size functionality.

package.json
lavr001: Not entirely sure how it's sneaked in, removed

reesscot:
@lavr001 You'll need to update snapshot tests for the ui package: https://github.com/aws-amplify/amplify-ui/actions/runs/6114903239/job/16597619003?pr=4414

```
yarn ui test -u
```

packages/react/src/primitives/Select/Select.tsx
lavr001: Here is the [docs] I was referencing to (https://www.w3schools.com/tags/att_size.asp), in the demo case our current placeholder is a part of the dropdown, that is shown when a `selectSize` is 1

calebpollman: So if i'm understanding correctly, to prevent a breaking change the default `selectSize` should be `1`, and we should show the placeholder when the `selectSize` is `1` or greater right?

lavr001: Yeap, that's my understanding of it as well

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -57,6 +59,8 @@ const SelectPrimitive: Primitive<SelectProps, 'select'> = (
             : defaultValue
         }
         isDisabled={isFieldsetDisabled ? isFieldsetDisabled : isDisabled}
+        multiple={isMultiple}
+        size={selectSize}
```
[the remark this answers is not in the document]
lavr001: Could you please specify what `view` file are you referring to? When I hover over that `size` prop I see that it expects a number, but may be I missed something.
Screenshot 2023-09-11 at 3 22 32 PM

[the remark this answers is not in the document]
calebpollman: Oh cool, was under the impression that all of our `size` properties mapped to `Sizes`

packages/react/src/primitives/Select/Select.tsx
```diff
@@ -57,6 +59,8 @@ const SelectPrimitive: Primitive<SelectProps, 'select'> = (
             : defaultValue
         }
         isDisabled={isFieldsetDisabled ? isFieldsetDisabled : isDisabled}
+        multiple={isMultiple}
+        size={selectSize || 1}
```
hbuchel, line 63:
```suggestion
        size={selectSize}
```

Since you added a default value you won't need this `||`

packages/react/src/primitives/Select/Select.tsx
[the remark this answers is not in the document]
hbuchel: FYI the `size` that we're passing here is an HTML attribute for `<select>`. For example:

```
<View size="small">I'm a div</View>
```

results in a type error because size doesn't exist on HTMLDivElement. But

```
<View as="select" size={2}>I'm a select</View>
```

works because size exists as a type on HTMLSelectElement.

Not to be confused with the `size` prop that exists on our *Field primitives which maps to Sizes ('small' | 'large') which unfortunately has the same name. That is why we're using `selectSize` as the name of the prop that our SelectField accepts to set this attribute.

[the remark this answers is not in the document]
calebpollman: Part of me feels like we are making usage DX confusing by using HTML attr names in a non-semantic way but discussed with @dbanksdesign and feel okay abt `selectSize`

.changeset/brave-timers-dream.md
```diff
@@ -0,0 +1,7 @@
+---
+"@aws-amplify/ui-docs": patch
```
hbuchel, line 2:
```suggestion
```
We should delete docs from this when we generate a changeset

.changeset/brave-timers-dream.md
```diff
@@ -0,0 +1,7 @@
...
+"@aws-amplify/ui-react": patch
+"@aws-amplify/ui": patch
+---
+
+added new isMultiple and selectSize props
```
hbuchel, line 7: Can we add an example usage and call out this is for SelectField?

hbuchel (commented):
Some nits for changeset:

.changeset/brave-timers-dream.md
lavr001: Good call, updated

hbuchel (approved):
LGTM! 🎉  Thank you for all the updates!

calebpollman (approved):

PR #4416, opened 2023-09-05
chore(ui): Update amplify dependencies and add no-op modules 

ioanabrooks:
#### Description of changes

* Updated all aws-amplify dependencies to `console-preview` tag
* Added no-op types and re-exported amplify modules to get all packages to build
* replaced Auth.changePassword with v6 counterpart
* updated Logger imports

Ran build locally.

#### Checklist

- [ ] PR description included


packages/ui/src/helpers/accountSettings/validator.ts
```diff
@@ -11,7 +11,7 @@ import {
 // gets password requirement from Amplify.configure data
 export const getPasswordRequirement = (): PasswordRequirement | null => {
   // need to cast to any because `Amplify.configure()` isn't typed properly
-  const config = Amplify.configure() as any;
+  const config = Amplify.configure({}) as any;
```
ioanabrooks, line 14:
```suggestion
  const config = Amplify.configure() as any;
```

dbanksdesign (commented):
Just one small question

canary/apps/angular/angularcli/package.json
```diff
@@ -19,7 +19,7 @@
     "@angular/platform-browser-dynamic": "^14.3.0",
     "@angular/router": "^14.3.0",
     "@aws-amplify/ui-angular": "latest",
-    "aws-amplify": "latest",
+    "aws-amplify": "console-preview",
```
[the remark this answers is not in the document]
ioanabrooks: `^6.0.0` doesn't resolve, we only have the custom tag to work with for now. It's not hard to change the deps, the canary apps specifically should be restored to latest before releasing anyway.

packages/react-liveness/package.json
```diff
@@ -43,7 +43,7 @@
     "typecheck": "tsc --noEmit"
   },
   "peerDependencies": {
-    "aws-amplify": "^5.0.1",
+    "aws-amplify": "console-preview",
```
thaddmt, line 46: Can we keep this pinned to v5? We want to test and see if we can keep liveness on v5

packages/ui/src/types/authenticator/stateMachine/authMachine.ts
```diff
@@ -30,7 +30,7 @@ export type AuthMachineSend = AuthInterpreter['send'];
  * xstate events.
  */
 export type AuthMachineHubHandler = (
-  data: HubCapsule,
+  data: HubCapsule<any, any>, //TODO: add types
```
reesscot, line 33: Do we need a task to come back to this?

reesscot (approved):

packages/configs/ts/tsconfig.base.json
```diff
@@ -2,28 +2,28 @@
   "compilerOptions": {
     // https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports
```
calebpollman, line 3: Why are we removing these?

ioanabrooks: We are not, not sure why they're highlighted with dark red base but they are still there.

packages/ui/src/types/authenticator/stateMachine/authMachine.ts
ioanabrooks: Added!

packages/react-liveness/package.json
ioanabrooks: don't think we can have 2 versions of `aws-amplify` without conflicts...

canary/apps/angular/angularcli/package.json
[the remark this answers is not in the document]
dbanksdesign: Ah its because they aren't publishing valid semvers for 6.0, they are publishing versions like `6.0.1-console-preview.f63250a.0` can we see if they can just publish `6.0.1` to the `console-preview` tag?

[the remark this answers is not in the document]
ioanabrooks: So...ask them to publish v6 early? :)

packages/configs/ts/tsconfig.base.json
reesscot: I think github highlights them since comments aren't allowed in json?

canary/apps/angular/angularcli/package.json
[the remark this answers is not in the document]
reesscot: I think we should ask if they will be publishing to 6.0 for dev preview

packages/react-liveness/package.json
reesscot: @thaddmt Can you try this in a follow up pr?

reesscot (approved):

calebpollman (approved):

PR #4424, opened 2023-09-07
chore: Update Logger imports

ioanabrooks:
#### Description of changes

Updated all imports to point to logger wrapper util in the ui package, that way regardless of what logging solution we use in the future we have to make changes in only one place.

Built packages locally

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/ui/src/helpers/logger/logger.ts
+++ b/packages/ui/src/helpers/logger/logger.ts
@@ -1,0 +1,62 @@
+type LoggerCategory =
+  | 'AccountSettings'
+  | 'Auth'
+  | 'Geo'
+  | 'Icon'
+  | 'Notifications'
+  | 'StateMachine'
+  | 'StorageImage'
+  | 'StorageManager';
+
+export enum LOG_LEVEL {
+  DEBUG = 'DEBUG',
+  ERROR = 'ERROR',
+  INFO = 'INFO',
+  WARN = 'WARN',
+}
+
+class Logger {
+  private name: string;
+
+  constructor(name: string) {
+    this.name = name;
+  }
+
+  private log(level: LOG_LEVEL, message: string, error?: Error): void {
+    switch (level) {
+      case LOG_LEVEL.INFO:
+        console.info(`[${level}] - ${this.name} ${message}`);
+        break;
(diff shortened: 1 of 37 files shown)


packages/ui/src/helpers/logger/logger.ts
```diff
@@ -0,0 +1,62 @@
...
+  | 'StorageManager';
+
+export enum LOG_LEVEL {
+  DEBUG = 'DEBUG',
+  ERROR = 'ERROR',
+  INFO = 'INFO',
+  WARN = 'WARN',
+}
+
+class Logger {
```
ioanabrooks, line 18: Simplistic version of a Logger to cover our existing usecases, open to add more if needed.

thaddmt:
Not sure we want to update this yet if all that is changing is where it is imported from. Open to discussion though

ioanabrooks:
> Not sure we want to update this yet if all that is changing is where it is imported from. Open to discussion though

Providing our own implementation aside, having a wrapper around it is good bc if the import changes in the future, we only have to modify in one place.

packages/ui/src/helpers/logger/logger.ts
```diff
@@ -0,0 +1,14 @@
...
+  | 'AccountSettings'
+  | 'Auth'
+  | 'Geo'
+  | 'Icon'
+  | 'Notifications'
+  | 'StateMachine'
+  | 'StorageImage'
+  | 'StorageManager';
+
+export const getLogger = (category: LoggerCategory) =>
```
reesscot, line 13: nit: could be `getUILogger` to indicate that it's prefixing with `AmplifyUI:`?
```suggestion
export const getLogger = (category: LoggerCategory) =>
```

ioanabrooks: Is that necessary? It's in the ui package, not seeing a future where we would have different or multiple prefixes.

reesscot: Yeah, it was a loosely held nit. I'm good with getLogger

calebpollman (approved):

PR #4433, opened 2023-09-11
feat(A.Next): add machine context

calebpollman:
#### Description of changes
Add `MachineProvider` and `useMachine` and state machine utils for `A.Next`:
- `MachineProvider` and `useMachine` correspond to existing `AuthenticatorProvider` and `useAuthenticator` with only required functionality
- add `overrideServices` to `createAuthenticatorMachine` to allow `A.Next` to bypass extraneous configuration updates

#### Issue #, if available
NA

Manual tests, some unit tests
#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-core-auth/src/components/Authenticator/Machine/MachineContext.tsx
+++ b/packages/react-core-auth/src/components/Authenticator/Machine/MachineContext.tsx
@@ -1,0 +1,53 @@
+import React, { useCallback } from 'react';
+import { useInterpret, useSelector } from '@xstate/react';
+import {
+  AuthenticatorMachineOptions,
+  AuthMachineState,
+  getNextServiceFacade,
+  createAuthenticatorMachine,
+} from '@aws-amplify/ui';
+
+import { MachineContextType, UseMachineSelector, UseMachine } from './types';
+import { defaultComparator, getComparator } from './utils';
+
+export const USE_MACHINE_ERROR =
+  '`useMachine` must be used inside an `Authenticator.Provider`.';
+
+const MachineContext = React.createContext<MachineContextType | null>(null);
+
+export function useMachine(selector?: UseMachineSelector): UseMachine {
+  const context = React.useContext(MachineContext);
+
+  if (!context) {
+    throw new Error(USE_MACHINE_ERROR);
+  }
+
+  const { service } = context;
+  const { send } = service;
+
+  const xstateSelector = useCallback(
+    (state: AuthMachineState) => getNextServiceFacade({ send, state }),
+    [send]
+  );
+
+  const comparator = selector ? getComparator(selector) : defaultComparator;
+  const facade = useSelector(service, xstateSelector, comparator);
+
+  return facade;
+}
+
+interface MachineProviderProps extends AuthenticatorMachineOptions {
+  children?: React.ReactNode;
+}
+
+export function MachineProvider({
+  children,
+  ...data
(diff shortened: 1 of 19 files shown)


packages/react-core-auth/src/components/Authenticator/Machine/MachineContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+  AuthMachineState,
+  getNextServiceFacade,
+  createAuthenticatorMachine,
+} from '@aws-amplify/ui';
+
+import { MachineContextType, UseMachineSelector, UseMachine } from './types';
+import { defaultComparator, getComparator } from './utils';
+
+export const USE_MACHINE_ERROR =
+  '`useMachine` must be used inside an `Authenticator.Provider`.';
```
reesscot, line 14: It shouldn't say to use it inside of MachineProvider?

packages/react-core-auth/src/components/Authenticator/Machine/MachineContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const xstateSelector = useCallback(
+    (state: AuthMachineState) => getNextServiceFacade({ send, state }),
+    [send]
+  );
+
+  const comparator = selector ? getComparator(selector) : defaultComparator;
+  const facade = useSelector(service, xstateSelector, comparator);
+
+  return facade;
```
reesscot, line 36: Now you're just showing off ;)

reesscot (approved):
LGTM

packages/react-core-auth/src/components/Authenticator/Machine/MachineContext.tsx
calebpollman: This is intentional, the customer would still be using `Authenticator.Provider` (can add a not explaining in a follow up)

packages/react-core-auth/src/components/Authenticator/Machine/MachineContext.tsx
calebpollman: Tbh we might be able to just return the `context` now since:
- 1) way less interactions are happening with the machine 
- 2) the machine data used is mostly javascript primitives that wouldn't cause re-renders unless they changed irregardless of a selector

packages/ui/src/types/authenticator/attributes.ts
```diff
@@ -45,6 +45,11 @@ export const LoginMechanismArray = [
   'phone_number',
 ] as const;
 
+/**
+ * Default supported federated sign sn providers
```
ioanabrooks, line 49:
```suggestion
 * Default supported federated sign in providers
```

packages/react-core-auth/src/components/Authenticator/Machine/MachineContext.tsx
ioanabrooks, line 36: Can we just return ` useSelector(service, xstateSelector, comparator);` ? What purpose does `facade` variable serve here?

calebpollman: No particular reason, just didn't update it when it was migrated from:
```
  const comparator = selector ? getComparator(selector) : defaultComparator;

  const facade = useSelector(service, xstateSelector, comparator);

  return { ...facade, authStatus };
```

ioanabrooks (approved):

PR #4438, opened 2023-09-13
fix(ui/react): option color/backgroundColor, style updates for [multiple][size]

hbuchel:
#### Description of changes

- Fixes the broken options on Windows Chrome. Chrome can now have the same styles we originally scoped to Firefox (which means we can remove the Firefox scope).
- Fixes the weird border that is currently in Firefox darkmode on open select elements.
- Also updates some of the theming for disabled options and disabled select
- Ensures these updates work with the new size/multiple functionality we're adding support for.

**Windows chrome issue (before)**
Screenshot (815)

**Windows chrome issue (after)**
Screenshot (816)

**Windows Firefox issue (before)**
Screenshot (813)

**Windows Firefox issue (after)**
Screenshot (814)

Tested in Firefox, Chrome, Edge on Windows.
Tested in Firefox, Chrome, Safari on Mac.

https://selectcssfixes.dvmvffzts1tcu.amplifyapp.com/react/components/selectfield

#### Checklist

- [ ] PR description included


packages/react/src/primitives/Select/Select.tsx
```diff
@@ -35,12 +35,16 @@ const SelectPrimitive: Primitive<SelectProps, 'select'> = (
...
+
+  const isExpanded = isMultiple || selectSize > 1;
+
   const componentClasses = classNames(
     ComponentClassNames.Select,
     ComponentClassNames.FieldGroupControl,
     classNameModifier(ComponentClassNames.Select, size),
     classNameModifier(ComponentClassNames.Select, variation),
     classNameModifierByFlag(ComponentClassNames.Select, 'error', hasError),
+    classNameModifierByFlag(ComponentClassNames.Select, 'expanded', isExpanded),
```
hbuchel, line 47: We were styling the "expanded" view using attribute selectors (`.amplify-selector[size]`, `.amplify-selector[multiple]`) but we don't want just any [size] attribute, we want it when it is greater than 1. Since we can't do those comparisons in CSS we'll use the `isExpanded` flag we defined earlier to create a class.

packages/ui/src/theme/tokens/components/select.ts
```diff
@@ -24,20 +28,27 @@ export type SelectTokens<Output extends OutputVariantKey> =
...
       Output
     > & {
       option?: DesignTokenProperties<'paddingBlock' | 'paddingInline', Output>;
     };
+    _disabled?: DesignTokenProperties<'backgroundColor' | 'color', Output>;
   };
 
 export const select: Required<SelectTokens<'default'>> = {
+  color: { value: '{components.fieldcontrol.color}' },
+  backgroundColor: { value: '{colors.background.primary.value}' },
```
hbuchel, line 46: Firefox on Windows plays nicer with option colors/background when a background is set on the select itself. Otherwise, you get these top and bottom "borders" (screenshot below). So I added new backgroundColor/color tokens for select. I've set the existing backgroundColor tokens for option to be transparent. I added a color token to pair it with to make sure we/developers can always provide a sufficient contrast.

Screenshot (813)

packages/ui/src/theme/css/component/select.scss
hbuchel, line 1: Seems like Chrome on Windows now behaves like Firefox on windows where both background color and color can be used to style options, so the previous firefox only CSS can now be the default option styling, with the addition of disabled styling so both a disabled select AND a disabled option can have appropriate styling.

hbuchel (commented):
Tour of changes:

packages/ui/src/theme/tokens/components/select.ts
```diff
@@ -59,12 +70,18 @@ export const select: Required<SelectTokens<'default'>> = {
       right: { value: '{space.medium.value}' },
     },
   },
-  // for Firefox only, to fix background color in darkmode
+  // It's important to test these option values on
+  // Chrome/FireFox/Edge on Windows and Firefox on
+  // Mac because they allow styling of the option element.
+  // Chrome/Safari on Mac uses the system ui.
```
hbuchel, line 76:
```suggestion
  // It's important to test these option values on Chrome/FireFox/Edge 
  // on Windows because they allow styling of the option element. 
  // Chrome/Safari/Firefox on Mac uses the system ui.
```

packages/ui/src/theme/css/component/select.scss
```diff
@@ -27,26 +27,30 @@
...
+      --amplify-components-select-disabled-background-color
+    );
+    color: var(--amplify-components-select-disabled-color);
     option {
-      background-color: var(
-        --amplify-components-select-option-background-color
-      );
-      color: var(--amplify-components-select-option-color);
+      color: inherit;
+      background-color: inherit;
```
hbuchel, line 53: Options that are part of a disabled select should inherit the selects styling

lavr001 (approved):
LGTM

Issue #4429, opened 2023-09-11
`LivenessError`, `LivenessErrorState` are not exported properly/usable


bradjones1:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React, React Native

### Which UI component?

Liveness

### How is your app built?

Expo web (Webpack 5)

### What browsers are you seeing the problem on?

Firefox, Android (React Native)

### Please describe your bug.

The change in https://github.com/aws-amplify/amplify-ui/pull/4134 to make better use of `LivenessErrorState` in the `onError` hook is a welcome change, however these enums are not exported in such a way that Webpack-built code can utilize them. This is because these types are not exported via the path in the `package.json`'s `exports`. I _think_ this is basically the issue described in https://github.com/webpack/webpack/issues/13865.

### What's the expected behaviour?

I can use the types `LivenessError` and `LivenessErrorState` directly from the library. Currently I have to copy these into my own code which is needless coupling and is not maintainable.

### Help us reproduce the bug!

Attempt to import these types dynamically from the library, you will get

```
Unable to resolve module @aws-amplify/ui-react-liveness/dist/types/components/FaceLivenessDetector/service
```

bradjones1:
> or changing these enums to be strings going forward.

In that case I'd still have to maintain like, my own list of them to match against, so I think exporting the Enum is better. They get flattened to strings by TypeScript anyway. Thanks.

PR #4450, opened 2023-09-15
chore(liveness): export error types

thaddmt:
#### Description of changes
- fixes https://github.com/aws-amplify/amplify-ui/issues/4429

#### Checklist

- [ ] PR description included


reesscot (dismissed):
LGTM

packages/react-liveness/src/components/FaceLivenessDetector/index.ts
```diff
@@ -10,4 +10,5 @@ export {
   AwsCredentialProvider,
   AwsTemporaryCredentials,
   AwsCredentials,
+  LivenessErrorState,
```
reesscot, line 13: What happens if we change this to a union type before exporting? Is that a breaking change?

ioanabrooks (approved):

calebpollman:
Unclear to me if exposing the error types via enum is the correct path here. Uasge of a string union is easier to handle for consumers IMO. Consuming functions handling the errors have to specify every key of the enum whereas a string union allows for a cleaner function signature and implementation:
```
// string union
const myErrorHandler = (errorKey: LivenessError) => {
  switch(errorKey) {
     case 'error_one': {
        // do error one stuff
     }
    case 'error_two': {
        // do error one stuff
     }
}

// enum
const myErrorHandler = (errorKey: LivenessError.error_one | LivenessError.error_two) => {
  switch(LivenessError.error_one) {
     case 'error_one': {
        // do error one stuff
     }
    case 'LivenessError.error_one': {
        // do error one stuff
     }
}
```

packages/react-liveness/src/components/FaceLivenessDetector/index.ts
calebpollman: Think regardless of whether we move forward with the union the namespace should be reconsidered

thaddmt: What namespace alternatives are you thinking of?

I think in terms of exporting it it wouldn't be breaking to my understanding since we don't export it at this time. What do you all think?

calebpollman: This just `LivenessError` would suffice, and changing the keys from SCREAMING_SNAKE_CASE to camelCase, also maybe dropping the extraneous usage of ERROR in the keys, example:

```ts
enum LivenessErrorState = {
  MOBILE_LANDSCAPE_ERROR = 'MOBILE_LANDSCAPE_ERROR',
  MULTIPLE_FACES_ERROR = 'MULTIPLE_FACES_ERROR',
}

type LivenessError = 'mobileLandscape' | 'multipleFaces'
```

thaddmt:
@calebpollman @reesscot as I was thinking about this more, I think scotts point of keeping the enum but exporting a type that is a union of all the enums seems to make sense and cover all cases. In the latest PR update I exported `LivenessErrorState` the enum and `ErrorState` union type (names are totally up for debate just couldn't think of good ones).

This solves the scenario brought up by Caleb where we can write a util function like `const myErrorHandler = (errorKey: LivenessError) => {` but also enables the scenario where developers can choose to do different behaviors on different error types instead of maintaining their own list of error strings. If we only exported the type they would have to maintain their own list of error strings which is what the original issue was complaining about. What do you all think?

calebpollman:
> @calebpollman @reesscot as I was thinking about this more, I think scotts point of keeping the enum but exporting a type that is a union of all the enums seems to make sense and cover all cases. In the latest PR update I exported `LivenessErrorState` the enum and `ErrorState` union type (names are totally up for debate just couldn't think of good ones).
> 
> This solves the scenario brought up by Caleb where we can write a util function like `const myErrorHandler = (errorKey: LivenessError) => {` but also enables the scenario where developers can choose to do different behaviors on different error types instead of maintaining their own list of error strings. If we only exported the type they would have to maintain their own list of error strings which is what the original issue was complaining about. What do you all think?

The union should cover both cases, might just be unclear to me what utility an enum provides in this scenario that is not handled by the union:
CleanShot 2023-09-19 at 11 59 46@2x

thaddmt:
> > @calebpollman @reesscot as I was thinking about this more, I think scotts point of keeping the enum but exporting a type that is a union of all the enums seems to make sense and cover all cases. In the latest PR update I exported `LivenessErrorState` the enum and `ErrorState` union type (names are totally up for debate just couldn't think of good ones).
> > This solves the scenario brought up by Caleb where we can write a util function like `const myErrorHandler = (errorKey: LivenessError) => {` but also enables the scenario where developers can choose to do different behaviors on different error types instead of maintaining their own list of error strings. If we only exported the type they would have to maintain their own list of error strings which is what the original issue was complaining about. What do you all think?
> 
> The union should cover both cases, might just be unclear to me what utility an enum provides in this scenario that is not handled by the union: CleanShot 2023-09-19 at 11 59 46@2x

I think the one thing this doesn't address is the main callout in the issue https://github.com/aws-amplify/amplify-ui/issues/4429

In your example union case as well customers need to maintain those lists of strings for `ERROR_ONE` and `ERROR_TWO` whereas if we export the utility they can use the one we provide.

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
```diff
@@ -459,7 +460,7 @@ export function isCameraDeviceVirtual(device: MediaDeviceInfo): boolean {
   return device.label.toLowerCase().includes('virtual');
 }
 
-export const LivenessErrorStateStringMap: Record<LivenessErrorState, string> = {
+export const LivenessErrorStateStringMap: Record<string, string> = {
```
calebpollman, line 463: Think we can remove this annotation now
```suggestion
export const LivenessErrorStateStringMap = {
```

calebpollman (approved):

calebpollman (approved):

reesscot (approved):

ioanabrooks (approved):

PR #4476, opened 2023-09-25
breaking: Expander -> Accordion

dbanksdesign:
#### Description of changes

`Expander` now becomes `Accordion` and `ExpanderItem` becomes `Accordion.Item`. Removing radix accordion and using `<details>` and `<summary>` elements with a bit of context logic sprinkled in to make it work like an accordion. 

https://next-release-expander.d32sljejpnv3v5.amplifyapp.com/react/components/accordion

#### Checklist

- [x] PR description included


reesscot:
What should we do about customers who navigate/search for today's Expander? should we put in a redirect, or add a note that it was previously called Expander so it's searchable?

yarn.lock
```diff
@@ -7107,21 +7107,6 @@
   dependencies:
     "@babel/runtime" "^7.13.10"
 
-"@radix-ui/react-accordion@1.0.0":
```
reesscot, line 7110: 🙌

packages/react/src/primitives/Accordion/types.ts
```diff
@@ -0,0 +1,55 @@
+import { ElementType, PrimitiveProps, BaseViewProps } from '../types/view';
+
+export interface BaseAccordionProps extends BaseViewProps {
```
reesscot, line 3: While you're here, aren't the Base* props the ones we don't want to export?

packages/react/src/primitives/Accordion/types.ts
```diff
@@ -0,0 +1,55 @@
...
+   * @description
+   * Event handler called when the expanded state of an item changes
+   */
+  onChange?: (value?: string[]) => void;
+}
+
+export type AccordionProps<Element extends ElementType = 'div'> =
+  PrimitiveProps<BaseAccordionProps, Element>;
+
+export interface BaseAccordionItemProps extends BaseViewProps {
```
reesscot, line 40: Same comment as above, I don't think we meant to export the Base* types

packages/react/src/primitives/Accordion/types.ts
dbanksdesign: Yup, but I just export them from this file. They aren't publicly exported.

docs/src/pages/[platform]/components/accordion/index.page.mdx
```diff
@@ -2,8 +2,9 @@
 title: Accordion
 description: The Accordion allows users to expand or collapse a set of sections.
 ariaPattern: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/
-themeSource: packages/ui/src/theme/tokens/components/accordion.ts
-reactSource: packages/react/src/primitives/Accordion/Accordion.tsx
+# TODO: uncomment after merge
+# themeSource: packages/ui/src/theme/tokens/components/accordion.ts
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/accordion/index.page.mdx:6]: We should make the link checker only check internal links on PRs.

hbuchel (commented):
Left some suggestions! Maybe this could be a later addition, but it might be nice to have a "quiet" variation; something like this:
Screenshot 2023-10-03 at 12 17 06 PM

packages/react/src/primitives/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { AccordionContext } from './AccordionContext';
+import { AccordionItem } from './AccordionItem';
+
+const AccordionPrimitive: Primitive<AccordionProps, 'div'> = (
+  {
+    children,
+    className,
+    defaultValue,
+    allowMultiple,
+    allowToggle,
```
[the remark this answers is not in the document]
dbanksdesign: Yea I was thinking that, I kinda wanted the boolean props to be false by default so you can just add them if you want them rather than turn them off, but didn't know what the opposite prop name for that would be...

hbuchel:
Re-reviewing this, is the demo working? It looks like the allowMultiple and isAlwaysOpen checkboxes don't do anything?

reesscot (approved):

packages/react/src/primitives/Accordion/types.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * @description
+   * The controlled value of the item(s) to expand. Must be used in conjunction with onValueChange.
+   */
+  value?: string[];
+
+  /**
+   * @description
+   * Determines whether one or multiple items can be opened at the same time.
```
calebpollman, line 19: IMO `Determines` reads as if `allowMultiple` is something that needs to be resolved

```suggestion
   * Allow multiple items to be expanded at the same time.
```

packages/react/src/primitives/Accordion/types.ts
```diff
@@ -0,0 +1,60 @@
...
+   * Default: false
+   */
+  allowMultiple?: boolean;
+
+  /**
+   * @description
+   * Determines whether the opened item can always be collapsed even if its the only one.
```
calebpollman, line 26:
```suggestion
   * Prevent collapse when a single item is expanded.

File: packages/react/src/primitives/Accordion/types.ts [hunk 8]
```diff
@@ -0,0 +1,60 @@
...
+   * Default: false
+   */
+  isAlwaysOpen?: boolean;
```
calebpollman, line 29: Tied to the suggestion above on line 26 - Would this property be more descriptive of its use case as `preventCollapse`?

packages/react/src/primitives/Accordion/types.ts
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { ElementType, PrimitiveProps, BaseViewProps } from '../types/view';
+
+export interface BaseAccordionProps extends BaseViewProps {
+  /**
+   * @description
+   * The value of the item(s) to expand.  Use on uncontrolled component.
+   */
+  defaultValue?: string[];
```
calebpollman, line 9: While this would increase complexity in the implementation, a DX improvement could be to conditionally type `defaultValue`/`value` based on whether `allowMultiple` is `true`. IMO the array is extraneous in the default use case

calebpollman (commented):
Great work, let me know your thoughts on the feedback

packages/react/src/primitives/Accordion/AccordionContainer.tsx
```diff
@@ -0,0 +1,96 @@
...
+       * the item, BUT if the Accordion only allows 1 item to be open at a time then
+       * replace the array with an array of only the item value.
+       */
+      const newValue = value.includes(_value)
+        ? !isAlwaysOpen || value.length > 1
+          ? value.filter((v) => v !== _value)
+          : value
+        : allowMultiple
+        ? [...value, _value]
+        : [_value];
```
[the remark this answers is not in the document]
dbanksdesign: But I made a very descriptive commenttttt

packages/react/src/primitives/Expander/types.ts
```diff
@@ -26,27 +23,19 @@ export interface BaseExpanderProps extends BaseViewProps {
    * @description
    * Determines whether one or multiple items can be opened at the same time.
    */
-  type?: ExpanderType;
-
-  /**
-   * @deprecated replace usage with `onValueChange`
-   */
-  onChange?: (value?: string | string[]) => void;
+  multiple?: boolean;
```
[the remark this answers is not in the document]
dbanksdesign: wtf why is Expander even in here

reesscot (dismissed):

packages/react/src/primitives/Accordion/AccordionContainer.tsx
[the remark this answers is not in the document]
dbanksdesign: Ok I cleaned it up so it is readable now

calebpollman (approved):
🚢

calebpollman (approved):

PR #4488, opened 2023-09-27
feat: git secrets as pre commit hook

esauerbo:
#### Description of changes

Added AWSLabs git secrets as a pre-commit, commit-msg, and prepare-commit-msg hook with husky.
Checks if git secrets is installed, and if not prompts user to install
Automatically registers git-secrets with aws 
Scans committed files, commit message, and commit history for secrets 

Tested by adding fake secrets to committed files/messages

#### Checklist

- [ ] PR description included

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,22 +1,6 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-# Check git-secrets installation
-if ! command -v git-secrets &> /dev/null
-then
-    echo "git-secrets is not installed. Please run 'brew install git-secrets' or visit https://github.com/awslabs/git-secrets#installing-git-secrets"
-    exit 1
-fi
-
-echo "commit-msg"
-# Initialize git-secrets configuration
-git secrets --register-aws > /dev/null
-# git secrets --add '[aA]pollo|[bB]razil|[cC]oral|[oO]din'
-# git secrets --add 'tt\.amazon\.com|issues\.amazon\.com|cr\.amazon\.com'
-# git secrets --add 'ic\.gov|sgov\.gov'
-# git secrets --add 'us-iso|aws-iso'
-# git secrets --add 'smil\.mil'
-
 echo "Scanning commit message for secrets..."
 # Scans the commit message.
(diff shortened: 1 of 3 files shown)


.husky/.secret.patterns
```diff
@@ -0,0 +1,6 @@
+smil\.mil
+(us|aws)-iso
+[aA]pollo|[bB]razil|[cC]oral|[oO]din
+tt\.amazon\.com|issues\.amazon\.com|cr\.amazon\.com
+ic\.gov|sgov\.gov
+(us|aws)-iso
```
ioanabrooks, line 6: Is this file needed? Some of these are internal and we shouldn't push in a public repo imo.

esauerbo: It's needed to check for those patterns, so if we don't need to worry about those then we could omit it. If we still want to check for those we could make adding this file a manual configuration step (but that sort of defeats the purpose). 
 
For the hook to look for those patterns without a configuration step they need to be somewhere, maybe if there's a secure place we could host it somewhere? Idk if these particular patterns are a different case but other patterns are listed on the awslabs public repo though https://github.com/awslabs/git-secrets#options-for-add-provider

ioanabrooks: I would assume we wouldn't have to deal with lines 3 and 4 in any way?

.husky/.secret.patterns
```diff
@@ -0,0 +1,3 @@
+smil\.mil
+(us|aws)-iso
+ic\.gov|sgov\.gov
```
calebpollman, line 3: Can you explain the use cases of each of these please? For context, no where in our codebase have we had a _.gov_ extension in the past nor do i foresee one popping up in the future

esauerbo: Yeah was discussed over slack and opted to get rid of them
It was included in opensource training which is why I had them in the first place

ioanabrooks (approved):

reesscot:
Testing this out, I'm finding it's taking almost 30seconds to scan on pre-commit check. We may want to consider running this pre-push instead of pre-commit, since pushing secrets to the repo is the real issue.

PR #4502, opened 2023-10-02
.

sowmya-kota07:


hbuchel (commented):
Hi @sowmyadev-kota! Thanks for opening this! In your PR, you've updated the snapshot file, but what really needs to be updated is the file where we define our type for inert and the description for it. [types/base.ts](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/primitives/types/base.ts) should be the right file.

After that, you can run `yarn docs test -u` This will then update the snapshot file based on your change.

Also a small nit, previously we didn't do a good job naming these links which results in a bunch of duplicate links labeled "MDN" which isn't great for accessibility. Can we make the link text say "Inert on MDN"?

sowmyadev-kota:
@hbuchel , sorry I didn't get a chance to test this, will make the update, thank you

calebpollman:
@sowmyadev-kota Is this PR still active?

calebpollman:
@sowmyadev-kota Going to close this for now but please re-open if still relevant

PR #4506, opened 2023-10-03
fix(liveness): use button/aria roles for photosensitivity warning

esauerbo:
#### Description of changes

- Changed 'i' toggle to button, making it keyboard navigable 
- Added `dialog` role to the popover content
- Popover content has `aria-hidden` set to `true` or `false` depending on whether it's open or closed
- Popover content has photosensitivity warning `aria-label`
- Button has an `aria-label` noting its purpose
- Button has `aria-haspopup` set to `dialog`
- Button has `aria-expanded` set to `true` or `false` depending on whether content is open or closed
- Button has `aria-controls` set to the content id (`popover-text`)

Visual inspection of accessibility settings on chrome
Used keyboard to navigate to and open/close popup

#### Checklist

- [ ] PR description included


.changeset/loud-dodos-crash.md
```diff
@@ -0,0 +1,10 @@
+---
+"@aws-amplify/ui-react-liveness": patch
+"@aws-amplify/ui-react": patch
```
thaddmt, line 3: Think this package can be removed here

thaddmt (approved):
LGTM

.changeset/loud-dodos-crash.md
```diff
@@ -0,0 +1,10 @@
...
+"@aws-amplify/ui": patch
+---
+
+breaking(liveness):
```
reesscot, line 7:
```suggestion
fix (liveness): 
```

.changeset/loud-dodos-crash.md
calebpollman: Think it should be there since the UI package was updated?

hbuchel: I think ui-react will get auto bumped with one of those "Updated dependencies" notes in the changelog since ui is bumped, so it doesn't need to explicitly be added here if it wasn't touched in the changed files.

packages/react-liveness/src/components/FaceLivenessDetector/shared/LivenessIconWithPopover.tsx
```diff
@@ -31,32 +32,44 @@ export const LivenessIconWithPopover: React.FC<LivenessIconWithPopoverProps> =
...
-              left={isMobileScreen ? -190 : -108}
-              data-testid="popover-text"
-            >
-              {children}
-            </Flex>
-          </>
-        )}
+      <Flex ref={wrapperRef}>
+        <Button
+          aria-controls="popover-text"
```
hbuchel, line 44: This needs to reference an `id` that's on the same element as `role="dialog"` We're basically saying "this button controls the showing and hiding of this element"

packages/react-liveness/src/components/FaceLivenessDetector/shared/LivenessIconWithPopover.tsx
```diff
@@ -5,9 +5,10 @@
 
 import * as React from 'react';
 
-import { Flex } from '@aws-amplify/ui-react';
+import { Button, Flex } from '@aws-amplify/ui-react';
 import { AlertIcon, useThemeBreakpoint } from '@aws-amplify/ui-react/internal';
 import { LivenessClassNames } from '../types/classNames';
+import { defaultLivenessDisplayText } from '../displayText';
```
hbuchel, line 11: Sorry I didn't really explain this part; we won't be able to import directly from this because these are just our _default_ displayText; our customers can optionally provide their own object of displayText matching our dictionary then and we merge them together at the [root of the component](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react-liveness/src/components/FaceLivenessDetector/FaceLivenessDetectorCore.tsx#L44).

I think we'll need to add new strings for the aria-labels (for Button at least, and maybe for the aria-label on the dialog element), so you'll need to update displayText.ts with the strings/type updates, and then update what strings as props we pass down to the [DefaultPhotosensitiveWarning](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react-liveness/src/components/FaceLivenessDetector/StartLiveness/StartLiveness.tsx#L42) component in order to use them in your changes below.

hbuchel (commented):
Great start! Thank you for looking up the roles/ARIA attributes to use. Left some comments for updates needed:

PR #4538, opened 2023-10-13
chore: Complete deep linking for social signin

ioanabrooks:
#### Checklist

- [ ] PR description included


examples/react-native/ios/ReactNative/Info.plist
```diff
@@ -1,64 +1,68 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
+<dict>
```
calebpollman, line 4: Curious abt the changes in this file, are they autogenerated by RN?

packages/react-native/package.json
```diff
@@ -42,7 +42,6 @@
   },
   "dependencies": {
     "@aws-amplify/react-native": "console-preview",
```
calebpollman, line 44: Not sure where this was added, but why do we need this package as a direct dep?

examples/react-native/ios/ReactNative/Info.plist
ioanabrooks: XCode. I think it's mostly just reordering , what was actually added is 

```
<array>
		<dict>
			<key>CFBundleTypeRole</key>
			<string>Editor</string>
			<key>CFBundleURLName</key>
			<string>myapp</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>myapp</string>
			</array>
		</dict>
	</array>
```

For deeplinking

packages/react-native/package.json
ioanabrooks: Good catch, also not sure how that got in there but definitely removing it.

examples/react-native/ios/ReactNative/Info.plist
calebpollman: Thanks! Bit confused on what the meaning of "add deep linking" is in this PR since the example app already has deep linking support?

ioanabrooks: I guess complete deep linking? Has some updates and aws-exports didn't point to the correct link so wasn't working.

calebpollman (approved):

thaddmt (approved):

PR #4548, opened 2023-10-18
chore: migrate in-app messaging

esauerbo:
#### Description of changes

Migrated in-app messaging to js v6

Ran unit and e2e tests locally  

#### Checklist

- [ ] PR description included


examples/next/pages/ui/components/in-app-messaging/demo/utils.ts
```diff
@@ -1,12 +1,12 @@
...
-import { Notifications } from '@aws-amplify/notifications';
 import * as Analytics from '@aws-amplify/analytics';
+import { syncMessages } from 'aws-amplify/in-app-messaging';
 import {
-  InAppMessage,
-  InAppMessageAction,
-  InAppMessageButton,
-  InAppMessageLayout,
-} from '@aws-amplify/notifications';
+  Message,
```
esauerbo, line 5: Wasn't sure if this was the right call but changed these to not expose the js types (because we were already reassigning them like type = jstype in the notifications package)

examples/react-native/src/features/InAppMessaging/Demo/Example.tsx
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
+import { Amplify } from 'aws-amplify';
 import { SafeAreaView, StyleSheet, Text, View } from 'react-native';
 
 import { withInAppMessaging } from '@aws-amplify/ui-react-native';
 import { Button, Checkbox, Radio, RadioGroup } from '../../../ui';
 import { useInAppDemo, ACTIONS, LAYOUTS, ORIENTATIONS } from './utils';
+import config from './aws-exports';
+
+Amplify.configure(config);
```
esauerbo, line 10: Added call to `Amplify.configure` here because InApp now requires `initializeInAppMessaging()` to use, which must be called with `Amplify.configure`

packages/react-core-notifications/src/InAppMessaging/hooks/useMessage/useMessage.ts
```diff
@@ -40,26 +36,17 @@ export default function useMessage<PlatformStyleProps>({
...
+    notifyMessageInteraction({ message, type: 'messageDismissed' });
     clearMessage();
   };
 
   const onDisplay = () => {
-    InAppMessaging.notifyMessageInteraction(
-      message,
-      InAppMessageInteractionEvent.MESSAGE_DISPLAYED
-    );
+    notifyMessageInteraction({ message, type: 'messageDisplayed' });
```
calebpollman, line 49: Formatting looks off here, might want to check if prettier is configured to run on save in your editor

packages/ui/src/machines/authenticator/signUp.ts
```diff
@@ -246,7 +246,6 @@ export function createSignUpMachine({ services }: SignUpMachineOptions) {
          * @migration data is Auth.SignUpOutput
          */
         shouldSkipConfirm: (context, { data }) => {
-
```
calebpollman, line 249: Looks like a stray found it's way in 😅

examples/next/pages/ui/components/in-app-messaging/demo/utils.ts
calebpollman: Do we export `InAppMessage` from ui-react-notifications? If not let's expose and extend from it to replace these types, example:
```ts
type MessageLayout = Message['layout']
```

calebpollman (approved):
Really great start Emma! Left a blocking comment on the initialization piece and some minor type/formatting issues but other than that LGTM

calebpollman (changes requested):
See previous review comment, goofed a button click 😆

examples/next/pages/ui/components/in-app-messaging/demo/utils.ts
esauerbo: We don't export it directly, it's like this: 
``` 
export type MessageLayout = InAppMessageLayout; 
```
Do you mean like just import Message into the example and then define the layout types etc in the example?

packages/react-core-notifications/src/InAppMessaging/types.ts
```diff
@@ -18,6 +18,7 @@ export type MessageImage = InAppMessageImage;
 export type MessageLayout = InAppMessageLayout;
 export type MessageStyle = InAppMessageStyle;
 export type MessageTextAlign = InAppMessageTextAlign;
+export type test = 'hello';
```
calebpollman, line 21: 👋

packages/react-core-notifications/src/InAppMessaging/hooks/useMessage/__tests__/useMessage.spec.ts
```diff
@@ -177,7 +177,8 @@ describe('useMessage', () => {
 
     beforeEach(() => {
       jest.clearAllMocks();
-
+      Amplify.configure(mockawsmobile);
+      initializeInAppMessaging();
```
calebpollman, line 181: Why do we need these?

packages/react-core-notifications/src/InAppMessaging/hooks/useMessage/__tests__/useMessage.spec.ts
```diff
@@ -31,6 +27,10 @@ const infoSpy = jest.spyOn(Logger.prototype, 'info');
 
 const mockUseInAppMessaging = useInAppMessaging as jest.Mock;
 const mockClearMessage = jest.fn();
+jest.mock('aws-amplify/in-app-messaging', () => ({
+  ...jest.requireActual('aws-amplify/in-app-messaging'),
+  notifyMessageInteraction: jest.fn(),
+}));
```
calebpollman, line 33: Prefer spies over mocks, spies retain type information and can catch breaking changes that a mock would hide

examples/next/pages/ui/components/in-app-messaging/demo/utils.ts
calebpollman: Close - what I'm asking is that we export `InAppMessage` from _@aws-amplify/ui-react-notifications_ (if we do not already) as it is a public package (and the type itself could provide value to consumers), and then create utility types from it as needed in this file

packages/react-core-notifications/src/InAppMessaging/hooks/useMessage/__tests__/useMessage.spec.ts
esauerbo: The calls to useMessage require initializing in-app :(

packages/react-core-notifications/src/InAppMessaging/hooks/useMessage/__tests__/useMessage.spec.ts
esauerbo: I couldn't figure out how to do that with the functional apis because spyOn expects an object and a method

examples/next/pages/ui/components/in-app-messaging/demo/utils.ts
esauerbo: But we're already exporting Message, MessageAction etc so should we stop doing that in favor of only exporting InAppMessage? I feel like it would be harder if you have to reconstruct the types every time you want to use them

calebpollman: Let's handle this as a follow up 🙂

packages/react-core-notifications/src/InAppMessaging/hooks/useMessage/__tests__/useMessage.spec.ts
calebpollman: Can we remove these calls by mocking?

esauerbo: Yes missed initializeInApp, but to my understanding Amplify.configure is being mocked here 
https://jestjs.io/docs/mock-functions#mocking-modules

packages/react-core-notifications/src/InAppMessaging/context/InAppMessagingProvider/InAppMessagingProvider.tsx
```diff
@@ -15,16 +15,14 @@ export interface InAppMessagingProviderProps {
...
   children,
 }: InAppMessagingProviderProps): JSX.Element {
   const [message, setMessage] =
     useState<InAppMessagingContextType['message']>(null);
 
   useEffect(() => {
-    const listener = InAppMessaging.onMessageReceived((message) => {
+    const listener = onMessageReceived((message) => {
       setMessage(message);
     });
```
calebpollman, line 27: Think you should be able to just pass `setMessage` directly to `onMessageReceived` here assuming the types don't get mad:
```suggestion
    const listener = onMessageReceived(setMessage);
```

calebpollman (approved):
🚢

PR #4549, opened 2023-10-18
fix: sign in with sms confirmation code

thaddmt:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/ui/src/machines/authenticator/actors/signIn.ts
+++ b/packages/ui/src/machines/authenticator/actors/signIn.ts
@@ -546,7 +546,9 @@ export function signInActor({ services }: SignInMachineOptions) {
         },
         shouldConfirmSignIn: (_, event): boolean => {
           groupLog('+++shouldConfirmSignIn', 'event', event);
-          return event.data.nextStep?.signInStep === 'CONFIRM_SIGN_IN';
+          return (
+            event.data.nextStep?.signInStep === 'CONFIRM_SIGN_IN_WITH_SMS_CODE'
+          );
         },
         shouldForceChangePassword: (_, event): boolean => {
           groupLog('+++shouldForceChangePassword', 'event', event);
--- a/packages/ui/src/machines/authenticator/index.ts
+++ b/packages/ui/src/machines/authenticator/index.ts
@@ -278,7 +278,11 @@ export function createAuthenticatorMachine(
             const cliLoginMechanisms = Object.entries(cliConfig.loginWith)
               .filter(([key, _value]) => key !== 'oauth')
               .filter(([_key, value]) => !!value)
-              .map((keyValueArray) => keyValueArray[0]) as LoginMechanism[];
+              .map((keyValueArray) => {
+                return keyValueArray[0] === 'phone' // the key for phone_number is phone in getConfig but everywhere else we treat is as phone_number
+                  ? 'phone_number'
+                  : keyValueArray[0];
+              }) as LoginMechanism[];
             const cliSignupAttributes = Object.entries(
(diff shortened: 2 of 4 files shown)


packages/ui/src/machines/authenticator/index.ts
```diff
@@ -278,7 +278,11 @@ export function createAuthenticatorMachine(
             const cliLoginMechanisms = Object.entries(cliConfig.loginWith)
               .filter(([key, _value]) => key !== 'oauth')
               .filter(([_key, value]) => !!value)
-              .map((keyValueArray) => keyValueArray[0]) as LoginMechanism[];
+              .map((keyValueArray) => {
+                return keyValueArray[0] === 'phone' // the key for phone_number is phone in getConfig but everywhere else we treat is as phone_number
```
thaddmt, line 282: If it's not clear, the value returned to us in `getConfig` is 

```
loginWith: {
email: false
phone: true
username: false
}
```

so we need to convert the key `phone` to `phone_number`

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -546,7 +546,9 @@ export function signInActor({ services }: SignInMachineOptions) {
         },
         shouldConfirmSignIn: (_, event): boolean => {
           groupLog('+++shouldConfirmSignIn', 'event', event);
-          return event.data.nextStep?.signInStep === 'CONFIRM_SIGN_IN';
+          return (
+            event.data.nextStep?.signInStep === 'CONFIRM_SIGN_IN_WITH_SMS_CODE'
+          );
         },
```
calebpollman, line 552: Assuming `shouldConfirmSignIn` covers flows that are not SMS related (TOTP for example)?

calebpollman: Maybe we should rename `shouldConfirmSignIn` to `shouldConfirmSignInWithSMS`?

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -14,7 +17,28 @@ export const defaultServices = {
   async getAmplifyConfig() {
     const result = Amplify.getConfig();
     groupLog('+++getAmplifyConfig', 'result', result);
-    return result;
+
+    const cliConfig = (result as ResourcesConfig).Auth.Cognito;
```
calebpollman, line 21: Curious why we need to cast the return type of `getConfig`?

thaddmt: Good call, this was just when I copy pasted from the other method as it needed the casting. Here it does not since it's already ResourcesConfig

packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -14,7 +17,28 @@ export const defaultServices = {
   async getAmplifyConfig() {
     const result = Amplify.getConfig();
     groupLog('+++getAmplifyConfig', 'result', result);
-    return result;
+
+    const cliConfig = result.Auth.Cognito;
+    const parsedLoginMechanisms = Object.entries(cliConfig.loginWith)
```
calebpollman, line 22: Can we handle this (and the below) in `getAuthenticatorConfig`?

thaddmt: Discussed offline, will punt on this for now

calebpollman (approved):
Approved, but do we think we need to follow up with something [here](https://github.com/aws-amplify/amplify-ui/pull/4549/files#r1364349861)

PR #4556, opened 2023-10-19
chore: call fetchUserAttributers with getUser

thaddmt:
#### Description of changes
- Previously getCurrentAuthenticatedUser returned user information as well as attributes
- Now that they are two separate calls combining them helps return a value similar to what was returned previously

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/ui/src/machines/authenticator/actors/signIn.ts
+++ b/packages/ui/src/machines/authenticator/actors/signIn.ts
@@ -835,7 +835,10 @@ export function signInActor({ services }: SignInMachineOptions) {
         },
         async getCurrentUserResolved(context) {
           groupLog('+++getCurrentUserResolved', context, event);
-          return await Auth.getCurrentUser();
+          return {
+            ...(await Auth.getCurrentUser()),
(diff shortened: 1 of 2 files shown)


packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -835,7 +835,10 @@ export function signInActor({ services }: SignInMachineOptions) {
         },
         async getCurrentUserResolved(context) {
           groupLog('+++getCurrentUserResolved', context, event);
-          return await Auth.getCurrentUser();
+          return {
+            ...(await Auth.getCurrentUser()),
+            attributes: { ...(await Auth.fetchUserAttributes()) },
```
calebpollman, line 840: Do we need to return the attributes for a specific reason?

thaddmt: Not that I know of but a few of our example apps previously used values like `user.attributes.email` and previously user contained attributes in v5

calebpollman (approved):

PR #4564, opened 2023-10-20
fix: various bug fixes found in bug bash

thaddmt:
#### Checklist

- [ ] PR description included

### DIFF
--- a/examples/angular/src/pages/ui/components/authenticator/sign-in-sms-mfa/sign-in-sms-mfa.component.html
+++ b/examples/angular/src/pages/ui/components/authenticator/sign-in-sms-mfa/sign-in-sms-mfa.component.html
@@ -4,7 +4,7 @@
     let-user="user"
     let-signOut="signOut"
   >
-    <h2>Welcome, {{ user.attributes?.name }}!</h2>
+    <h2>Welcome, {{ user.attributes?.email }}!</h2>
     <button (click)="signOut()">Sign Out</button>
   </ng-template>
 </amplify-authenticator>
--- a/packages/ui/src/machines/authenticator/actions.ts
+++ b/packages/ui/src/machines/authenticator/actions.ts
@@ -92,7 +92,7 @@ export const setChallengeName = assign({
 export const setRequiredAttributes = assign({
   requiredAttributes: (_, event: AuthEvent) => {
     groupLog('+++setRequiredAttributes', 'event', event);
-    return event.data?.challengeParam?.requiredAttributes;
+    return event.data?.nextStep?.missingAttributes;
   },
 });
 
--- a/packages/ui/src/machines/authenticator/actions.ts
+++ b/packages/ui/src/machines/authenticator/actions.ts
@@ -148,8 +148,9 @@ export const setUnverifiedContactMethods = assign({
       event.data as Auth.FetchUserAttributesOutput;
 
     return {
-      ...(email_verified === 'false' && { email }),
-      ...(phone_number_verified === 'false' && { phone_number }),
+      ...(email_verified === 'false' && email && { email }),
+      ...(phone_number_verified === 'false' &&
+        phone_number && { phone_number }),
     };
   },
 });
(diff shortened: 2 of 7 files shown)


reesscot (approved):

examples/angular/src/pages/ui/components/authenticator/sign-in-sms-mfa/sign-in-sms-mfa.component.html
```diff
@@ -4,7 +4,7 @@
     let-user="user"
     let-signOut="signOut"
   >
-    <h2>Welcome, {{ user.attributes?.name }}!</h2>
+    <h2>Welcome, {{ user.attributes?.email }}!</h2>
```
calebpollman, line 7: Does one of the available attributes map to what we are currently using for `name`? Maybe I'm mistaken, but reading this as a breaking change

packages/ui/src/machines/authenticator/actions.ts
```diff
@@ -148,8 +148,9 @@ export const setUnverifiedContactMethods = assign({
       event.data as Auth.FetchUserAttributesOutput;
 
     return {
-      ...(email_verified === 'false' && { email }),
-      ...(phone_number_verified === 'false' && { phone_number }),
+      ...(email_verified === 'false' && email && { email }),
+      ...(phone_number_verified === 'false' &&
+        phone_number && { phone_number }),
```
calebpollman, line 153: Is there a scenario where both line 151 and 152 resolve to meaningful values?

packages/vue/src/components/confirm-sign-in.vue
```diff
@@ -20,7 +20,7 @@ const facade: UseAuthenticator = useAuthenticator();
 const { submitForm, toSignIn, updateForm } = facade;
 const { user, error, isPending } = toRefs(facade);
 
-const challengeName = computed(() => user.value.challengeName);
+const challengeName = computed(() => user.value.nextStep.signInStep);
```
calebpollman, line 23: Think we can avoid a breaking change here if we map `nextStep.signInStep` to `challengeName`. Also avoids exposing JS return types directly on `user`

packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -576,9 +576,15 @@ export function signInActor({ services }: SignInMachineOptions) {
...
-            email_verified === 'false' && phone_number_verified === 'false'
-          );
+          // email/phone_verified is returned as a string
+          const emailNotVerified =
+            email_verified === undefined || email_verified === 'false';
+          const phoneNotVerified =
+            phone_number_verified === undefined ||
+            phone_number_verified === 'false';
+
+          // only request verification if both email and phone are not verified
```
calebpollman, line 586: Think we should stick with the current behavior here at least until the migration is complete. Concerned abt updating UX while we still have moving pieces

calebpollman (commented):
Thank you for looking in to these! Left some call outs/questions related to avoiding breaking behavior

packages/ui/src/machines/authenticator/actors/signIn.ts
thaddmt: Yep this changes it to match the current behavior better. Previously there was a bug that if just email was verified and there was no value for `phone_number_verified` it would try to request verification

packages/ui/src/machines/authenticator/actions.ts
thaddmt: There is a case, this was just the short hand way of conditionally assigning the value only if the value exists. E.g. avoiding adding `{phone_number: undefined}`

packages/vue/src/components/confirm-sign-in.vue
thaddmt: yep, can change that as we discussed

examples/angular/src/pages/ui/components/authenticator/sign-in-sms-mfa/sign-in-sms-mfa.component.html
thaddmt: I believe in this case the app was broken in prod as well

calebpollman (approved):

PR #4574, opened 2023-10-23
chore: add data plane tracking

esauerbo:
#### Description of changes

Updating data plane tracking for v6 with Authenticator
Adding data plane tracking to AccountSettings, InAppMessaging, LocationSearch, StorageManager

#### Checklist

- [ ] PR description included


packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
```diff
@@ -21,6 +22,15 @@ function DeleteUser({
   const [state, setState] = React.useState<DeleteUserState>('IDLE');
   const [errorMessage, setErrorMessage] = React.useState<string | null>(null);
 
+  React.useEffect(() => {
+    const clearUserAgent = setUserAgent({
```
esauerbo, line 26: Headers
Screenshot 2023-10-23 at 6 05 17 PM

dbanksdesign: Should this be 'AccountSettings' rather than 'Authenticator'?

dbanksdesign: Actually this looks correct in the code, the screenshot is probably from a different request

dbanksdesign (approved):
LGTM!

packages/react/src/components/AccountSettings/DeleteUser/DeleteUser.tsx
esauerbo: My bad 
DeleteUser

packages/react-native/src/Authenticator/Authenticator.tsx
```diff
@@ -10,8 +10,7 @@ import {
   useAuthenticatorInitMachine,
 } from '@aws-amplify/ui-react-core';
 
-import { configureComponent } from '@aws-amplify/ui';
-
+import { useSetUserAgent } from '@aws-amplify/ui-react/internal';
```
calebpollman, line 13: `@aws-amplify/ui-react` is not a dependency of `@aws-amplify/ui-react-native` nor should it be

calebpollman (changes requested):
Approach to `useSetUserAgent` looks correct but should be in the ui-react-core React utility package that's shared between React and React Native. 

Also rather than lumping the `setUserAgent` changes in this PR, prefer keeping those changes in their own PR like you had done initially, just without removing `configureComponent` so the consumers don't break

packages/vue/src/components/authenticator.vue
```diff
@@ -15,12 +15,11 @@ import {
   AuthenticatorRoute,
   SocialProvider,
   authenticatorTextUtil,
-  configureComponent,
+  setUserAgent,
 } from '@aws-amplify/ui';
 
 import { useAuth, useAuthenticator } from '../composables/useAuth';
 import { UseAuthenticator } from '../types';
-import { VERSION } from '../version';
```
calebpollman, line 23: Not an official official style rule or anythiing, but we try to group imports in the following order:

_Framework/Framework 3pp_

```ts
import React from 'react';
import { something } from 'react-hook-form';
```

_Internal imports_
```ts
import { somethingBetter } from '@aws-amplify/ui-react-core';
```
_relative paths_
1) by relative proximity of the path
2) alphabetize by import path

```ts
import { stuff } from '../../stuff';
import { things } from '../../thngs';

import { nearby } from './nearby';
```

packages/react-core/src/hooks/useSetUserAgent.ts
```diff
@@ -0,0 +1,20 @@
...
+  useEffect(() => {
+    const clearUserAgent = setUserAgent({
+      componentName,
+      packageName,
+      version,
+    });
+
+    return () => {
+      clearUserAgent();
+    };
```
calebpollman, line 18: Instead of wrapping the call to `clearUserAgent` in anonymous function can simplify the return to:
```suggestion
    return clearUserAgent;
```

calebpollman (commented):
Left some style related comments but other than that LGTM

calebpollman (approved):
🚢

PR #4576, opened 2023-10-24
[fix][vue/angular] reinitialize state machine when signInWithRedirect event occurs

thaddmt:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/ui/src/helpers/authenticator/utils.ts
+++ b/packages/ui/src/helpers/authenticator/utils.ts
@@ -91,6 +91,9 @@ export const defaultAuthHubHandler: AuthMachineHubHandler = async (
         send('TOKEN_REFRESH');
       }
       break;
+    case 'signInWithRedirect':
+      send('REFRESH');
+      break;
     case 'signedIn':
       if (isFunction(onSignIn)) {
         onSignIn();
--- a/packages/ui/src/helpers/authenticator/utils.ts
+++ b/packages/ui/src/helpers/authenticator/utils.ts
@@ -3,7 +3,7 @@
  * implementations can use.
  */
 
-import { Hub } from '@aws-amplify/core';
+import { Hub } from 'aws-amplify/utils';
 // import { appendToCognitoUserAgent } from '@aws-amplify/auth';
 import { waitFor } from 'xstate/lib/waitFor.js';
 
--- a/packages/ui/src/machines/authenticator/defaultServices.ts
+++ b/packages/ui/src/machines/authenticator/defaultServices.ts
@@ -41,6 +41,8 @@ export const defaultServices = {
     };
   },
   async getCurrentUser() {
+    console.log('+++getCurrentUser');
+    await new Promise((resolve) => setTimeout(resolve, 4));
     return {
       ...(await Auth.getCurrentUser()),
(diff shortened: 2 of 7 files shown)


packages/ui/src/machines/authenticator/defaultServices.ts
```diff
@@ -41,6 +41,8 @@ export const defaultServices = {
     };
   },
   async getCurrentUser() {
+    console.log('+++getCurrentUser');
+    await new Promise((resolve) => setTimeout(resolve, 4));
```
thaddmt, line 45: - There is a race condition between the machine being reinitialized after redirect and the JS API logging in the user
- If the user attempts to call an authenticated API in this time it seems to hang here but adding this short wait resolves this issue.
- This seems to not occur on react because on the signedIn event react re-renders whereas this does not seem to occur in angular - https://github.com/aws-amplify/amplify-ui/blob/next-release/migration/packages/ui/src/helpers/authenticator/utils.ts#L85

calebpollman: Is this an issue in the current Authenticator or just with the migration?

thaddmt: I'm not able to get the angular example app for v5 working with federated sign in, I consistently get an error and instpecting the network log for the 400 it shows

Partially omitted the error for saftey
```
{"__type":"ValidationException","message":"1 validation error detected: Value '{cognito-idp.us-east-2.amazonaws.com/us-east-2********}' at 'logins' failed to satisfy constraint: Map value must satisfy constraint: [Member must have length less than or equal to 50000, Member must have length greater than or equal to 1]"}
```

calebpollman: What is the impact of removing the call to `getCurrentUser` after redirect?

thaddmt: This call to `getCurrentUser` happens everytime the state machine is instantiated - https://github.com/aws-amplify/amplify-ui/blob/next-release/fix-vue-federated/packages/ui/src/machines/authenticator/index.ts#L70

It's how we check for whether the user is authenticated or not so removing it here might be complicated to say the least

calebpollman: Would it be possible not to call it on redirect though?

thaddmt: Ended up attempting something else which might be the fix

packages/ui/src/helpers/authenticator/utils.ts
```diff
@@ -91,6 +91,9 @@ export const defaultAuthHubHandler: AuthMachineHubHandler = async (
         send('TOKEN_REFRESH');
       }
       break;
+    case 'signInWithRedirect':
+      send('REFRESH');
+      break;
```
calebpollman, line 96: Concern here is that we may have introduced a race condition somewhere else in the machine (@esauerbo has tested Vue and Angular with the current machine and had no issue) and that this might mask a deeper issue that will bite us somewhere else

calebpollman:
Closing as the issue that required this PR was found to be a bug in dependency code that is now fixed

PR #4603, opened 2023-10-27
fix(e2e): fix storage manager test 

esauerbo:
#### Description of changes

Fixed failing e2e test in storage-manager 

Ran e2e tests 

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/e2e/cypress/integration/common/fileuploader.ts
+++ b/packages/e2e/cypress/integration/common/fileuploader.ts
@@ -39,15 +39,15 @@ When(
 );
 
 When('I drag and drop a file with file name {string}', (fileName: string) => {
-  cy.get('input[type=file]').selectFile(
-    {
-      contents: Cypress.Buffer.from('file contents'),
-      fileName,
-      lastModified: Date.now(),
+  cy.get('input[type=file]').trigger('drop', {
+    dataTransfer: {
+      files: [
+        new File(['file contents'], fileName, { lastModified: Date.now() }),
+      ],
     },
     /**
      *  Since the input is hidden, this will need to be forced through Cypress
      */
-    { force: true, action: 'drag-drop' }
-  );
+    force: true,
+  });
 });


ioanabrooks (approved):

packages/e2e/cypress/integration/common/fileuploader.ts
```diff
@@ -39,15 +39,15 @@ When(
 );
 
 When('I drag and drop a file with file name {string}', (fileName: string) => {
-  cy.get('input[type=file]').selectFile(
-    {
-      contents: Cypress.Buffer.from('file contents'),
-      fileName,
-      lastModified: Date.now(),
+  cy.get('input[type=file]').trigger('drop', {
```
ioanabrooks, line 42: Curious about this change. Looks like only one storage manager test is failing but this change affects all tests, could you explain a bit more about it? In [CI](https://github.com/aws-amplify/amplify-ui/actions/runs/6674643826/job/18141881100) failed test screenshot reveals failure due to the fact that 2 files were uploaded somehow, when the component is configured to only accept one. Maybe some cleanup from a previous test missing.

ioanabrooks (changes requested):

packages/e2e/cypress/integration/common/fileuploader.ts
reesscot: That's a good point, can you take a look @ioanabrooks?

reesscot: Right but we still want to test both, can we find a way to cleanup between tests or split tests?

calebpollman: Splitting makes sense to me, imo would improve our usage scenario coverage

esauerbo: Do you mean add an additional test for successfuly uploading to public access but with select instead of drag/drop? I'm confused because there was only one test using this code and the scenario was "I should be able to click and drag any files" so I thought it was only testing drag/drop

reesscot: Actually it looks like this assertion is only used in one file, so I think the change makes sense.
CleanShot 2023-10-31 at 09 34 25

reesscot (approved):

ioanabrooks (approved):

calebpollman (approved):

PR #4666, opened 2023-11-08
fix(storagemanager): upload action now displays state properly

dbanksdesign:

Before

CleanShot 2023-11-07 at 20 48 59

After

CleanShot 2023-11-07 at 20 44 52 |

#### Checklist

- [x] PR description included


reesscot (dismissed):

packages/react-storage/src/components/StorageManager/hooks/useStorageManager/reducer.ts
```diff
@@ -48,7 +48,9 @@ export function storageManagerStateReducer(
           ...files,
           {
             ...currentFile,
-            status: FileStatus.QUEUED,
+            ...(currentFile.status === FileStatus.ADDED
```
reesscot, line 51: If I understand this right, the problem was the previously queued files where being re-added here? The fix is to only make them queued if they were just added?

ioanabrooks: I saw a couple of fixes for storage manager that look like big misses, some from moving things from one branch to another, others just misses. Can we add some tests so that in future we catch these issues?

reesscot (approved):

calebpollman (approved):

PR #4667, opened 2023-11-08
fix(storageimage): handle onStorageGetError properly

dbanksdesign:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageImage/StorageImage.tsx
+++ b/packages/react-storage/src/components/StorageImage/StorageImage.tsx
@@ -23,7 +23,10 @@ export const StorageImage = ({
 
   const url = useStorageURL({
     key: imgKey,
-    options,
+    options: {
+      ...options,
+      validateObjectExistence: true,
+    },
     fallbackURL: fallbackSrc,
     onStorageGetError,
   });
--- a/packages/react-storage/src/components/StorageImage/StorageImage.tsx
+++ b/packages/react-storage/src/components/StorageImage/StorageImage.tsx
@@ -14,19 +14,17 @@ export const StorageImage = ({
   identityId,
   imgKey,
   onStorageGetError,
+  validateObjectExistence = true,
   ...rest
 }: StorageImageProps): JSX.Element => {
   const options = React.useMemo(
-    () => ({ accessLevel, identityId }),
-    [accessLevel, identityId]
+    () => ({ accessLevel, identityId, validateObjectExistence }),
+    [accessLevel, identityId, validateObjectExistence]
   );
 
   const url = useStorageURL({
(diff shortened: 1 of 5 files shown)


packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -23,7 +23,10 @@ export const StorageImage = ({
 
   const url = useStorageURL({
     key: imgKey,
-    options,
+    options: {
+      ...options,
+      validateObjectExistence: true,
```
calebpollman, line 28: Don't feel strongly but should `validateObjectExistence` be a prop with a default value of `true`?

dbanksdesign: Done

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -14,11 +14,12 @@ export const StorageImage = ({
   identityId,
   imgKey,
   onStorageGetError,
+  validateObjectExistence = true,
```
calebpollman, line 17: What if the provided value is `false` or `null` 😅

dbanksdesign: It works? I just tried it. Or do you mean should we change this boolean prop to be the opposite so the default is false?

reesscot: Yeah, if they provide `false` or `null`, I think it would default to `true` right?

reesscot (approved):
LGTM, thanks for fixing this

reesscot:
Missing changeset for new behavior

packages/react-storage/src/components/StorageImage/StorageImage.tsx
calebpollman: > Yeah, if they provide false or null, I think it would default to true right

This 🤠 Maybe just add something like:
```
const resolvedValidateObjectExistence = isUndefined(validateObjectExistence) ? true : validateObjectExistence;
```

dbanksdesign: ugh fine. a bit wordy, but changed

calebpollman: Mean you could've renamed it 😏

calebpollman (dismissed):

reesscot (approved):

reesscot (approved):
LGTM

calebpollman (approved):

reesscot (approved):

PR #4699, opened 2023-11-13
chore(react): fixing console errors on react primitives unit tests

esauerbo:
React unit tests pass with no errors printed to console 

#### Checklist

- [ ] PR description included


packages/react/src/hooks/__tests__/useAuth.test.ts
```diff
@@ -78,7 +78,7 @@ describe('useAuth', () => {
 
       const { result, waitForNextUpdate } = renderHook(() => useAuth());
 
-      await waitForNextUpdate();
+      await act(async () => await waitForNextUpdate());
```
calebpollman, line 81: Think `waitForNextUpdate` returns `Promise<void>`, does the callback need to return?

calebpollman (commented):
Just left a question but LGTM other than that. Thanks for the cleanup!

packages/react/src/hooks/__tests__/useAuth.test.ts
esauerbo: In some cases yes but took out the async code where it wasn't necessary

packages/react/src/primitives/Accordion/__tests__/Accordion.test.tsx
```diff
@@ -1,4 +1,5 @@
 import { render, screen } from '@testing-library/react';
+import { act } from '@testing-library/react-hooks';
```
calebpollman, line 2: I don't think there are any hooks being tested in this file. Prefer to import act  from `'@testing-library/react'`

packages/react/src/hooks/__tests__/useAuth.test.ts
calebpollman: Think my question may have been lost in translation, an arrow function with curly brackets creates an implicit return:
```
const returnThing = () => 'thing';

console.log(returnThing());
// 'thing';
```

In the case of:

```
await act(async () => await waitForNextUpdate());
```

`act` do anything with the return of its callback, so the implicit return is extraneous

calebpollman (commented):
Still seeing imports of `act` from _@testing-library/react-hooks_ in files that are not testing React hooks, but ready to approve once those are cleaned up!

packages/react/src/hooks/__tests__/useAuth.test.ts
```diff
@@ -54,7 +54,9 @@ describe('useAuth', () => {
 
     const { result, waitForNextUpdate } = renderHook(() => useAuth());
 
-    await waitForNextUpdate();
+    await act(async () => {
+      await waitForNextUpdate();
+    });
```
calebpollman, line 59: Non-blocking but we can probably remove the `act` and `await` here if we remove the `async` declaration on line 52 (same with line 69)

calebpollman (approved):
🚢

packages/react/src/hooks/__tests__/useAuth.test.ts
esauerbo: I tried that but then the result comes back undefined and the test fails

reesscot (approved):

PR #4757, opened 2023-11-16
feat: Added external link icon

lavr001:
#### Description of changes

Added external link icon when `isExternal` prop is true

Before:

Link_before

After:

Link_after

#### Checklist

- [x] PR description included


packages/react/src/primitives/Link/Link.tsx
```diff
@@ -10,6 +10,7 @@ import {
 } from '../types';
 import { View } from '../View';
 import { primitiveWithForwardRef } from '../utils/primitiveWithForwardRef';
+import { BiLinkExternal } from 'react-icons/bi';
```
dbanksdesign, line 13: We don't usually pull in icons from an external source, we add them manually in the ui package. You'll also need to add it to the IconsProvider so it can be overridden.

packages/ui/src/theme/css/component/link.scss
```diff
@@ -4,6 +4,10 @@
   // all:unset removes this, so we have to add it back.
   cursor: pointer;
 
+  display: flex;
```
dbanksdesign, line 7: I think this should be `inline-flex` because a link is an inline element

packages/ui/src/theme/css/component/link.scss
```diff
@@ -4,6 +4,10 @@
...
+  align-items: center;
+  gap: 2px;
```
dbanksdesign, line 9: Let's use a design token here.

dbanksdesign (changes requested):

packages/react/src/primitives/Link/Link.tsx
calebpollman: +1 to the above. Additionally, `react-icons` is not a dependency of `@aws-amplify/ui-react` and shipping this would cause runtime failures

packages/ui/src/theme/css/component/link.scss
lavr001: Done

packages/react/src/primitives/Icon/icons/IconBoxArrowUpRight.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { classNames } from '@aws-amplify/ui';
+import { ComponentClassName } from '@aws-amplify/ui';
```
calebpollman, line 3: This should only be a single import line

packages/react/src/primitives/Link/Link.tsx
```diff
@@ -25,6 +24,7 @@ const LinkPrimitive: Primitive<LinkProps, 'a'> = (
       {...rest}
     >
       {children}
+      {isExternal && <IconBoxArrowUpRight />}
```
calebpollman, line 27: Few things here:
- Prefer ternary returning `null` if `isExternal` is falsy to prevent this expression from evaluating to `false`
- How do i hide the icon?
- How do i position the icon to the left of the `children`?

lavr001: 1. Updated to use a ternary operator
2. You can hide the icon when `isExternal` prop is set to `false` (default)
3. It's important to start with a word that indicates the action that results if customers click the link. Putting an icon at the beginning of the link disrupts the readability of the link quite massively, therefore it is a common convention to put icons after the actual link text

packages/ui/src/icons/box_arrow_up_right.svg
reesscot, line 1: Where did this icon come from and why is it not centered like the other icons?

packages/react/src/primitives/Link/Link.tsx
calebpollman: > You can hide the icon when isExternal prop is set to false (default)

Sure, but what if the link is external and i don't want the icon?

> It's important to start with a word that indicates the action that results if customers click the link. Putting an icon at the beginning of the link disrupts the readability of the link quite massively, therefore it is a common convention to put icons after the actual link text

1. What abt RTL languages?
2. This response is an opinion on the usage of the `Link` component but does not address the actual question

packages/ui/src/icons/box_arrow_up_right.svg
lavr001: Since we don't typically use icons from the external libraries, I just looked online for a SVG path that forms the icon. I've added `align-items: center` to make the icon vertically aligned, but I can take another look to see what else could be done to make it more vertically centered

packages/react/src/primitives/Link/Link.tsx
lavr001: Good call on RTL languages. I'll work on adding a position prop, so devs can control the icon position

lavr001: There is no need to create a position prop since the direction='rtl' can be set in the `ThemeProvider` and it will automatically applied to the `Link` component
Screenshot 2023-12-04 at 4 00 24 PM

packages/react/src/primitives/Link/Link.tsx
lavr001: Done

packages/ui/src/icons/box_arrow_up_right.svg
lavr001: I checked SVG icons from material and they didn't really have a specific external link icon, so I grabbed another SVG

packages/react/src/primitives/Icon/icons/IconBoxArrowUpRight.tsx
calebpollman: Unresolved, there are two import lines for the same package still. In the future, please refrain from resolving a conversation until the feedback has been addressed

packages/ui/src/icons/box_arrow_up_right.svg
calebpollman: Where is this SVG taken from and what was the reasoning used in choosing it?

packages/react/src/primitives/Link/Link.tsx
calebpollman: @lavr001 The above question is still unanswered:

> > You can hide the icon when isExternal prop is set to false (default)
> 
> Sure, but what if the link is external and i don't want the icon?

calebpollman (changes requested):
Concerned overall with the approach taken here. 

Rendering an icon by default for any `Link` components that have an external link could break customers layouts that have added an icon themselves, nor are we providing a means to override the icon or hide it. Blocking this until all feedback is addressed and the PR no longer contains breaking changes

packages/react/src/primitives/Icon/icons/IconBoxArrowUpRight.tsx
lavr001: Looks like there are two import lines for the same package across multiple primitives, I'll make sure to make it one line when done with other updates on this PR.

packages/react/src/primitives/Link/Link.tsx
lavr001: @calebpollman Are you suggesting having a new prop called `hasIcon` that would be visible only when the link is external, so users can control if they want the icon or not?

packages/ui/src/icons/box_arrow_up_right.svg
lavr001: I simply googled svg path for external icon, since we don't usually pull in icons from an external source. Not sure what do you mean by reasoning. Are you suggesting some other approaches here?

packages/react/src/primitives/Link/Link.tsx
lavr001: @calebpollman Added new prop `hideIcon`, so now if the Link is external you have an option to show/hide the icon. Also added a new prop `linkIconPosition`, so you can position the icon left or right to the Link
Screenshot 2024-01-09 at 2 54 22 PM

PR #4775, opened 2023-11-20
vue skipped unit tests are fixed

lavr001:
#### Description of changes

vue skipped unit tests are fixed, successfully tested out locally

#### Checklist

- [ ] PR description included


thaddmt (dismissed):

packages/vue/src/components/__tests__/confirm-reset-password.spec.ts
```diff
@@ -135,14 +135,24 @@ describe('ConfirmResetPassword', () => {
...
+    const confirmPasswordField =
+      await screen.findByLabelText('Confirm Password');
+    await fireEvent.input(codeField, { target: codeInputPrams });
+    await fireEvent.input(newPasswordField, {
+      target: newPasswordInputParams,
+    });
+    await fireEvent.input(confirmPasswordField, {
+      target: confirmPasswordInputParams,
+    });
+    // then we can submit the form
```
calebpollman, line 152: Curious why we need so much additional code in this test? Same question applies to other unit test in this PR

calebpollman (commented):
Thanks for doing this @lavr001! Left a question, additionally would like to see a commit message following [conventional commit format](https://www.conventionalcommits.org/en/v1.0.0/), with a contextual message

packages/vue/src/components/__tests__/confirm-reset-password.spec.ts
lavr001: My thinking was since basically forms can only be submitted if they are filled out properly, so first we need fill the form. I can have a second look, just to double check if you think that it's a blocker.

calebpollman: What changed that caused the need for the additional code?

lavr001: That test was failing because it was testing the action of submitting the form, that's why I came up with this approach, so the form can only be submitted after all the fields have been filled out first.

calebpollman: Previous to the MV, this test passed without the additional actions in the unit test. I am curious as to what caused the unit test to no longer pass, which will give the required context to determine if the changes made here are appropriate

calebpollman (changes requested):
Please see comments

packages/vue/src/components/confirm-reset-password.vue
```diff
@@ -57,6 +57,7 @@ function onBlur(e: Event) {
         @input="onInput"
         @blur.capture="onBlur"
         @submit.prevent="onConfirmResetPasswordSubmit"
+        role="form"
```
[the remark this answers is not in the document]
lavr001: Acknowledged, I'll work on it.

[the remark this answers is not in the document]
lavr001: Removed `role="form"` as requested.

packages/vue/src/components/__tests__/confirm-reset-password.spec.ts
lavr001: I looked at the migration commit and the Vue package doesn't seem to be affected by much, only a few dependency updates, so the current solution that passes all the tests is the only one I could come up with.

calebpollman: The open question is still why this test requires so many changes, where as other Vue Authenticator test files with the same test did not. Not looking for an "optimized" solution here, looking to understand the need for this solution since it differs from the aforementioned test files

Issue #4776, opened 2023-11-21
isDisabled not working in Breadcrumbs.Item alongside Breadcrums.Link


regularpersonn:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Primitive components

### How is your app built?

Create React

### What browsers are you seeing the problem on?

Chrome, Safari

### Please describe your bug.

isDisabled flag is not being respected despite confirming that the framework is rendering <a disabled... component in HTML.

### What's the expected behaviour?

Breadcrumb item should be disabled when isDisabled is true for either Item or Link

### Help us reproduce the bug!

Can be reproduced by using the provided code snippet (note I am even hard coding true and it doesn't disable the link).

### Code Snippet

```typescript
       <Breadcrumbs.Container>
          <Breadcrumbs.Item >
            <Breadcrumbs.Separator />
            <Breadcrumbs.Link onClick={() => setStep(1)} isCurrent={step === 1}>Select Blueprint</Breadcrumbs.Link>
            <Breadcrumbs.Separator />
          </Breadcrumbs.Item>
          <Breadcrumbs.Item isDisabled={true}>
            <Breadcrumbs.Link onClick={() => setStep(2)} isCurrent={step === 2} isDisabled={true}>Select 2</Breadcrumbs.Link>
            <Breadcrumbs.Separator />
          </Breadcrumbs.Item>
        </Breadcrumbs.Container> 

```

### Additional information and screenshots

<img width="1216" alt="image" src="https://github.com/aws-amplify/amplify-ui/assets/11780366/4c1324fb-de32-49a6-a605-8d0193d6d241">

esauerbo:
Hi @regularpersonn thanks for raising this. Could you clarify exactly what the desired behavior is, e.g. are you thinking the `onClick` handlers shouldn't run if disabled?

regularpersonn:
[...]

PR #4783, opened 2023-11-27
fix: add disabled styling to breadcrumbs component 

esauerbo:
#### Issue #, if available
#4776 

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/ui/src/theme/css/component/breadcrumbs.scss
+++ b/packages/ui/src/theme/css/component/breadcrumbs.scss
@@ -15,6 +15,13 @@
   line-height: var(--amplify-components-breadcrumbs-item-line-height);
   color: var(--amplify-components-breadcrumbs-item-color);
   font-size: var(--amplify-components-breadcrumbs-item-font-size);
+  &[disabled] {
+    background-color: var(
+      --amplify-components-breadcrumbs__item-disabled-color
+    );
+    color: var(--amplify-components-breadcrumbs__item-disabled-color);
+    cursor: not-allowed;
+  }
 }
 
 .amplify-breadcrumbs__separator {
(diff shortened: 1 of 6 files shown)


packages/ui/src/theme/css/component/breadcrumbs.scss
```diff
@@ -15,6 +15,20 @@
   line-height: var(--amplify-components-breadcrumbs-item-line-height);
   color: var(--amplify-components-breadcrumbs-item-color);
   font-size: var(--amplify-components-breadcrumbs-item-font-size);
+  &[disabled] {
```
ioanabrooks, line 18: nit for consistency, think we use the `&--disabled` convention, is the disabled suffix not getting applied? if not we should fix that

packages/ui/src/theme/css/component/breadcrumbs.scss
```diff
@@ -15,6 +15,16 @@
...
   font-size: var(--amplify-components-breadcrumbs-item-font-size);
+  &--disabled {
+    cursor: not-allowed;
+    .amplify-breadcrumbs__link {
+      background-color: var(
+        --amplify-components-breadcrumbs__link-disabled-color
+      );
+      color: var(--amplify-components-breadcrumbs__link-disabled-color);
+      pointer-events: none;
+    }
```
ioanabrooks, line 26: Is this needed here? Should be covered by lines 45-51.

esauerbo: I guess it depends, the customer who filed the bug report mentioned wanting the ability to disable `Link` or `Item` so this allows you to pass the prop to either. But if we only want the prop to work with `Item` then this isn't needed.

esauerbo: Actually think I misunderstood, so without this you'd need to add `isDisabled` to `Breadcrumsb.Link` in addition to `Breadcrumsb.Item` otherwise the link would still be active. 

IMO it makes sense the way it is because you can also create the component like `<Breadcrumbs items=...` so it might be confusing if you could pass `isDisabled` in with `items` in only that case but you'd need to put it in both `Link` and `Item` if using `Breadcrumbs.Items`

hbuchel (commented):
Is this a fix or a feature request? I remember having some discussion with @dbanksdesign about the usefulness of adding "disabled" links to Breadcrumbs; namely that they should probably just be a text element not a link with a disabled attribute. Some things to consider:
- A link with a disabled attribute doesn't disable it the way it does for a button or input, and it doesn't convey any disabled state to assistive tech. If you [test a link](https://codepen.io/hbuchel/live/bGzxbjK) with a disabled attribute with a screen reader, nothing is announced about it being disabled.
- Using `pointer-events: none` in CSS still allows the link to be keyboard operable.

PR #4799, opened 2023-11-30
chore: move start screen skipping to ui and away from state machine

thaddmt:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/FaceLivenessDetectorCore.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/FaceLivenessDetectorCore.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useInterpret } from '@xstate/react';
+import { useActor, useInterpret } from '@xstate/react';
 import {
   livenessMachine,
   FaceLivenessDetectorCoreProps as FaceLivenessDetectorPropsFromUi,
@@ -23,7 +23,7 @@ export interface FaceLivenessDetectorCoreProps
 export default function FaceLivenessDetectorCore(
   props: FaceLivenessDetectorCoreProps
 ): JSX.Element {
-  const { components, config, displayText } = props;
+  const { components, config, displayText, disableStartScreen } = props;
   const currElementRef = React.useRef<HTMLDivElement>(null);
   const {
     hintDisplayText,
@@ -42,6 +42,21 @@ export default function FaceLivenessDetectorCore(
       },
     },
   });
+
+  const [state, send] = useActor(service);
+  const isStartView = state.matches('start');
+
+  const beginLivenessCheck = React.useCallback(() => {
+    send({
+      type: 'BEGIN',
+    });
+  }, [send]);
+
+  React.useLayoutEffect(() => {
+    if (disableStartScreen && isStartView) {
+      beginLivenessCheck();
+    }
+  }, [beginLivenessCheck, disableStartScreen, isStartView]);
 
   return (
     <View className={DETECTOR_CLASS_NAME} testId={DETECTOR_CLASS_NAME}>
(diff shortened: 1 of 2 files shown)


esauerbo (approved):

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCheck.tsx
```diff
@@ -38,6 +38,7 @@ interface LivenessCheckProps {
   streamDisplayText: Required<StreamDisplayText>;
   errorDisplayText: Required<ErrorDisplayText>;
   components?: FaceLivenessDetectorComponents;
+  beginLivenessCheck: () => void;
```
reesscot, line 41: Does this new prop affect customers, or is this only used internally?

thaddmt: only used internally

packages/react-liveness/src/components/FaceLivenessDetector/FaceLivenessDetectorCore.tsx
```diff
@@ -43,6 +43,21 @@ export default function FaceLivenessDetectorCore(
...
+  const [state, send] = useActor(service);
+  const isStartView = state.matches('start');
+
+  const beginLivenessCheck = React.useCallback(() => {
+    send({
+      type: 'BEGIN',
+    });
+  }, [send]);
+
+  React.useLayoutEffect(() => {
```
ioanabrooks, line 55: Why is `useLayoutEffect` needed here? (as opposed to `useEffect`)

packages/react-liveness/src/components/FaceLivenessDetector/FaceLivenessDetectorCore.tsx
```diff
@@ -43,6 +43,21 @@ export default function FaceLivenessDetectorCore(
     },
   });
 
+  const [state, send] = useActor(service);
+  const isStartView = state.matches('start');
+
+  const beginLivenessCheck = React.useCallback(() => {
```
ioanabrooks, line 49: Can we add a unit test for this?

packages/react-liveness/src/components/FaceLivenessDetector/FaceLivenessDetectorCore.tsx
thaddmt: Ah no particular reason, i think this was just copied from somewhere else in the code which required useLayoutEffect, good catch!

packages/react-liveness/src/components/FaceLivenessDetector/FaceLivenessDetectorCore.tsx
```diff
@@ -43,6 +43,21 @@ export default function FaceLivenessDetectorCore(
     },
   });
 
+  const [state, send] = useActor(service);
+  const isStartView = state.matches('start');
```
calebpollman, line 47: `disableStartScreen` and `isStartView` read as somewhat contradicting one another imo, and since the existing declarations of `isStartView` are declared as:
```
const isStartView = state.matches('start') || state.matches('userCancel');
```

Maybe we can have a variable name spaced as something like `shouldBeginLivenessCheck` , example:
```
const shouldBeginLivenessCheck = disableStartScreen && state.matches('start');
```

Which would simplify the logic inside/deps required in the callback passed to `useLayoutEffect`

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -237,12 +239,6 @@ export const LivenessCameraModule = (
     setIsCameraReady(true);
   };
 
-  const beginLivenessCheck = React.useCallback(() => {
-    send({
-      type: 'BEGIN',
-    });
-  }, [send]);
```
calebpollman, line 244: As noted in `FaceLivenessDetectorCore`, think the `beginLivenessCheck` declaration should remain here. Additionally, we don't need `useCallback` here, `beginLivenessCheck` is just being passed to a `Button` which is not an expensive re-render, and the `Button` is only rendered when `isStartView` is `true`

PR #4803, opened 2023-12-01
fix(liveness): add accessibility attributes to error modal

esauerbo:
#### Description of changes
Added `aria-labelledby` and `role` to the modal element
Screenshot 2023-12-01 at 3 09 39 PM

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
@@ -69,7 +69,12 @@ const renderToastErrorModal = (props: {
     <>
       <Flex className={LivenessClassNames.ErrorModal}>
         <AlertIcon ariaHidden variation="error" />
-        <Text className={LivenessClassNames.ErrorModalHeading}>{heading}</Text>
+        <Text
+          className={LivenessClassNames.ErrorModalHeading}
+          id="timeout-error-heading"
+        >
+          {heading}
+        </Text>
       </Flex>
       {message}
     </>
@@ -116,7 +121,7 @@ export const FaceLivenessErrorModal: React.FC<FaceLivenessErrorModalProps> = (
 
   return (
     <Overlay className={LivenessClassNames.OpaqueOverlay}>
-      <Toast>
+      <Toast ariaLabelledBy="timeout-error-heading" role="dialog">
         {children}
         <Flex justifyContent="center">
           <Button variation="primary" type="button" onClick={onRetry}>
--- a/packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
@@ -71,7 +71,7 @@ const renderToastErrorModal = (props: {
         <AlertIcon ariaHidden variation="error" />
         <Text
           className={LivenessClassNames.ErrorModalHeading}
-          id="timeout-error-heading"
+          id="error-heading"
         >
           {heading}
         </Text>
@@ -121,7 +121,7 @@ export const FaceLivenessErrorModal: React.FC<FaceLivenessErrorModalProps> = (
 
   return (
     <Overlay className={LivenessClassNames.OpaqueOverlay}>
-      <Toast ariaLabelledBy="timeout-error-heading" role="dialog">
+      <Toast ariaLabelledBy="error-heading" role="dialog">
         {children}
         <Flex justifyContent="center">
           <Button variation="primary" type="button" onClick={onRetry}>
--- a/packages/react-liveness/src/components/FaceLivenessDetector/shared/__tests__/FaceLivenessErrorModal.test.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/shared/__tests__/FaceLivenessErrorModal.test.tsx
@@ -26,12 +26,27 @@ describe('FaceLivenessErrorModal', () => {
         {renderErrorModal({ errorState })}
       </FaceLivenessErrorModal>
     );
-
-    expect(screen.getByText(serverHeaderText)).toBeInTheDocument();
-    expect(screen.getByText(serverMessageText)).toBeInTheDocument();
+    const timeoutModal = screen.getByRole('dialog');
+    expect(timeoutModal).toBeInTheDocument();
+    expect(timeoutModal).toHaveAttribute(
+      'aria-labelledby',
+      'timeout-error-heading'
+    );
   });
 
   it('should render the timeout message appropriately', () => {
+    const errorState = LivenessErrorState.TIMEOUT;
+    render(
+      <FaceLivenessErrorModal onRetry={() => {}}>
+        {renderErrorModal({ errorState })}
+      </FaceLivenessErrorModal>
+    );
+
+    expect(screen.getByText(timeoutHeaderText)).toBeInTheDocument();
+    expect(screen.getByText(timeoutMessageText)).toBeInTheDocument();
+  });
+
+  it('should render the timeout message with proper accessibility attributes', () => {
     const errorState = LivenessErrorState.TIMEOUT;
     render(
       <FaceLivenessErrorModal onRetry={() => {}}>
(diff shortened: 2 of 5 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
```diff
@@ -116,7 +121,7 @@ export const FaceLivenessErrorModal: React.FC<FaceLivenessErrorModalProps> = (
 
   return (
     <Overlay className={LivenessClassNames.OpaqueOverlay}>
-      <Toast>
+      <Toast ariaLabelledBy="error-heading" role="dialog">
```
hbuchel, line 124: Can you verify what this experience is with a screen reader? I think the intent is to expose this error message when it appears to the user so I would run through it with VoiceOver (really we should be doing that any time we're adding roles and adding ARIA attributes). I think we'll need a few changes for that to happen:

- This should probably be an `alertdialog` since it's preventative of the experience proceeding and has important information the user needs to know about the state of the UI. There is an example in the [APG here](https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/). I also like that they include the additional error text in aria-describedby. That means when the dialog appears it announce the title, the additional text in the dialog, and then the focusable element (in their case, the "no" button, in our case the "Try again" button). I think you'll want to make sure those elements always exist, though; I don't know if every error message has additional text.
- We'll need to move focus to the dialog when it's added to the DOM. In that example, they focus the button which I think would work for us, too.

So if you run through this in VoiceOver, I would expect if you fail the challenge (like ducking out of view of it):
- The dialog appears
- The "Try again" button is focused,
- VoiceOver announces "Time out" (or w/e the title is), followed by the additional text "Face didn't fit inside oval in time limit. Try again and completely fill the oval with face in it." (or w/e it is), then the text of the button "Try again"

hbuchel (commented):
Left a comment about the View prop type changes. Additionally, left some higher level comments about the ARIA changes with some suggestions and things to verify/test. The role and label you added would probably pass an automated test but I don't think they (alone) actually convey the context we want to assistive tech users.

packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
esauerbo: Added the `aria-describedby` property and tested with VoiceOver. Button focuses and VoiceOver announces the heading and dialog description then the text of the button.

thaddmt (approved):

packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
```diff
@@ -69,7 +69,12 @@ const renderToastErrorModal = (props: {
     <>
       <Flex className={LivenessClassNames.ErrorModal}>
         <AlertIcon ariaHidden variation="error" />
-        <Text className={LivenessClassNames.ErrorModalHeading}>{heading}</Text>
+        <Text
+          className={LivenessClassNames.ErrorModalHeading}
+          id="error-heading"
```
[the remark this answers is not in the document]
esauerbo: Yeah makes sense

packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
```diff
@@ -69,7 +69,12 @@ const renderToastErrorModal = (props: {
     <>
       <Flex className={LivenessClassNames.ErrorModal}>
         <AlertIcon ariaHidden variation="error" />
-        <Text className={LivenessClassNames.ErrorModalHeading}>{heading}</Text>
+        <Text
+          className={LivenessClassNames.ErrorModalHeading}
+          id="amplify-liveness-error-heading"
```
calebpollman, line 74: Should this be a `testId`? `id` reads style related imo

calebpollman: Looked up how `aria-labelledby` should work, please ignore the above

packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
```diff
@@ -116,8 +121,12 @@ export const FaceLivenessErrorModal: React.FC<FaceLivenessErrorModalProps> = (
...
   return (
     <Overlay className={LivenessClassNames.OpaqueOverlay}>
-      <Toast>
-        {children}
+      <Toast
+        aria-labelledby="amplify-liveness-error-heading"
+        aria-describedby="amplify-liveness-error-description"
+        role="alertdialog"
+      >
+        <div id="amplify-liveness-error-description">{children}</div>
```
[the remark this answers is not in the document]
calebpollman: Hmmmm, how does `aria-describedby` work if `children` is not a string?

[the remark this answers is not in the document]
esauerbo: Actually nvm to above we can put the id in the modal. But I think it will just go through any nested elements and read the text, and if there is none anywhere then does nothing. Like here it reads the heading and message
```
<>
      <Flex
        className={LivenessClassNames.ErrorModal}
        id="amplify-liveness-error-description"
      >
        <AlertIcon ariaHidden variation="error" />
        <Text
          className={LivenessClassNames.ErrorModalHeading}
          id="amplify-liveness-error-heading"
        >
          {heading}
        </Text>
      </Flex>
      {message}
    </>
```

packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
```diff
@@ -67,9 +67,17 @@ const renderToastErrorModal = (props: {
 
   return (
     <>
-      <Flex className={LivenessClassNames.ErrorModal}>
+      <Flex
+        className={LivenessClassNames.ErrorModal}
+        id="amplify-liveness-error-description"
```
calebpollman, line 72: It feels off to me that a "description" element is wrapping a "heading" element?

thaddmt (approved):

packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
```diff
@@ -69,9 +69,14 @@ const renderToastErrorModal = (props: {
...
-        <Text className={LivenessClassNames.ErrorModalHeading}>{heading}</Text>
+        <Text
+          className={LivenessClassNames.ErrorModalHeading}
+          id="amplify-liveness-error-heading"
+        >
+          {heading}
+        </Text>
       </Flex>
-      {message}
+      <div id="amplify-liveness-error-description">{message}</div>
```
calebpollman, line 79: Correct me if I am wrong, but it looks like `message` will always be a `string` so maybe we should use a `Text` element here? Also, is there a reason to use "description" instead of "message" in the `id`?

esauerbo: Yeah that would make sense. And not really, I was thinking it was a description of the error but I can change to message.

packages/react-liveness/src/components/FaceLivenessDetector/shared/FaceLivenessErrorModal.tsx
```diff
@@ -69,9 +69,14 @@ const renderToastErrorModal = (props: {
...
-        <Text className={LivenessClassNames.ErrorModalHeading}>{heading}</Text>
+        <Text
+          className={LivenessClassNames.ErrorModalHeading}
+          id="amplify-liveness-error-heading"
+        >
+          {heading}
+        </Text>
       </Flex>
-      {message}
+      <text id="amplify-liveness-error-message">{message}</text>
```
calebpollman, line 79: Can we use our `Text` here?

calebpollman (commented):
LGTM, approval pending update of the new `text` to a `Text` component

thaddmt (dismissed):

.changeset/plenty-snakes-try.md
```diff
@@ -0,0 +1,6 @@
+---
+"@aws-amplify/ui-react-liveness": patch
+"@aws-amplify/ui-react": patch
```
reesscot, line 3: Does this affect the ui-react package, or just ui-react-liveness?

reesscot, line 3:
```suggestion
```

esauerbo: It was just autogenerated

reesscot (approved):

hbuchel (approved):

PR #4804, opened 2023-12-02
feat: Text formatting in PhoneNumberField

lavr001:
#### Description of changes

Added some logic to format user's input in the PhoneNumberField component. More cases for other countries codes could be added as needed.

Tested out changes with multiple country codes from the new file `utils.ts`, all numbers were successfully formatted after being fully typed.

#### Checklist

- [x] PR description included


packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -28,10 +29,40 @@ const PhoneNumberFieldPrimitive: Primitive<PhoneNumberFieldProps, 'input'> = (
...
+      formatPhoneNumber(dialCode, (value as string) ?? defaultValue ?? '')
+    );
+  const onNumberChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const phoneNumber = event.target.value;
+    // format the phone number with the current dial code
+    const formatted = formatPhoneNumber(dialCode, phoneNumber);
+    setFormattedPhoneNumber(formatted);
+    onChange?.(event);
+  };
+  const onDialCodeChangeWrapper = (
```
reesscot, line 55: These functions will be recreated on every render

reesscot:
Please add unit tests

packages/react/src/primitives/PhoneNumberField/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+    case '+81':
+      formatted = cleaned.replace(/(\d{3})(\d{4})(\d{4})/, '$1 $2 $3');
+      break;
+    case '+91':
+      formatted = cleaned.replace(/(\d{5})(\d{5})/, '$1 $2');
+      break;
+    case '+234':
+      formatted = cleaned.replace(/(\d{3})(\d{4})(\d{4})/, '$1 $2 $3');
+      break;
+    // Add more cases for other country codes as needed
```
reesscot, line 41: How many country codes should we support?

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -28,10 +29,40 @@ const PhoneNumberFieldPrimitive: Primitive<PhoneNumberFieldProps, 'input'> = (
...
+    event: React.ChangeEvent<HTMLSelectElement>
+  ) => {
+    const code = event.target.value;
+    // Call the onDialCodeChange prop if it exists
+    setDialCode(code);
+    // Reformat the phone number with the new dial code
+    const formatted = formatPhoneNumber(code, formattedPhoneNumber);
+    setFormattedPhoneNumber(formatted);
```
reesscot, line 63: Will this be formatted on every keystroke? Can you share a GIF/video showing the UX?

packages/react/src/primitives/PhoneNumberField/utils.ts
lavr001: The related GItHub issue did not specify the amount of dial codes. Currently I've added some regex logic to support 10 dial codes, we can add more if needed OR we can use the libphonenumber-js library, so we don't need to add manually every single dial code, however I'm not entirely sure if that library won't cause any runtime failures or/and require additional testing.

packages/react/src/primitives/PhoneNumberField/utils.ts
```diff
@@ -0,0 +1,46 @@
+export const formatPhoneNumber = (
+  dialCode: string,
+  phoneNumberString: string
+): string => {
+  const cleaned = ('' + phoneNumberString).replace(/\D/g, '');
```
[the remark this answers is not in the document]
lavr001: Yes, to remove all non digit

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
lavr001: Sure, see the attached.

https://github.com/aws-amplify/amplify-ui/assets/19784835/756a19b8-27a5-4bbf-b8ec-8e5a58df7ccc

thaddmt: Do we have some default debounce logic here? I would have thought the same thing as Scott

thaddmt:
Make sure to add a changeset as well please!

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
lavr001: I believe the debounced logic means no on the fly formatting. The formatting will happen after the debounced, so if users type fast it will happen after they stop typing. 
Could you please provide more details about the desired debounce logic here?

calebpollman: Is there debouncing here? Don't see anything added in this PR, nor do i believe there is debouncing in the existing in the existing code

packages/react/src/primitives/PhoneNumberField/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+  let formatted: string;
+  switch (dialCode) {
+    case '+1':
+      formatted = cleaned.replace(/(\d{3})(\d{3})(\d{4})/, '($1) $2-$3');
+      break;
```
calebpollman, line 10: Where are these regex patterns coming from and what if A) I don't want this formatting B) I want a format of `'$1-$2-$3'` or `'$1.$2.$3'`?

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -56,6 +88,9 @@ const PhoneNumberFieldPrimitive: Primitive<PhoneNumberFieldProps, 'input'> = (
       isDisabled={isDisabled}
       isReadOnly={isReadOnly}
       onInput={onInput}
+      // Pass the onChange function to the TextField and use the formattedPhoneNumber state as the value
+      onChange={onNumberChange}
+      value={formattedPhoneNumber}
```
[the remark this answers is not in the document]
calebpollman: Also think this handling breaks controlled usage of the component since we are ignoring `value`?

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
```diff
@@ -28,10 +29,40 @@ const PhoneNumberFieldPrimitive: Primitive<PhoneNumberFieldProps, 'input'> = (
...
+    // format the phone number with the current dial code
+    const formatted = formatPhoneNumber(dialCode, phoneNumber);
+    setFormattedPhoneNumber(formatted);
+    onChange?.(event);
+  };
+  const onDialCodeChangeWrapper = (
+    event: React.ChangeEvent<HTMLSelectElement>
+  ) => {
+    const code = event.target.value;
+    // Call the onDialCodeChange prop if it exists
```
calebpollman, line 59: Huh?

.changeset/new-fans-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aws-amplify/ui-react": patch
```
calebpollman, line 2: Feature commits should be a `minor` change, not a `patch`

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
calebpollman: Think we are trying to clarify if there is debouncing happening in your example video, not requesting it. The lack of formatting until you have completed typing gives the appearance that there is some form of debounce happening, can you just clarify if the example video includes debouncing or not?

lavr001: There is no debouncing, the phone number gets formatted after being fully typed. I do agree that it might be not intuitive for the user to wait until the end to see the number formatted. I can work on updating the logic so it would be formatted on the fly while you're typing, to make the formatting more intuitive and interactive, if it makes sense

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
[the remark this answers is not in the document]
lavr001: I have a logic that removes all non digit to avoid any symbols gets entered and there is `+1` already included in the dropdown of dial codes, so if you enter `8001231212` it will be formatted to `(800) 123-1212`

packages/react/src/primitives/PhoneNumberField/utils.ts
lavr001: Are you suggesting we should have different formats for the same dial code?

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
lavr001: This is used to detect when dialCode changes to apply the formatting with the new Code, you can see it in action in the video I attached above.

calebpollman: FYI code comments are usually placed directly above what the comment is referencing, suggest applying that here

packages/react/src/primitives/PhoneNumberField/utils.ts
calebpollman: Suggesting that the current approach is inflexible for consumers and will be difficult to maintain from our side. What can we do that will allow the scenarios outlined in A and B above to be handled?

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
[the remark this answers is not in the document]
calebpollman: Say I currently have a form in my application that uses a `PhoneNumberField`, using the example values above the value of the field submitted to the form would go from `8001231212` to `(800) 123-1212` which is a breaking change right?

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
calebpollman: Would expect the formatting on each change event

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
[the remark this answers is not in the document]
lavr001: Yes, I think so, any formatting of the `8001231212` would be a breaking change from the data that gets submitted with a form perspective

packages/react/src/primitives/PhoneNumberField/utils.ts
lavr001: Scenario A: we can add a new prop `isFormatted` that could be off by default, so users can control if they want to format the phone number or not.

Scenario B: I think we should not have different formats for let's say the same dial code, because then the regex logic would be very complex, unless you're ok with using a third party library like `libphonenumber` which is a library that parses, formats, stores and validates international phone numbers.

Let me know if you have other suggestions.

packages/react/src/primitives/PhoneNumberField/utils.ts
```diff
@@ -0,0 +1,46 @@
+export const formatPhoneNumber = (
+  dialCode: string,
+  phoneNumberString: string
+): string => {
+  const cleaned = ('' + phoneNumberString).replace(/\D/g, '');
+  let formatted: string;
+  switch (dialCode) {
```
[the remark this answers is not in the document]
lavr001: Created a separate `constants.ts` file to make it easier to support more dial codes in the future

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
lavr001: Wrapped up `onValueChange` and `onDialCodeChangeWrapper` in the `useCallback` function

packages/react/src/primitives/PhoneNumberField/utils.ts
lavr001: Unit tests have been added

packages/react/src/primitives/PhoneNumberField/PhoneNumberField.tsx
lavr001: Added the formatting on each change event

PR #4810, opened 2023-12-05
remove dynamic imports from primitive docs

jacoblogan:
#### Description of changes
This PR removes the dynamic imports from the primitive component docs.  This is in an effort to move to a fully static rendered docs site and improve the Content Layout Shift of the existing docs.

Staging Site: https://static-rendering.d1beoyo3clqxv6.amplifyapp.com/

When loading a primitive component page the page content can been seen as part of the initial page response in the network tab.

#### Checklist

- [x] PR description included


docs/src/components/FilterChildren.tsx
```diff
@@ -0,0 +1,15 @@
+import { useRouter } from 'next/router';
+
+export const FilterChildren = ({ children, filterKey = '' }) => {
```
dbanksdesign, line 3: Do we need 2 separate components? Could we just have a single component? I don't think this one would ever be used on its own right?

docs/src/pages/[platform]/components/input/index.page.mdx
```diff
@@ -27,7 +29,11 @@ export async function getStaticProps() {
 <PageTabLayout 
   tabComponents={[{
     title: "Documentation", 
-    children: <Fragment>{({ platform }) => import(`./${platform}.mdx`)}</Fragment>
+    children: <>
+      <InlineFilter filters={["react"]}>
```
Comment (reesscot) [inline docs/src/pages/[platform]/components/input/index.page.mdx:33]: Is this necessary since we don't have anything but React here? I don't think Next would generate pages based on the static paths?

docs/src/components/InlineFilter.tsx
```diff
@@ -0,0 +1,21 @@
+import { Fragment } from 'react';
+import { FilterChildren } from './FilterChildren';
+
+type InlineFilterProps = {
+  children: React.ReactNode;
+  filters: string[];
+};
+
+export const InlineFilter = ({ filters, children }: InlineFilterProps) => {
```
dbanksdesign, line 9: Can we add a comment on this component to what it is doing and how to use it? I have a feeling in 2 years we are going to be like "wtf is this thing doing"

docs/src/components/InlineFilter.tsx
```diff
@@ -0,0 +1,21 @@
...
+  if (!filters || !Array.isArray(filters) || filters.length < 1) {
+    return <></>;
+  }
+
+  const filteredChildren: Array<React.JSX.Element> = [];
+
+  filters.forEach((filter) => {
+    filteredChildren.push(<Fragment key={filter}>{children}</Fragment>);
```
dbanksdesign, line 17: I don't think I get this, if `filters` is an array, this rendering _all_ the children n times where n=filters.length. So if I have `<InlineFilter filters=["react","react-native"]>` the entire contents would rendered for both, which I don't think thats what we want right?

docs/src/pages/[platform]/components/input/index.page.mdx
Comment (jacoblogan) [reply docs/src/pages/[platform]/components/input/index.page.mdx]: Nope the `InlineFilter` isn't needed here because there is just the single supported framework.  I've removed it and am just returning the react page but once these components support additional frameworks we'll need to add the `<InlineFilter>` to only display the content for the current framework.

docs/src/components/FilterChildren.tsx
jacoblogan: Nope 2 separate components are not needed for this, they have been combined and simplified into a single component.

docs/src/components/InlineFilter.tsx
```diff
@@ -1,21 +1,40 @@
...
-  filters: string[];
+  /**
+   * List of platforms that will render this content. If the current platform
+   * is not in this list then the content will not be render
+   */
+  filters: (Framework | 'all')[];
 };
 
+/**
+ *
```
reesscot, line 14: Use 
```suggestion
 * Used to only show content for specific platforms
```

docs/src/components/InlineFilter.tsx
```diff
@@ -1,21 +1,40 @@
...
+  let filterKey = '';
 
-  filters.forEach((filter) => {
-    filteredChildren.push(<Fragment key={filter}>{children}</Fragment>);
-  });
+  if ('platform' in router.query) {
+    filterKey = router.query.platform as string;
+  }
+
+  for (let i = 0; i < filters.length; i++) {
```
reesscot: nit: could make use of for of here, so we don't have to track indexes:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

reesscot (approved):
LGTM

reesscot (dismissed):

dbanksdesign (approved):
LGTM!

great job

reesscot (approved):

dbanksdesign (dismissed):

docs/src/pages/[platform]/components/autocomplete/index.page.mdx
```diff
@@ -28,7 +30,7 @@ export async function getStaticProps() {
 <PageTabLayout 
   tabComponents={[{
     title: "Documentation", 
-    children: <Fragment>{({ platform }) => import(`./${platform}.mdx`)}</Fragment>
+    children: <ReactPage />
```
Comment (calebpollman) [inline docs/src/pages/[platform]/components/autocomplete/index.page.mdx:33]: From a cursory glance, it looks like every primitive file repeats:
```

<PageTabLayout 
  tabComponents={[{
    title: "Documentation", 
    children: <ReactPage />
  },
  {
    title: "Props", 
    children: <PropsTableTab 
                componentName={frontmatter.title} 
                PropsData={data} 
                htmlElement={frontmatter.htmlElement} 
                mdnUrl={frontmatter.mdnUrl}
              />
  }]}
/>
```

Is that correct and if so can we just have a single component reused in each of these files?

calebpollman (commented):
Thanks for doing this @jacoblogan! Left some questions/feedback, additionally would like to see unit tests added for `InlineFilter`

docs/src/components/InlineFilter.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  if (!filters || !Array.isArray(filters) || filters.length < 1) {
+    return null;
+  }
+
+  let filterKey = '';
+
+  if ('platform' in router.query) {
+    filterKey = router.query.platform as string;
+  }
```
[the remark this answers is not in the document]
jacoblogan: added these checks to the early return statement above so here we are guaranteed that router.query exists.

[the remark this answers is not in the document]
calebpollman: Would prefer not to see casting here

docs/src/pages/[platform]/components/autocomplete/index.page.mdx
Comment (calebpollman) [reply docs/src/pages/[platform]/components/autocomplete/index.page.mdx:32]: @jacoblogan Can you please address this question?

Comment (jacoblogan) [reply docs/src/pages/[platform]/components/autocomplete/index.page.mdx:32]: Yes every primitive file is repeating this pattern but the abstraction is already in place with `PageTabLayout` while we could remove a little more by creating another abstraction it feels like this wouldn't be worth it because `<ReactPage />`, `data`, and all of the `frontmatter` information is coming from the page itself so these would still need to be passed into the newly created component.

docs/__tests__/inline-filter.test.tsx
```diff
@@ -0,0 +1,136 @@
+import { render, screen } from '@testing-library/react';
+import { InlineFilter } from '../src/components/InlineFilter';
+// import * as React from 'react';
```
calebpollman, line 3: Commented out code
```suggestion
```

docs/src/components/Layout/index.tsx
```diff
@@ -54,6 +54,7 @@ export default function Page({
         }))
       );
     });
+    updateHeaders(); // with static rendering the mutation observer is no longer triggered on initial page view because the content has already been rendered
```
jacoblogan, line 57: This line was added because we need to generate the initial page headings, this was being handled previously by the MutationObserver after the dynamic content was loaded into the page, but now that the dynamic content has been removed the observer is no longer triggering on the initial load.

calebpollman: Do we still need `observer`?

docs/__tests__/inline-filter.test.tsx
```diff
@@ -0,0 +1,135 @@
+import { render } from '@testing-library/react';
```
calebpollman, line 1: Let's colocate this file/directory as a sibling to `InlineFilter` and use PascalCase for the test file name, e.g.
```
components/
  InlineFilter.tsx
  __tests__/
    InlineFilter.test.tsx
```

docs/__tests__/inline-filter.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    );
+    const reactNode = container.querySelector('#react-content');
+    const filter1 = container.querySelector('#filter1');
+    const filter2 = container.querySelector('#filter2');
+    expect(reactNode).not.toBeNull();
+    expect(filter1).not.toBeNull();
+    expect(filter2).toBeNull();
+  });
+
+  it('should not render anything if no or an invalid filter is passed in', () => {
```
calebpollman, line 82: Rather than declaring multiple cases within the same test, each case be isolated using `it.each`, example:
```
it.each([undefined, true]).('renders nothing when provided with %s as the provided value of filters`, (filters) => {
  const { container } = render(
     <InlineFilter filters={filters}>
        <span id="filtered">Content</span>
      </InlineFilter>
  );

  const filtered = container.querySelector('#filtered');
  expect(filtered).toBeNull();
});
```

docs/src/components/InlineFilter.tsx
```diff
@@ -0,0 +1,44 @@
...
+  let filterKey = '';
+
+  if ('platform' in router.query) {
+    filterKey = router.query.platform;
+  }
+
+  const showContent = filters.some(
+    (filter) => filter === filterKey || filter === 'all'
+  );
+  return showContent ? children : null;
```
calebpollman, line 43: Think this can be simplifed to:
```suggestion
  const platform = router?.query?.platform;
  
  const showContent =
    typeof platform === 'string' &&
    Array.isArray(filters) &&
    filters.some((filter) => filter === platform || filter === 'all');
  return showContent ? children : null;
```

docs/src/components/Layout/index.tsx
calebpollman: Is it problematic to call it here and from `observer`?

docs/src/pages/[platform]/components/autocomplete/index.page.mdx
Comment (calebpollman) [reply docs/src/pages/[platform]/components/autocomplete/index.page.mdx:32]: Messaged you offline

docs/src/components/Layout/index.tsx
jacoblogan: I don't think its problematic calling `updateHeaders` in both places because this one is being called with `useLayoutEffect` at the same time as the observer is being created but the observer doesn't trigger until the page content is changed at which time updateHeaders is called again.

calebpollman (approved):

dbanksdesign (approved):

PR #4816, opened 2023-12-06
chore(docs): bump glob version used in docs 

esauerbo:
#### Description of changes

Added glob as a dependency in packages where it was being used to keep it at latest version 
`docs/src/utils/getContentPaths.ts`
`packages/ui/scripts/generateCSS.ts`

#### Checklist

- [ ] PR description included


packages/ui/package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "csstype": "^3.1.1",
+    "glob": "^10.3.10",
```
reesscot, line 48: I think it's a dev dependency in ui package, not a main dep

docs/package.json
```diff
@@ -65,6 +65,7 @@
     "canvas": "^2.11.2",
     "dotenv-safe": "^8.2.0",
     "eslint-config-next": "^13.5.5",
+    "glob": "^10.3.10",
```
reesscot, line 68: How is it used in docs? I think this is only used during build. I think it's a dev dependency here too

reesscot (commented):
I think it's a devDependency for both of these. Also need to run yarn install to update yarn lock file

docs/package.json
esauerbo: Yeah it's just used in the one file `docs/src/utils/getContentPaths.ts`, did you mean if it's only used in build we don't need to declare it here?

docs/__tests__/__snapshots__/sitemap.test.ts.snap
```diff
@@ -1,166 +1,43 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Sitemap Snapshot 1`] = `
-"/, 
-/android,
```
reesscot, line 5: looks like upgrading glob broke our sitemap script?

esauerbo: Actually I don't think this is related to this change specifically, not sure what caused it but I noticed the sitemap snapshot was failing in a few other docs prs

docs/src/utils/getContentPaths.ts
```diff
@@ -10,8 +10,7 @@ export type GetContentPaths = (pattern?: string) => Promise<string[]>;
 export const getContentPaths: GetContentPaths = async (
```
reesscot, line 10: Is it still an async function if we use globSync?

reesscot (approved):
LGTM!

ioanabrooks (approved):

PR #4833, opened 2023-12-09
chore(e2e): add separate workflows for hotfix prs

esauerbo:
#### Description of changes

Blocks docs e2e tests from running PRs to hotfix branch 

#### Checklist

- [ ] PR description included

### DIFF
--- a/.github/workflows/reusable-e2e-hotfix.yml
+++ b/.github/workflows/reusable-e2e-hotfix.yml
@@ -1,0 +1,610 @@
+# Description: this is a reusable workflow that runs e2e tests.
+#
+# See `./test-prs.yml` to see an example usage.
+
+on:
+  workflow_call:
+    inputs:
+      commit:
+        # commit to run tests against
+        required: true
+        type: string
+      repository:
+        # repository to run tests against (possibly a fork of amplify-ui)
+        required: true
+        type: string
+      skip-changed-packages-check:
+        # skip checking changed packages job if true
+        required: false
+        default: 'false'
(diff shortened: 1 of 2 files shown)


thaddmt (dismissed):

calebpollman (commented):
Are these commits meant to be stay in the repo after the patch?

esauerbo:
> Are these commits meant to be stay in the repo after the patch?

Is the hotifx branch only used for patching older versions? If so I think we can keep this for future hotifxes bc the docs e2e test would fail on anything that isn't the latest version. But if we want to use the hotfix branch for latest then we probably should take this out after patching.

calebpollman:
> > Are these commits meant to be stay in the repo after the patch?
> 
> Is the hotifx branch only used for patching older versions? If so I think we can keep this for future hotifxes bc the docs e2e test would fail on anything that isn't the latest version. But if we want to use the hotfix branch for latest then we probably should take this out after patching so we can make sure to test docs.

A bit concerned abt drift between these workflow files and the files they are copied from, for example the commented out code for the RN e2e tests in [.github/workflows/test-internal-prs-hotfix.yml](https://github.com/aws-amplify/amplify-ui/pull/4833/files#diff-7c4bd0b20025d318b6cfa8a008f7465c3f691cd961799869cd2f1339788c4299) could be missed  when it is uncommented in the originating file. Would prefer an approach where we can leverage the same workflow for both hot fix and standard PRs

thaddmt (dismissed):

PR #4855, opened 2023-12-15
Static rendering getting started

jacoblogan:
#### Description of changes
remove dynamic imports from all getting started pages

#### Checklist

- [ ] PR description included


docs/src/pages/[platform]/getting-started/installation/dependencies.mdx
```diff
@@ -0,0 +1,185 @@
...
+<InlineFilter filters={['react']}>
+  Install `@aws-amplify/ui-react` with [npm](https://www.npmjs.com/) or [yarn](https://yarnpkg.com/):
+
+  <InstallScripts framework="react" />
+
+  After adding the `aws-amplify` and `@aws-amplify/ui-react` dependencies you are now ready to add any of our [components](../components) to your application.
+</InlineFilter>
+
+<InlineFilter filters={['swift']}>
+  Add info here on how to install FaceLivenessDetector for Swift.
```
Comment (calebpollman) [inline docs/src/pages/[platform]/getting-started/installation/dependencies.mdx:145]: Know this line copy pasta but it reads as a todo, so i looked in to whether it is actually currently rendered or not and unable to find it on the current docs (might just be me missing it). How are we verifying whether theses changes are going to lead to any content updates?

Comment (jacoblogan) [reply docs/src/pages/[platform]/getting-started/installation/dependencies.mdx:145]: You are correct that this particular line is never rendered.  The getting-started instillation section doesn't have `swift` as one of its supported platforms, but there were swift files that would have appeared if swift was ever added as a supported platform so to be consistent I moved it into this file under a filter that will currently just never display.
To your larger question of how these changes are being validated, I don't have a good answer for that.  I've been manually validating that content isn't changing but I know that is not ideal and can be very error prone.  I'm open to suggestions for validating this content.

docs/src/pages/[platform]/getting-started/troubleshooting/troubleshooting.react.mdx
```diff
@@ -238,17 +239,11 @@ When working with a [Vite](https://vitejs.dev) project you must make a few modif
     <Tabs.Item value="JavaScript">JavaScript</Tabs.Item>
   </Tabs.List>
   <Tabs.Panel value="TypeScript">
-    <Fragment>
-      {({ platform }) => import(`./shared/vite-polyfill.mdx`)}
-    </Fragment>
-    <Fragment>
-      {({ platform }) => import(`./shared/vite-polyfill-ts.mdx`)}
-    </Fragment>
+    <VitePolyfill />
```
Comment (calebpollman) [inline docs/src/pages/[platform]/getting-started/troubleshooting/troubleshooting.react.mdx:242]: Should this be here?

calebpollman (commented):
LGTM, just left a question regarding places where `VitePolyfill` and `VitePolyfillTS` are rendered together

docs/src/pages/[platform]/getting-started/troubleshooting/troubleshooting.react.mdx
Comment (jacoblogan) [reply docs/src/pages/[platform]/getting-started/troubleshooting/troubleshooting.react.mdx:242]: Yes the `<VitePolyfill />` component is currently being used for both `js` and `ts` and then the `<VitePolyfillTS />` component is additional information that is only relevant for `ts`.  This could be refactored so that only one or the other is used but instead I decided to just follow the pattern that is already in use.

Comment (calebpollman) [reply docs/src/pages/[platform]/getting-started/troubleshooting/troubleshooting.react.mdx:242]: Can we add `VitePolyfill` inside `VitePolyfillTS`? Think the current approach might be confusing in the future since the names read as being two variants

jacoblogan:
No longer needed

Issue #1660, opened 2022-04-11
Unhandled Authenticator Route - Please open an issue: null


nmlynch94:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Create React App

### Please describe your bug.

When attempting to create a user in an existing Cognito User Pool, I am receiving the error: 

```
'Unhandled Authenicator route – please open an issue: '
```

And redirected to the login page instead of being presented the appropriate MFA screen

### What's the expected behaviour?

Am expecting to be presented the screen to setup a User's MFA

### Help us reproduce the bug!

To make sure the issue wasn't with my application, I started a new app with `npx create-react-app my-app ` using node `14.19.0`. 

Next, I followed the getting started guide here: https://ui.docs.amplify.aws/getting-started/installation?platform=react as well as the Authenticator backend setup guide here: https://docs.amplify.aws/lib/auth/start/q/platform/js/#re-use-existing-authentication-resource. 

Next, I simply tried to create a user with the Create Account tab in the Authenticator component after running with `npm start`. After getting my confirmation code via Email and entering it, it returns to the login screen with the error in the console.

### Code Snippet

Here is the package.json I ended up with:

```
{
  "name": "my-app",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "@aws-amplify/ui-react": "^2.15.1",
    "@testing-library/jest-dom": "^5.16.4",
    "@testing-library/react": "^12.1.4",
    "@testing-library/user-event": "^13.5.0",
    "aws-amplify": "^4.3.19",
    "react": "^18.0.0",
    "react-dom": "^18.0.0",
    "react-scripts": "5.0.0",
    "web-vitals": "^2.1.4"
  },
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test",
    "eject": "react-scripts eject"
  },
  "eslintConfig": {
    "extends": [
      "react-app",
      "react-app/jest"
    ]
  },
  "browserslist": {
    "production": [
[...]

PR #4892, opened 2024-01-08
fix(authenticator): unhandled authenticator route console warning 

esauerbo:
#### Description of changes
Adding previously unhandled cases to routing logic 

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/1660

#### Checklist

- [ ] PR description included


packages/ui/src/helpers/authenticator/getRoute.ts
```diff
@@ -39,18 +41,20 @@ export const getRoute = (
...
       /**
        * This route is needed for autoSignIn to capture both the
        * autoSignIn.pending and the resolved states when the
        * signIn actor is running.
        */
       return 'transition';
+    case state?.matches('signInActor.runActor'):
+      return 'signIn';
+    case state?.matches('signUpActor.runActor'):
+      return 'signUp';
```
esauerbo, line 54: Putting these at the end so they don't take precedence over more relevant cases (e.g. `confirmSignUp` also has `signUpActor.runActor` state`

packages/ui/src/helpers/authenticator/getRoute.ts
```diff
@@ -17,16 +17,18 @@ export const getRoute = (
     case state.matches('authenticated'):
       return 'authenticated';
     case actorState?.matches('confirmSignUp'):
-    case actorState?.matches('confirmSignUp.resendConfirmationCode'):
```
esauerbo, line 20: Don't see this being used anymore

packages/ui/src/helpers/authenticator/getRoute.ts
calebpollman: Rather than use `signUpActor.runActor` as the case here can we look at the value of `actorState`? The fact that other steps (such as the noted `confirmSignUp`) have a state value of `signUpActor.runActor` makes this solution fragile IMO

esauerbo: Yeah so all of the other cases I added are values of `actorState` where state is either `signUpActor.runActor` or `signInActor.runActor`. 

I included these because some customers who saw this error didn't give steps for reproduction and our current error message doesn't show `actorState` so I can't tell if the other cases are exhaustive. I did update the error message so maybe we can take these out and if anyone still sees the warning we can add a more specific case?

packages/ui/src/helpers/authenticator/__tests__/getRoute.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        return this.value === targetState;
+      },
+    } as AuthMachineState;
+
+    mockActorState = {
+      value: 'unknown',
+      matches: function (this: AuthActorState, targetState: string) {
+        return this.value === targetState;
+      },
+    } as AuthActorState;
```
calebpollman, line 25: Since `state` and `actorState` are xstate utils, prefer creating spies over mocks that could potentially drift from their base implementation

packages/ui/src/helpers/authenticator/getRoute.ts
```diff
@@ -39,18 +41,14 @@ export const getRoute = (
     case actorState?.matches('confirmVerifyUserAttribute'):
       return 'confirmVerifyUser';
     case state.matches('getCurrentUser'):
-    case actorState.matches('fetchUserAttributes'):
+    case actorState?.matches('fetchUserAttributes'):
```
calebpollman, line 44: Why is the optional chaining now required here?

esauerbo: Didn't cause any issues for me but each of the other cases for actorState use optional chaining so assumed it was a mistake

packages/ui/src/helpers/authenticator/__tests__/getRoute.test.ts
esauerbo: Discussed offline and decided to create state/actor state within each test with both values set to what they would actually be if running.

calebpollman (approved):
🚢

thaddmt (dismissed):
LGTM, make sure to add a changeset

thaddmt (approved):

calebpollman (approved):

PR #4900, opened 2024-01-10
fix: reduce re-renders on hint component

thaddmt:
#### Description of changes
- Previously the hint component was re-rendering the hint message when checking face is on screen or when face is far enough, this pr makes it so we show the same hint for both those states
- Removed a unnecessary function for rendering the toast component as it was also causing extra renders

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
@@ -107,7 +107,7 @@ export const Hint: React.FC<HintProps> = ({ hintDisplayText }) => {
   }
 
   if (errorState ?? (isCheckFailed || isCheckSuccessful)) {
-    return;
+    return null;
   }
 
   if (!isRecording) {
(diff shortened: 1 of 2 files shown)


esauerbo (approved):

packages/react-liveness/src/components/FaceLivenessDetector/shared/Hint.tsx
```diff
@@ -95,113 +95,104 @@ export const Hint: React.FC<HintProps> = ({ hintDisplayText }) => {
...
+          {hintDisplayText.hintCenterFaceInstructionText}
+        </VisuallyHidden>
+        <DefaultToast text={hintDisplayText.hintCenterFaceText} isInitial />
+      </>
+    );
+  }
+
+  if (errorState ?? (isCheckFailed || isCheckSuccessful)) {
+    return;
+  }
```
calebpollman, line 111: Think we need to return `null` here? Pretty sure React below 18 doesn't gracefully handle `undefined` returned from JSX

esauerbo (approved):

calebpollman (approved):

PR #5047, opened 2024-03-07
chore(liveness/nolight): pass in multiple challenges

esauerbo:
#### Description of changes

- Updates the sdk to the newer version which accepts multiple challenge types, and pass in multiple challenges in the StartFaceLivenessSession API call. 

- In the state machine, retrieves the selected challenge type from the websocket and event and adds it to the context. Add some utils related to this.

- Update the `CustomWebSocketFetchHandler` to the later version included with the new sdk. (The current version of this handler on main is copied directly from the sdk; it was updated in the later version so re-copied it over.)

- Refactor `liveness.ts` to work with multiple challenge types.

*This code won't work on it's own. It sends + receives the challenge type, checks for the challenge type, and gets the session information and oval details, but will need the no-light state machine to prevent it from entering the light flashing stage (where it errors out currently). 

#### Checklist

- [ ] PR description included


packages/react-liveness/src/components/FaceLivenessDetector/service/machine/index.ts
```diff
@@ -1289,7 +1301,15 @@ const responseStreamActor = async (callback: StreamActorCallback) => {
   try {
     const stream = await responseStream;
     for await (const event of stream) {
-      if (isServerSesssionInformationEvent(event)) {
+      if (isChallengeEvent(event)) {
```
esauerbo, line 1304: New SDK has new `ChallengeEvent` where we can retrieve the challenge type.

PR #5138, opened 2024-04-08
fix(react-liveness): remove retry status code to allow for timeout event

hbuchel:
#### Description of changes

This removes the 500 status code we were previously setting during timeout errors because it is causing the AWS SDK to retry with expired session id. It also adds an error message "Websocket connection timeout". 

Because we are no longer sending a '500' status, the SDK will no longer retry during it's 60 second timeout, and instead we will timeout after 10 seconds.

Currently, the timeout will throw a generic `error: Error`:

Screenshot 2024-04-09 at 2 22 00 PM

Now it will show `Error: Websocket connection timeout`

Screenshot 2024-04-10 at 2 42 49 PM

In Chrome, I added a 70k ms latency network throttle:

Screenshot 2024-04-09 at 2 10 38 PM

Then I enabled this throttle before clicking "Start video check". The connection should eventually timeout and throw an error in the console:
```
{
    "state": "SERVER_ERROR",
    "error": {
        "name": "ValidationException",
        "$fault": "client",
        "$metadata": {},
        "message": "End event timeoutExceeded"
    }
}
```

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/createStreamingClient/CustomWebSocketFetchHandler.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/createStreamingClient/CustomWebSocketFetchHandler.ts
@@ -159,14 +159,9 @@ export class CustomWebSocketFetchHandler {
     socket: WebSocket,
     connectionTimeout: number
   ): Promise<void> {
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve) => {
       const timeout = setTimeout(() => {
         this.removeNotUsableSockets(socket.url);
-        reject({
-          $metadata: {
-            httpStatusCode: 500,
-          },
-        });
       }, connectionTimeout);
 
       socket.onopen = () => {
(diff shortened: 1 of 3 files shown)


reesscot:
nit on PR description:
This removes the 500 status code we were previously setting during timeout errors because it is causing the AWS SDK to retry with expired session id.

hbuchel:
> nit on PR description: This removes the 500 status code we were previously setting during timeout errors because it is causing the AWS SDK to retry with expired session id.

thanks, updated!

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/createStreamingClient/CustomWebSocketFetchHandler.ts
```diff
@@ -159,14 +159,9 @@ export class CustomWebSocketFetchHandler {
     socket: WebSocket,
     connectionTimeout: number
   ): Promise<void> {
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve) => {
       const timeout = setTimeout(() => {
         this.removeNotUsableSockets(socket.url);
-        reject({
```
reesscot, line 165: Not sure we want to fully remove the reject? Did we try just removing the metadata? Does the error still come through?

PR #5136, opened 2024-04-05
chore(gen2-storage): Update StorageImage API

esauerbo:
#### Checklist

- [ ] Have read the [Pull Request Guidelines](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md)
- [ ] PR description included
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] `yarn test` passes and tests are updated/added
- [ ] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5140, opened 2024-04-10
chore(ge2 storage): Add useGetUrl API

esauerbo:
#### Description of changes
- Add a `useGetUrl` api to be used by `StorageImage` component in place of `useStorageURL`. 
- This makes the same call to Storage `getUrl` as `useStorageURL`, but accepts either path or key props. 
- Logic for converting the url to a string and conditionally using the `fallbackUrl` will be moved to `StorageImage`.
```
interface UseStorageUrlOutput {
   url: string | undefined
}
```
->
```
interface UseGetUrlOutput {
  isLoading: boolean;
  url: URL | undefined;
  expiresAt: Date | undefined;
}
```

From https://github.com/aws-amplify/amplify-ui/pull/5136 

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
+++ b/packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
@@ -12,12 +12,12 @@ const onError = jest.fn();
 const KEY_INPUT: UseGetUrlInput = {
   key: 'file.jpg',
   options: { accessLevel: 'guest' },
-  onError: onError,
+  onError,
 };
 
 const PATH_INPUT: UseGetUrlInput = {
   path: 'guest/file.jpg',
-  onError: onError,
+  onError,
 };
 
 describe('useGetUrl', () => {
@@ -30,13 +30,14 @@ describe('useGetUrl', () => {
     it('should return a Storage URL', async () => {
       getUrlSpy.mockResolvedValue({ url, expiresAt: new Date() });
 
-      const { onError, ...getUrlParams } = KEY_INPUT;
-
       const { result, waitForNextUpdate } = renderHook(() =>
         useGetUrl(KEY_INPUT)
       );
 
-      expect(getUrlSpy).toHaveBeenCalledWith(getUrlParams);
+      expect(getUrlSpy).toHaveBeenCalledWith({
+        key: KEY_INPUT.key,
+        options: KEY_INPUT.options,
+      });
       expect(result.current.isLoading).toBe(true);
       expect(result.current.url).toBe(undefined);
 
@@ -50,14 +51,16 @@ describe('useGetUrl', () => {
 
     it('should invoke onStorageGetError when getUrl fails', async () => {
       const customError = new Error('Something went wrong');
-      const { onError, ...getUrlParams } = KEY_INPUT;
       getUrlSpy.mockRejectedValue(customError);
 
       const { result, waitForNextUpdate } = renderHook(() =>
         useGetUrl(KEY_INPUT)
       );
 
-      expect(getUrlSpy).toHaveBeenCalledWith(getUrlParams);
+      expect(getUrlSpy).toHaveBeenCalledWith({
+        key: KEY_INPUT.key,
+        options: KEY_INPUT.options,
+      });
 
       // Next update will happen when getUrl resolves
       await waitForNextUpdate();
@@ -73,13 +76,11 @@ describe('useGetUrl', () => {
     it('should return a Storage URL', async () => {
       getUrlSpy.mockResolvedValue({ url, expiresAt: new Date() });
 
-      const { onError, ...getUrlParams } = PATH_INPUT;
-
       const { result, waitForNextUpdate } = renderHook(() =>
         useGetUrl(PATH_INPUT)
       );
 
-      expect(getUrlSpy).toHaveBeenCalledWith(getUrlParams);
+      expect(getUrlSpy).toHaveBeenCalledWith({ path: PATH_INPUT.path });
       expect(result.current.isLoading).toBe(true);
       expect(result.current.url).toBe(undefined);
 
@@ -93,14 +94,13 @@ describe('useGetUrl', () => {
 
     it('should invoke onGetUrlError when getUrl fails', async () => {
       const customError = new Error('Something went wrong');
-      const { onError, ...getUrlParams } = PATH_INPUT;
       getUrlSpy.mockRejectedValue(customError);
 
       const { result, waitForNextUpdate } = renderHook(() =>
         useGetUrl(PATH_INPUT)
       );
 
-      expect(getUrlSpy).toHaveBeenCalledWith(getUrlParams);
+      expect(getUrlSpy).toHaveBeenCalledWith({ path: PATH_INPUT.path });
 
       // Next update will happen when getUrl resolves
       await waitForNextUpdate();
@@ -121,13 +121,11 @@ describe('useGetUrl', () => {
       .mockResolvedValueOnce({ url, expiresAt: new Date() })
       .mockResolvedValueOnce({ url: secondUrl, expiresAt: new Date() });
 
-    const { onError, ...getUrlParams } = PATH_INPUT;
-
     const { result, waitForNextUpdate, rerender } = renderHook(
       (input: UseGetUrlInput = PATH_INPUT) => useGetUrl(input)
     );
 
-    expect(getUrlSpy).toHaveBeenCalledWith(getUrlParams);
+    expect(getUrlSpy).toHaveBeenCalledWith({ path: PATH_INPUT.path });
     expect(result.current.isLoading).toBe(true);
     expect(result.current.url).toBe(undefined);
 
@@ -139,7 +137,6 @@ describe('useGetUrl', () => {
     await waitForNextUpdate();
 
     expect(getUrlSpy).toHaveBeenCalledWith({
-      ...getUrlParams,
       path: 'guest/second-file.jpg',
     });
 
@@ -152,7 +149,6 @@ describe('useGetUrl', () => {
     const firstError = new Error('Something went wrong');
     const secondError = new Error('Something went wrong again');
 
-    const { onError } = PATH_INPUT;
     getUrlSpy
       .mockRejectedValueOnce(firstError)
       .mockRejectedValueOnce(secondError);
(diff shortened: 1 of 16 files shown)


packages/react/src/hooks/useGetUrl.ts
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+
+import { getUrl, GetUrlInput } from 'aws-amplify/storage';
+
+import { isFunction } from '@aws-amplify/ui';
+import { useHasValueUpdated } from '@aws-amplify/ui-react-core';
+
+export type UseGetUrlInput = GetUrlInput & {
+  onError?: (error: Error) => void;
```
esauerbo, line 9: Initially tried interfaces over types but get this ts error `An interface can only extend an object type or intersection of object types with statically known members.`

packages/react/src/hooks/useGetUrl.ts
```diff
@@ -0,0 +1,68 @@
...
+      })
+      .catch((error: Error) => {
+        if (ignore) {
+          return;
+        }
+        if (isFunction(onError)) {
+          onError(error);
+        }
+
+        return () => {
```
esauerbo, line 59: A lot of this logic is reused from useStorageURL. Believe this line is meant to prevent unnecessary re-renders so if the component unmounts before resolving the url the state doesn't get set. https://github.com/aws-amplify/amplify-ui/blob/22f8c98607a9e77e767ffbe22fca174f54bd3381/packages/react/src/hooks/useStorageURL.ts#L53

calebpollman (commented):
Looking good! Left some minor feedback, think we should be able to merge after addressing it

packages/react/src/hooks/useGetUrl.ts
esauerbo: In this case we do want it to be returned conditionally though right? Because it should only run if the promise fails to resolve?

packages/react-storage/jest.config.ts
```diff
@@ -11,7 +11,7 @@ const config: Config = {
   ],
   coverageThreshold: {
     global: {
-      branches: 80,
+      branches: 50,
```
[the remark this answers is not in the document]
esauerbo: Yeah, I think because StorageImage uses getUrl and currently the path props aren't covered.

packages/react/src/hooks/useGetUrl.ts
calebpollman: Looked at the original hook and saw that cleanup effect is returned from the `catch`, which is incorrect as well. The `ignore` should ensure that only the results of the most recent call are returned, and should always be returned unconditionally

packages/react/src/hooks/useGetUrl.ts
```diff
@@ -0,0 +1,68 @@
...
+          ignore = true;
+        };
+      })
+      .finally(() => {
+        setResult((prevResult) => ({ ...prevResult, isLoading: false }));
```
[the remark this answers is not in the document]
esauerbo: Makes sense, updated

packages/react-core/src/hooks/useGetUrl.ts
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        if (isFunction(onError)) {
+          onError(error);
+        }
+
+        setResult({ ...INIT_STATE, isLoading: false });
+      });
+
+    return () => {
```
zchenwei, line 48: This won't work as expected I think because `let ignore = false;` defines a brand new one on each call

zchenwei (commented):
Left a comment

packages/react-core/src/hooks/useGetUrl.ts
esauerbo: `ignore` is set to `true` in the cleanup so if the component unmounts before the promise is resolved, the result won't be set. Then when the component re-renders we expect it to be set back to `false` so we can continue with setting the state and the most recent result will be used.

zchenwei: I see. That makes sense

zchenwei (approved):
LGTM

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
```diff
@@ -0,0 +1,195 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { waitFor } from '@testing-library/react';
+import { getUrl } from 'aws-amplify/storage';
+
+import { useGetUrl, UseGetUrlInput } from '../useGetUrl';
+
+jest.mock('aws-amplify/storage');
```
calebpollman, line 7: Let's remove this and add a spy with a mock implementation

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    options: { accessLevel: 'guest' },
+    onError: onStorageGetError,
+  };
+
+  const PATH_INPUT: UseGetUrlInput = {
+    path: 'guest/file.jpg',
+    onError: onGetUrlError,
+  };
+
+  const paramType = [
```
calebpollman, line 30: `paramType` as well as the above test variables should live outside the `describe` block

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    },
+  ];
+
+  it.each(paramType)(
+    `should return true for isLoading at initialization $description`,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    async ({ useGetUrlParams, description }) => {
+      (getUrl as jest.Mock).mockResolvedValue({ url });
+      const { result } = renderHook(() => useGetUrl(useGetUrlParams));
+      await waitFor(() => {
```
calebpollman, line 49: Can we replace `waitFor` with `waitForNextUpdate`? Prefer not using multiple test libraries unless actually required

packages/react-core/src/hooks/index.ts
```diff
@@ -2,6 +2,7 @@ export {
   default as useDeprecationWarning,
   UseDeprecationWarning,
 } from './useDeprecationWarning';
+export * from './useGetUrl';
```
calebpollman, line 5: Please follow existing patterns and only export what is needed, export all syntax shorthand can lead to unintentional exports

packages/react-storage/jest.config.ts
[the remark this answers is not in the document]
calebpollman: Let's remove this change, this branch does not touch the Storage package outside of the `@ts-expect-error` additions

calebpollman (changes requested):
Implementation looks good! Left a blocking comment on eslint disables and general feedback on cleaning up the test file

packages/react-storage/jest.config.ts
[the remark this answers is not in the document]
esauerbo: I think removing this will break CI because the threshold isn't met 
Screenshot 2024-04-22 at 1 51 11 PM

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
```diff
@@ -0,0 +1,218 @@
+import { renderHook } from '@testing-library/react-hooks';
+import * as Storage from 'aws-amplify/storage';
+
+import useGetUrl, { UseGetUrlInput } from '../useGetUrl';
+
+const getUrlSpy = jest.spyOn(Storage, 'getUrl');
+
+const url = new URL('https://amplify.s3.amazonaws.com/path/to/the/file.jpg');
+
+const onErrorMock = jest.fn();
```
calebpollman, line 10: Nit: This can just be `onError`

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
```diff
@@ -0,0 +1,218 @@
...
+      expect(getUrlSpy).toHaveBeenCalledTimes(1);
+      expect(result.current.isLoading).toBe(false);
+      expect(result.current.url).toBe(url);
+
+      getUrlSpy.mockClear();
+    });
+
+    it('should invoke onStorageGetError when getUrl fails', async () => {
+      const customError = new Error('Something went wrong');
+      const { onError, ...getUrlParams } = KEY_INPUT;
```
calebpollman, line 53: No need to destructure `onError` here (same for other tests). You can just reference `onErrorMock` directly in the `expect` calls

calebpollman (dismissed):
Left some small non-blocking feedback, can be addressed in this PR or as an immediate follow up

reesscot (dismissed):

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
esauerbo: The desctructuring here is actually to assert `getUrl` was called with `getUrlParams`

calebpollman: In that case the assertion should be:
```
expect(getUrlSpy).toHaveBeenCalledWith({ key: KEY_INPUT.key, options: KEY_INPUT.options });
```

Right?

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
```diff
@@ -0,0 +1,206 @@
...
+const getUrlSpy = jest.spyOn(Storage, 'getUrl');
+
+const url = new URL('https://amplify.s3.amazonaws.com/path/to/the/file.jpg');
+
+const onError = jest.fn();
+
+const KEY_INPUT: UseGetUrlInput = {
+  key: 'file.jpg',
+  options: { accessLevel: 'guest' },
+  onError: onError,
```
calebpollman, line 15: Prefer object short hand:
```suggestion
  onError,
```

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
```diff
@@ -0,0 +1,206 @@
...
+};
+
+const PATH_INPUT: UseGetUrlInput = {
+  path: 'guest/file.jpg',
+  onError: onError,
```
calebpollman, line 20: Prefer object short hand:
```suggestion
  onError,
```

packages/react-core/src/hooks/__tests__/useGetUrl.test.ts
esauerbo: Yeah isn't that kind of the same thing though?

calebpollman (approved):

reesscot (approved):

PR #5141, opened 2024-04-10
fix(react-liveness): add websocket connection timeout and related error messaging

hbuchel:
#### Description of changes

This PR 
- removes the 500 status code we were previously setting during timeout errors because it is causing the AWS SDK to retry with expired session id.
- Rejects with a `Websocket connection timeout` error when this timeout occurs.
- Adds new `CONNECTION_TIMEOUT` to `LIvenessErrorState` so the error can be identified separately from other error states and presented in the UI.
- Adds related displayText updates for showing the new error state, updates tests, types.

CONNECT_TIMEOUT error thrown in console

Connection timeout modal

Tested with next-example app by increasing latency to 20k ms. Currently our [connection times out at 10k ms](https://github.com/aws-amplify/amplify-ui/blob/505a4db44e500551aaaca21626829736f75a8d49/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/createStreamingClient/createStreamingClient.ts#L20) so this is enough to trigger the Websocket connection timeout error which should show the error in a modal, and also log to the console (since our next-example app is logging errors via onError prop)

#### Checklist

- [ ] PR description included


packages/react-liveness/src/components/FaceLivenessDetector/service/machine/__tests__/machine.test.ts
```diff
@@ -516,6 +516,26 @@ describe('Liveness Machine', () => {
...
+      await flushPromises();
+      jest.advanceTimersToNextTimer();
+      expect(service.state.value).toEqual('error');
+      expect(service.state.context.errorState).toBe(
+        LivenessErrorState.CONNECTION_TIMEOUT
+      );
+      expect(mockcomponentProps.onError).toHaveBeenCalledTimes(1);
+      const livenessError = (mockcomponentProps.onError as jest.Mock).mock
+        .calls[0][0];
+      expect(livenessError.state).toBe(LivenessErrorState.CONNECTION_TIMEOUT);
```
reesscot, line 536: Should we also check to see if the 'test' message text came through?

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -98,17 +99,27 @@ const responseStreamActor = async (callback: StreamActorCallback) => {
...
         'Invalid region in FaceLivenessDetector or credentials are scoped to the wrong region.'
       );
     }
 
     if (returnedError instanceof Error) {
-      callback({
-        type: 'SERVER_ERROR',
-        data: { error: returnedError },
-      });
+      if (isConnectionTimeout) {
```
reesscot, line 112: nit, since this is the only place we use the `isConnectionTimeout` const, could probably just call the function?
```suggestion
      if (isConnectionTimeoutError(error)) {
```

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/__tests__/machine.test.ts
hbuchel: Yes, updated with an `expect()` for `error.message`

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
hbuchel: Thanks, updated!

esauerbo (approved):
LGTM -- smoke tested locally and seeing the Connection time out with "Websocket connection timeout" error message when throttling the network to 10k ms latency.

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -98,17 +99,25 @@ const responseStreamActor = async (callback: StreamActorCallback) => {
...
+        callback({
+          type: 'CONNECTION_TIMEOUT',
+          data: { error: returnedError },
+        });
+      } else {
+        callback({
+          type: 'SERVER_ERROR',
+          data: { error: returnedError },
+        });
+      }
```
calebpollman, line 120: Think the logic in this `catch` could be simplified a bit for future readability:
```suggestion
    if (isInvalidSignatureRegionException(error)) {
      callback({ type: 'SERVER_ERROR', error: new Error(
        'Invalid region in FaceLivenessDetector or credentials are scoped to the wrong region.'
      ) })
    } else if (error instanceof Error) {
      callback({
        type: isConnectionTimeoutError(error) ? 'CONNECTION_TIMEOUT' : 'SERVER_ERROR',
        data: { error },
      });
    }
```

calebpollman (approved):
🚢

esauerbo (approved):

reesscot (approved):

PR #5173, opened 2024-04-19
chore(gen2-storage): Add gen2 examples and e2e tests

esauerbo:
#### Description of changes
Will take out of draft and change base branch to `gen2-storage/main` once https://github.com/aws-amplify/amplify-ui/pull/5171 is merged. 
- This adds `StorageImage` examples using `path` props, as well as additional e2e tests. 
- The e2e tests are exactly the same as existing StorageImage tests except they use both string and callback variations of `path` 

```
path: string | ((input: { identityId?: string }) => string);
```

#### Checklist

- [ ] PR description included


zchenwei (commented):
This is awesome. IMO, we probably want to test the new prop against a Gen2 app because that the major use case we want to cover. Plus, it should already cover the scenario of using the new prop in a Gen1 app. If it works for a Gen2 app, it should intrinsically play nice with a Gen1 app. That way, we do not need to run the same test suites twice and shorten the time to run tests. Thoughts?

PR #5194, opened 2024-04-29
chore(storage-manager): add first-class support for path

calebpollman:
#### Description of changes
Add first class support for `path` as callback
- remove `prefix` prop
- allow `accessLevel` to be `undefined` to support `path` only use cases
- remove default `accessLevel` value

#### Issue #, if available
NA

Manual tests, unit tests
#### Checklist

- [x] PR description included

### DIFF
--- a/.changeset/fair-tips-deliver.md
+++ b/.changeset/fair-tips-deliver.md
@@ -3,4 +3,6 @@
 "@aws-amplify/ui-react": patch
 ---
 
-feat(storage-manager): add the `prefix` prop to support gen2 use cases
+feat(storage-manager): update `StorageManager` to support differing usages of `path` prop:
+- existing: `accessLevel` prop provided with or without optional `path` prop
+- additive: no `accessLevel` prop provided with required `path` as either a `string` or a callback provided the current `identityId`
--- a/packages/react-storage/src/components/StorageManager/StorageManager.tsx
+++ b/packages/react-storage/src/components/StorageManager/StorageManager.tsx
@@ -63,7 +63,7 @@ const StorageManagerBase = React.forwardRef(function StorageManager(
 ): JSX.Element {
   if (!maxFileCount) {
     // eslint-disable-next-line no-console
-    console.warn(MISSING_REQUIRED_PROPS_MESSAGE);
+    logger.warn(MISSING_REQUIRED_PROPS_MESSAGE);
   }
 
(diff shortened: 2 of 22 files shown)


packages/react-storage/src/components/StorageManager/types.ts
```diff
@@ -58,10 +56,10 @@ export interface StorageManagerProps {
    */
   acceptedFileTypes?: string[];
   /**
-   * @deprecated
```
esauerbo, line 61: Why are we removing the deprecation warning?

packages/react-storage/src/components/StorageManager/StorageManager.tsx
```diff
@@ -26,63 +27,53 @@ import {
...
-    );
-  }
-
-  if (!prefix && !accessLevel) {
-    // set accessLevel to private to respect the default before adding the `prefix` prop
-    accessLevel = 'private';
-    // accessLevel has been depreacted so only guide the user to pass the `prefix` prop
-    logger.warn('`StorageManager` requires the `prefix` prop to be specified');
+    // eslint-disable-next-line no-console
+    console.warn(MISSING_REQUIRED_PROPS_MESSAGE);
```
calebpollman, line 66:
```suggestion
    logger.warn(MISSING_REQUIRED_PROPS_MESSAGE);
```

reesscot (approved):

reesscot (approved):

reesscot (approved):

packages/react-storage/src/components/StorageManager/types.ts
reesscot: Let's chat about this offline, there are pros and cons both ways.

.changeset/fair-tips-deliver.md
```diff
@@ -3,4 +3,6 @@
 "@aws-amplify/ui-react": patch
```
esauerbo, line 3: would this be considered an additive change? 
```suggestion
"@aws-amplify/ui-react": minor
```

esauerbo (approved):

.changeset/fair-tips-deliver.md
calebpollman: Good catch! Think `patch` is okay cause it's the internal exports but if public `minor` for sure

PR #5234, opened 2024-05-09
chore(liveness): move starting of ws connection to beginning of state…

thaddmt:
… machine

#### Description of changes
- Moved ws connection opening to right after camera initialization at the start of the machine
- Adding a loading screen per the figma designs

Screenshot 2024-05-09 at 12 53 58 PM

Screenshot 2024-05-09 at 12 54 06 PM

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -71,14 +71,6 @@ export interface LivenessCameraModuleProps {
   testId?: string;
 }
 
-const centeredLoader = (
-  <Loader
-    size="large"
-    className={LivenessClassNames.Loader}
-    data-testid="centered-loader"
-  />
-);
-
 const showMatchIndicatorStates = [
   FaceMatchState.TOO_FAR,
   FaceMatchState.CANT_IDENTIFY,
@@ -136,8 +128,12 @@ export const LivenessCameraModule = (
   const freshnessColorRef = useRef<HTMLCanvasElement | null>(null);
 
   const [isCameraReady, setIsCameraReady] = useState<boolean>(false);
-  const isCheckingCamera = state.matches('cameraCheck');
-  const isWaitingForCamera = state.matches('waitForDOMAndCameraDetails');
+  const isInitCamera = state.matches('initCamera');
+  const isInitWebsocket = state.matches('initWebsocket');
+  const isCheckingCamera = state.matches({ initCamera: 'cameraCheck' });
+  const isWaitingForCamera = state.matches({
+    initCamera: 'waitForDOMAndCameraDetails',
+  });
   const isStartView = state.matches('start') || state.matches('userCancel');
   const isDetectFaceBeforeStart = state.matches('detectFaceBeforeStart');
   const isRecording = state.matches('recording');
@@ -274,7 +270,7 @@ export const LivenessCameraModule = (
       >
         <Loader
           size="large"
-          className={LivenessClassNames.Loader}
+          className={LivenessClassNames.CenteredLoader}
           data-testid="centered-loader"
           position="unset"
         />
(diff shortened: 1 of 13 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -290,14 +286,32 @@ export const LivenessCameraModule = (
     );
   }
 
+  const shouldShowCenteredLoader = isInitCamera || isInitWebsocket;
```
calebpollman, line 289: is there a "non-centered" loader?

thaddmt: Yep, there is an inline loader that runs when the camera is loading only

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -200,24 +200,57 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
...
+            },
+          },
+          waitForSessionInfo: {
+            after: {
+              0: {
+                target: '#livenessMachine.start',
+                cond: 'hasServerSessionInfo',
+              },
+              100: { target: 'waitForSessionInfo' },
+            },
```
calebpollman, line 252: Does this need to poll? Wondering if we can just treat it as a service call

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -200,24 +200,57 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
...
+          },
+          waitForDOMAndCameraDetails: {
+            after: {
+              0: {
+                target: '#livenessMachine.initWebsocket',
+                cond: 'hasDOMAndCameraDetails',
+              },
+              10: { target: 'waitForDOMAndCameraDetails' },
+            },
```
calebpollman, line 226: Any idea how complicated it would be to remove the polling here and have the UI send an update to the machine once the end user has added camera permissions?

calebpollman (commented):
Think this makes sense, just left some questions

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
thaddmt: Let me look into both of these cases to see if we can simplify here :)

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
thaddmt: If we were to change this to a service call, wouldnt somewhere down the line we need to poll for the event to come in? I can't currently think of a way that we can make an function that will await until we receive an event on the websocket connection, unless that function involved a poll

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
thaddmt: Was able to remove this polling, good catch!

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
calebpollman: Think it can be done by refactoring the chain of state updates related to the `SET_SESSION_INFO` event, but that much of a refactor but might equate to scope creep for this PR

calebpollman (approved):
🚢

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -561,11 +566,6 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
       }),
       startRecording: assign({
         videoAssociatedParams: (context) => {
-          if (!context.serverSessionInformation) {
```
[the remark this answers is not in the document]
thaddmt: Yeah I think that was not a necessary check, there are other checks / guard rails to verify that that information exists already

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -290,14 +286,32 @@ export const LivenessCameraModule = (
...
+
   // We don't show full screen camera on the pre check screen (isStartView/isWaitingForCamera)
   const shouldShowFullScreenCamera =
-    isMobileScreen && !isStartView && !isWaitingForCamera;
+    isMobileScreen &&
+    !isStartView &&
+    !isWaitingForCamera &&
```
esauerbo, line 295: nit but could we remove `isWaitingForCamera` here since it's a subset of `isInitCamera`?
```suggestion
```

thaddmt: good catch, tested locally and this works!

esauerbo (approved):
LGTM

calebpollman (approved):

PR #5251, opened 2024-05-13
test(auth): add gen2 auth environments

zchenwei:
#### Checklist

- [ ] PR description included


packages/e2e/features/ui/components/authenticator/gen2/sign-up-with-phone.feature
```diff
@@ -0,0 +1,36 @@
+Feature: Sign Up with Phone 
+
+  Create a new user in the Amazon Cognito UserPool by passing the new user’s phone number and password.
+
+  Background:
+    Given I'm running the example "ui/components/authenticator/gen2/sign-up-with-phone/"
+    Then I intercept '{ "headers": { "X-Amz-Target": "AWSCognitoIdentityProviderService.SignUp" } }' with fixture "sign-up-with-phone"
+
+  @angular @react @vue @react-native
```
reesscot, line 9: Do you know if cucumber has a way to share fragments of feature files? Might be something to look into so we don't have to copy the entire files.

environments/auth/auth-with-email-lambda-signup-trigger/package.json
```diff
@@ -3,6 +3,6 @@
   "name": "@aws-amplify/ui-auth-with-email-lambda-signup-trigger-environment",
   "version": "0.0.1",
   "scripts": {
-    "pull": "amplify pull --appId d1ytpi9jtd8llt --envName staging"
```
reesscot, line 6: I think we'll want to continue testing with Gen1 as well, right? I may be missing something, but it looks like this is removing that capability

zchenwei: I combine both to reuse the shell script. It was moved to the end

zchenwei: `xxx && echo n | amplify pull --appId d1ytpi9jtd8llt --envName staging"`

packages/e2e/features/ui/components/authenticator/gen2/sign-up-with-phone.feature
zchenwei: I can check but likely no way. Even though we can change the route, I am afraid we would have to copy the same scenarios for gen2 in the file

PR #5254, opened 2024-05-14
chore(liveness): add more dynamic configs

thaddmt:
#### Description of changes
- Updated Liveness flow to use more config values from the server session Info
  - Added values for FaceDistanceThresholdMin, FaceDistanceThreshold, OvalHeightWidthRatio

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
@@ -657,6 +657,20 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
       clearErrorState: assign({ errorState: (_) => undefined }),
       updateSessionInfo: assign({
         serverSessionInformation: (_, event) => {
+          const challengeConfig = (
+            event.data!.sessionInfo as SessionInformation
+          )?.Challenge?.FaceMovementAndLightChallenge?.ChallengeConfig;
+
+          if (
+            !challengeConfig ||
+            !challengeConfig.FaceDistanceThreshold ||
+            !challengeConfig.FaceDistanceThresholdMin ||
+            !challengeConfig.OvalHeightWidthRatio
+          ) {
+            throw new Error(
+              'Challenge config not returned from session information.'
+            );
+          }
           return event.data!.sessionInfo as SessionInformation;
         },
       }),
@@ -1009,22 +1023,15 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
           serverSessionInformation?.Challenge?.FaceMovementAndLightChallenge
             ?.ChallengeConfig;
 
-        if (!challengeConfig || !challengeConfig.OvalHeightWidthRatio) {
-          throw new Error(
-            'Challenge config not returned from session information.'
-          );
-        }
-
-        const { OvalHeightWidthRatio } = challengeConfig;
         const ovalDetails = getStaticLivenessOvalDetails({
           width: width!,
           height: height!,
-          ovalHeightWidthRatio: OvalHeightWidthRatio,
+          ovalHeightWidthRatio: challengeConfig!.OvalHeightWidthRatio!,
         });
 
         const { isDistanceBelowThreshold: isFaceFarEnoughBeforeRecording } =
           await isFaceDistanceBelowThreshold({
-            sessionInformation: serverSessionInformation,
+            sessionInformation: serverSessionInformation!,
             faceDetector: faceDetector!,
             videoEl: videoEl!,
             ovalDetails,
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
@@ -1100,9 +1100,13 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
           videoWidth: videoEl!.width,
         });
 
+        const challengeConfig =
+          serverSessionInformation?.Challenge?.FaceMovementAndLightChallenge
+            ?.ChallengeConfig;
+
         // renormalize initial face
         const renormalizedFace = generateBboxFromLandmarks({
-          sessionInformation: serverSessionInformation!,
+          ovalHeightWidthRatio: challengeConfig!.OvalHeightWidthRatio,
           face: initialFace,
           oval: ovalDetails,
           frameHeight: videoEl!.videoHeight,
@@ -1140,8 +1144,12 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
         let detectedFace: Face | undefined;
         let illuminationState: IlluminationState | undefined;
 
+        const challengeConfig =
+          serverSessionInformation?.Challenge?.FaceMovementAndLightChallenge
+            ?.ChallengeConfig;
+
         const initialFaceBoundingBox = generateBboxFromLandmarks({
-          sessionInformation: serverSessionInformation!,
+          ovalHeightWidthRatio: challengeConfig!.OvalHeightWidthRatio,
           face: initialFace!,
           oval: ovalDetails!,
           frameHeight: videoEl!.videoHeight,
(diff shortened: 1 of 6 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
```diff
@@ -12,10 +12,8 @@ import {
 import { ColorSequence, SequenceColorValue } from './ColorSequenceDisplay';
 import {
   FACE_HEIGHT_WEIGHT,
+  OVAL_HEIGHT_WIDTH_RATIO,
   PUPIL_DISTANCE_WEIGHT,
-  FACE_DISTANCE_THRESHOLD,
```
esauerbo, line 16: So are these values now coming from the config instead?

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -995,63 +995,44 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
       },
       async detectFaceDistance(context) {
         const {
+          serverSessionInformation,
           isFaceFarEnoughBeforeRecording: faceDistanceCheckBeforeRecording,
         } = context;
-        const { videoEl, videoMediaStream, isMobile } =
```
esauerbo, line 1000: Why is `isMobile` no longer used?

esauerbo (commented):
LGTM just had a few questions. Could you also add a description? I think I'm missing the context behind this change

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
thaddmt: yep!

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
thaddmt: Because the value is coming from the server now, previously we would change the values based on mobile or not but now the server provides a distance threshold during before face is detected and a distance threshold during afterwards.

esauerbo (dismissed):

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -995,63 +995,44 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
...
 
+        const challengeConfig =
+          serverSessionInformation?.Challenge?.FaceMovementAndLightChallenge
+            ?.ChallengeConfig;
+
+        if (!challengeConfig || !challengeConfig.OvalHeightWidthRatio) {
+          throw new Error(
+            'Challenge config not returned from session information.'
+          );
+        }
```
calebpollman, line 1016: How does the machine reach this state without the `challengeConfig` being defined and what is the impact on the end user UX?

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
```diff
@@ -326,11 +326,28 @@ function getPupilDistanceAndFaceHeight(face: Face) {
...
+  sessionInformation: SessionInformation;
+  face: Face;
+  oval: LivenessOvalDetails;
+  frameHeight: number;
+}): BoundingBox {
+  const challengeConfig =
+    sessionInformation?.Challenge?.FaceMovementAndLightChallenge
+      ?.ChallengeConfig;
+
+  if (!challengeConfig || !challengeConfig.OvalHeightWidthRatio) {
```
calebpollman, line 344: Do we need to perform this check here if we already have done the same check [here](https://github.com/aws-amplify/amplify-ui/pull/5254/files#diff-cca8afb5864b273535f7e509efb269bdfc349fa9f544501a16443ce86a7e9680R1012)?

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
```diff
@@ -326,11 +326,28 @@ function getPupilDistanceAndFaceHeight(face: Face) {
...
+    throw new Error('Challenge config not returned from session information.');
+  }
+
+  const { OvalHeightWidthRatio } = challengeConfig;
+  const ovalHWRatio = OvalHeightWidthRatio ?? OVAL_HEIGHT_WIDTH_RATIO;
```
calebpollman, line 349: Since `OvalHeightWidthRatio` is the only value eventually read from `sessionInformation`, we can add `ovalHWRatio`  as a parameter with a default value of `OVAL_HEIGHT_WIDTH_RATIO`?

zchenwei (dismissed):

calebpollman (commented):
No major concerns but we can simplify here and looking to understand the new error case in the machine

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
thaddmt: I believe for the most part this case should never be reached. We have other checks before this to ensure that the session information exists. This is here in the case the server decides to return a session information without challenge config, which would be an error on the rekognition service side. In the case that this is triggered they end user would see the usual error experience and the developer could print this error from the `onError` cb

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
thaddmt: Yeah I think i did this at first but forgot why I changed it back, I think i was thinking in case we needed other values from session information in the future. Will take your suggestion and just update it to session Info if we ever need it later :)

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
thaddmt: removed!

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
```diff
@@ -624,21 +630,34 @@ export async function getFaceMatchState(
...
+  const challengeConfig =
+    sessionInformation?.Challenge?.FaceMovementAndLightChallenge
+      ?.ChallengeConfig;
+  if (
+    !challengeConfig ||
+    !challengeConfig.FaceDistanceThreshold ||
+    !challengeConfig.FaceDistanceThresholdMin
+  ) {
+    throw new Error('Challenge config not returned from session information.');
+  }
```
calebpollman, line 657: Can we handle this validation before calling `isFaceDistanceBelowThreshold`? Ideally we would handle this validation where the challenge is returned before updating the state machine with the `challengeConfg`

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/liveness.ts
```diff
@@ -326,11 +326,17 @@ function getPupilDistanceAndFaceHeight(face: Face) {
   return { pupilDistance, faceHeight };
 }
 
-export function generateBboxFromLandmarks(
-  face: Face,
-  oval: LivenessOvalDetails,
-  frameHeight: number
-): BoundingBox {
+export function generateBboxFromLandmarks({
+  ovalHeightWidthRatio = OVAL_HEIGHT_WIDTH_RATIO,
```
calebpollman, line 330: 🙌

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
calebpollman: Thanks for the explanation! Noted [here](https://github.com/aws-amplify/amplify-ui/pull/5254/files#diff-4dc9d74828ef070874870649a51d5a4dd622a7ab2eca61a4d761f36075cffd9dR657) as well, but these validations could be run against `challengeConfig` before setting in the machine?

calebpollman (dismissed):

esauerbo (approved):

calebpollman (approved):
🚢

Issue #5253, opened 2024-05-14
SelectField SVG accessibility


aleshh:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

vite

### What browsers are you seeing the problem on?

Firefox

### Please describe your bug.

SelectField has a nifty drop-down icon SVG. This gets flagged in accessibility scan because there is no accessible name. I think the SVG should have `aria-hidden="true"`, perhaps as an option. 

Work-around is to pass in an custom icon with the appropriate props.

### What's the expected behaviour?

SelectField should have `aria-hidden="true"` as an option or maybe even default.

### Help us reproduce the bug!

This is visible on the Amplify docs demo.

### Code Snippet

```typescript
// Put your code below this line.

```

PR #5256, opened 2024-05-15
fix(ui-react): add aria-hidden to Select expand icon

zchenwei:
#### Description of changes

Add `aria-hidden="true"` to the Select drop-down icon to remove it from the accessibility tree.

#### Issue #, if available

fix https://github.com/aws-amplify/amplify-ui/issues/5253

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react/src/primitives/Select/Select.tsx
+++ b/packages/react/src/primitives/Select/Select.tsx
@@ -81,6 +81,7 @@ const SelectPrimitive: Primitive<SelectProps, 'select'> = (
             classNameModifier(ComponentClassName.SelectIcon, size)
           )}
           color={iconColor}
+          aria-hidden="true"
         >
           {icon ?? icons?.expand ?? <IconExpandMore />}
         </Flex>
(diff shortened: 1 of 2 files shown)


packages/react/src/primitives/Select/Select.tsx
```diff
@@ -81,6 +81,7 @@ const SelectPrimitive: Primitive<SelectProps, 'select'> = (
             classNameModifier(ComponentClassName.SelectIcon, size)
           )}
           color={iconColor}
+          aria-hidden="true"
```
reesscot, line 84: Can we do a quick test with a screen reader to make sure this prevents this from being read out as "image"?

hbuchel (approved):
Verified in screen reader that it accessibly hides the SVG and no longer reads out "image", LGTM

reesscot (approved):

packages/react/src/primitives/Select/Select.tsx
zchenwei: Verified

PR #5261, opened 2024-05-17
chore(liveness): Added Liveness telemetry

thaddmt:
#### Description of changes
- Added a telemetry middleware to the streaming client which appends values to the query
- Appends `attempt-count` and `precheck-view-enabled`
- Attempt count is counted as the number of attempts that are less than 5 minutes apart, once 5 minutes has passed without an attempt we will reset the count

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
@@ -2,6 +2,7 @@ import {
   StartFaceLivenessSessionCommandInput,
   StartFaceLivenessSessionCommandOutput,
 } from '@aws-sdk/client-rekognitionstreaming';
+import { HttpRequest } from '@smithy/protocol-http';
 import {
   BuildHandler,
   BuildHandlerArguments,
@@ -53,14 +54,10 @@ export const createTelemetryReporterMiddleware =
   async (
     args: BuildHandlerArguments<StartFaceLivenessSessionCommandInput>
   ): Promise<BuildHandlerOutput<StartFaceLivenessSessionCommandOutput>> => {
-    // @ts-ignore
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-    args.request.query['attempt-count'] = attemptCount.toString();
-    // @ts-ignore
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-    args.request.query['precheck-view-enabled'] = preCheckViewEnabled
-      ? '1'
-      : '0';
+    (args.request as HttpRequest).query['attempt-count'] =
+      attemptCount.toString();
+    (args.request as HttpRequest).query['precheck-view-enabled'] =
+      preCheckViewEnabled ? '1' : '0';
 
     const result = await next(args);
     return result;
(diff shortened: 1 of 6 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
```diff
@@ -0,0 +1,69 @@
...
+type AttemptCountData = { timestamp: number; count: number };
+
+function getAttemptCountData(): AttemptCountData {
+  return JSON.parse(localStorage.getItem(ATTEMPT_KEY) ?? '{}') as {
+    timestamp: number;
+    count: number;
+  };
+}
+
+function setAttemptCountData(data: AttemptCountData) {
```
thaddmt, line 25: Opted to use local storage so that the attempt count could be tracked between attempts. In cases where the component gets unmounted and then re mounted the liveness state machine would be reset and we have no way of tracking attempts. This is a common use case in even in our default example app as we show a loading spinner in place of the component while we grab the session id

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
```diff
@@ -0,0 +1,69 @@
...
+    next: BuildHandler<
+      StartFaceLivenessSessionCommandInput,
+      StartFaceLivenessSessionCommandOutput
+    >
+  ) =>
+  async (
+    args: BuildHandlerArguments<StartFaceLivenessSessionCommandInput>
+  ): Promise<BuildHandlerOutput<StartFaceLivenessSessionCommandOutput>> => {
+    // @ts-ignore
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
calebpollman, line 59: What's the issue we are covering up here?

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
calebpollman: Let's discuss oflline

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
```diff
@@ -0,0 +1,69 @@
+// type TelemetryDataKey = `${Lowercase<string>}-${Lowercase<string>}`;
+// type TelemetryData = Record<TelemetryDataKey, string>;
```
calebpollman, line 2:
```suggestion
```

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
thaddmt: whoops forgot i added this while testing, can remove :)

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/TelemetryReporter/TelemetryReporter.ts
```diff
@@ -0,0 +1,50 @@
...
+  BuildHandler,
+  BuildHandlerArguments,
+  BuildHandlerOutput,
+} from '@smithy/types';
+
+const DEFAULT_ATTEMPT_COUNT_TIMEOUT = 300000; // 5 minutes / 300000 ms
+
+export class TelemetryReporter {
+  static attemptCount = 0;
+  static timestamp = Date.now();
```
thaddmt, line 16: Switched to static class vars here, this works for unmount and mount but does not work for refresh. What do you think?

esauerbo (approved):
LGTM

calebpollman (approved):
🚢

PR #5274, opened 2024-05-25
Gen2 auth environment v3

zchenwei:
#### Checklist

- [ ] PR description included


thaddmt (dismissed):

esauerbo (dismissed):
LGTM!

environments/auth/auth-with-federated/amplify/backend/auth/authwithfederatedd9f06071/cli-inputs.json
```diff
@@ -1,72 +0,0 @@
-{
-  "version": "1",
```
reesscot, line 2: should this be getting deleted?

examples/angular/src/app/app-routing.module.ts
```diff
@@ -105,10 +105,6 @@ const routes: Routes = [
     path: 'ui/components/authenticator/sign-up-with-phone',
     component: SignUpWithPhoneComponent,
   },
-  {
-    path: 'ui/components/authenticator/sign-up-with-email',
```
reesscot, line 109: Is this not in use?

environments/datastore/action-hooks/package.json
```diff
@@ -3,6 +3,7 @@
   "name": "@aws-amplify/ui-action-hooks-environment",
   "version": "0.0.1",
   "scripts": {
-    "pull": "amplify pull --appId d1c6vwjkmuz6ml --envName dev"
+    "pull": "amplify pull --appId d1c6vwjkmuz6ml --envName dev",
+    "generate": "echo 'Not available yet. Generating a placeholder file for now...' && touch amplify_outputs.json && echo '{}' > amplify_outputs.json"
```
calebpollman, line 7: Let's rethink these "Not available yet" messages, "yet" creates an assumption that support is on the way which may not be true, but even if there is support coming that allows this message to be removed in the future, how do readers correlate that support to this script?

environments/pull-environments.sh
```diff
@@ -27,9 +27,12 @@ while getopts ':r:i:e:' OPTION; do
   e)
     exclude="$OPTARG"
     ;;
+  v)
+    version="$OPTARG"
+    ;;
 
   ?)
-    echo "Usage: $(basename $0) [-r region] [-i include] [-e exclude]"
+    echo "Usage: $(basename $0) [-r region] [-i include] [-e exclude] [-v version]"
```
calebpollman, line 35: Why do we need to add a version flag to the legacy pull command?

examples/angular/angular.json
```diff
@@ -11,8 +11,11 @@
       "prefix": "app",
       "architect": {
         "build": {
-          "builder": "@angular-devkit/build-angular:browser",
+          "builder": "@angular-builders/custom-webpack:browser",
```
calebpollman, line 14: Can you provide context on this change?

examples/angular/src/pages/ui/components/authenticator/auth-status/auth-status.component.ts
```diff
@@ -4,15 +4,17 @@ import { Amplify } from 'aws-amplify';
 
 import { signIn, signOut } from 'aws-amplify/auth';
 
-import awsExports from './aws-exports';
+const amplifyOutputs = (
+  await import(`@environments/auth/auth-with-email/${process.env.PATH}`)
+).default;
```
calebpollman, line 9: Are any of these rendered in the docs?

examples/angular/tsconfig.json
```diff
@@ -10,8 +10,9 @@
     "experimentalDecorators": true,
     "moduleResolution": "node",
     "importHelpers": true,
+    "resolveJsonModule": true,
     "target": "es2020",
-    "module": "es2020",
+    "module": "esnext",
```
calebpollman, line 15: Why are we updating this?

examples/vue/src/pages/ui/components/authenticator/sign-in-with-phone/index.vue
```diff
@@ -3,9 +3,23 @@ import { Amplify } from 'aws-amplify';
...
 import { Authenticator, translations } from '@aws-amplify/ui-vue';
 import '@aws-amplify/ui-vue/styles.css';
-import aws_exports from './aws-exports';
 
-Amplify.configure(aws_exports);
+const amplifyOutputs =
+  import.meta.env.VITE_VERSION === 'gen1'
+    ? (
+        await import(
+          // @ts-ignore
```
calebpollman, line 11: We should just have a util that resolves the import prefix (e.g. `@environments/auth/auth-with-phone-number/` and appends the output file type. This pattern of repetition is fragile

.github/workflows/reusable-e2e.yml
```diff
@@ -229,9 +262,14 @@ jobs:
...
           aws configure set default.region $AWS_REGION
 
-      - name: Pull down Storage AWS environments
+      - name: Pull down Gen1 Storage AWS environments
+        if: ${{ matrix.version == 'gen1' }}
         run: yarn environments storage pull
 
+      - name: Pull down Gen2 Storage AWS environments
+        if: ${{ matrix.version == 'gen2' }}
+        run: yarn environments storage pull -v gen2
```
calebpollman, line 271: Can we just iterate over the values of `matrix.version` and run these by passing the version to the pull script? Something like:

```
      - name: Pull down ${{ matrix.version }} Storage AWS environments
        if: ${{ matrix.version == 'gen2' }}
        run: yarn environments storage pull -v ${{ matrix.version }}
```

.github/workflows/reusable-e2e.yml
```diff
@@ -280,17 +324,25 @@ jobs:
...
 
       - name: Build ${{ matrix.example }} example
         run: yarn ${{ matrix.example }}-example build
+        env:
+          VERSION: ${{ matrix.version }}
+          VITE_VERSION: ${{ matrix.version }}
 
       - name: Start ${{ matrix.example }} example
-        run: yarn ${{ matrix.example }}-example start & npx wait-on -c waitOnConfig.json -t 20000 http-get://localhost:3000/ui/components/authenticator/sign-in-with-username
+        run: yarn ${{ matrix.example }}-example start & npx wait-on -c waitOnConfig.json -t 20000 http-get://localhost:3000/ui/components/authenticator/sign-in-with-email
```
calebpollman, line 345: Can you provide a screenshot in the PR description showing passing Android and iOS e2e test runs?

calebpollman (commented):
Great direction here! Left some questions and a request for an RN e2e passing test artifact

environments/datastore/action-hooks/package.json
zchenwei: I will update those to the real commands. When I opened this PR, I did not finish the creation for some of the components 😅

.github/workflows/reusable-e2e.yml
zchenwei: Good idea. That way we do not need the if conditional as well. I will update and give it a try 👍

examples/angular/src/pages/ui/components/authenticator/auth-status/auth-status.component.ts
zchenwei: Double checked. They are not

examples/angular/tsconfig.json
zchenwei: Dynamic import requires top level await, which is supported after `module: "es2022"`

.github/workflows/reusable-e2e.yml
zchenwei: Sure, I will run against Android and iOS e2e tests

environments/pull-environments.sh
zchenwei: This the script to pull all the environments.
```sh
if [ "$version" == "gen2" ]; then
  printf $dirs | xargs -P $numParallelTasks -I {} sh -c ""$shell_path"/pull-environment-gen2.sh {}"
else
  printf $dirs | xargs -P $numParallelTasks -I {} sh -c ""$shell_path"/pull-environment.sh {} "$region""
fi
```
`pull-enviroment.sh` and `pull-enviroment-gen2.sh` are used to pull a single environment for gen1 and gen2 respectively.

examples/angular/angular.json
zchenwei: yeah, so this allows us to add extra webpack config to the default, which resolves the json config(`amplify_outputs.json`) and the value of `process.env.PATH` at build time.

```js
// examples/angular/extra-webpack.config.js
const webpack = require('webpack');

module.exports = {
  resolve: {
    extensions: ['.json'],
  },
  plugins: [
    new webpack.DefinePlugin({
      'process.env.PATH': JSON.stringify(
        process.env.VERSION === 'gen1' ? 'src/aws-exports' : 'amplify_outputs'
      ),
    }),
  ],
};
```

More info: https://www.npmjs.com/package/@angular-builders/custom-webpack

examples/vue/src/pages/ui/components/authenticator/sign-in-with-phone/index.vue
zchenwei: Yeah, originally I was doing something to simplify it but Vite has some trouble doing import analysis when it comes to variables in dynamic imports due to the limitations of the [plugin](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations) it is using. e.g.:
- Variables only represent file names one level deep (if file is `src/aws-exports.js`, the import would fail)
- Imports must end with a file extension
- Imports must start with `./` or `../`
etc...

But I like the idea to create a util. Let me try to see if I can create a util and simplify this as much as possible without annoying Vite.

calebpollman: Got it, thanks for confriming

examples/react-native/src/features/InAppMessaging/Demo/Example.tsx
```diff
@@ -6,9 +6,16 @@ import { SafeAreaView, StyleSheet, Text, View } from 'react-native';
...
 import { Button, Checkbox, Radio, RadioGroup } from '../../../ui';
 import { useInAppDemo, ACTIONS, LAYOUTS, ORIENTATIONS } from './utils';
-import config from './aws-exports';
+import { VERSION } from '@env';
 
-Amplify.configure(config);
+const AMPLIFY_CONFIG_PATH =
+  VERSION === 'gen1' ? 'src/amplifyconfiguration' : 'amplify_outputs';
+
+const amplifyOutputs = require(
```
zchenwei, line 14: Have to use `require()` for dynamic import because `import()` is only supported by react native bundler when used  in conjunction with `React.lazy()` to load a lazy component.

examples/react-native/src/features/InAppMessaging/Demo/Example.tsx
```diff
@@ -6,9 +6,16 @@ import { SafeAreaView, StyleSheet, Text, View } from 'react-native';
 import { withInAppMessaging } from '@aws-amplify/ui-react-native';
 import { Button, Checkbox, Radio, RadioGroup } from '../../../ui';
 import { useInAppDemo, ACTIONS, LAYOUTS, ORIENTATIONS } from './utils';
-import config from './aws-exports';
+import { VERSION } from '@env';
 
-Amplify.configure(config);
+const AMPLIFY_CONFIG_PATH =
```
zchenwei, line 11: Using `src/amplifyconfiguration` - the json equivalent  of `src/aws-exports` because `require` will return the exact config object when loading a json file versus putting it behind a `default` key when loading a js file

examples/react-native/src/features/InAppMessaging/Demo/Example.tsx
zchenwei: Note: we cannot extract `const AMPLIFY_CONFIG_PATH` and reuse it elsewhere because that will cause trouble for the bundler to do import analysis. It has to stay along with the `require` statement below.

PR #5284, opened 2024-05-31
(Fix): Whitespace in Tab Labels

### LINKED ISSUE #5220: bug(react/a11y): Tabs primitive uses malformed id values with spaces in them

- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies

React

### Which UI component?

Primitive components

### How is your app built?

Next.js

_No response_

_No response_

### Please describe your bug.

The Tabs primitive generates `id`s with spaces in them.

The tabs primitive generates `id` attributes for the TabsPanel and TabsItems elements using the `value` prop which is the name of the Tab. This leads to the following markup (using the [docs demo](https://ui.docs.amplify.aws/react/components/tabs#composable) example): Note that the `id` for the button has a space in it as does the `id` it references in the `aria-controls`. Similarly, the `id` and `aria-labelledby` attributes have spaces in their values as well.

```
<div class="amplify-tabs">
  <div
    role="tablist"
    class="amplify-tabs__list"
    style="justify-content: flex-start"
  >
    <button
      role="tab"
      id="Tab 1-tab"
      aria-selected="true"
      aria-controls="Tab 1-panel"
      class="amplify-tabs__item amplify-tabs__item--active"
    >
      Tab 1</button
    >
    // ... other tab buttons
  </div>
  <div
    role="tabpanel"
    id="Tab 1-panel"
    aria-labelledby="Tab 1-tab"
    class="amplify-tabs__panel amplify-tabs__panel--active"
  >
    Tab content #1
  </div>
  // ... other tab panels
</div>
```

This can cause issues for assistive tech because then the tabs aren't properly associated to each other, and just generally other issues because they are not valid IDs that can be used with other javascript or CSS correctly. For example you can see in the accessibility tab for this tab panel that it can't find the correct element that aria-labelledby references so instead uses "First":

Screenshot 2024-05-06 at 6 40 46 PM

 

You could probably:
1. Trim the white space for each instance where the value is used for an id `id={`${value.trim()}-tab`}` (just ok)
1. Generate an `id` for each `item` using `React.useId()` to ensure each Tabs usage has unique ids for it's tab elements (probably better than using just the name of the tab)

Tabs `id`s don't have spaces in them and tabs are programmatically linked to each other correctly 

Inspect the markup for Tabs on https://ui.docs.amplify.aws/react/components/tabs

### Code Snippet

```typescript
// Put your code below this line.

```

### Console log output

_No response_

_No response_

axelEandrews:
#### Description of changes
Generates ID without whitespace for each tab component. 

Example: A tab declared with `value="Tab 1"` produces the following HTML output `<button role="tab" id="Tab-1-tab" aria-selected="true" aria-controls="Tab-1-panel" class="amplify-tabs__item amplify-tabs__item--active">Tab 1</button>`

#### Issue #, if available
Fixes #5220

Inspected elements in Tabs Docs, IDs now contain no whitespace and are unique.

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/src/primitives/Tabs/TabsPanel.tsx
+++ b/packages/react/src/primitives/Tabs/TabsPanel.tsx
@@ -8,14 +8,13 @@ import { View } from '../View';
 import { BaseTabsPanelProps, TabsPanelProps } from './types';
 import { primitiveWithForwardRef } from '../utils/primitiveWithForwardRef';
 import { TabsContext } from './TabsContext';
-import { useStableId } from '../utils/useStableId';
 
 const TabPanelPrimitive: Primitive<TabsPanelProps, 'div'> = (
   { className, value, children, role = 'tabpanel', ...rest },
   ref
 ) => {
   const { activeTab, isLazy } = React.useContext(TabsContext);
-  const idValue = useStableId();
+  const idValue = value.replace(' ', '-');
 
   if (isLazy && activeTab !== value) return null;
 
--- a/packages/react/src/primitives/Tabs/TabsPanel.tsx
+++ b/packages/react/src/primitives/Tabs/TabsPanel.tsx
@@ -8,12 +8,14 @@ import { View } from '../View';
 import { BaseTabsPanelProps, TabsPanelProps } from './types';
 import { primitiveWithForwardRef } from '../utils/primitiveWithForwardRef';
 import { TabsContext } from './TabsContext';
+import { useStableId } from '../utils/useStableId';
 
 const TabPanelPrimitive: Primitive<TabsPanelProps, 'div'> = (
   { className, value, children, role = 'tabpanel', ...rest },
   ref
 ) => {
   const { activeTab, isLazy } = React.useContext(TabsContext);
+  const idValue = useStableId();
 
   if (isLazy && activeTab !== value) return null;
 
@@ -21,8 +23,8 @@ const TabPanelPrimitive: Primitive<TabsPanelProps, 'div'> = (
     <View
       {...rest}
       role={role}
-      id={`${value}-panel`}
-      aria-labelledby={`${value}-tab`}
+      id={`${idValue}-panel`}
+      aria-labelledby={`${idValue}-tab`}
       className={classNames(
         ComponentClassName.TabsPanel,
         classNameModifierByFlag(
--- a/packages/react/src/primitives/Tabs/TabsItem.tsx
+++ b/packages/react/src/primitives/Tabs/TabsItem.tsx
@@ -12,6 +12,7 @@ import { View } from '../View';
 import { primitiveWithForwardRef } from '../utils/primitiveWithForwardRef';
 import { BaseTabsItemProps, TabsItemProps } from './types';
 import { TabsContext } from './TabsContext';
+import { useStableId } from '../utils/useStableId';
 
 const TabsItemPrimitive: Primitive<TabsItemProps, 'button'> = (
   { className, value, children, onClick, as = 'button', role = 'tab', ...rest },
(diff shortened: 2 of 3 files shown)


zchenwei (commented):
This looks good. Could we add unit tests for the updates?

axelEandrews:
@zchenwei I added tests to confirm that the IDs are unique and that they follow the naming convention to end in '-tab'.

reesscot:
Can you please add an HTML output example in the description to show the full id that's generated?

packages/react/src/primitives/Tabs/TabsPanel.tsx
```diff
@@ -8,21 +8,23 @@ import { View } from '../View';
...
   if (isLazy && activeTab !== value) return null;
 
   return (
     <View
       {...rest}
       role={role}
-      id={`${value}-panel`}
-      aria-labelledby={`${value}-tab`}
+      id={`${idValue}-panel`}
+      aria-labelledby={`${idValue}-tab`}
```
dbanksdesign, line 27: If we are using `useStableId` in 2 different components, wouldn't that generate 2 different IDs? We need `aria-labelledby` in this component to be the `id` of the `TabItem` and vice-versa with `aria-controls` and this `id`

zchenwei: oh, thats true

axelEandrews: I switched it back to using the given value, with spaces replaced with `-` to ensure the all of the labels work cohesively. This relies on the user giving each tab a unique value.

hbuchel: Can we think of another solution to providing a unique id? Maybe passing a `React.useId` or `useStableId` down through context so it is shared? @zchenwei any thoughts on that?

I think it creates sort of a "hidden" requirement that each value has to be unique because it is also happens to be used to generate the ID. It also creates a scenario where the developer has to track down the cause of duplicate IDs or potentially inaccessible markup. This should be something we can safely back into the component. An example is if you inspect that markup and usage of the Radix Tabs component https://www.radix-ui.com/primitives/docs/components/tabs

axelEandrews: Looking at the Radix implementation, they still rely on the user-inputed `value` to create their full IDs. They feed the value into functions to create the trigger ID and the content ID, both based off of one `baseID` for the entire tab group. In their implementation, so if the user chose to input the same value for multiple tabs in the same group they'd run into the same issue. What their solution does address is having multiple tab groups on the same page, because the `baseID` will be different, and thus two tabs with value "Tab 1" will result in trigger ids: `{baseID}-trigger-Tab 1`, where the `baseID` will be different. 

Essentially, it looks they are guaranteeing that so long as you declare unique values for each tab in a tab list, you won't run into any issues, but I believe that there is still that dependance on the user entering unique values. 

Screenshot 2024-06-04 at 12 16 54 PM

hbuchel: Right, the issue is not so much with tabs within the same tab group, but for multiple Tab components in the DOM at the same time that use the same tab names. Within the same tab group, I think it's reasonable that users should specify unique tab names. But for multiple sets of the same tab names on a page, that should be something we can build into this component by generating a unique ID (like the baseId that Radix generates) that is used to build out the ID string.

An example of this is our Docs site that sometimes uses multiple tab components on a page, like [the docs for adding Facebook login, Google login, etc](https://docs.amplify.aws/gen1/react/build-a-backend/auth/add-social-provider/).

PR #5296, opened 2024-06-04
fix(ui-react): prevent invalid HTML IDs in Tabs

### LINKED ISSUE #5220: bug(react/a11y): Tabs primitive uses malformed id values with spaces in them

axelEandrews:
#### Description of changes
Adds `groupId` to `TabsContext` such that unique IDs can be created for each tab group. If whitespace is present in the given `value` for a tab, the whitespace is replaced with `-`. IDs are of the form `${groupId}-tab-${value}` for tabs, and 
`${groupId}-panel-${value}` for panels. 

Outer html output for a tablist (tabs given values: ['Tab 1', 'Tab 2', 'Tab 3']: `<div role="tablist" class="amplify-tabs__list" style="justify-content: flex-start;"><button role="tab" id="amplify-id-:rl:-tab-Tab-1" aria-selected="true" aria-controls="amplify-id-:rl:-panel-Tab-1" class="amplify-tabs__item amplify-tabs__item--active">Tab 1</button><button role="tab" id="amplify-id-:rl:-tab-Tab-2" aria-selected="false" aria-controls="amplify-id-:rl:-panel-Tab-2" tabindex="-1" class="amplify-tabs__item">Tab 2</button><button disabled="" role="tab" id="amplify-id-:rl:-tab-Tab-3" aria-selected="false" aria-controls="amplify-id-:rl:-panel-Tab-3" tabindex="-1" class="amplify-tabs__item">Disabled tab</button></div>`

#### Issue #, if available
Fixes #5220 

#### Checklist

- [x] PR description included


calebpollman:
Can you please update the PR description to follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) syntax? e.g.
```
fix(ui-react): prevent invalid HTML IDs in Tabs
```

calebpollman:
Also need to add a change-set

hbuchel (commented):
Thanks @axelEandrews! LGTM. Just leaving my input but will let UI folks handle approvals:
- Reviewed the output DOM and ran the HTML on the Tabs doc page through an HTML validator to show there are no duplicate ID issues.
- Also verified tab panels are now correctly associated with their labels (before this would show a warning for broken id). Fixed output:
Screenshot 2024-06-05 at 2 09 37 PM
- Also ran it through VoiceOver and verified it no longer reads "First, Second, etc" as the tab panel labels.

packages/react/src/primitives/Tabs/TabsItem.tsx
```diff
@@ -17,23 +17,24 @@ const TabsItemPrimitive: Primitive<TabsItemProps, 'button'> = (
...
-  const { activeTab, setActiveTab } = React.useContext(TabsContext);
+  const { activeTab, setActiveTab, groupID } = React.useContext(TabsContext);
   const isActive = activeTab === value;
   const handleOnClick = (e: React.MouseEvent<HTMLButtonElement>) => {
     if (isTypedFunction(onClick)) {
       onClick?.(e);
     }
     setActiveTab(value);
   };
+  const idValue = value.replace(' ', '-');
```
calebpollman, line 28: Couple questions:
- Can we just a stable id here as well?
- Any impact if `value` changes between renders?

packages/react/src/primitives/Tabs/TabsPanel.tsx
```diff
@@ -13,16 +13,17 @@ const TabPanelPrimitive: Primitive<TabsPanelProps, 'div'> = (
   { className, value, children, role = 'tabpanel', ...rest },
   ref
 ) => {
-  const { activeTab, isLazy } = React.useContext(TabsContext);
+  const { activeTab, isLazy, groupID } = React.useContext(TabsContext);
+  const idValue = value.replace(' ', '-');
```
calebpollman, line 17: See previous [comment](https://github.com/aws-amplify/amplify-ui/pull/5296/files#diff-3c1290d578cef1df38109c904cb19c8f50191b84d42cbdba52cf2a06158d4e76R28)

packages/react/src/primitives/Tabs/types.ts
```diff
@@ -38,6 +38,7 @@ export interface BaseTabsProps extends BaseTabsListProps {
   items?: Array<{
     value: string;
     isDisabled?: boolean;
+    groupID: string;
```
calebpollman, line 41: Think this is a breaking change?

calebpollman (commented):
Great start @axelEandrews! Left some questions

packages/react/src/primitives/Tabs/TabsItem.tsx
axelEandrews: I think we would not want to use a stable id here, because then when switching between tabs a new stable id would be created each time we do that switch. I believe that would lead to the id changing each time we switch off of a tab and back to it. 

I'm not sure about the impact of value changing between renders, but I'm a bit confused on how that could happen. The `value` should only be interacted with in the initial creation of the tab.

packages/react/src/primitives/Tabs/types.ts
axelEandrews: I could be wrong, but I don't think so. I have `groupID` initialized as an empty string in TabsContext, and there is no user interaction with `groupID` when creating tab groups.

calebpollman: Think requiring a new prop in the `items` definition will break existing usage? Example:
```
// current required properties
<Tabs items={[{ content: 'content', label: 'label', value: 'value' }]} />

// updated required properties
<Tabs items={[{ content: 'content', groupId: 'groupId', label: 'label', value: 'value' }]} />
```

Also `groupId` is an internal detail and not something that we want to expose as overridable right?

packages/react/src/primitives/Tabs/TabsItem.tsx
calebpollman: Does the value of `idValue` in `TabsItem` need to match the `idValue` in the corresponding `TabsPanel`?

packages/react/src/primitives/Tabs/TabsContext.tsx
```diff
@@ -3,10 +3,12 @@ import * as React from 'react';
 export interface TabsContextInterface {
   activeTab: string;
   isLazy?: boolean;
+  groupID?: string;
```
calebpollman, line 6: Prefer `groupId` over `groupID`

packages/react/src/primitives/Tabs/TabsItem.tsx
axelEandrews: Yes, because the aria labels are generated based on the idValue, and the tab will point at the panel and vice-versa.

calebpollman: Need to ensure that there is a single source of truth rather than evaluate the `idValues` individually, this approach is fragile since an update to `idValue` resolution in one component (but not the other) will break functionality

axelEandrews: Ok, I think this somewhat goes back to the original question of the user creating tab items and tab panels with accurately matching `value`. The only way for idValue to be different between the `TabItem` and `TabLabel` is if the user assigns mismatching `value` to them. There is also the issue that they _could_ enter the same value for every item and panel in one tab group, which would also break things. Could we just assign numeric `value` to each created `TabItem` and `TabPanel`, and assume that they are defined in the correct order? The current solution was relying on the assumptions that for a given tab group: each tab/panel pair must share the same `value`, and each pair must have a unique `value`.

calebpollman: Suggest creating unique per item ids in the Tabs context so there is a single source of truth

PR #5297, opened 2024-06-05
fix(liveness): add a check to see if recording works before choosing …

thaddmt:
…an encoding

#### Description of changes
- There is a known bug on Chrome 125 where the the H264 encoder fails silently - https://issues.chromium.org/issues/343199623
- This has caused issues on certain devices where it will attempt to record video but nothing will be recorded in memory
- Added this solution where we attempt to do a small 200ms recording before hand to see if it can successfully create chunks
- If recording does not successfully add chunks we will use an alternate encoding type from the common default which appears to be `'video/x-matroska;codecs=avc1` (h264 === avc)

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
@@ -41,6 +41,7 @@ import {
   isCameraDeviceVirtual,
   FreshnessColorDisplay,
   drawStaticOval,
+  doesDefaultMimeTypeWork,
 } from '../utils';
 
 import { getStaticLivenessOvalDetails } from '../utils/liveness';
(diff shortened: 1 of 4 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -60,6 +61,7 @@ import { WS_CLOSURE_CODE } from '../utils/constants';
 
 const CAMERA_ID_KEY = 'AmplifyLivenessCameraId';
 const DEFAULT_FACE_FIT_TIMEOUT = 7000;
+const ALTERNATE_MIME_TYPE = 'video/x-matroska;codecs=vp8';
```
calebpollman, line 64: Is this supported in safari?

thaddmt: Seems like certain versions support it https://caniuse.com/?search=vp8

but we might not be going with this solution anyways

PR #5302, opened 2024-06-07
fix(ui-react): Whitespace in Tab IDs

axelEandrews:
Adds `groupId` to TabsContext such that unique IDs can be created for each tab group. IDs are of the form `${groupId}-tab-${value}` for tabs, and`${groupId}-panel-${value}` for panels. Also adds `whitespaceValue` to TabsContext, which is assigned a value in TabsContainer, then used to remove whitespace from `value` in TabsItem and TabsPanel (ensures proper operation regardless of how Tabs are created by user).

Outer html output for a tablist with given values: ['Tab 1', 'Tab 2', 'Tab 3']: `<div role="tablist" class="amplify-tabs__list" style="justify-content: flex-start;"><button role="tab" id="amplify-id-:rl:-tab-Tab-1" aria-selected="true" aria-controls="amplify-id-:rl:-panel-Tab-1" class="amplify-tabs__item amplify-tabs__item--active">Tab 1</button><button role="tab" id="amplify-id-:rl:-tab-Tab-2" aria-selected="false" aria-controls="amplify-id-:rl:-panel-Tab-2" tabindex="-1" class="amplify-tabs__item">Tab 2</button><button disabled="" role="tab" id="amplify-id-:rl:-tab-Tab-3" aria-selected="false" aria-controls="amplify-id-:rl:-panel-Tab-3" tabindex="-1" class="amplify-tabs__item">Disabled tab</button></div>`

#### Issue #, if available
#5220 

#### Checklist

- [x] PR description included
- [] Relevant documentation is changed or added (and PR referenced)


packages/react/src/primitives/Tabs/Tabs.tsx
```diff
@@ -24,8 +24,8 @@ const TabsPrimitive: Primitive<TabsProps, 'div'> = (
           </TabsItem>
         ))}
       </TabList>
-      {items?.map(({ value, content, isDisabled }) => (
-        <TabPanel key={value} value={value} isDisabled={isDisabled}>
+      {items?.map(({ value, content, isDisabled, ...rest }) => (
+        <TabPanel {...rest} key={value} value={value} isDisabled={isDisabled}>
```
reesscot, line 28: qq: are there any items in rest that we don't want to spread into TabPanel? Geniunely asking as I don't know. I've seen cases in the past where we have passed down properties that aren't appropriate, which can throw a react warning in the console.

packages/react/src/primitives/Tabs/TabsContainer.tsx
```diff
@@ -20,6 +21,8 @@ const TabsContainerPrimitive: Primitive<TabsProps, 'div'> = (
   }: BaseTabsProps,
   ref
 ) => {
+  const groupId = useStableId();
+  const whitespaceValue = '-'; // used to create valid HTML IDs
```
reesscot, line 25: Since this never changes, it doesn't need to be declared inside of the component and should be moved outside of the component (near the top of file, under imports)

packages/react/src/primitives/Tabs/Tabs.tsx
axelEandrews: `rest` just contains `label: React.ReactNode` for the TabPanel, and for the TabList it just has `isDisabled: Boolean | undefined`. We may as well just pass them directly at that point if that's intended, right?

calebpollman: If additional parameters are added to the parent interface they'll bleed through. Is that behavior what we want here?

packages/react/src/primitives/Tabs/TabsContext.tsx
```diff
@@ -3,10 +3,14 @@ import * as React from 'react';
 export interface TabsContextInterface {
   activeTab: string;
   isLazy?: boolean;
+  groupId?: string;
```
calebpollman, line 6: Why is `groupId` optional?

packages/react/src/primitives/Tabs/TabsItem.tsx
```diff
@@ -17,23 +17,24 @@ const TabsItemPrimitive: Primitive<TabsItemProps, 'button'> = (
   { className, value, children, onClick, as = 'button', role = 'tab', ...rest },
   ref
 ) => {
-  const { activeTab, setActiveTab } = React.useContext(TabsContext);
+  const { activeTab, setActiveTab, groupId, whitespaceValue } =
+    React.useContext(TabsContext);
```
calebpollman, line 21: Since `whitespaceValue` is a static value, there is no need to pass it through context as it will not update

packages/react/src/primitives/Tabs/TabsContainer.tsx
```diff
@@ -7,6 +7,9 @@ import { primitiveWithForwardRef } from '../utils/primitiveWithForwardRef';
 import { BaseTabsProps, TabsProps } from './types';
 import { View } from '../View';
 import { TabsContext } from './TabsContext';
+import { useStableId } from '../utils/useStableId';
+
+const whitespaceValue = '-'; // used to create valid HTML IDs
```
calebpollman, line 12: Module scoped constants should be declared in SCREAMING_SNAKE_CASE

packages/react/src/primitives/Tabs/TabsContainer.tsx
```diff
@@ -20,6 +23,7 @@ const TabsContainerPrimitive: Primitive<TabsProps, 'div'> = (
   }: BaseTabsProps,
   ref
 ) => {
+  const groupId = useStableId();
```
calebpollman, line 26: It would be helpful to add a small comment explaining the use case for `groupId` for future maintainers

packages/react/src/primitives/Tabs/Tabs.tsx
axelEandrews: I removed this spread, wasn't needed so I think it's safer to just get rid of it.

packages/react/src/primitives/Tabs/TabsContainer.tsx
dbanksdesign: Love the name SCREAMING_SNAKE_CASE

dbanksdesign (approved):
LGTM

hbuchel (commented):
Just took a look at the Tabs docs page from this branch; it looks like somewhere we've created a breaking change because the Tabs no longer default to showing an open tab if the existing tab name had a space in it. This is what I'm seeing locally, can you please verify? 
Screenshot 2024-07-09 at 3 02 39 PM

hbuchel (changes requested):
Just throwing a block on this for now until the breaking change is addressed. Also, I re-ran the Docs e2e test (not sure why at first glance it failed)

PR #5306, opened 2024-06-11
chore: update flutter examples to use Amplify Flutter v2

Jordan-Nelson:
#### Description of changes
- Update the Flutter docs and example to use Amplify Flutter v2
- Update the background color of the Flutter authenticator example to match the docs site
- Update the Android device used in the Flutter authenticator example to a more modern device

Current UI
Screenshot 2024-06-11 at 4 10 38 PM

Updated UI
Screenshot 2024-06-11 at 4 11 00 PM

I deployed the changes to the URL below and manually confirmed the docs site and the flutter example built successfully.

https://chore-amplify-flutter-v2.d4urt9m0exbgc.amplifyapp.com/flutter/connected-components/authenticator

#### Checklist

- [x] PR description included

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -10,8 +10,8 @@ applications:
             - nvm use 20.13.1
             - node -v
             - export FLUTTER_HOME=${HOME}/sdks/flutter
-            # pin to flutter v3.10.6
-            - git clone -b 3.10.6 --depth 1 https://github.com/flutter/flutter.git ${FLUTTER_HOME}
+            # pin to flutter v3.22.2
+            - git clone -b 3.22.2 --depth 1 https://github.com/flutter/flutter.git ${FLUTTER_HOME}
             - export PATH="$PATH:${FLUTTER_HOME}/bin"
             # Skip cypress binary installation, as it's unneeded for docs and often fails transiently
             # https://docs.cypress.io/guides/references/advanced-installation#Skipping-installation
@@ -19,7 +19,7 @@ applications:
             - (cd .. && yarn install && yarn build)
         build:
           commands:
-            - nvm install 20.13.1 
+            - nvm install 20.13.1
             - nvm use 20.13.1
             - node -v
             - yarn flutter:build
--- a/docs/flutter/authenticator/lib/authenticator_config.dart
+++ b/docs/flutter/authenticator/lib/authenticator_config.dart
@@ -221,7 +221,7 @@ enum Device {
       case Device.ios:
         return Devices.ios.iPhone13;
       case Device.android:
-        return Devices.android.mediumPhone;
+        return Devices.android.samsungGalaxyS20;
       case Device.iPad:
         return Devices.ios.iPad;
       case Device.web:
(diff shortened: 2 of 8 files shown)


Equartey (dismissed):
Your changes look good to me. 

I did notice when the color scheme is set to system default, and when system is set to dark mode, it doesn't use the dark color scheme. This doesn't seem related to your changes tho. Here's a screenshot of my dark mode chrome, use system pref, and white background.
Screenshot 2024-06-12 at 9 46 16 AM

calebpollman (approved):
🚢

Jordan-Nelson:
> I did notice when the color scheme is set to system default, and when system is set to dark mode, it doesn't use the dark color scheme. This doesn't seem related to your changes tho. Here's a screenshot of my dark mode chrome, use system pref, and white background.

Good catch. I missed this initially. It seems to only happen if you have the color scheme is set to system default on page load. If it starts as dark mode and is then changed the background is dark. Agreed it is not new issue but I do want to resolve it. I'll see if it is a simple fix.

docs/src/components/FlutterAuthenticatorExample.tsx
```diff
@@ -12,15 +14,20 @@ export function FlutterAuthenticatorExample({
...
+    );
+    if (themeMode === 'system') {
+      if (window.matchMedia('(prefers-color-scheme:dark)').matches) {
+        themeMode = 'dark';
+      } else {
+        themeMode = 'light';
+      }
+    }
+    setThemeMode(themeMode);
+  }, []);
```
calebpollman, line 30: Without full context might be missing something but it looks like `themeMode` could be resolved with a ternary, e.g:
```ts
 const themeMode = window.matchMedia('(prefers-color-scheme:dark)').matches) ? 'dark' : 'light';
```

calebpollman, line 30: Without full context might be missing something but it looks `lightMode` could be resolved sync. Is there not a hook that already provides this information though?

Jordan-Nelson: I am not aware of an existing hook.

I have updated this because the original implementation wasn't working as expected, but I am wondering if there is a better way to do this.

The current implementation in main is flawed because `data-amplify-color-mode` is not set when this code runs on a page refresh. That results in themeMode always being "light" after a page refresh.

I initially used `useLayoutEffect` to wait until DOM mutations are complete. This also is flawed because themeMode never updates.

The current change in the PR uses `useLayoutEffect` but passes in a fn that returns the value of `getAttribute('data-amplify-color-mode')` as a dependency. This forces the fn to run when that attribute updates. I am not sure if there is a better way to do in react.

Jordan-Nelson: I could update the PR to change `useLayoutEffect` to `useEffect`. `useLayoutEffect` is not needed now that the getter for data-amplify-color-mode is a dependency.

Jordan-Nelson: I reverted these changes as the issue they aimed to fix were not related to this PR and it doesn't seem to be reproducible in prod.

calebpollman (approved):

Equartey (approved):

PR #5311, opened 2024-06-12
fix(liveness): fix camera switching after ws connection established

thaddmt:
#### Description of changes
- Issue: with no-light feature we decided to open the websocket connection on the camera check screen, the websocket connection takes in a readable stream derived from the MediaRecorder as an input, and we don't have a way to change that mid connection
- Potential solution: Now when users change camera we create a new MediaRecorder object and feed it's even stream into the originally set MediaRecorder that was used as an input to the websocket connection. Any events that are sent to the new camera / MediaRecorder are passed through to the original mediaRecorder which we maintain a reference to

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/StreamRecorder/StreamRecorder.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/StreamRecorder/StreamRecorder.ts
@@ -12,6 +12,7 @@ export class StreamRecorder {
   #recordingStartTimestamp: number | undefined;
   #recorderStopped!: Promise<void>;
   #videoStream: VideoStream;
+  #eventListeners: { [key: string]: (args: any) => void };
 
   constructor(stream: MediaStream) {
     if (typeof MediaRecorder === 'undefined') {
@@ -22,6 +23,7 @@ export class StreamRecorder {
     this.#recorder = new MediaRecorder(stream, { bitsPerSecond: 1000000 });
     this.#initialRecorder = this.#recorder;
     this.#videoStream = this.#createReadableStream();
+    this.#eventListeners = {};
 
     this.#setupCallbacks();
   }
@@ -31,8 +33,9 @@ export class StreamRecorder {
   }
 
   setNewVideoStream(stream: MediaStream): void {
+    this.#cleanUpEventListeners();
     this.#recorder = new MediaRecorder(stream, { bitsPerSecond: 1000000 });
-    this.#createPassThroughRecorder();
+    this.#eventListeners = this.#createPassThroughRecorder();
   }
 
   dispatchStreamEvent<T extends StreamResultType>(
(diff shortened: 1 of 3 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/utils/StreamRecorder/StreamRecorder.ts
```diff
@@ -123,6 +130,43 @@ export class StreamRecorder {
...
+   * @param oldRecorder
+   */
+  #createPassThroughRecorder(): void {
+    this.#recorder.ondataavailable = ({ data }) => {
+      this.#initialRecorder.dispatchEvent(
+        new MessageEvent('dataavailable', { data })
+      );
+    };
+
+    this.#recorder.addEventListener('sessionInfo', (e: any) => {
```
reesscot, line 145: Do we need some sort of cleanup to remove the event listeners if they switch to a different recorder?

thaddmt: added event listener clean up when they switch recorders

esauerbo (dismissed):

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/StreamRecorder/StreamRecorder.ts
```diff
@@ -28,6 +32,12 @@ export class StreamRecorder {
     return this.#videoStream;
   }
 
+  setNewVideoStream(stream: MediaStream): void {
```
calebpollman, line 35: Why can't instantiate a new `StreamRecorder` on camera change?

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/StreamRecorder/StreamRecorder.ts
```diff
@@ -123,6 +133,55 @@ export class StreamRecorder {
...
+    this.#recorder.addEventListener('endStream', onEndStreamHandler);
+
+    this.#setupCallbacks();
+
+    return {
+      endStream: onEndStreamHandler,
+      closeCode: onCloseCodeHandler,
+      streamStop: onStreamStopHandler,
+      sessionInfo: onSessionInfoHandler,
+      dataavailable: onDataAvailableHandler,
```
calebpollman, line 181: Can just set the handlers on `this.#eventHandlers` instead of returning the handlers?

thaddmt: haha yes, i'm not sure why i didn't do that

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/StreamRecorder/StreamRecorder.ts
thaddmt: We could do something like that but I figured it would be more complicated. In this solution we can keep track of the initial and the current `MediaRecorder` with the current one feeding data into the initial one. 

With instantiating multiple `StreamRecorders` we would have to instantiate a different type of `StreamRecorder`class which feeds streams into another `StreamRecorder`.

esauerbo (dismissed):

reesscot (dismissed):

calebpollman (approved):

PR #5374, opened 2024-07-09
feat(react): useUserAttributes Multipurpose Hook

axelEandrews:
#### Changes by File (Added 7/10):

useUserAttrActionTypes: 
Defines interfaces and types to be used within the useUserAttributes multipurpose hook. Each action has a defined type, and each action has a useAction type which leverages useDataState to fit the inputs/outputs necessary for that action. Where needed, default outputs are defined (`sendVerificationCode`, `updateUserAttributes`) such that useDataState has default data to use. Interface `UseActions` maps each action to its respective type, allowing for correct assignment of return types based on the input action in useUserAttributes. Several functions are defined for input type-predication, narrowing the input from `AttributeManagementInputs` which contains the five possible input types to the singular input type required for the given action. 

useUserAttributes:
Defines the multipurpose hook to be used for User Attribute management. Imports the `useAction` hooks from useUserAttrActionTypes, then initializes each within the hook (TASK: only create needed hook dependent on input action). A Hub listener is present only if the input action was `'fetch'`, which listens for the UI-Hub events dispatched by the actions defined in useUserAttrActionTypes. The appropriate `[state, handleState]` pair dependent on the input action type is returned. Something to note about this current implementation is that `null` must be explicitly passed to `handleFetch()`, something that would be best to avoid in the final implementation. 

useDataState/tests:
This version of amplify-ui was forked before useDataState was added, so it wasn't present in my cloned copy. I'm sure there's a better way to do what I did, but I essentially just copied in the up-to-date version to the correct directory. 

constants:
Consolidated into useUserAttrActionTypes.

##### Original PR (7/9):
Adds confirmUserAttributeAction, deleteUserAttributesAction, fetchUserAttributesAction, sendUserAttributeVerificationCodeAction, and updateUserAttributesAction, all to be used with the useDataState hook to achieve User Attribute Management functionality. Also, several interfaces are included with the intent of defining the inputs and outputs of each action (the inputs to handleAction and outputs to state.data). No changeset included yet - need to confirm correct location of hook installment. 

Also adds an initial implementation of useUserAttributes multipurpose hook. This version explicitly defines each `[state, handleAction]` return pair, and there are several typing issues. When hovering over the returned objects, the typing is what is should be (or missing undefined in some cases). This hook is functional in localhost when built on top of the `amplify-vite-react-template`. 

`constants.ts` features some default inputs to the actions used by `useDataState`, and they can should be condensed into the same file as `useUserAttrsActions` later on. 

Also added a copy of `useDataState.ts` so that it is present in this repo. 

#### Issue #, if available
No issue #, working towards full code implementation of [Amplify UI User Attribute Management](https://quip-amazon.com/lUUwA2aJEufP/Amplify-UI-User-Attribute-Management-APITech-Design).

#### Checklist

- [x] PR description included


packages/react/src/hooks/useDataState.ts
```diff
@@ -0,0 +1,42 @@
...
+  message: string | undefined;
+}
+
+const getActionState = <T>(data: T): ActionState<T> => ({
+  data,
+  isLoading: false,
+  message: undefined,
+});
+
+export function useDataState<T, K>(
```
calebpollman, line 15: `useDataState` is available in `@aws-amplify/ui-react-core`, no need to duplicate it

packages/react/src/hooks/constants.ts
```diff
@@ -0,0 +1,272 @@
...
+  preferred_username: undefined,
+  profile: undefined,
+  sub: undefined,
+  updated_at: undefined,
+  website: undefined,
+  zoneinfo: undefined,
+};
+
+// Default object to be passed into useDataState
+export const defaultUpdateUserAttributesOutput: UpdateUserAttributesOutput = {
```
calebpollman, line 33: Does the output of `updateUserAttributes` always include every field?

packages/react/src/hooks/constants.ts
```diff
@@ -0,0 +1,272 @@
+import {
+  SendUserAttributeVerificationCodeOutput,
+  UpdateUserAttributesOutput,
+  UserAttributeKey,
+} from "aws-amplify/auth";
+
+export type UserAttributes = Record<UserAttributeKey, string | undefined>;
```
calebpollman, line 7: Think we probably have this type previously defined somewhere in the codebase

packages/react/src/hooks/constants.ts
```diff
@@ -0,0 +1,272 @@
...
+  profile: undefined,
+  sub: undefined,
+  updated_at: undefined,
+  website: undefined,
+  zoneinfo: undefined,
+};
+
+// Default object to be passed into useDataState
+export const defaultUpdateUserAttributesOutput: UpdateUserAttributesOutput = {
+  ["custom:placeholder"]: {
```
calebpollman, line 34: What is this for?

packages/react/src/hooks/constants.ts
```diff
@@ -0,0 +1,272 @@
...
+    },
+  },
+  address: {
+    isUpdated: true,
+    nextStep: {
+      codeDeliveryDetails: {
+        attributeName: undefined,
+        deliveryMedium: undefined,
+        destination: undefined,
+      },
```
calebpollman, line 52: Does `updateUserAttributes` return a `codeDeliveryDetails` object for every attribute?

packages/react/src/hooks/constants.ts
```diff
@@ -0,0 +1,272 @@
...
+      updateAttributeStep: "DONE",
+    },
+  },
+  address: {
+    isUpdated: true,
```
calebpollman, line 46: Need clarity on why we would provide a default value that points to a successful update action, same with providing `updateAttributeStep: "DONE"` as a default value

packages/react/src/hooks/constants.ts
```diff
@@ -0,0 +1,272 @@
...
+    },
+  },
+};
+// Default object to be passed into useDataState
+export const defaultSendUserAttributeVerificationCodeOutput: SendUserAttributeVerificationCodeOutput =
+  {
+    attributeName: "email",
+    deliveryMedium: "EMAIL",
+    destination: "a***@b.com",
+  };
```
calebpollman, line 272: What if the configuration of the consuming project has only enabled `phone_number` as a verifiable attribute?

packages/react/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [sendVerificationCodeState, handleSendVerificationCode] = useDataState(
+    sendUserAttributeVerificationCodeAction,
+    defaultSendUserAttributeVerificationCodeOutput
+  );
+  const [fetchState, handleFetch] = useDataState(
+    fetchUserAttributesAction,
+    DefaultAttributes
+  );
+
+  const fetchHub: HubCallback = React.useCallback(
```
calebpollman, line 76: You can just define the listener inside the useEffect on ln 104

packages/react/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,128 @@
...
+  // Hub subscriptions
+  // Declare both subscriptions in same useEffect
+  // signOut 'auth', FETCH_ATTRIBUTES 'ui'
+
+  // I DON"T KNOW IF WE REALLY NEED TO LISTEN TO AUTH ON TOP OF UI, COULD JUST FOCUS ON OUR EVENTS
+  // React.useEffect(() => {
+  //   const unsubscribeAuth = Hub.listen("auth", fetchHub);
+  //   handleFetch();
+  //   return unsubscribeAuth;
+  // }, [fetchHub, handleFetch]);
```
calebpollman, line 102: Unless noting a PR explicitly WIP, commented out code should be removed prior to requesting review

packages/react/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,128 @@
...
+import {
+  HandlerInputs,
+  StateDataTypes,
+  confirmUserAttributeAction,
+  deleteUserAttributesAction,
+  fetchUserAttributesAction,
+  sendUserAttributeVerificationCodeAction,
+  updateUserAttributesAction,
+  Actions,
+} from "./actionTypes";
```
calebpollman, line 23: Where are these defined?

packages/react/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,128 @@
...
+import React from "react";
```
calebpollman, line 24: External deps should be imported before imports from our packages, followed by local imports. See https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#imports (Could probably use some updating, feel free to reach out offline with questions)

packages/react/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,128 @@
...
+  message: string | undefined;
+}
+
+// interface Actions {
+//   confirm: typeof confirmUserAttributeAction;
+//   delete: typeof deleteUserAttributesAction;
+//   sendVerificationCode: typeof sendUserAttributeVerificationCodeAction;
+//   update: typeof updateUserAttributesAction;
+//   fetch: typeof fetchUserAttributesAction;
+// }
```
calebpollman, line 47: Remove commented out code

packages/react/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,128 @@
...
+const useUserAttributes = <T extends keyof Actions>(
+  action: T
+): [
+  state: ActionState<StateDataTypes[T]>,
+  handleAction: (...input: HandlerInputs[T][]) => void
+] => {
+  const [deleteState, handleDelete] = useDataState(
+    deleteUserAttributesAction,
+    undefined
+  );
```
calebpollman, line 58: Let's break these in to individual hooks to isolate unit testing and recompose them here. Example:
```ts
 const useDeleteUserAttributes = useDataState(
    deleteUserAttributesAction,
    undefined
  );

const useUserAttributes = <T extends keyof Actions>(
  action: T
): [
  state: ActionState<StateDataTypes[T]>,
  handleAction: (...input: HandlerInputs[T][]) => void
] => {
  const delete = useDeleteUserAttributes();
  const update = useUpdateUserAttributes();
  // ...etc
```

packages/react/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,128 @@
...
+  // React.useEffect(() => {
+  //   const unsubscribeAuth = Hub.listen("auth", fetchHub);
+  //   handleFetch();
+  //   return unsubscribeAuth;
+  // }, [fetchHub, handleFetch]);
+
+  React.useEffect(() => {
+    if (action === "fetch") {
+      const unsubscribe = Hub.listen("ui", fetchHub);
+      // handleFetch()
```
calebpollman, line 107: Remove commented code

calebpollman (commented):
Great start! Left some feedback to address, also unclear where the action handlers are defined

packages/react/src/hooks/constants.ts
axelEandrews: Originally I'd seen a type error if I didn't include at least one custom attribute on the defaultUpdate type, but now I've removed the custom one. Looking to the comment below too, I've removed the codeDeliveryDetails because it's not required. One weird thing still is that updateUserAttributes only returns the attributes and the nextStep for the attributes you submitted, but if I don't include every standard attribute in the defaultUpdateOutput I see type errors.

packages/react/src/hooks/constants.ts
axelEandrews: No, it only includes the fields you submit. However, when feeding it into useDataState, I see an error unless all the standard attributes are included. Likely some typing issue, using a `Partial<UpdateUserAttributesOutput>` does not resolve it.

calebpollman (commented):
Great progress! A little worried that the size of this PR is going to make it complicated to continue reviewing as we need to begin adding unit tests. Suggest addressing the feedback here and then opening a separate PR with just `useFetchUserAttributes`, `fetchUserAttributesAction`, related types, and relevant unit tests

packages/react-core/src/hooks/useUserAttrActionTypes.ts
```diff
@@ -0,0 +1,411 @@
...
+): input is SendUserAttributeVerificationCodeInput {
+  return (
+    typeof input === 'object' && input !== null && 'userAttributeKey' in input
+  );
+}
+function isFetchUserAttributesInput(
+  input: AttributeManagementInputs
+): input is null {
+  return input === null;
+}
```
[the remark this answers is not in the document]
axelEandrews: Yes, fixing `Actions` to be the actual actions removed the need for this.

packages/react-core/src/hooks/useUserAttrActionTypes.ts
```diff
@@ -0,0 +1,411 @@
...
+  preferred_username: undefined,
+  profile: undefined,
+  sub: undefined,
+  updated_at: undefined,
+  website: undefined,
+  zoneinfo: undefined,
+};
+
+// Default object to be passed into useDataState
+export const defaultUpdateUserAttributesOutput: UpdateUserAttributesOutput = {
```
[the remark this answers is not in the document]
axelEandrews: `isUpdated: false` only occurs if `nextStep.updateAttributeStep === 'CONFIRM_ATTRIBUTE_WITH_CODE"` from what I've seen, so I'm not sure that setting these to be false would make more sense. The way I have it set up now is the state returned from a submission of all attributes where nothing needs to be confirmed, but I'll keep thinking about that.

packages/react-core/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,64 @@
...
+  useConfirmUserAttribute,
+  useDeleteUserAttributes,
+  useFetchUserAttributes,
+  useSendUserAttributeVerificationCode,
+  useUpdateUserAttributes,
+} from './useUserAttrActionTypes';
+
+const useUserAttributes = <T extends keyof Actions>(
+  action: T
+): [state: UseActions[T][0], handleAction: UseActions[T][1]] => {
```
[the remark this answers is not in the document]
axelEandrews: I thought that made sense too, but see errors when actually doing on the `return []` in the final `switch`.

packages/react-core/src/hooks/useUserAttrActionTypes.ts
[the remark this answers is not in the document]
calebpollman: Rather than setting them to `false`, can you think of how this interface could be re-written to simplify the default value and usage by consumers?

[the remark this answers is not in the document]
axelEandrews: Got it, I think the latest commit makes more sense. See the new type `UpdateAttributesOutput` in `useUserAttributes/types.ts`.

packages/react-core/src/hooks/useUserAttributes.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const useSendCode = useSendUserAttributeVerificationCode();
+
+  const handleFetch = useFetch[1];
+
+  useEffect(() => {
+    if (action === 'fetch') {
+      const fetchHub: HubCallback = ({ payload }) => {
+        switch (payload.event) {
+          case 'FETCH_ATTRIBUTES': {
+            handleFetch(null);
```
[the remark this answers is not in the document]
axelEandrews: This was also caused by incorrect `Actions` interface.

packages/react-core/src/hooks/useUserAttrActionTypes.ts
```diff
@@ -0,0 +1,411 @@
...
+  } else {
+    try {
+      const result = await updateUserAttributes(input);
+      Hub.dispatch('ui', {
+        event: 'FETCH_ATTRIBUTES',
+        message: 'attributes updated successfully',
+      });
+      return result;
+    } catch (error) {
+      Hub.dispatch('ui', {
```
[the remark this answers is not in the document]
axelEandrews: Removed these

packages/react-core/src/hooks/useUserAttrActionTypes.ts
```diff
@@ -0,0 +1,411 @@
...
+];
+
+export const useDeleteUserAttributes = (): UseDeleteUserAttributes =>
+  useDataState(deleteUserAttributesAction, undefined);
+
+export const updateUserAttributesAction = async (
+  _prev: Awaited<ReturnType<typeof updateUserAttributes>>,
+  input: UpdateUserAttributesInput
+): Promise<UpdateUserAttributesOutput> => {
+  if (!isUpdateUserAttributesInput(input)) {
```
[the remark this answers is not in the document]
axelEandrews: Removed the input verification after fixing `Actions` interface.

PR #5378, opened 2024-07-09
fix(ui-react): Whitespace in Tab IDs

axelEandrews:
Adds `groupId` to TabsContext such that unique IDs can be created for each tab group. IDs are of the form `${groupId}-tab-${idValue}` for tabs, and`${groupId}-panel-${idValue}` for panels. `idValue` is the given `value` stripped of its whitespace using the added constant `WHITESPACE_VALUE`. `idValue` is created in TabsPanel and TabsItem, so assuming the developer enters the same values for these two elements the functionality works. `defaultValue` and controlled tab functionality still works because the actual `value` of the tabs is not altered in any way, but it is highly recommended to not assign `value` to a string containing whitespace. 

Outer html output for a tablist with given values: ['Tab 1', 'Tab 2', 'Tab 3']: `<div role="tablist" class="amplify-tabs__list" style="justify-content: flex-start;"><button role="tab" id="amplify-id-:rl:-tab-Tab-1" aria-selected="true" aria-controls="amplify-id-:rl:-panel-Tab-1" class="amplify-tabs__item amplify-tabs__item--active">Tab 1</button><button role="tab" id="amplify-id-:rl:-tab-Tab-2" aria-selected="false" aria-controls="amplify-id-:rl:-panel-Tab-2" tabindex="-1" class="amplify-tabs__item">Tab 2</button><button disabled="" role="tab" id="amplify-id-:rl:-tab-Tab-3" aria-selected="false" aria-controls="amplify-id-:rl:-panel-Tab-3" tabindex="-1" class="amplify-tabs__item">Disabled tab</button></div>`

Also updated the ComposedTabsExample.tsx file such that each Tab has unique content (previously each was "Tab 1 Content"). 

#### Issue #, if available
#5220 

#5302 is an old version of this PR that had errors

Also, edited the Amplify UI Docs examples to explore cases where `value` and `defaultValue` are passed with whitespace to ensure composability, control, and disable still work as intended. User experience perfectly matches the production version of Amplify UI Docs. 

#### Checklist

- [x] PR description included
- [] Relevant documentation is changed or added (and PR referenced)

### DIFF
--- a/packages/react/src/primitives/Tabs/TabsContainer.tsx
+++ b/packages/react/src/primitives/Tabs/TabsContainer.tsx
@@ -7,6 +7,8 @@ import { primitiveWithForwardRef } from '../utils/primitiveWithForwardRef';
 import { BaseTabsProps, TabsProps } from './types';
 import { View } from '../View';
 import { TabsContext } from './TabsContext';
+import { useStableId } from '../utils/useStableId';
+import { WHITESPACE_VALUE } from './constants';
 
 const TabsContainerPrimitive: Primitive<TabsProps, 'div'> = (
   {
@@ -20,7 +22,11 @@ const TabsContainerPrimitive: Primitive<TabsProps, 'div'> = (
   }: BaseTabsProps,
   ref
 ) => {
+  const groupId = useStableId(); // groupId is used to ensure uniqueness between Tab Groups in IDs
   const isControlled = controlledValue !== undefined;
+  if (defaultValue && typeof defaultValue === 'string') {
+    defaultValue = defaultValue.replace(' ', WHITESPACE_VALUE); // remove whitespace from defaultValue
+  }
   const [localValue, setLocalValue] = React.useState(() =>
     isControlled ? controlledValue : defaultValue
   );
@@ -44,8 +50,9 @@ const TabsContainerPrimitive: Primitive<TabsProps, 'div'> = (
       activeTab,
       isLazy,
       setActiveTab,
+      groupId,
     };
-  }, [activeTab, setActiveTab, isLazy]);
+  }, [activeTab, setActiveTab, isLazy, groupId]);
 
   return (
     <TabsContext.Provider value={_value}>
--- a/packages/react/src/primitives/Tabs/TabsContainer.tsx
+++ b/packages/react/src/primitives/Tabs/TabsContainer.tsx
@@ -24,9 +24,6 @@ const TabsContainerPrimitive: Primitive<TabsProps, 'div'> = (
 ) => {
   const groupId = useStableId(); // groupId is used to ensure uniqueness between Tab Groups in IDs
   const isControlled = controlledValue !== undefined;
-  if (defaultValue && typeof defaultValue === 'string') {
-    defaultValue = defaultValue.replace(' ', WHITESPACE_VALUE); // remove whitespace from defaultValue
-  }
   const [localValue, setLocalValue] = React.useState(() =>
     isControlled ? controlledValue : defaultValue
   );
--- a/packages/react/src/primitives/Tabs/TabsContainer.tsx
+++ b/packages/react/src/primitives/Tabs/TabsContainer.tsx
@@ -8,7 +8,6 @@ import { BaseTabsProps, TabsProps } from './types';
 import { View } from '../View';
 import { TabsContext } from './TabsContext';
 import { useStableId } from '../utils/useStableId';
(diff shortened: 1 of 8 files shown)


hbuchel (dismissed):
I re-ran through pa11y, axe and VoiceOver and looks good to me!

timngyn (dismissed):

packages/react/src/primitives/Tabs/TabsContainer.tsx
```diff
@@ -20,7 +22,11 @@ const TabsContainerPrimitive: Primitive<TabsProps, 'div'> = (
   }: BaseTabsProps,
   ref
 ) => {
+  const groupId = useStableId(); // groupId is used to ensure uniqueness between Tab Groups in IDs
   const isControlled = controlledValue !== undefined;
+  if (defaultValue && typeof defaultValue === 'string') {
+    defaultValue = defaultValue.replace(' ', WHITESPACE_VALUE); // remove whitespace from defaultValue
```
calebpollman, line 28: We can't trim the `defaultValue` since that os overwriting the value that the customer passed in

calebpollman (changes requested):

packages/react/src/primitives/Tabs/TabsContainer.tsx
axelEandrews: That's the same thing we are doing to the values they pass in, this ensures the same change to the `defaultValue` that was done to the `value`. The original fix made it so that all `value` produced legal html IDs, but created the case where `defaultValue` also had whitespace and then could not reference the correct `value` generated ID.

calebpollman: Modifying provided values breaks the contract of `onValueChange` - consumers cannot reliably trust that the values they have provided to the `Tabs` are the values that are provided to their `onValueChange` callback

axelEandrews: Ok, switched back to using `idValue = value.replace()` and doing nothing to `defaultValue`. Preserves user-inputted values while maintaining legal html IDs.

timngyn (dismissed):

hbuchel (dismissed):

packages/react/src/primitives/Tabs/TabsItem.tsx
```diff
@@ -12,28 +12,32 @@ import { View } from '../View';
...
   return (
     <View
       {...rest}
       role={role}
       as={as}
-      id={`${value}-tab`}
+      id={`${groupId}-tab-${idValue}`}
       aria-selected={isActive}
-      aria-controls={`${value}-panel`}
+      aria-controls={`${groupId}-panel-${idValue}`}
```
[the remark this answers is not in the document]
axelEandrews: `value` for `TabsPanel` and `TabsItem` is required to be a string

packages/react/src/primitives/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -86,6 +86,61 @@ describe('Tabs', () => {
...
+          <Tabs.Item value="3">Tab 3</Tabs.Item>
+        </Tabs.List>
+        <Tabs.Panel value="1">Tab 1</Tabs.Panel>
+        <Tabs.Panel value="2">Tab 2</Tabs.Panel>
+        <Tabs.Panel value="3">Tab 3</Tabs.Panel>
+      </Tabs.Container>
+    );
+    const tabs = await screen.findAllByRole('tab');
+    expect(tabs[0].id === tabs[1].id).toBeFalsy;
+    expect(tabs[0].id === tabs[2].id).toBeFalsy;
```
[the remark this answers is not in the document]
axelEandrews: Fixed this

packages/react/src/primitives/Tabs/TabsItem.tsx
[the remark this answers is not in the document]
axelEandrews: From the testing I've done this change does not seem to be breaking. At this stage of the Tabs composition `value` is required to be defined, so creating `idValue` based on it does not seem to introduce any errors. Because `value` itself is not changed, as long as `defaultValue` matches the intended target Tab's `value` the controllability and default functionality seems to work as intended. Tested by changing `value` and `defaultValue` for the examples to include whitespace.

timngyn (approved):
lgtm!

calebpollman (approved):

PR #5390, opened 2024-07-11
Feat(identityProviders): useIsSignedIn Hook

AhmetAhunbayAWS:
#### Description of changes

Creates a new react hook that gives customers access to whether or not the end user is signed in. Creates a tester file and updates index files to expose the hook to the customer. Adds new hook to jest snapshot file. This react hook is decoupled from the existing Authenticator.

I created a test file to go over possible cases and ran yarn test.

#### Checklist

- [x] PR description included


packages/react-core/src/hooks/useIsSignedIn.ts
```diff
@@ -0,0 +1,45 @@
...
+      });
+  }, []);
+
+  useEffect(() => {
+    const listener: HubCallback = ({ payload }) => {
+      if (payload.event === 'signedIn') {
+        setIsSigned(true);
+      } else if (payload.event === 'signedOut') {
+        setIsSigned(false);
+      }
```
thaddmt, line 34: Does this work with other auth hub events? I recall there being events like `tokenRefresh_failure `, does `signedOut` get called after those events and does the hook successfully show signed in status for something like that?

thaddmt (commented):
LGTM other than a question!

packages/react-core/src/hooks/useIsSignedIn.ts
AhmetAhunbayAWS: I think yes, I have to catch that event as well. 

If you look at [useAuth](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/src/hooks/useAuth.ts), it seems like tokenRefresh_failure and signIn_failure have the same functionality, which is different from signedOut and doesn't update the useAuth user state. 

But if you look at [defaultHubHandler](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/helpers/authenticator/defaultAuthHubHandler.ts), both tokenRefresh_failure and signedOut send out the same event that triggers a signOutActor in the state machine. But I did not find any references to this defaultHubHandler anywhere else in the codebase. 

Also, I asked someone on JS and they said they do not send a signedOut event along with the tokenRefresh_failure event.

thaddmt (approved):

packages/react-core/src/hooks/useIsSignedIn.ts
thaddmt: Sounds good to me

axelEandrews (approved):
Looks good to me!

PR #5400, opened 2024-07-15
chore(react-core): update useDataState async logic and add hasError

calebpollman:
#### Description of changes
- abstract forced async resolving in to `resolveMaybeAsync`
- add `hasError` to `DataState`
- add/update unit tests

#### Issue #, if available
NA

unit tests
#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-core/src/hooks/useDataState.ts
+++ b/packages/react-core/src/hooks/useDataState.ts
@@ -1,42 +1,50 @@
 import React from 'react';
 
-interface ActionState<T> {
+export interface DataState<T> {
   data: T;
+  hasError: boolean;
   isLoading: boolean;
   message: string | undefined;
 }
 
-const getActionState = <T>(data: T): ActionState<T> => ({
-  data,
-  isLoading: false,
-  message: undefined,
-});
+// default state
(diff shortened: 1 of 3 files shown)


packages/react-core/src/hooks/useDataState.ts
```diff
@@ -1,42 +1,50 @@
...
   message: string | undefined;
 }
 
-const getActionState = <T>(data: T): ActionState<T> => ({
-  data,
-  isLoading: false,
-  message: undefined,
-});
+// default state
+const INITIAL_STATE = { hasError: false, isLoading: false, message: undefined };
```
axelEandrews, line 11: Would it make sense to have a default message for the initial state? I'm just thinking that could be useful to distinguish between an action that didn't add a message and having not performed the action yet, but I can see how it might obfuscate customer experience.

axelEandrews (approved):
Left one comment on the `INITIAL_STATE`, the rest looks good to me.

thaddmt (approved):

packages/react-core/src/hooks/useDataState.ts
calebpollman: That's an interesting proposal! Will consider it in the future

PR #5420, opened 2024-07-17
chore: update types for createAIConversation inputs, added an AutoHid…

thaddmt:
…ablePrompt

#### Description of changes
- Updated createAIConversation to get more values from the input
- Updated AIConversation component to accept messages and other inputs
- Added an AutoHidablePrompt control for internal use

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-ai/src/components/AIConversation/createAIConversation.tsx
+++ b/packages/react-ai/src/components/AIConversation/createAIConversation.tsx
@@ -1,12 +1,17 @@
 import React from 'react';
-import { Controls, AIConversationInput, AIConversation } from './types';
+import {
+  Controls,
+  AIConversationInput,
+  AIConversation,
+  AIConversationProps,
+} from './types';
 import {
   ActionsBarControl,
   AvatarControl,
   Conversation,
   HeaderControl,
   MessagesControl,
-  SuggestedPrompts,
+  PromptControl,
 } from './views';
 import { AIConversationElements } from './context/elements';
 import createProvider from './createProvider';
@@ -16,13 +21,14 @@ export function createAIConversation<T extends Partial<AIConversationElements>>(
 ): {
   AIConversation: AIConversation<T>;
 } {
-  const { elements } = input;
+  const { elements, suggestedPrompts, actions } = input;
 
-  const Provider = createProvider({ elements });
+  const Provider = createProvider({ elements, actions, suggestedPrompts });
 
-  function AIConversation(): JSX.Element {
+  function AIConversation(props: AIConversationProps): JSX.Element {
+    const { messages, avatars } = props;
     return (
-      <Provider>
+      <Provider messages={messages} avatars={avatars}>
         <Conversation />
       </Provider>
     );
@@ -37,11 +43,12 @@ export function createAIConversation<T extends Partial<AIConversationElements>>(
     Header: HeaderControl,
     // @ts-expect-error TODO fix type error
     Messages: MessagesControl,
+    // @ts-expect-error TODO fix type error
+    SuggestedPrompts: PromptControl,
   };
 
   AIConversation.Provider = Provider;
   AIConversation.Conversation = Conversation;
-  AIConversation.SuggestedPrompts = SuggestedPrompts;
   AIConversation.Controls = Controls;
 
   return { AIConversation };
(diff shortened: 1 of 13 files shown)


packages/react-ai/src/components/AIConversation/createAIConversation.tsx
```diff
@@ -16,13 +21,14 @@ export function createAIConversation<T extends Partial<AIConversationElements>>(
...
 
-  const Provider = createProvider({ elements });
+  const Provider = createProvider({ elements, actions, suggestedPrompts });
 
-  function AIConversation(): JSX.Element {
+  function AIConversation(props: AIConversationProps): JSX.Element {
+    const { messages, avatars } = props;
     return (
-      <Provider>
+      <Provider messages={messages} avatars={avatars}>
```
thaddmt, line 31: Moved all these values that we previously mocked to be passed in from the example app

packages/react-ai/src/components/AIConversation/views/Controls/PromptControl.tsx
```diff
@@ -111,6 +112,14 @@ export const PromptControl: PromptControl = () => (
...
 );
 
+export const AutoHidablePromptControl = (): JSX.Element | undefined => {
+  const messages = React.useContext(MessagesContext);
+
+  if (!messages || messages.length === 0) {
+    return <PromptControl />;
+  }
+};
+
```
thaddmt, line 122: @esauerbo I wasn't really sure if this was the best way to do this, but I figured internally our SuggestedPromptControl would show or hide automatically based on whether messages exist or not.

But if a customer were to compose it themselves they might find that automated behavior strange, so wanted to separate the auto hiding feature to it's own control. What do you think?

packages/react-ai/src/components/AIConversation/createAIConversation.tsx
esauerbo: What's the rationale behind passing `messages` and `avatars` as props versus `suggestedPrompts` and `actions` as input to `createAIConversation`?

packages/react-ai/src/components/AIConversation/views/Controls/PromptControl.tsx
esauerbo: I thinks that makes sense! 

Just spitballing here but one consideration is that there are a few other UI aspects (like showing the Avatar, disabling the `send` button, input placeholder text) that depend on the messages "state". Would we want to do the same thing in those cases? would it make sense to keep track of any messages metadata like # of messages in the context, or just perform that check within each component?

packages/react-ai/src/components/AIConversation/createAIConversation.tsx
thaddmt: This is because suggestedPrompts and actions can be determined at build time and can be defined outside the scope of the React component.

Messages will be fetched async so must be passed in inside the react component and can't be done in `createAIConversation`.

Avatar is for a similar reason in that Avatar information can be updated based on react context or anything the developer decides at runtime, so should be changable with react props as well

packages/react-ai/src/components/AIConversation/views/Controls/PromptControl.tsx
thaddmt: Yeah I think it might make sense to expand the messages context to include that type of metadata if we find it helpful? Seems like a bit of a case by case scenario but I can see that making sense for showing/hiding the avatar. Not sure about `send` button or placeholder text.

packages/react-ai/src/components/AIConversation/views/Controls/PromptControl.tsx
```diff
@@ -111,6 +112,14 @@ export const PromptControl: PromptControl = () => (
...
 PromptControl.Container = Container;
```
esauerbo, line 123: Are you intentionally not exposing the Icon/Header/Buttons on `PromptControl`?

packages/react-ai/src/components/AIConversation/createAIConversation.tsx
esauerbo: Got it

packages/react-ai/src/components/AIConversation/views/Controls/PromptControl.tsx
thaddmt: Ah was just a miss on my part, will add

esauerbo (approved):
LGTM!

hbuchel (approved):

PR #5421, opened 2024-07-18
[WIP] chore(storage-browser): add default styles export and default styles

hbuchel:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/rollup.config.ts
+++ b/packages/react-storage/rollup.config.ts
@@ -51,6 +51,15 @@ const config = defineConfig([
     },
     plugins: [styles({ mode: ['extract'] })],
   },
+  {
+    input: 'src/storage-browser-styles.ts',
+    output: {
+      dir: 'dist',
+      format: 'cjs',
+      assetFileNames: '[name][extname]',
+    },
+    plugins: [styles({ mode: ['extract'] })],
+  },
 ]);
 
 export default config;
--- a/packages/react-storage/src/components/StorageBrowser/Views/LocationsListView.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Views/LocationsListView.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { useDataState } from '@aws-amplify/ui-react-core';
 import { listLocationsAction } from '../context/actions';
+import { RefreshControl, TitleControl, DividerControl } from './Controls';
+import { CLASS_BASE } from './constants';
 
 export default function LocationsListView(): JSX.Element {
   const [{ data, isLoading }, handleListLocations] = useDataState(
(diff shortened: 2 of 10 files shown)


packages/react-storage/src/components/StorageBrowser/Views/LocationsListView.tsx
```diff
@@ -17,5 +19,22 @@ export default function LocationsListView(): JSX.Element {
     ? null
     : data.locations.map(({ name }) => <p key={name}>{name}</p>);
 
-  return <>{!hasLocations && isLoading ? 'loading...' : listLocations}</>;
+  return (
+    <>
+      <div className={`${CLASS_BASE}__header`}>
```
hbuchel, line 24: Need to determine where these divs with classNames actually need to live, for now they are just here for styling purposes.

packages/react-storage/src/storage-browser.css
```diff
@@ -0,0 +1,51 @@
+.amplify-storagebrowser__header {
+  display: flex;
+  flex-direction: column;
+  gap: 12px;
```
calebpollman, line 4: Are these hardcoded `gap` values just place holders?

packages/react-storage/rollup.config.ts
```diff
@@ -51,6 +51,15 @@ const config = defineConfig([
     },
     plugins: [styles({ mode: ['extract'] })],
   },
+  {
+    input: 'src/storage-browser-styles.ts',
```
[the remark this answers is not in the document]
hbuchel: styles.css is already exported from ui-react-storage, and is an export of the entire UI library's CSS (essentially the same thing if you imported `import '@aws-amplify/ui-react/styles.css';` into your app)

packages/react-storage/src/storage-browser.css
hbuchel: Mostly a placeholder; I could make them a css property (`var(--gap)`) and use a REM value like we do in UI.

hbuchel: Updated in latest to use css properties

timngyn (approved):

packages/react-storage/rollup.config.ts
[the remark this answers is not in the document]
thaddmt: Was wondering, since react-ai is new should we follow the same pattern? we currently only h ave the scoped styles of `ai-conversation-styles.css`

thaddmt (approved):

packages/react-storage/rollup.config.ts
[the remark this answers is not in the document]
calebpollman: Got it. So we don't bloat the _src_ directory, think it would make sense to nest these inside a _src/styles_ directory, but that can be done in a follow up

calebpollman (approved):

PR #5426, opened 2024-07-19
chore(storage-browser): scaffold TableControl

timngyn:
#### Checklist

- [ ] PR description included


examples/next/pages/ui/components/storage/storage-browser/composable-playground/index.page.tsx
```diff
@@ -20,24 +25,144 @@ const Title = React.forwardRef<HTMLHeadingElement>(
...
+        }
+
+        setRowData(sorted);
+
+        return { key: column.key, direction };
+      });
+    },
+    [rowData]
+  );
+
```
timngyn, line 165: I'm guessing this would be handled in context? Was just trying to set up the table control so that it was not doing any sorting logic

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,7 +1,135 @@
...
+  className: `${BLOCK_NAME}__sort-icon--descending`,
+  ...iconAttributes,
+});
+
+export interface ILocation {
+  /**
+   * Name of the specified location or object.
+   */
+  name: string;
+}
```
timngyn, line 105: I wanted to create a base interface for the table where users implementing the TableControl component could extend from it. I wanted to have `name` since this needs to be a button for navigating into the location, does that sound right?

packages/react-storage/src/components/StorageBrowser/hooks/useSortHandler.ts
timngyn, line 1: This is what I have for the sorting at the moment, but I'm not sure where it should go or if this is the right approach. 

Any feedback on this is appreciated!

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,7 +1,135 @@
...
+  rows: T[] | null;
+}
+
+export type SortDirection = 'ascending' | 'descending' | 'none';
+
+export interface TableControlProps<T extends ILocation> {
+  data: TableData<T>;
+  ariaLabel?: string;
+  onSort: (columnKey: Column<T>) => void;
+  sortState: { key: keyof T; direction: SortDirection };
```
timngyn, line 127: Added these as table props to help pass down info for sorting since we didn't want to sort within the component

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -10,5 +147,101 @@ export interface TableControl<T extends StorageBrowserElements>
...
+                column.key === sortState?.key ? sortState.direction : 'none'
+              }
+            >
+              {column.sortable && onSort ? (
+                <TableHeaderButton
+                  aria-label={`${column.header}, ${
+                    sortState.key === column.key
+                      ? sortAriaLabel[sortState.direction]
+                      : sortAriaLabel['none']
+                  }`}
```
timngyn, line 198: I originally thought `aria-sort` in the `th` element would be announced when it changed, but that wasn't the case. As a result I added this aria label to get read by the screen reader when you press the button. With VoiceOver, I was able to press the button with `ctrl+option+space` and have it say, `Name, sorted ascending` for instance

When moving through the table cells, it also reads this out which I thought was nice, but open to feedback on it

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,7 +1,144 @@
...
 
-export interface TableControl<T extends StorageBrowserElements>
-  extends Pick<
+const {
+  Table: BaseTable,
+  TableBody: BaseTableBody,
+  TableData: BaseTableData,
+  TableHead: BaseTableHead,
+  TableHeader: BaseTableHeader,
+  TableRow: BaseTableRow,
```
timngyn, line 12: Should I add other elements like `caption`?

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

packages/react-storage/src/index.ts
```diff
@@ -10,4 +10,5 @@ export {
   FilePickerProps,
   FileListHeaderProps,
   FileListFooterProps,
+  useSortHandler,
```
calebpollman, line 13: Anything that gets exported here becomes a public export - does this hook need to be?

packages/react-storage/src/components/StorageBrowser/hooks/useSortHandler.ts
```diff
@@ -0,0 +1,96 @@
...
+    // Otherwise, compare them as regular strings
+    return a.localeCompare(b);
+  } else if (typeof a === 'number' && typeof b === 'number') {
+    // Compare numbers
+    return a - b;
+  }
+  throw new Error('Invalid arguments');
+}
+
+interface ISortState<T> {
```
calebpollman, line 28: Why `ISortState`, none of our existing types have the `I` prefix?

timngyn: ah I can remove the `I`. I previously exported this interface and thought it was kind of confusing since I had `sortState` being used

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -10,5 +147,101 @@ export interface TableControl<T extends StorageBrowserElements>
...
+  props: TableControlProps<U>
+) => {
+  const { data, ariaLabel } = props;
+  const { rows, columns } = data;
+
+  let onSort: (columnKey: Column<U>) => void;
+  let sortState: { key: keyof U; direction: SortDirection };
+
+  if (isSortableTableControlProps(props)) {
+    ({ onSort, sortState } = props);
```
calebpollman, line 178: This re-assignment is hard to grok from a glance

packages/react-storage/src/index.ts
timngyn: With how I set up the table, this hook is used to help sort the data outside of the table. I'm not sure if this the right way to do that though. It is set up as a hook to move this code in a specific place for this review

I guess firstly, do you think having it sort outside of the table is still the approach we should take?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -10,5 +147,101 @@ export interface TableControl<T extends StorageBrowserElements>
...
+};
+
+export const TableControl: TableControl = <U extends ILocation>(
+  props: TableControlProps<U>
+) => {
+  const { data, ariaLabel } = props;
+  const { rows, columns } = data;
+
+  let onSort: (columnKey: Column<U>) => void;
+  let sortState: { key: keyof U; direction: SortDirection };
```
calebpollman, line 175: The Table doesn't need to know anything abt sorting data, it's responsibility is to render the data provided to it

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,7 +1,144 @@
...
+      clipRule="evenodd"
+      d="M3.16512 10.4442C2.72085 11.1086 3.19712 12 3.99642 12H12.0037C12.8021 12 13.2785 11.1104 12.8361 10.4458L8.84116 4.44491C8.44589 3.85117 7.5739 3.85029 7.17744 4.44324L3.16512 10.4442Z"
+      fill="#16191F"
+    />
+  ),
+  className: `${BLOCK_NAME}__sort-icon--descending`,
+  ...iconAttributes,
+});
+
+export interface ILocation {
```
calebpollman, line 100: Left another comment on this as well but we have not traditionally prefix types with `I`. Also, there is no need to use `Location` in the type name here, the `Table` is agnostic to that information

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,7 +1,144 @@
...
+  rows: T[] | null;
+}
+
+export type SortDirection = 'ascending' | 'descending' | 'none';
+
+interface SortableTableControlProps<T extends ILocation> {
+  data: TableData<T>;
+  ariaLabel?: string;
+  onSort: (columnKey: Column<T>) => void;
+  sortState: { key: keyof T; direction: SortDirection };
```
calebpollman, line 127: This information is not required by the `Table` to just render data

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -10,5 +147,101 @@ export interface TableControl<T extends StorageBrowserElements>
...
+    ({ onSort, sortState } = props);
+  }
+
+  return (
+    <Table aria-label={ariaLabel}>
+      <TableHead>
+        <TableRow>
+          {columns?.map((column) => (
+            <TableHeader
+              key={column.key as string}
```
calebpollman, line 187: Why does this need to be casted?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -10,5 +147,101 @@ export interface TableControl<T extends StorageBrowserElements>
...
+              )}
+            </TableHeader>
+          ))}
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {rows?.map((row, rowIndex) => (
+          <TableRow key={rowIndex}>
+            {columns.map((column) => (
+              <TableData key={`${rowIndex}-${column.key as string}`}>
```
calebpollman, line 222: Cast?

calebpollman (commented):
Left some feedback but we should sync offline on the approach

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
timngyn: That makes sense, I guess I wasn't sure how to have the `th` elements show that a column was being sorted though. Should this instead be included inside the data that gets passed to the table or something?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,8 +1,124 @@
...
+}
+
+export interface TableData<T extends Data> {
+  columns: Column<T>[];
+  rows: T[] | null;
+}
+
+interface TableControlProps<T extends Data> {
+  data?: TableData<T>;
+  ariaLabel?: string;
```
calebpollman, line 120: Needs to be removed per offline conversation

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -15,11 +131,63 @@ export interface TableControl<
...
+    columns: [],
+    rows: [],
+  };
+
+  return (
+    <Table aria-label={ariaLabel}>
+      <TableHead>
+        <TableRow>
+          {columns?.map((column) => (
+            <TableHeader key={String(column.key)} aria-sort="none">
```
calebpollman, line 155: Why does `column.key` need go be coerced in to a `string`? Same question for line 176

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,7 +1,128 @@
...
+const iconAttributes = {
+  'aria-hidden': true,
+  width: '24',
+  height: '24',
+  viewBox: '0 0 24 24',
+  fill: 'none',
+  xmlns: 'http://www.w3.org/2000/svg',
+};
+
+const SortIndeterminateIcon = withBaseElementProps(Icon, {
```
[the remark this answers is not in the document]
timngyn: I exported them as svg from the figma file

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
timngyn: I get a typescript error when not casting it because of my `Column` interface:

```
export interface Column<T extends Data> {
  header: string;
  key: keyof T;
  sortable: boolean; // placeholder
  sortType?: 'string' | 'number' | 'date'; // placeholder
}
```

`Type 'string | number | symbol' is not assignable to type 'Key | null | undefined'.
  Type 'symbol' is not assignable to type 'Key | null | undefined'.`
  
  Do you think this is okay to leave or should I update interface?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,7 +1,128 @@
...
+  children: (
+    <path
+      fillRule="evenodd"
+      clipRule="evenodd"
```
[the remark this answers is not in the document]
timngyn: I got them from the svg file after exporting them from figma so it was just a copy paste. Let me see if I can remove them

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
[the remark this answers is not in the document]
calebpollman: We need to pull all needed icons from https://fonts.google.com/icons with consistent fills, weights, etc so they are consistent

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
calebpollman: Need to update the interface, please remove anything related to sort as well

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -1,8 +1,103 @@
...
+};
+
+const SortIndeterminateIcon = withBaseElementProps(Icon, {
+  children: (
+    <path
+      d="M480-360 280-560h400L480-360Z"
+      fill="none"
+      stroke="currentColor"
+      strokeWidth="1"
+      vectorEffect="non-scaling-stroke"
```
timngyn, line 67: Added these attributes to create a similar triangle from the figma file:

image

calebpollman: Use https://fonts.google.com/icons?selected=Material+Symbols+Outlined:check_indeterminate_small:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=indeter&icon.size=24&icon.color=%23e8eaed for the indeterminate icon

calebpollman (approved):

reesscot (approved):

PR #5430, opened 2024-07-19
chore(raven): Add aria attributes to messages subcomponents

esauerbo:
#### Description of changes
Update message and action button elements with aria attributes 

#### Checklist

- [x] PR description included


packages/react-ai/src/components/AIConversation/views/Controls/MessagesControl.tsx
```diff
@@ -16,7 +16,7 @@ const MESSAGE_BLOCK = 'ai-message';
 
 const MediaContent = withBaseElementProps(Image, {
   className: `${MESSAGE_BLOCK}__image`,
-  src: 'test.jpg',
+  alt: 'AI Image',
```
hbuchel, line 19: Is this a placeholder? What is the context for this image (like how is it used)? Wondering if there is something more descriptive we should use.

esauerbo, line 19:
```suggestion
  alt: 'Image attachment',
```

esauerbo: The image would be something the user attaches and sends. Since they upload using file input I'm not sure there's a way to really add descriptive alt text without adding a separate field for that. What do you think about `Image attachment` as the alt text?

hbuchel: Yeah, that might make more sense. Whatever lines up with the file upload label maybe so the user can associate it to be the thing they just uploaded.

hbuchel (approved):

thaddmt (approved):

PR #5457, opened 2024-07-24
chore(raven): Add default icons

esauerbo:
#### Description of changes
- Create IconElement that returns default icon variants
- Add DEFAULT_ICON_PATHS that correspond to material icons 

#### Checklist

- [ ] PR description included


packages/react-ai/src/components/AIConversation/views/Controls/ActionsBarControl.tsx
```diff
@@ -10,35 +10,18 @@ const { Button, Span, View } = AIConversationElements;
 
 const ACTIONS_BAR_BLOCK = 'ai-actions-bar';
 
-const iconAttributes = {
-  'aria-hidden': true,
+const ActionIcon = withBaseElementProps(Span, {
```
esauerbo, line 13: The actual icon gets passed in by the user

packages/react-ai/src/components/AIConversation/views/Controls/AvatarControl.tsx
```diff
@@ -4,40 +4,42 @@ import { withBaseElementProps } from '@aws-amplify/ui-react-core/elements';
...
-  width: '100%',
-};
+const DEFAULT_USER_ICON = withBaseElementProps(Icon, {
+  variant: 'user-avatar',
+});
 
-const avatarIconProps = () => ({
-  children: (
-    <svg viewBox="0 0 24 24">
+const DEFAULT_AI_ICON = () => (
```
esauerbo, line 15: The Raven icon is custom so not including it as a default bc it would require overriding almost every attribute but curious if anyone has opinions on that

thaddmt: By custom do you mean it doesnt share most of it's attributes with the other icons? Why is that an issue?

esauerbo: Yeah I just meant it's not an icon you could get from material icons. I was thinking using the IconElement and overriding all the default props would probably be more verbose than just hard coding the svg but curious if we do want to use that util here instead.

thaddmt (approved):

timngyn (approved):

Issue #5467, opened 2024-07-25
processFile in StorageManager causing endless uploads


OperationalFallacy:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Storage (Storage Manager)

### How is your app built?

Next.js

### What browsers are you seeing the problem on?

Chrome, Safari

### Which region are you seeing the problem in?

us-east-1

### Please describe your bug.

I'm developing the Amplify Gen2 project, and I used StorageManager. It was working fine.

Now, if there is `processFile={processFile}` - it goes into an endless upload loop. Without `processFile` it works fine.

![image](https://github.com/user-attachments/assets/f95bacf1-c331-47ef-8931-6de11412a6cb)
![image](https://github.com/user-attachments/assets/924af501-1ac7-43ad-b5d6-dc6fc28705fb)

This has to do with a weird local re-rendering, not the component itself, but I don't know how to approach it. Any advice is appreciated!

### What's the expected behaviour?

It should upload a file with the processFile function.

### Help us reproduce the bug!

I added some code samples, but it's probably a specific issue with the local environment.

### Code Snippet

_app.tsx
```typescript
"use client";
import type { AppProps } from "next/app";
import "@aws-amplify/ui-react/styles.css";
import "@fontsource-variable/inter";
import { Authenticator, useAuthenticator } from "@aws-amplify/ui-react";
import { Toaster } from "react-hot-toast";
import AuthForm from "../components/Auth";
import Layout, { Loader } from "../components/Layout";
import { useAuth } from "../hooks/useAuth";

const MyCustomApp = ({ Component, pageProps }: AppProps) => {
  const { handleLoginClick, handleCancelAuth, handleLogoutClick } = useAuth();

  const { authStatus } = useAuthenticator((context) => [context.authStatus]);
  console.log("MyCustomApp loaded...");
  return (
    <Layout
      handleLoginClick={handleLoginClick}
[...]

PR #5477, opened 2024-07-25
fix(storage): fix looping calls to processFile

thaddmt:
#### Description of changes
- Using a combination of `processFile` and `path` would cause the processFile to be called on loop
- Seems to be caused by the `setProcessKey` being called before the getting the `identityId` for use with the path callback
- Moved the call to setPRocessedKey` after the call to `fetchAuthSession` and that seems to resolve the issue
- Updated e2e test to use `path` with `processFile`
- fixes https://github.com/aws-amplify/amplify-ui/issues/5467

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageManager/utils/__tests__/getInput.spec.ts
+++ b/packages/react-storage/src/components/StorageManager/utils/__tests__/getInput.spec.ts
@@ -3,7 +3,9 @@ import { UploadDataWithPathInput, UploadDataInput } from 'aws-amplify/storage';
 import { getInput, GetInputParams } from '../getInput';
 
 const identityId = 'identity-id';
-jest.spyOn(AuthModule, 'fetchAuthSession').mockResolvedValue({ identityId });
+const fetchAuthSpy = jest
+  .spyOn(AuthModule, 'fetchAuthSession')
+  .mockResolvedValue({ identityId });
 
 const file = new File(['hello'], 'hello.png', { type: 'image/png' });
 const key = file.name;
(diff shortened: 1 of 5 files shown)


packages/react-storage/src/components/StorageManager/utils/getInput.ts
```diff
@@ -63,6 +58,11 @@ export const getInput = ({
       hasCallbackPath ? path({ identityId }) : path
     }${processedKey}`;
 
+    if (processFile) {
```
calebpollman, line 61: Do we also need to add a call to `onProcessFileSuccess` inside the conditional block on [line 47](https://github.com/aws-amplify/amplify-ui/pull/5477/files#diff-04342a0ffee2005ce1b867d03610c88d0b63ec3b73a9c475fef0f66ef92d2f4aR47) for `key` inputs?

thaddmt: Yep you're totally right

reesscot (approved):

packages/react-storage/src/components/StorageManager/utils/__tests__/getInput.spec.ts
```diff
@@ -186,6 +189,33 @@ describe('getInput', () => {
...
+    });
+
+    await input();
+
+    const fetchAuthSessionCallOrder = fetchAuthSpy.mock.invocationCallOrder[0];
+    const onProcessFileSuccessCallORder =
+      onProcessFileSuccess.mock.invocationCallOrder[0];
+    expect(fetchAuthSessionCallOrder).toBeLessThan(
+      onProcessFileSuccessCallORder
+    );
```
calebpollman, line 211: 🙏

calebpollman (approved):
🚢

PR #5484, opened 2024-07-26
feat(storage-browser): add base actions, integrate ActionProvider, add ConfigContext

calebpollman:
#### Description of changes
- add base actions: `createListLocations`, `listLocationItems`
- integrate `ActionProvider`/`useAction` with `listLocationItems`
- add `ConfigContext`/`useConfig`
- add `listLocations` initial load to `LocationsView`
- add `listLocationItems` initial load to `LocationDetailView`
- update unit tests, add temp TS workarounds
- add initial Amplify "listLocations" handler
- use JS "storage-manager" tag

#### Checklist

- [ ] PR description included


timngyn (approved):

examples/next/package.json
```diff
@@ -10,7 +10,6 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@aws-amplify/geo": "3.0.31",
```
reesscot, line 13: nice!

packages/react-storage/src/components/StorageBrowser/context/config.ts
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import { ListLocations } from './actions/listLocationsAction';
+import { Permission } from './actions/types';
+
+export interface Config<T = Permission> {
+  listLocations: ListLocations<T>;
+}
+
+export const USE_CONFIG_ERROR_MESSAGE = 'Add me later';
```
reesscot, line 10: Maybe put a //@TODO here to find it easier later?

packages/react-storage/src/components/StorageBrowser/context/actions/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  key: string;
+  lastModified: Date;
+  size: number;
+  type: 'FILE';
+}
+
+export type LocationItem = FileItem | FolderItem;
+
+export type Permission = 'READ' | 'READWRITE' | 'WRITE';
+export type LocationType = 'OBJECT' | 'PREFIX' | 'BUCKET';
```
reesscot, line 16: Will this always be the end part of the full location? Just curious for a prefix type, if you'll want to store the bucket name separately from the prefix.

packages/react-storage/src/components/StorageBrowser/context/actions/types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface LocationData<T = Permission> {
+  scope: string;
```
reesscot, line 19: Will this be the full location path?

packages/react-storage/src/components/StorageBrowser/context/actions/listLocationItemsAction.ts
```diff
@@ -0,0 +1,65 @@
...
+  NonNullable<
+    (ListPaginateWithPathInput | ListAllWithPathInput)['options']
+  >['subpathStrategy']
+>;
+
+interface ListLocationItemsActionInput {
+  prefix: string;
+  options?: { delimiter?: string; pageSize?: number } & (
+    | { nextToken?: string; refresh?: never }
+    | { nextToken?: never; refresh?: boolean }
```
reesscot, line 21: If I'm reading this right, this is to make sure we don't poss both  refresh and nextToken at the same times since a refresh should never use a nextToken. Am I understanding that right?

packages/react-storage/src/components/StorageBrowser/context/actions/listLocationItemsAction.ts
```diff
@@ -0,0 +1,65 @@
...
+        { key, lastModified, size: size!, type: 'FILE' }
+      : { key, type: 'FOLDER' }
+  );
+
+export default async function listLocationItemsAction(
+  prevState: ListLocationItemsActionOutput,
+  input: ListLocationItemsActionInput
+): Promise<ListLocationItemsActionOutput> {
+  const { options = {}, prefix: path } = input ?? {};
+  const { delimiter, nextToken: _nextToken, pageSize, refresh } = options;
```
reesscot, line 44: I may be wrong about this, but doesn't destructuring throw if one of the variables is undefined? I'm just thinking if nextToken is undefined.

packages/react-storage/src/components/StorageBrowser/context/actions/__tests__/listLocationsAction.spec.ts
```diff
@@ -1,33 +1,75 @@
...
-      await listLocationsAction({
-        locations,
-        nextToken,
-      });
+      await listLocationsAction(
+        {
+          locations,
+          nextToken,
+        },
+        { pageSize: 100 }
```
reesscot, line 71: Doesn't need to be in this PR, but probably move this to a const?

reesscot (approved):
LGTM. Left some nonblocking comments/questions

packages/react-storage/src/components/StorageBrowser/context/config.ts
calebpollman: Normally yes but need to continue working in this file and will just handle it when working on the store

packages/react-storage/src/components/StorageBrowser/context/actions/types.ts
calebpollman: Good question, will think abt it

packages/react-storage/src/components/StorageBrowser/context/actions/types.ts
calebpollman: Yes, not sure if I'm really aligned with `scope` as the name since it feels kinda access grant-y but it does seem to align well with the underlying service as a whole

packages/react-storage/src/components/StorageBrowser/context/actions/listLocationItemsAction.ts
calebpollman: Nope, they'll just be `undefined`

packages/react-storage/src/components/StorageBrowser/context/actions/__tests__/listLocationsAction.spec.ts
calebpollman: For the unit tests not worried since there is no specific behavior for `pageSize` at the action level but will be adding a configuration constant for page size

PR #5494, opened 2024-07-30
chore(storage-browser): Add Navigate control to LocationDetailView

timngyn:
#### Description of changes
- Updated `NavigateState` 
   - to have `history`, which is a string array of the folder names we are navigating through
- Updated `Navigate` control to use the navigate context to create the `NavigateItems` 
- Moved `listLocationItems` loading logic out of `Controller.tsx` to `LocationDetailView`

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/Controller.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Controller.tsx
@@ -33,7 +33,6 @@ export function Controller(): null {
     type: 'LIST_LOCATION_ITEMS',
   });
   const [{ location }] = useControl({ type: 'NAVIGATE' });
-  const { current } = location;
 
   const loadLocations = shouldLoadLocations(locationsState);
 
@@ -63,11 +62,11 @@ export function Controller(): null {
 
   const loadLocationItems = shouldLoadLocationItems(locationItemsState);
   React.useEffect(() => {
-    if (current && loadLocationItems) {
+    if (location && loadLocationItems) {
       // TODO: update to exhaustive call
-      listLocationItems(current);
+      listLocationItems(location);
     }
-  }, [current, listLocationItems, loadLocationItems]);
+  }, [location, listLocationItems, loadLocationItems]);
 
   return null;
 }
--- a/packages/react-storage/src/components/StorageBrowser/Controller.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Controller.tsx
@@ -18,10 +18,10 @@ const shouldLoadLocations = ({
 }: LocationsDataState[0]) => !data.locations.length && !isLoading && !hasError;
 
 const shouldLoadLocationItems = ({
-  data,
+  // data,
   hasError,
   isLoading,
-}: LocationItemsState[0]) => !data.items.length && !isLoading && !hasError;
+}: LocationItemsState[0]) => !isLoading && !hasError;
 
 /**
  * Handles fetching of list data
@@ -40,40 +40,65 @@ export function Controller(): null {
     if (loadLocations) {
       // TODO: update to exhaustive call
       handleListLocations();
-    } else if (location.shouldRefresh) {
-      // TODO: update to exhaustive call
-      handleListLocations({ refresh: true });
     }
-  }, [location.shouldRefresh, handleListLocations, loadLocations]);
+    // else if (location.shouldRefresh) {
+    //   // TODO: update to exhaustive call
+    //   handleListLocations({ refresh: true });
+    // }
+  }, [handleListLocations, loadLocations]);
+
+  const { bucket, permission, scope, prefix } = location ?? {};
 
   const listLocationItems = React.useCallback(
-    ({ bucket, permission, scope }: LocationData, refresh: boolean = false) =>
-      handleListLocationItems({
+    ({ prefix = '', scope }: { prefix: string | undefined; scope: string }) => {
+      // s3://SOME_STRING/*
+      // eslint-disable-next-line no-console
+      console.log('listLocationItems scope', scope);
+
+      return handleListLocationItems({
         // uncomment to test managed auth config
-        prefix: '',
+        // prefix: '',
         // uncomment to test with amplify config with public access level
-        // prefix: 'public/'
+
+        prefix,
         config: {
-          bucket,
+          bucket: bucket!,
           credentialsProvider: async () =>
-            await getLocationCredentials({ permission, scope }),
+            await getLocationCredentials({ permission: permission!, scope }),
           region,
         },
-        options: { refresh },
-      }),
-    [getLocationCredentials, handleListLocationItems, region]
+        options: { refresh: true },
+      });
+    },
+    [
+      getLocationCredentials,
+      bucket,
+      permission,
+      handleListLocationItems,
+      region,
+    ]
   );
 
-  const loadLocationItems = shouldLoadLocationItems(locationItemsState);
   React.useEffect(() => {
-    if (location.current && loadLocationItems) {
+    // eslint-disable-next-line no-console
+    console.log('location changed', location);
+  }, [location]);
+
+  const loadLocationItems = shouldLoadLocationItems(locationItemsState);
+
+  React.useEffect(() => {
+    if (scope && loadLocationItems) {
       // TODO: update to exhaustive call
-      listLocationItems(location.current);
-    } else if (location.current && history.shouldRefresh) {
-      // TODO: update to exhaustive call
-      listLocationItems(location.current, true);
+      listLocationItems({ prefix, scope });
     }
-  }, [location, history.shouldRefresh, listLocationItems, loadLocationItems]);
+    // if (location.current && loadLocationItems) {
+    //   // TODO: update to exhaustive call
+    //   listLocationItems(location.current);
+    // } else if (location.current && history.shouldRefresh) {
+    //   // TODO: update to exhaustive call
+    //   listLocationItems(location.current, true);
+    // }
+  }, [prefix, scope, listLocationItems, loadLocationItems]);
 
   return null;
 }
--- a/packages/react-storage/src/components/StorageBrowser/Controller.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Controller.tsx
@@ -41,6 +41,7 @@ export function Controller(): null {
       // TODO: update to exhaustive call
       handleListLocations();
     } else if (location.shouldRefresh) {
+      // TODO: update to exhaustive call
       handleListLocations({ refresh: true });
     }
   }, [location.shouldRefresh, handleListLocations, loadLocations]);
@@ -69,6 +70,7 @@ export function Controller(): null {
       // TODO: update to exhaustive call
       listLocationItems(location.current);
     } else if (location.current && history.shouldRefresh) {
+      // TODO: update to exhaustive call
       listLocationItems(location.current, true);
     }
   }, [location, history.shouldRefresh, listLocationItems, loadLocationItems]);
--- a/packages/react-storage/src/components/StorageBrowser/Controller.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Controller.tsx
@@ -1,15 +1,6 @@
 import React from 'react';
 
-import {
-  LocationData,
-  LocationItemsState,
-  LocationsDataState,
-  useAction,
-  useLocationsData,
-} from './context/actions';
-
-import { useControl } from './context/controls';
-import { useConfig } from './context/config';
+import { LocationsDataState, useLocationsData } from './context/actions';
 
 const shouldLoadLocations = ({
   data,
@@ -17,22 +8,11 @@ const shouldLoadLocations = ({
   isLoading,
 }: LocationsDataState[0]) => !data.locations.length && !isLoading && !hasError;
 
-const shouldLoadLocationItems = ({
-  // data,
-  hasError,
-  isLoading,
-}: LocationItemsState[0]) => !isLoading && !hasError;
-
 /**
  * Handles fetching of list data
  */
 export function Controller(): null {
-  const { getLocationCredentials, region } = useConfig();
   const [locationsState, handleListLocations] = useLocationsData();
-  const [locationItemsState, handleListLocationItems] = useAction({
-    type: 'LIST_LOCATION_ITEMS',
-  });
-  const [{ location, history }] = useControl({ type: 'NAVIGATE' });
 
   const loadLocations = shouldLoadLocations(locationsState);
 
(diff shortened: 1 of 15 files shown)


packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
```diff
@@ -93,23 +88,53 @@ export interface NavigateControl<
...
+              }
+            }}
+          >
+            {state.location.current.bucket}
+          </NavigateButton>
+          {state.history.list && state.history.list.length > 0 ? (
+            <Separator />
+          ) : null}
+        </NavigateItemContainer>
+      )}
```
timngyn, line 121: Defining the "Home" button and the "Bucket" manually here since there is only one instance of each. I did this so that `RenderItemProps` could have the `item` property only be a string and not a mix of types

packages/react-storage/src/components/StorageBrowser/context/controls/Navigate/Navigate.tsx
```diff
@@ -40,13 +41,53 @@ export function navigateReducer(
...
+      const updatedHistory = state.history.list?.slice(0, index + 1);
+
+      return {
+        ...state,
+        history: {
+          list: updatedHistory,
+          shouldRefresh: true,
+        },
+      };
+    }
```
calebpollman, line 89: Adding values to our control state increases likelihood of long term complexity and tech debt due to  exponential increases in possible event variations and usage by consumers, and in this case impacts the `StorageBrowser` interface which needs to be as minimal as possible.

A value conveying intent (e.g. `shouldRefresh`) has to be "cleaned up" (e.g. set to `false`) after any side effects are run to avoid non-deterministic behavior in state consumers.

On selection of a `location`, we rely on React to be React - updating `state.location.current` with a new  reference signals the `useEffect` in `Controller` responsible for fetching `location` data to run with the new values inside `location.state.current`. If `state` updates from a different action, but `state.location` stays untouched, the `useEffect` will ignore the `state` update - no need for a `shouldRefresh` flag.

Think there are probably some simplifications we can do here 😅

packages/react-storage/src/components/StorageBrowser/Views/LocationDetailView/LocationDetailView.tsx
```diff
@@ -18,7 +21,28 @@ export const LocationDetailView: LocationDetailView = () => {
 
   const listItems = !hasItems
     ? null
-    : data.items.map(({ key }) => <p key={key}>{key}</p>);
+    : data.items.map((item) => {
+        if (item.type === 'FOLDER') {
+          return (
+            <button
+              onClick={() => {
+                if (item.key.endsWith('/')) {
```
calebpollman, line 29: What happens on click events here if the key doesn't end in `/`?

packages/react-storage/src/components/StorageBrowser/Views/LocationDetailView/LocationDetailView.tsx
```diff
@@ -18,7 +21,28 @@ export const LocationDetailView: LocationDetailView = () => {
 
   const listItems = !hasItems
     ? null
-    : data.items.map(({ key }) => <p key={key}>{key}</p>);
+    : data.items.map((item) => {
```
calebpollman, line 24: Destructuring simplifies reading `code` and writing boilerplate and helps React with handling reference equality:
```
data.items.map(({ key }) => {
```

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
```diff
@@ -93,23 +88,53 @@ export interface NavigateControl<
...
+      {state.history &&
+        state.history.list?.map((folder, index) => (
```
calebpollman, line 123: `state.history?.list?.map` 🤔

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
```diff
@@ -93,23 +88,53 @@ export interface NavigateControl<
...
+            item={folder}
+            current={
+              folder === state.history.list?.[state.history.list.length - 1]
+            }
+            onClick={() => {
+              handleUpdateState({
+                type: 'EXIT_FOLDER',
+                index,
+              });
+            }}
```
calebpollman, line 135: Let's sync on behavior here, if we need to pass something here it should probably be the `folder`

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
calebpollman: Need to have deterministic behavior here for flexibility

calebpollman (changes requested):
Nice job getting this out! Left some blocking feedback on the store additions ( e.g `shouldRefresh`), but think we should be to able to work through them quickly. Will review unit tests once we get refactoring complete

packages/react-storage/src/components/StorageBrowser/Controller.tsx
```diff
@@ -32,20 +32,22 @@ export function Controller(): null {
...
-  const { current } = location;
+  const [{ location, history }] = useControl({ type: 'NAVIGATE' });
 
   const loadLocations = shouldLoadLocations(locationsState);
 
   React.useEffect(() => {
     if (loadLocations) {
       // TODO: update to exhaustive call
       handleListLocations();
+    } else if (location.shouldRefresh) {
```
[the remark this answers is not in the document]
timngyn: I think I needed a way to call `handleListLocations()` again when we navigate back to "home". Even though I was updating the `location`, `loadLocations` was still false and I wasn't sure how to update it. I was thinking that I shouldn't update `locationsState` and so I added `shouldRefresh`

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
timngyn: Sounds good. I might be misunderstanding how the `Navigate` control should be working. I thought it was supposed to work like breadcrumbs and so I think it needs a way to navigate to a specific folder that isn't on the top of the stack.

For instance: Home / bucket / "folder1" / "folder2" / "folder3", if I wanted to go back to "folder1", should that be allowed or are we only going back up one folder at a time?

packages/react-storage/src/components/StorageBrowser/Controller.tsx
```diff
@@ -57,17 +59,21 @@ export function Controller(): null {
...
+    if (location.current && loadLocationItems) {
+      // TODO: update to exhaustive call
+      listLocationItems(location.current);
+    } else if (location.current && history.shouldRefresh) {
       // TODO: update to exhaustive call
-      listLocationItems(current);
+      listLocationItems(location.current, true);
     }
-  }, [current, listLocationItems, loadLocationItems]);
+  }, [location, history.shouldRefresh, listLocationItems, loadLocationItems]);
```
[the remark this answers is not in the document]
timngyn: I wanted to pass `refresh` to the `options` object from  the `listLocationsItems` call to the `handleListLocationItems` call you set up. Don't know if it's necessary though. I  wasn't sure how to let `handleListLocationItems` know that we needed to refresh the list items list when we navigate back to a previous `NavigateItem` that needed to reload location items, hopefully that makes sense

I think I just need to sync with you on how to call `listLocationsItems` again because when we navigate up to a parent folder from the `Navigate` control, location doesn't change technically and so I don't think this `useEffect` would run. Do you have any thoughts on this?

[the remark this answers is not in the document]
calebpollman: Sounds good, we can sort out any required changes to `loadLocationItems`/`listLocationItems`

packages/react-storage/src/components/StorageBrowser/Controller.tsx
timngyn, line 1: I think LocationsView is not refreshing when navigating back to home, currently looking into it.

timngyn: I was thinking of doing this:
```
export function Controller(): null {
  const [_, handleListLocations] = useLocationsData();
  const [{ location }] = useControl({ type: 'NAVIGATE' });

  const { bucket } = location ?? {};

  React.useEffect(() => {
    if (bucket === undefined) {
      // TODO: update to exhaustive call
      handleListLocations({ refresh: true });
    }
  }, [handleListLocations, bucket]);

  return null;
}
```

I think bucket would only be undefined on the first load and when we go back to "home", but we can come back to this later

packages/react-storage/src/components/StorageBrowser/context/actions/listLocationItemsAction.ts
```diff
@@ -40,13 +40,13 @@ export interface ListLocationItemsActionOutput {
...
 const parseResultItems = (items: ListOutputItem[]): LocationItem[] =>
-  items.map(({ path: key, lastModified, size }) =>
-    lastModified
-      ? // `size` is marked as potentially `undefined` in `ListOutputItem`
-        // but is populated when the item is a file
-        { key, lastModified, size: size!, type: 'FILE' }
-      : { key, type: 'FOLDER' }
-  );
+  items.map(({ path: key, lastModified, size }) => {
+    if (size === 0 && key.endsWith('/')) {
```
timngyn, line 44: Added this just for now. In one of the items that gets returned, there is a `txt` file that is 0 bytes. So far it looks like the items that are folders do end with `/`, but is there something else we need to do to ensure this?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
```diff
@@ -93,23 +91,46 @@ export interface NavigateControl<
...
+        current={!state.location}
+        onClick={() => handleUpdateState({ type: 'DESELECT_LOCATION' })}
       />
+      {state.location && (
+        <>
+          <NavigateItem
+            item={state.location.bucket}
+            current={!state.history || state.history.length === 0}
+            onClick={() => {
+              if (state?.location) {
```
calebpollman, line 109: Just so i understand correctly, is the optional chaining here just due to types or is it possible that a `NavigateItem` can be rendered that won't do anything on click?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
```diff
@@ -93,23 +91,46 @@ export interface NavigateControl<
...
+              }
+            }}
+          />
+          {state.history?.map((folder, index) => (
+            <NavigateItem
+              key={index}
+              item={folder}
+              current={folder === state.history?.[state.history.length - 1]}
+              onClick={() => {
+                if (isFolderName(folder)) {
```
calebpollman, line 123: Same question as https://github.com/aws-amplify/amplify-ui/pull/5494/files#r1700996963 for this conditional

packages/react-storage/src/components/StorageBrowser/Views/LocationDetailView/LocationDetailView.tsx
```diff
@@ -4,21 +4,90 @@ import { useAction } from '../../context/actions';
...
+    }: {
+      bucket: string;
+      prefix: string | undefined;
+      scope: string;
+    }) =>
+      handleListLocationItems({
+        // uncomment to test managed auth config
+        // prefix: '',
+        // uncomment to test with amplify config with public access level
+
```
calebpollman, line 46:
```suggestion
```

packages/react-storage/src/components/StorageBrowser/context/actions/listLocationItemsAction.ts
calebpollman: Think we need something a bit more resilient but this is fine for now IMO

packages/react-storage/src/components/StorageBrowser/context/controls/Navigate/Navigate.tsx
```diff
@@ -40,13 +27,73 @@ export function navigateReducer(
...
-    };
+  switch (action.type) {
+    case 'SELECT_LOCATION': {
+      /** 
+        This action comes from the navigate item specifying the location
+        It uses the current state location and so this has the "old" scope
+        which could include the previous folder we were at
+        We'll need to update the scope here
+        We need to update the prefix too?
+      */
```
calebpollman, line 38: Nit: prefer to only use either JSDoc syntax for comments that provide information to consumers or `//` comments to explain behaviors

packages/react-storage/src/components/StorageBrowser/context/controls/Navigate/Navigate.tsx
```diff
@@ -40,13 +27,73 @@ export function navigateReducer(
...
+    }
+    case 'ENTER_FOLDER': {
+      const { name } = action;
+      const { location, history } = state;
+
+      if (name === history?.[history.length - 1]) {
+        // Don't add the same folder into history again
+        // For some reason listLocationItems also returns the current folder again
+        return state;
+      }
```
calebpollman, line 64: Let's make sure we're tracking this as a TODO somewhere

packages/react-storage/src/components/StorageBrowser/createStorageBrowser.tsx
```diff
@@ -29,9 +29,9 @@ interface ResolvedStorageBrowserElements<
  */
 function DefaultStorageBrowser(): React.JSX.Element {
   const [{ location }] = useControl({ type: 'NAVIGATE' });
-  const { current } = location;
 
-  if (current) {
+  if (location) {
+    // if (current) {
```
calebpollman, line 34:
```suggestion
```

calebpollman (commented):
LGTM, we can address the tech debt issues in follow ups

calebpollman (approved):

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
timngyn: ah sorry, I just removed the optional chaining locally and it looks like it's not complaining. I do have to add the null check for `state.location` though because typescript says it could be undefined

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
timngyn: When trying to pass `folder` to `handleUpdateState`, typescript complains that `Type 'string' is not assignable to type '`${string}/`'.ts(2322)` since `name` is of type `FolderName`

Added this type guard so that it doesn't complain

packages/react-storage/src/components/StorageBrowser/context/controls/Navigate/Navigate.tsx
timngyn, line 38:
```suggestion
        // This action comes from the navigate item specifying the location
        // It uses the current state location and so this has the "old" scope
        // which could include the previous folder we were at
        // We'll need to update the scope here
        // We need to update the prefix too?
```

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
timngyn, line 109:
```suggestion
              if (state.location) {
```

calebpollman (approved):

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
calebpollman: Okay cool, it might make sense to remove that type but we can worry abt that later 😅

calebpollman (approved):

hbuchel (approved):

PR #5496, opened 2024-07-30
Feat(IdentityProviders): useIsSignedIn with data state

AhmetAhunbayAWS:
#### Description of changes

Updated the useIsSignedIn hook to use the data state hook. This allows for an isLoading field in the state, as well as an error message.
This hook uses the useDataState hook in a unique way, since the action function (getCurrentUser) is not returned to the customer. On top of this, signedIn state needs to use hub events to update state. Therefore, I decided to call getCurrentUser after each hub event update. Also, since the false state throws an error, I needed to add a condition to the action that allowed for setting the data state vs setting the error state, and made 2 calls to the action to update each.

I created a test file to go over possible cases and ran yarn test.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-core/src/hooks/__tests__/useIsSignedIn.test.ts
+++ b/packages/react-core/src/hooks/__tests__/useIsSignedIn.test.ts
@@ -7,8 +7,8 @@ import { Hub } from 'aws-amplify/utils';
 
 const getCurrentUserSpy = jest.spyOn(AuthModule, 'getCurrentUser');
 
-const mockError = new Error('Authorization error');
-const mockSuccess: GetCurrentUserOutput = {
+const errorResult = new Error('Authorization error');
+const successResult: GetCurrentUserOutput = {
   username: 'username',
   userId: '123',
 };
@@ -42,7 +42,7 @@ describe('useIsSignedIn', () => {
     getCurrentUserSpy.mockImplementationOnce(async () => {
       return new Promise((_, reject) => {
         setTimeout(() => {
-          reject(mockError);
+          reject(errorResult);
         }, 10);
       });
     });
@@ -60,7 +60,7 @@ describe('useIsSignedIn', () => {
     getCurrentUserSpy.mockImplementationOnce(() => {
       return new Promise((resolve) => {
         setTimeout(() => {
-          resolve(mockSuccess);
+          resolve(successResult);
         }, 10);
       });
     });
@@ -75,7 +75,7 @@ describe('useIsSignedIn', () => {
   });
 
   it('should be true if receiving a signedIn event', async () => {
-    getCurrentUserSpy.mockRejectedValueOnce(mockError);
+    getCurrentUserSpy.mockRejectedValueOnce(errorResult);
 
     const { result } = renderHook(() => useIsSignedIn());
     expect(result.current).toEqual(expectedLoadingState);
@@ -92,7 +92,7 @@ describe('useIsSignedIn', () => {
   });
 
   it('should be false if receiving a signedOut event', async () => {
-    getCurrentUserSpy.mockResolvedValueOnce(mockSuccess);
+    getCurrentUserSpy.mockResolvedValueOnce(successResult);
 
     const { result } = renderHook(() => useIsSignedIn());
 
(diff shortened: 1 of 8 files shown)


packages/react-core/src/hooks/__tests__/useIsSignedIn.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import { act, renderHook } from '@testing-library/react-hooks';
+import { waitFor } from '@testing-library/react';
+import * as AuthModule from 'aws-amplify/auth';
+import { GetCurrentUserOutput } from 'aws-amplify/auth';
+import { Hub } from 'aws-amplify/utils';
+
+const getCurrentUserSpy = jest.spyOn(AuthModule, 'getCurrentUser');
+
+const mockError = new Error('Authorization error');
+const mockSuccess: GetCurrentUserOutput = {
```
axelEandrews, line 11: One thing @calebpollman had mentioned on my PRs is that these aren't really 'mocks', the mocks are the functions being mocked. I changed these names to `errorResult` and `successResult` or something like that.

axelEandrews (approved):
Left one comment about variable names in your unit tests, but aside from that Looks Good To Me, LGTM.

packages/react-core/src/hooks/useIsSignedIn.ts
```diff
@@ -0,0 +1,46 @@
+import { getCurrentUser } from 'aws-amplify/auth';
+import { Hub, HubCallback } from '@aws-amplify/core';
+import { useEffect, useState } from 'react';
+import useDataState, { DataState } from './useDataState';
+
+const action = async (_: { isSignedIn: boolean }, __: undefined) => {
+  await getCurrentUser();
+  return { isSignedIn: true };
```
thaddmt, line 8: Do we need to catch here and return the `defaultState`? I know the `defaultState` is set initially but I'm more concerned about an uncaught error here when user is not signed in

AhmetAhunbayAWS: The idea is that the uncaught error will be reflected in the dataState:

const expectedErrorState = {
    data: { isSignedIn: false },
    hasError: true,
    isLoading: false,
    message: 'Authorization error',
};

thaddmt: I might be wrong but doesn't `getCurrentUser()` always throw if the user is not signed in? And in that case it's not really an error scenario for the customer

packages/react-core/src/hooks/useIsSignedIn.ts
```diff
@@ -2,10 +2,23 @@ import { getCurrentUser } from 'aws-amplify/auth';
...
 const action = async (_: { isSignedIn: boolean }, __: undefined) => {
-  await getCurrentUser();
-  return { isSignedIn: true };
+  try {
+    await getCurrentUser();
+    return { isSignedIn: true };
+  } catch (error) {
+    if (
+      error instanceof AuthError &&
+      error.name === USER_UNAUTHENTICATED_EXCEPTION
```
thaddmt, line 16: Perfect!

thaddmt (dismissed):

axelEandrews (dismissed):

packages/react-core/src/hooks/useIsSignedIn.ts
```diff
@@ -0,0 +1,59 @@
...
+    ) {
+      return { isSignedIn: false };
+    }
+    throw error;
+  }
+};
+
+const defaultState: DataState<{ isSignedIn: boolean }> = {
```
calebpollman, line 24: Constants should be SREAMING_SNAKE_CASE

thaddmt (approved):

axelEandrews (dismissed):

timngyn (approved):

dbanksdesign (approved):

packages/react-core/src/hooks/useIsSignedIn.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  useEffect(() => {
+    handler();
+
+    const listener: HubCallback = ({ payload }) => {
+      if (payload.event === 'signedIn') {
+        handler();
+      } else if (payload.event === 'signedOut') {
+        handler({ setState: true });
+        handler();
```
calebpollman, line 32: What is requiring two calls to `handler` here?

AhmetAhunbayAWS: The first one sets the data and the other sets the error. With useDataState you cannot change both in the same call since the data is set through the return value and message/hasError is set through a thrown error. Is there a better way to do this that you can think of?

PR #5502, opened 2024-07-31
chore(Liveness-V2): patch fast-xml-parser 

esauerbo:
#### Description of changes
- Upgrades `@aws-sdk/client-rekognitionstreaming` dependency in liveness V2 package to latest (3.621.0) which brings in `fast-xml-parser@4.4.1`, which contains the patch 
- Refactored CustomWebSocketFetchHandler to work with new sdk 
Screenshot 2024-08-01 at 5 19 41 PM
Screenshot 2024-08-01 at 5 20 36 PM

#### Issue #, if available
https://github.com/aws-amplify/amplify-ui/security/dependabot/149

Manually smoke tested liveness app. Tested happy path and several error scenarios including the custom websocket timeout error. 

#### Checklist

- [ ] PR description included


thaddmt (approved):
Does v2 send out the custom websocket closure codes?

esauerbo:
> Does v2 send out the custom websocket closure codes?

No, I just updated the custom handler to have our message. So it will still show "Server Error" but the actual error message will say "Websocket connection timeout". Do you think we should also update the codes?

reesscot (approved):
LGTM

reesscot:
> > Does v2 send out the custom websocket closure codes?
> 
> No, I just updated the custom handler to have our message. So it will still show "Server Error" but the actual error message will say "Websocket connection timeout" instead of the 500 error. Do you think we should also update the codes?

This PR matches the behavior of v3 though, right?

esauerbo:
> This PR matches the behavior of v3 though, right?

@reesscot not exactly. In v3 we added new display text and a `CONNECTION_TIMEOUT` error type in https://github.com/aws-amplify/amplify-ui/pull/5141
https://github.com/aws-amplify/amplify-ui/blob/e86f3ae853a123f60aaee0b6d74ce14f7ae6cf60/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts#L112
so the error modal will say "Connection Error: Connection Timeout" and it will print like "CONNECTION_TIMEOUT: Websocket connection timeout" but in v2 it's counted as a serverError. So it'll print "SERVER_ERROR: Websocket connection timeout" 

Do we want them to align completely?

PR #5520, opened 2024-08-03
chore(storage-browser): set up create folder action view 

timngyn:
#### Description of changes
- Add scaffolding for CREATE_FOLDER ui
- Add tests to help with coverage

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/__tests__/createStorageBrowser.spec.tsx
```diff
@@ -12,9 +13,30 @@ const config = {
...
 
 describe('createStorageBrowser', () => {
-  it('returns a StorageBrowser', () => {
+  it('returns a StorageBrowser', async () => {
     const { StorageBrowser } = createStorageBrowser({ config });
 
-    expect(render(<StorageBrowser />).container).toBeDefined();
+    await waitFor(() => {
+      expect(render(<StorageBrowser />).container).toBeDefined();
+    });
```
timngyn, line 21: Added this because there was a console.error being logged during tests about testing components that updated state in `act`. Saw this post: https://stackoverflow.com/a/68331605/6995752 and saw that `waitFor` was being used elsewhere so I added it here

packages/react-storage/src/components/StorageBrowser/Views/CreateFolderActionView/Controls.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <Target.Field.Input type="text" />
+        <Target.Field.Submit
+          onClick={() => {
+            // eslint-disable-next-line no-console
+            console.log('create folder');
+          }}
+        >
+          Create folder
+        </Target.Field.Submit>
+      </Target.Field.Container>
```
timngyn, line 28: Am I using this right? I couldn't add some extra HTML elements to the `label` and `input` that I wanted to for accessibility. I was trying to do this:

```tsx
<form
  onSubmit={(event) => {
    // @TODO: we will probably want to use context and handleUpdateState()
    // to send the folder name to so that it can be sent off to be created?

    event.preventDefault();

    // eslint-disable-next-line no-console
    console.log('create folder', folderName);
  }}
>
  <label htmlFor="folder-name-input">Enter folder name:</label>
  <input
    type="text"
    id="folder-name-input"
    value={folderName}
    onChange={(event) => setFolderName(event.target.value)}
    required
  />
  <button type="submit">Create folder</button>
</form>
```

For instance, for `htmlFor`, typescript was complaining:
```
Type '{ children: string; htmlFor: string; }' is not assignable to type 'IntrinsicAttributes & Omit<BaseElementProps<never, string, DetailedHTMLProps<LabelHTMLAttributes<HTMLLabelElement>, HTMLLabelElement>>, "ref"> & RefAttributes<...>'.
  Property 'htmlFor' does not exist on type 'IntrinsicAttributes & Omit<BaseElementProps<never, string, DetailedHTMLProps<LabelHTMLAttributes<HTMLLabelElement>, HTMLLabelElement>>, "ref"> & RefAttributes<...>'.
  ```

File: packages/react-storage/src/components/StorageBrowser/__tests__/createStorageBrowser.spec.tsx [hunk 3]
```diff
@@ -12,9 +13,30 @@ const config = {
...
+    const { StorageBrowser } = createStorageBrowser({ config });
+
+    // Used 'mockReturnValueOnce` three times to mock each time it's called
+    // to render LocationActionView
+    // Twice in `DefaultStorageBrowser` and then once in `LocationActionView`
+    jest
+      .spyOn(controlsModule, 'useControl')
+      .mockReturnValueOnce([{ location: {} }])
+      .mockReturnValueOnce([{ selected: { actionType: 'CREATE_FOLDER' } }])
+      .mockReturnValueOnce([{ selected: { actionType: 'CREATE_FOLDER' } }]);
```
timngyn, line 34: this seems really fragile, does anyone have thoughts on how to make this better? The coverage was off by .04% and I just wanted to get it through without having to lower the coverage for now

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/LocationActionView.tsx
```diff
@@ -34,7 +35,12 @@ export const LocationActionView: LocationActionView = () => {
       >
         Cancel
       </button>
-      <LocationActionViewControls />
+      {actionType === 'CREATE_FOLDER' ? (
+        <CreateFolderActionViewControls />
```
timngyn, line 39: At the moment, this isn't styled or anything. Was thinking of wrapping the controls grid around the ui in `LocationActionView`. I think this could be a follow up, but at the moment create folder looks like this:

image

packages/react-storage/src/components/StorageBrowser/__tests__/createStorageBrowser.spec.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 import { createStorageBrowser } from '../createStorageBrowser';
+import * as controlsModule from '../context/controls';
```
calebpollman, line 4: Prefer ordering the file to be asserted against  (.e.g. `../createStorageBrowser.tsx`) as the final import and  `PascalCase` for Module names

```suggestion

import * as ControlsModule from '../context/controls';
import { createStorageBrowser } from '../createStorageBrowser';
```

packages/react-storage/src/components/StorageBrowser/__tests__/createStorageBrowser.spec.tsx
calebpollman: There's no need to complicate the logic here by considering child component behavior since `location` does not apply in the case of `ActionView` rendering, can make the assertion against:
```suggestion
 jest
      .spyOn(controlsModule, 'useControl')
      .mockReturnValueOnce([{ selected: { actionType: 'CREATE_FOLDER' } }])
```

packages/react-storage/src/components/StorageBrowser/Views/CreateFolderActionView/Controls.tsx
```diff
@@ -0,0 +1,30 @@
...
+> extends Pick<Controls<T>, 'Target'> {
+  (): React.JSX.Element;
+}
+
+const { Target } = Controls;
+
+// @ts-expect-error TODO: add Controls assignment
+export const CreateFolderActionViewControls: CreateFolderActionViewControls =
+  () => {
+    return (
```
calebpollman, line 16: Can you please add a React `state` value here for `folderName`, then wire the state update dispatch in the `onClick` of the `Submit` button?
```tsx
const [folderName, setFolderName] = React.useState('');
// pass as `ref` to `Target.Field.Input`, non-prescriptive
const inputRef = React.useRef<HTMLInputElement>(null);



<Target.Field.Input ref={inputRef} />
<Target.Field.Submit
  onClick={() => {
    setFolderName(() => { 
      // no idea if this lookup is valid
      return { folderName:  inputRef.current.value }
    })
  }
/>
```

This will allow us to conditionally disable the `Primary` button (which has yet to be implemented, out of scope for this PR) that will call the handler from `useAction`

calebpollman (dismissed):
Looking good, just some minor feedback and a request that can be done in a follow up

calebpollman (approved):

PR #5540, opened 2024-08-06
chore(storage-browser): create ActionStart/ActionCancel controls, update styles

hbuchel:
#### Description of changes

Adds a `<Primary />` and `<Exit />` control. Exit is for backing out of the current action. Primary handles whatever act we need to take for the selected Action (start upload, create folder, etc)

**Upload files**
Screenshot 2024-08-08 at 12 20 16 PM

**Create folder**
Screenshot 2024-08-08 at 12 20 00 PM

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/Controls.tsx
```diff
@@ -3,42 +3,20 @@ import React from 'react';
...
-    </NavigateItem.Button>
-  );
-};
-
-export function Primary(props: {
-  children?: React.ReactNode;
-  onClick?: React.MouseEventHandler<HTMLButtonElement>;
-}): React.JSX.Element {
-  return <Button {...props} className={`${CLASS_BASE}__refresh`} />;
-}
```
calebpollman, line 40: `Primary` should handle `action` begin, cancellation and "done"/exit once the `action` has completed, can we update `ActionStart` to reflect those use cases?

`Navigate` is used for just exiting/closing the `LocationActionView`, think that is what you are addressing with `ActionCancel`? Don't feel strongly abt the `Navigate` name here (might be somewhat misleading vs `Navigate` in the `LocationDetailView`) but think we should avoid naming it `ActionCancel` since the naming implies a use case that is covered by `Primary`. Thoughts?

hbuchel: Yeah, I don't think we should re-use the Navigate name/control since it will likely have unique styling for it's use in the breadcrumbs-type control. So it's something that would need it's own variant or way to theme separately. But yes, that is what is meant to be addressed by ActionCancel (maybe ActionExit?). Would exiting the action view also cancel the action if the action is in progress?

For `Primary`, is the idea that it's a single Button that will change it's children/props/use case depending on the state?

hbuchel: Also wondering if instead of "Cancel", the exit button would just say "<- Back" (and handle EXIT, or in the case that the action has started it will EXIT and cancel). Then, the `Primary` control wouldn't need to handle exit because it's kind of weird/doesn't make sense having a button that says "Done" and does something. Probably when it's "done" it can convert to a disabled state.

hbuchel: After our discussion, renamed ActionCancel -> Exit and ActionStart -> Primary

timngyn (dismissed):

timngyn (approved):

esauerbo (approved):

thaddmt (approved):

PR #5541, opened 2024-08-06
feat(storage-browser): add create folder support

calebpollman:
#### Description of changes
Add create folder support, refactor navigation to have deterministic behavior location types

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
@@ -6,14 +6,8 @@ import { StorageBrowserElements } from '../../context/elements';
 import { DownloadControl } from './Download';
 import { CLASS_BASE } from '../constants';
 import { useControl } from '../../context/controls';
-import {
-  LocationAccess,
-  LocationData,
-  LocationItem,
-  Permission,
-} from '../../context/types';
+import { LocationAccess, LocationItem, Permission } from '../../context/types';
 import { useAction, useLocationsData } from '../../context/actions';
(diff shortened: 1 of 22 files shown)


timngyn (approved):

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -257,25 +247,22 @@ export const LocationDetailViewTable = (): JSX.Element => {
     type: 'LIST_LOCATION_ITEMS',
   });
 
-  const { prefix: initialPrefix } = location
-    ? parseLocationAccess(location)
-    : ({} as LocationData);
-
-  const prefix =
-    history.length === 1 ? initialPrefix : history[history.length - 1];
+  const prefix = history.join('');
```
reesscot, line 250: What does this do?

timngyn: I think with Caleb's latest changes, it updated how the entries in the navigation history is stored to just be specific folders, like `'folder1/` > `nestedFolder1/` for instance.

Before this change, we had entries in the history that looked like this: `public/folder1/` > `public/folder1/nestedFolder1/`

I think he needs to join it together in order to get the complete prefix again so that we can fetch the list items when we navigate back to the specific folder

calebpollman: @timngyn Nailed it

reesscot (approved):

PR #5551, opened 2024-08-07
chore(storage-browser): Update TargetControl to have Error message/validation

hbuchel:
#### Description of changes

Adds a field error message to the TargetControl, uses the Message component to show a success message.

The validation still feels wonky; I think we want to change this UI a lot more so that aspect might just be temporary. Currently, the "Create folder" button is not disabled. It will validate the current folder name onClick; if invalid the inline field error message will show. The field also validates onBlur (so if you keyboard navigate). The field resets the validation message onFocus.

**Screenshots**
Screenshot 2024-08-13 at 3 27 36 PM

Screenshot 2024-08-13 at 3 30 04 PM

Screenshot 2024-08-13 at 3 30 14 PM

Screenshot 2024-08-13 at 3 22 40 PM

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/CreateFolderControls.tsx
```diff
@@ -52,35 +79,30 @@ export const CreateFolderControls = (): React.JSX.Element => {
...
-            }
-            // clear error
-            setFieldValidationError(undefined);
-            setData(() => inputRef.current?.value ?? '');
-          }}
-        >
-          Set Folder Name
-        </Target.Field.Button>
+        {fieldValidationError ? (
+          <Target.Field.Error id="fieldError">
```
calebpollman, line 100: Is camelCase standard for `id` values?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/CreateFolderControls.tsx
```diff
@@ -6,7 +6,14 @@ import { useControl } from '../../context/controls';
 import { Controls } from '../Controls';
 import { Title } from './Controls';
 
-const { Exit, Primary, Target } = Controls;
+const { Exit, Message, Primary, Target } = Controls;
+
+export const isValidFolderName = (name: string | undefined): boolean => {
+  return name !== undefined && name.length > 1 && name.endsWith('/');
```
calebpollman, line 12: Think this can be simplified to:
```suggestion
  return !!name?.length && name.endsWith('/');
```

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/CreateFolderControls.tsx
hbuchel: I think our usage has been kind of all over the place. I'm not sure there is a standard given it's not needed to be human readable. 

In any case, it should proooooobably use useStableId utility (or just React.useId()) when we're passing id's meant for accessibility.

calebpollman: Yeah that makes sense, can we please add a backlog item for auditing/moving to stable/unique ids?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/CreateFolderControls.tsx
hbuchel: I added the length > 1 incase we don't want folks to be able to upload folders named '/' (not sure if that's valid? It definitely still worked when I tried it). But I can take that out if so and just check for it being defined and ending with '/'

calebpollman: An empty string evaluates to `false`:
CleanShot 2024-08-13 at 13 32 28@2x

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/CreateFolderControls.tsx
hbuchel: Added to our backlog

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/CreateFolderControls.tsx
hbuchel: Right, I think my question is more like if '/' is a valid name; it has length, and has a trailing slash, and is not undefined, but then the folder created would just be named '/' like in this case (where I added a folder with just '/' inside of the existing Cloudberry folder). 

Screenshot 2024-08-13 at 4 38 39 PM

Adding a check that the length is > 1 means you wouldn't be able to do that. Do we need to check for that?

calebpollman: Oh i see, yeah that makes sense. Think we can still simplify it to:
```
const isValidFolderName = (name: string | undefined): boolean => !!name?.length > 1 && name.endsWith('/');
```
But good either way, just less code paths to test (maybe)

calebpollman (approved):

timngyn (approved):

PR #5552, opened 2024-08-08
chore(storage-browser): use getUrl to download

timngyn:
#### Description of changes
- Use `getUrl` to create a signed url and download the file when clicked

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/Views/Controls/Download.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Views/Controls/Download.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
-import { useAction } from '../../context/actions';
+import { downloadAction } from '../../context/actions';
 import { StorageBrowserElements } from '../../context/elements';
 import { CLASS_BASE } from '../constants';
+import { useDataState } from '@aws-amplify/ui-react-core';
 
 const { Button: ButtonElement, Icon: IconElement } = StorageBrowserElements;
 
@@ -11,7 +12,7 @@ const BLOCK_NAME = `${CLASS_BASE}__download`;
 export interface DownloadControl<
   T extends StorageBrowserElements = StorageBrowserElements,
 > {
-  (props: { key: string }): React.JSX.Element;
+  (props: { fileKey: string }): React.JSX.Element;
   Button: T['Button'];
   Icon: T['Icon'];
 }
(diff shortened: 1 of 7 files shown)


packages/react-storage/src/components/StorageBrowser/Views/Controls/__tests__/Download.spec.tsx
```diff
@@ -63,22 +56,30 @@ describe('DownloadControl', () => {
     expect(icon).toHaveAttribute('aria-hidden', 'true');
   });
 
-  it('calls downloadData onClick', async () => {
-    // @ts-expect-error
-    downloadSpy.mockResolvedValueOnce({ key: 'a_key' });
+  it('calls getUrl onClick', async () => {
```
timngyn, line 47: This test runs the `download` action too, but I couldn't get the mocks to correctly work for `document.location` when the `click()` happens and so there is a console error that gets printed during the tests

packages/react-storage/src/components/StorageBrowser/Views/Controls/Download.tsx
```diff
@@ -42,18 +43,39 @@ const DownloadButton: typeof ButtonElement = React.forwardRef(
...
+
+  a.click();
+
+  document.body.removeChild(a);
+}
+
+export const DownloadControl: DownloadControl = ({ fileKey }) => {
+  const [{ data }, handleDownload] = useDataState(downloadAction, {
+    signedUrl: '',
   });
```
timngyn, line 62: Using `useDataState` here instead because using context state was causing all the buttons in the table to re-render when I clicked a specific row to download. I think this encapsulates the state into each component individually

hbuchel (dismissed):

packages/react-storage/src/components/StorageBrowser/context/actions/downloadAction.ts
```diff
@@ -15,12 +15,18 @@ export function downloadAction(
...
-    },
-  });
-  return Promise.resolve({ key: path });
+  try {
+    const signedUrl = await getUrl({
+      path,
+      options: {
+        bucket,
+        locationCredentialsProvider: credentialsProvider,
+        validateObjectExistence: true,
```
timngyn, line 24: Added this `validateObjectExistence`, which throws an error if the item doesn't exist anymore so that the download link doesn't try to download something that doesn't exist.

At the moment if you try to download something that doesn't exist, there is no UI update to let the user know that the file they were trying to download doesn't exist. Can only see the error the dev tools console or network tab

hbuchel (approved):

reesscot (approved):

PR #5566, opened 2024-08-12
Feat(react): FederatedIdentities context providers

AhmetAhunbayAWS:
#### Description of changes

This adds in the main context providers for the FederatedIdentities component. There are also a few fixes:

- Added a new text baseElement
- Removed the toProviderData logic from the context provider

I wrote tests for each of the context providers and the toProviderData utility

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/src/components/Authenticator/FederatedIdentities/context/DisplayTextContext.tsx
+++ b/packages/react/src/components/Authenticator/FederatedIdentities/context/DisplayTextContext.tsx
@@ -4,13 +4,11 @@ export function displayText(displayName: string): string {
   return `Sign in with ${displayName}`;
 }
 
-const DisplayTextContext = React.createContext<typeof displayText | undefined>(
-  undefined
-);
+const DisplayTextContext = React.createContext<typeof displayText>(displayText);
 
 export const DisplayTextProvider = ({
   children,
-  customDisplayText,
+  customDisplayText = displayText,
 }: {
   children?: React.ReactNode;
   customDisplayText?: typeof displayText;
@@ -23,11 +21,5 @@ export const DisplayTextProvider = ({
 };
 
 export const useDisplayTextContext = (): typeof displayText => {
-  const _displayText = React.useContext(DisplayTextContext);
-
-  if (!_displayText) {
-    return displayText;
-  }
-
-  return _displayText;
+  return React.useContext(DisplayTextContext);
 };
--- a/packages/react/src/components/Authenticator/FederatedIdentities/context/DisplayTextContext.tsx
+++ b/packages/react/src/components/Authenticator/FederatedIdentities/context/DisplayTextContext.tsx
@@ -1,0 +1,35 @@
+import React from 'react';
+
+export function displayText(displayName: string): string {
+  return `Sign in with ${displayName}`;
+}
+
+const DisplayTextContext = React.createContext<typeof displayText | undefined>(
+  undefined
+);
+
+export const DisplayTextProvider = ({
+  children,
+  customDisplayText,
+}: {
+  children?: React.ReactNode;
+  customDisplayText?: typeof displayText;
+}): JSX.Element => {
+  return (
+    <DisplayTextContext.Provider value={customDisplayText}>
+      {children}
+    </DisplayTextContext.Provider>
+  );
+};
+
+export const useDisplayTextContext = (): typeof displayText => {
+  const context = React.useContext(DisplayTextContext);
+
+  if (!context) {
+    return displayText;
+  }
+
+  const _displayText = context;
+
+  return _displayText;
+};
--- a/packages/react/src/components/Authenticator/FederatedIdentities/context/DisplayTextContext.tsx
+++ b/packages/react/src/components/Authenticator/FederatedIdentities/context/DisplayTextContext.tsx
@@ -23,11 +23,11 @@ export const DisplayTextProvider = ({
 };
 
 export const useDisplayTextContext = (): typeof displayText => {
-  const context = React.useContext(DisplayTextContext);
+  const _displayText = React.useContext(DisplayTextContext);
 
-  if (!context) {
+  if (!_displayText) {
     return displayText;
   }
 
-  return context;
+  return _displayText;
 };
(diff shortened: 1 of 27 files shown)


packages/react/src/components/Authenticator/FederatedIdentities/controls/FederatedIdentitiesControl.tsx
```diff
@@ -28,13 +34,76 @@ export interface FederatedIdentities<T extends string = string>
...
+): ProviderData<T>[] {
+  validateProviderTypes(providers);
+  return providers.map((provider) => {
+    if (DefaultFederatedProviderList.includes(provider as FederatedProvider)) {
+      return getSupportedProviderData<T>(provider as T);
+    } else {
+      return provider as ProviderData<T>;
+    }
+  });
+}
```
thaddmt, line 73: Nit: Might make sense to move all of these to a utils file

thaddmt (approved):
LGTM, one small nit

packages/react/src/components/Authenticator/FederatedIdentities/context/DisplayTextContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <DisplayTextContext.Provider value={customDisplayText}>
+      {children}
+    </DisplayTextContext.Provider>
+  );
+};
+
+export const useDisplayTextContext = (): typeof displayText => {
+  const _displayText = React.useContext(DisplayTextContext);
+
+  if (!_displayText) {
```
dbanksdesign, line 28: Nit: would it be more clear to just pass `displayText` to `React.createContext` on line 7 instead of `undefined`? Then we wouldn't need to do this check as displayText from the context would always be defined.

packages/react/src/components/Authenticator/FederatedIdentities/context/HandleSignInWithRedirectContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+  );
+};
+
+export const useHandleSignInWithRedirectContext =
+  (): typeof signInWithRedirect => {
+    const handleSignInWithRedirect = React.useContext(
+      HandleSignInWithRedirectContext
+    );
+
+    if (!handleSignInWithRedirect) {
```
dbanksdesign, line 28: Same comment here

packages/react/src/components/Authenticator/FederatedIdentities/context/ProviderDataContext.tsx
```diff
@@ -0,0 +1,29 @@
...
+      {children}
+    </ProviderDataContext.Provider>
+  );
+};
+
+export const useProviderDataContext = (): ProviderData => {
+  const providerData = React.useContext(ProviderDataContext);
+
+  if (providerData === undefined) {
+    throw new Error();
```
dbanksdesign, line 26: We should include a helpful error message here

dbanksdesign (commented):
Looks good, just a few suggestions!

packages/react/src/components/Authenticator/FederatedIdentities/context/HandleSignInWithRedirectContext.tsx
```diff
@@ -0,0 +1,25 @@
...
+  customRedirect: SignInWithRedirectAction;
+}): JSX.Element => {
+  return (
+    <HandleSignInWithRedirectContext.Provider value={customRedirect}>
+      {children}
+    </HandleSignInWithRedirectContext.Provider>
+  );
+};
+
+export const useHandleSignInWithRedirectContext =
```
thaddmt, line 22: I think I'm confused between this file and `UseHandleSigninWithRedirectContext.tsx`, they also both export `useHandleSignInWithRedirectContext` and `UseHandleSignInWithRedirectContext`.

Can they be combined in some way?

packages/react/src/components/Authenticator/FederatedIdentities/context/UseHandleSignInWithRedirectContext.tsx
```diff
@@ -0,0 +1,46 @@
...
+  async function _signInWithRedirectAction(
+    _: void | undefined,
+    input: SignInWithRedirectInput
+  ): Promise<void> {
+    await handleSignInWithRedirect(input);
+  }
+
+  return (
+    <UseHandleSignInWithRedirectContext.Provider
+      value={useDataState(_signInWithRedirectAction, undefined)}
```
thaddmt, line 26: If we're passing the hook outputs here can we move it outside the props definition and just pass the outputs of the hook here?

packages/react/src/components/Authenticator/FederatedIdentities/controls/createProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  providers,
+  handleSignInWithRedirect = signInWithRedirect,
+  displayText,
+}: CreateProviderInput<T>) {
+  return function Provider({
+    children,
+  }: {
+    children?: React.ReactNode;
+  }): React.JSX.Element {
+    const context = React.useContext(UseHandleSignInWithRedirectContext);
```
thaddmt, line 26: Should this be `useHandleSignInWithRedirect`?

packages/react/src/components/Authenticator/FederatedIdentities/context/UseHandleSignInWithRedirectContext.tsx
AhmetAhunbayAWS: Yep, I can do that. Good catch, would definitely make things clearer.

packages/react/src/components/Authenticator/FederatedIdentities/controls/createProvider.tsx
AhmetAhunbayAWS: I just didn't want an error to be thrown here from the useHandleSignInWithRedirect function. But I could do the function with a try/catch if you think that would be better.

packages/react/src/components/Authenticator/FederatedIdentities/context/HandleSignInWithRedirectContext.tsx
AhmetAhunbayAWS: No, they are two different context providers. I can change the names to be more clear.

The idea is that this useHandleSignInWithRedirectContext is getting the handleSignInWithRedirectContext, which is used by the buttons. the UseHandleSignInWithRedirectContext is providing context for the useHandleSignInWithRedirect hook when it is called by the customer. I don't think there is a good way to combine these.

thaddmt (approved):

jordanvn (approved):
This all looks good to me

PR #5584, opened 2024-08-14
fix(storage-browser): set flag to specify when to download file so that it only happens on click

timngyn:
#### Description of changes
- Noticed a bug where a download would automatically happen when you navigate between prefixes. Added a flag to make sure that the download only happens on a click.

**Steps to reproduce bug**:
1. Navigate to a nested prefix with files to download
2. Download a file
3. Navigate up a prefix using the navigate breadcrumbs
4. Observe that the file that was just downloaded, gets downloaded again
   - I think this might be happening because of the `useEffect` in the `DownloadControl` component and how it will call the function to download if the`signedUrl` is valid. When we first download a file, it saves the `signedUrl` in its state and so navigating will cause each of the `DownloadControls` to re-render which causes it to download the file again since it has a valid `signedUrl`?
   
   

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/Views/Controls/Download.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Views/Controls/Download.tsx
@@ -48,13 +48,16 @@ export const DownloadControl: DownloadControl = ({ fileKey }) => {
     signedUrl: '',
   });
 
+  const [shouldDownload, setShouldDownload] = React.useState(false);
+
   const { signedUrl } = data;
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


hbuchel (approved):
This fixes it for me. I tried navigating back and forth with a few files using the breadcrumbs after downloading a few files and with this change it no longer auto-downloads.

calebpollman (approved):
Approving to get a fix out in the immediate term but we need to readdress this logic here to avoid usage of `React.useEffect`. @timngyn Can you please add a backlog item for refactoring?

PR #5585, opened 2024-08-14
chore(storage-browser): Add file/folder icons to table, some css cleanup

hbuchel:
#### Description of changes

- Adds File/Folder icons depending on the key/row type to LocationsView and LocationDetailView tables
- Adds a container for table cells. Since buttons inside of the table have padding, developers can use this to make a consistent padding around cells that don't have a button (so now the icons can line up)
- Added var(--storage-browser-gap-small) to use in cases like the Navigate item gap, and smaller gaps between text and icons.
- Some adjustments to the next-example custom UI

**LocationsView**
Screenshot 2024-08-14 at 12 48 22 PM

**LocationDetailView**
Screenshot 2024-08-14 at 12 48 41 PM

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
@@ -17,56 +17,74 @@ const {
   TableRow: BaseTableRow,
   Button,
   Icon,
+  View,
 } = StorageBrowserElements;
 
-const BLOCK_NAME = 'table';
+const BLOCK_NAME = `${CLASS_BASE}__table`;
 
 const Table = withBaseElementProps(BaseTable, {
-  className: `${CLASS_BASE}__${BLOCK_NAME}`,
+  className: `${BLOCK_NAME}`,
 });
 
 const TableBody = withBaseElementProps(BaseTableBody, {
-  className: `${CLASS_BASE}__${BLOCK_NAME}__body`,
+  className: `${BLOCK_NAME}__body`,
 });
 
 const TableHead = withBaseElementProps(BaseTableHead, {
-  className: `${CLASS_BASE}__${BLOCK_NAME}__head`,
+  className: `${BLOCK_NAME}__head`,
 });
 
 const TableHeader = withBaseElementProps(BaseTableHeader, {
-  className: `${CLASS_BASE}__${BLOCK_NAME}__header`,
+  className: `${BLOCK_NAME}__header`,
 });
 
 const TableHeaderButton = withBaseElementProps(Button, {
-  className: `${CLASS_BASE}__${BLOCK_NAME}__header-button`,
+  className: `${BLOCK_NAME}__header__button`,
   variant: 'sort',
 });
 
 const TableData = withBaseElementProps(BaseTableData, {
-  className: `${CLASS_BASE}__${BLOCK_NAME}__data`,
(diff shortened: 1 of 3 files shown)


packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -282,10 +307,12 @@ export const LocationDetailViewTable = (): JSX.Element => {
                     }}
                     key={`${index}-${row.key}`}
                   >
-                    {row.key}
+                    <TableDataIcon variant="folder" /> {row.key}
                   </TableDataButton>
                 ) : (
-                  <>{parseTableData(row, column)}</>
+                  <TableDataContainer>
```
calebpollman, line 313: Why do we need this?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -258,7 +276,14 @@ export const LocationDetailViewTable = (): JSX.Element => {
         } else if (column.key === ('download' as keyof LocationItem)) {
           return <DownloadControl fileKey={row.key} />;
         } else {
-          return row[column.key];
+          return (
+            <>
+              {column.key === 'key' && row.type === 'FILE' ? (
+                <TableDataIcon variant="file" />
```
calebpollman, line 282: Can we just use `IconElement` and pass the `classname`/`variant` directly here?

hbuchel: Yeah that works, updated in the latest.

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
hbuchel: This is to provide a way for there to be consistent spacing/padding in the table. Some table cells have buttons in them which have their own padding. So without *something* to apply padding to table cells that just have text in them (like a file name), the text doesn't line up with cells that have a button, like here Where the folder names look indented):

Screenshot 2024-08-14 at 2 14 12 PM

or here (where text doesn't line up to table headers, because they have sort buttons):

Screenshot 2024-08-14 at 2 14 21 PM

We'll probably also need this wrapper to do some text truncation also.

hbuchel: It could also just be a base View element? It really just needs a `<div>` with a class on it.

calebpollman: Hmmm, should we be able to use flex positioning on the `TableRow` and `TableCell` components to center content vertically? It feels off to me that the padding of the `button` is forcing changes to parent HTML structure

hbuchel: It won't be just the vertical height that is different, but also the horizontal space without it (like this screenshot with no wrapper element to provide padding to the text elements):

Screenshot 2024-08-14 at 2 43 01 PM

I could use :has() in css to remove excess padding if the td has a button in it, but I think we'll need a wrapper for text truncation anyways.

calebpollman (approved):

reesscot (approved):

PR #5593, opened 2024-08-15
chore(storage-browser): grid updates

hbuchel:
#### Description of changes

- use auto rows so vertical gaps aren't present for empty rows
- use less columns

**Before**
Screenshot 2024-08-26 at 12 11 07 PM

**After**

Screenshot 2024-08-26 at 12 12 22 PM

#### Checklist

- [ ] PR description included


packages/react-storage/src/styles/storage-browser.css
```diff
@@ -3,29 +3,21 @@
   --storage-browser-gap: 0.6rem;
   display: grid;
   gap: var(--storage-browser-gap);
-  grid-template-columns: repeat(32, 1fr);
-  grid-template-rows: repeat(24, min-content);
-}
-
-.storage-browser__header {
```
hbuchel, line 10: old css we're not using anymore

calebpollman:
Will using removing the empty `rows` create a scenario where the space above the Table is inconsistent between Views?

hbuchel:
> Will removing the empty `rows` create a scenario where the space above the Table is inconsistent between Views?

Sort of? In the current implementation, it does the same thing except the gaps from empty rows are still visible (so not the empty row itself, but the gap created by that row). This makes it so both the empty rows and the gaps from those rows are not visible which was creating weird inconsistent spacing.

timngyn (approved):
lgtm!

hbuchel:
> > Will removing the empty `rows` create a scenario where the space above the Table is inconsistent between Views?
> 
> Sort of? In the current implementation, it does the same thing except the gaps from empty rows are still visible (so not the empty row itself, but the gap created by that row). This makes it so both the empty rows and the gaps from those rows are not visible which was creating weird inconsistent spacing.

Added a Before/After image to show the difference in gaps.

calebpollman (approved):

Issue #5590, opened 2024-08-15
Unable to resolve "style-dictionary/lib/utils/deepExtend"


Felix1772:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React Native

### Which UI component?

Authenticator

### How is your app built?

Expo

### What browsers are you seeing the problem on?

Chrome, Firefox, Microsoft Edge, Safari, iOS (React Native), Android (React Native)

### Please describe your bug.

After following the installation guide of AmplifyUI and installing all necessary modules I experience the following error:
`Unable to resolve "style-dictionary/lib/utils/deepExtend" from "node_modules/@aws-amplify/ui-react-native/src/theme/createTheme.ts"`

Then I tried to install `style-dictionary` which fixed this specific error, but created a new one:
`Unable to resolve "style-dictionary/lib/utils/references/usesReference" from "node_modules/@aws-amplify/ui-react-native/src/theme/createTheme.ts`

After looking into `createTheme.ts` and the node_module of `style-dictionary` I noticed that it is called `usesReferences`.
After changing to that I got the error that `structuredClone` is not available on native platforms as it is only available on the web.

### What's the expected behaviour?

Working Authenticator without a `style-dictionary`-error and not having to install it because it is not described in the installation guide.

### Help us reproduce the bug!

1. Initialize expo project
2. Installation of:
    ```
    npm add \
      @aws-amplify/ui-react-native \
      @aws-amplify/react-native \
      aws-amplify \
      @react-native-community/netinfo \
      @react-native-async-storage/async-storage \
      react-native-safe-area-context \
      react-native-get-random-values \
      react-native-url-polyfill
    ```

### Code Snippet

```typescript
// Put your code below this line.

```

### Console log output

`Unable to resolve "style-dictionary/lib/utils/deepExtend" from "node_modules/@aws-amplify/ui-react-native/src/theme/createTheme.ts`

justinmakaila:
[...]

PR #5594, opened 2024-08-15
fix: add style-dictionary dependency to docs and react-native package

jordanvn:
#### Description of changes

`style-dictionary` is a required dependency of docs/ and packages/react-native/

Previously, this dependency had been indirectly provided via ui/, which caused regression when `style-dictionary` was reclassified as a dev dependency in ui/

These changes seek to resolve the issue by adding `style-dictionary` as a dependency to docs/ and packages/react-native/

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/5590

This was confirmed by validating attached issue using latest version of `@aws-amplify/ui-react-native`, followed by validating the resolution of the issue by rolling back to a version of `@aws-amplify/ui-react-native` in which the dependency was provided.

#### Checklist

- [x] PR description included

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -53,7 +53,8 @@
     "remark-mdx-images": "^1.0.2",
     "rimraf": "^3.0.2",
     "sass": "^1.43.4",
-    "slugify": "^1.6.3"
+    "slugify": "^1.6.3",
+    "style-dictionary": "3.9.1"
   },
   "license": "MIT",
   "devDependencies": {
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -31,7 +31,8 @@
   "dependencies": {
     "@aws-amplify/ui": "6.1.0",
     "@aws-amplify/ui-react-core": "3.0.18",
-    "@aws-amplify/ui-react-core-notifications": "2.0.18"
+    "@aws-amplify/ui-react-core-notifications": "2.0.18",
+    "style-dictionary": "3.9.1"
   },
   "peerDependencies": {
     "aws-amplify": "^6.3.2",
--- a/.changeset/tame-toys-agree.md
+++ b/.changeset/tame-toys-agree.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ui-react-native': patch
+---
+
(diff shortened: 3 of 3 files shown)


docs/package.json
```diff
@@ -53,7 +53,8 @@
     "remark-mdx-images": "^1.0.2",
     "rimraf": "^3.0.2",
     "sass": "^1.43.4",
-    "slugify": "^1.6.3"
+    "slugify": "^1.6.3",
+    "style-dictionary": "3.9.1"
```
calebpollman, line 57: Do the docs also rely on `style-dictionary`? Seems like we wouldn't have been able to deploy without this if so?

packages/react-native/package.json
```diff
@@ -31,7 +31,8 @@
   "dependencies": {
     "@aws-amplify/ui": "6.1.0",
     "@aws-amplify/ui-react-core": "3.0.18",
-    "@aws-amplify/ui-react-core-notifications": "2.0.18"
+    "@aws-amplify/ui-react-core-notifications": "2.0.18",
+    "style-dictionary": "3.9.1"
```
reesscot, line 35: We've confirmed this is the version used by previous ui and this fixes the build issue when building an expo app?

calebpollman (commented):
RN LGTM! Just to confirm, have we manually tested installation of the RN package in a standalone app? Also will need to add a changeset to bump the RN version

docs/package.json
jordanvn: `style-dictionary` referenced in docs/src/components/home/themes/index.preval.ts and not included in docs/package.json

We didn't directly observe a failure though

packages/react-native/package.json
dbanksdesign: it is the version in the current yarn.lock file

dbanksdesign (approved):

packages/react-native/package.json
jordanvn: We have confirmed that this fixes the build issue when building an expo app

jacoblogan (approved):

PR #5614, opened 2024-08-20
chore(storage-browser): add LoadingControl

hbuchel:
#### Description of changes

- Adds `<Loading />` control. 
- Removes placeholder loading elements from `LocationDetailViewTable` and `LocationsViewTable`

**Loading locations**

https://github.com/user-attachments/assets/fa241f7e-29ab-4e2c-a169-9e0fdc198c95

**Loading items**

https://github.com/user-attachments/assets/c8f29ffe-285d-492b-9ab3-0584ae65755e

#### Checklist

- [ ] PR description included


timngyn (dismissed):

packages/react-storage/src/components/StorageBrowser/Views/Controls/Loading.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export interface LoadingControl<
+  T extends StorageBrowserElements = StorageBrowserElements,
+> extends OmitElements<_LoadingControl<T>, 'Icon' | 'Text' | 'View'> {
+  ({ loadingText }: LoadingControlProps): React.JSX.Element;
+}
+
+export const LoadingControl: LoadingControl = ({ loadingText }) => (
+  <View className={BLOCK_NAME}>
```
calebpollman, line 28: It is possible to use a fragment here and use flex properties to center the icon and expand the margins of the Icon to fill the parent container?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Loading.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <Icon className={`${BLOCK_NAME}__icon`} variant="loading" />
+    <Span className={`${BLOCK_NAME}__text`} aria-live="polite">
+      {loadingText ?? 'Loading'}
+    </Span>
```
calebpollman, line 32: Is the text here required for ally, if so is it possible to use `aria-live` with hidden content? IMO there is not much context provided by a loading message that is not already conveyed in the controls and each additional piece of text we add requires further configuration properties

packages/react-storage/src/components/StorageBrowser/Views/Controls/Loading.tsx
```diff
@@ -0,0 +1,34 @@
...
+  ({ loadingText }: LoadingControlProps): React.JSX.Element;
+}
+
+export interface LoadingControl<
+  T extends StorageBrowserElements = StorageBrowserElements,
+> extends OmitElements<_LoadingControl<T>, 'Icon' | 'Text' | 'View'> {
+  ({ loadingText }: LoadingControlProps): React.JSX.Element;
+}
+
+export const LoadingControl: LoadingControl = ({ loadingText }) => (
```
calebpollman, line 27: Text content should be provided as `children`

packages/react-storage/src/components/StorageBrowser/Views/Controls/Loading.tsx
hbuchel: Yes, we need some sort of context for screen readers so accessible text of some sort is needed. I have simplified this to just say "Loading" and not accept props in latest commit, because I don't think the custom text on each view provides that much more context. But, even if we accessibly hide this text, it will eventually need to be configurable for i18n.

packages/react-storage/src/components/StorageBrowser/Views/Controls/Loading.tsx
hbuchel: Removed this in favor of using the default text.

packages/react-storage/src/components/StorageBrowser/Views/Controls/Loading.tsx
hbuchel: I think that would make the layout of this component a little too dependent on the parent; and a little more brittle should you change any of the styling on the parent (which in this case would be the top level `.storage-browser` and it's grid/alignment styling). I'd prefer to keep the internal layout of this (the flex positioning, centering) dependent on it's own wrapper. That way if developers want to change the internal layout, like stacking the icon on top of the text for mobile, adding a background/padding, etc. it will be easier to do.

packages/react-storage/src/components/StorageBrowser/Views/Controls/__tests__/Table.spec.tsx
```diff
@@ -195,7 +195,7 @@ describe('LocationsViewTable', () => {
     await waitFor(() => {
       render(
         <Provider>
-          <LocationDetailViewTable />
```
hbuchel, line 198: Not sure how this happened but this was testing the wrong table so the test was failing on this PR.

timngyn (approved):

packages/react-storage/src/components/StorageBrowser/Views/Controls/__tests__/Table.spec.tsx
timngyn: I'll look into it, sorry that this happened

dbanksdesign (approved):

packages/react-storage/src/components/StorageBrowser/Views/Controls/__tests__/Table.spec.tsx
hbuchel: No worries!

PR #5632, opened 2024-08-20
feat(primitives): add avatar primitive

dbanksdesign:
#### Description of changes

Adding Avatar primitive to be use in the AIConversation component. 

```jsx
{/* Avatar with image */}
<Avatar src="/cats/5.jpg" />
{/* Avatar with default placeholder icon */}
<Avatar />
{/* Avatar with initials */}
<Avatar>DB</Avatar>
{/* Avatar with custom icon */}
<Avatar>
  <FiSmile style={{ width: '60%', height: '60%' }} />
</Avatar>
```

CleanShot 2024-08-20 at 15 23 35

CleanShot 2024-08-20 at 14 19 05@2x

#### Checklist

- [x] PR description included


docs/src/pages/[platform]/components/avatar/examples/DefaultAvatarExample.tsx
```diff
@@ -0,0 +1,19 @@
...
+      {/* Avatar with image */}
+      <Avatar src="/cats/5.jpg" />
+      {/* Avatar with default placeholder icon */}
+      <Avatar />
+      {/* Avatar with initials */}
+      <Avatar>DB</Avatar>
+      {/* Avatar with custom icon */}
+      <Avatar>
+        <FiSmile style={{ width: '60%', height: '60%' }} />
+      </Avatar>
```
Comment (thaddmt) [inline docs/src/pages/[platform]/components/avatar/examples/DefaultAvatarExample.tsx:16]: Just clarifying that for our use case with AIConversation we will just pass two separate instances of this `Avatar` with a custom icon for the user and the assistance right?

thaddmt (dismissed):

docs/src/pages/[platform]/components/avatar/demo.tsx
```diff
@@ -0,0 +1,43 @@
...
+    (src ? `\n  src=${JSON.stringify(src)}` : '') +
+    (size ? `\n  size=${JSON.stringify(size)}` : '') +
+    (colorTheme ? `\n  colorTheme=${JSON.stringify(colorTheme)}` : '') +
+    (variation ? `\n  variation=${JSON.stringify(variation)}` : '') +
+    `\n/>`
+  );
+};
+
+const defaultAvatarProps = {
+  src: '/cats/1.jpg',
```
Comment (hbuchel) [inline docs/src/pages/[platform]/components/avatar/demo.tsx:22]: Just a nit: should we use a transparent avatar or svg icon so the demo can show the variations better? Currently the demo doesn't make it clear that filled/default variations do much when you change them.

docs/src/pages/[platform]/components/avatar/examples/DefaultAvatarExample.tsx
Comment (dbanksdesign) [reply docs/src/pages/[platform]/components/avatar/examples/DefaultAvatarExample.tsx:16]: Yup

hbuchel (commented):
LGTM! 

Can we include an accessibility section with an example of using alt text with an image? It wasn't immediately clear to me that I could pass an alt attribute to `<Avatar />` and it would find it's way to the ``

The default placeholder Avatar is essentially non-existent to screen readers. A developer could overcome this by passing an `aria-label` to `<Avatar />` with whatever placeholder text makes sense for their application. I don't think that needs to be "braked in" but maybe we could show that in a11y example, too?

hbuchel (approved):

thaddmt (approved):

PR #5640, opened 2024-08-21
feat(scrollview): add autoScroll prop

dbanksdesign:
#### Description of changes

Added a `autoScroll` prop to the `ScrollView` primitive that can be set to 'smooth', 'instant', or 'auto'. This will be useful for a chat history to always show the latest message. 

CleanShot 2024-08-20 at 22 29 03

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/src/primitives/ScrollView/ScrollView.tsx
+++ b/packages/react/src/primitives/ScrollView/ScrollView.tsx
@@ -23,7 +23,7 @@ const ScrollViewPrimitive: Primitive<ScrollViewProps, 'div'> = (
     internalRef,
   });
 
-  React.useLayoutEffect(() => {
+  React.useEffect(() => {
     if (autoScroll) {
       internalRef.current?.scrollTo({
         top: internalRef.current?.scrollHeight,
@@ -31,7 +31,10 @@ const ScrollViewPrimitive: Primitive<ScrollViewProps, 'div'> = (
         behavior: autoScroll,
       });
     }
-  }, [children, autoScroll]);
+  }, [
+    children, // include children in the dependency array so that when children changes the scrollview scrolls
+    autoScroll,
+  ]);
 
   return (
     <View
--- a/packages/react/src/primitives/ScrollView/ScrollView.tsx
+++ b/packages/react/src/primitives/ScrollView/ScrollView.tsx
@@ -11,23 +11,42 @@ import {
 } from '../types';
 import { View } from '../View';
 import { primitiveWithForwardRef } from '../utils/primitiveWithForwardRef';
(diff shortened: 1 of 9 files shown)


thaddmt (approved):

packages/react/src/primitives/ScrollView/ScrollView.tsx
```diff
@@ -11,23 +11,42 @@ import {
...
+  { children, className, orientation, autoScroll, ...rest },
+  externalRef
+) => {
+  const internalRef = React.useRef<HTMLDivElement | null>(null);
+  const composedRefs = useComposeRefsCallback<HTMLDivElement | null>({
+    externalRef,
+    internalRef,
+  });
+
+  React.useLayoutEffect(() => {
```
calebpollman, line 26: What's the reasoning for `useLayoutEffect` rather than `useEffect` and do we need to support SSR?

packages/react/src/primitives/ScrollView/ScrollView.tsx
```diff
@@ -11,23 +11,42 @@ import {
...
+    if (autoScroll) {
+      internalRef.current?.scrollTo({
+        top: internalRef.current?.scrollHeight,
+        left: internalRef.current?.scrollWidth,
+        behavior: autoScroll,
+      });
+    }
+  }, [children, autoScroll]);
```
calebpollman, line 34: Without context in to why `children` is included in the dependency array, a future maintainer might think this was a typo and remove. 

Can we add a short comment here explaining the impact on behavior and a unit test that makes an  assertion that will fail if `children` is removed?

packages/react/src/primitives/types/scrollView.ts
```diff
@@ -9,6 +9,12 @@ export interface BaseScrollViewProps extends BaseViewProps {
    * make the width or height of the ScrollView component smaller than the content.
    */
   orientation?: ScrollViewOrientation;
+
+  /**
+   * @description
+   * Scroll to the end as the children of it change
+   */
+  autoScroll?: 'smooth' | 'instant' | 'auto';
```
calebpollman, line 17: Any potential in the future that we would want to support auto scroll behavior in addition to scrolling to the end of the view?

packages/react/src/primitives/ScrollView/ScrollView.tsx
dbanksdesign: TBH I dunno, changed it to useEffect and it works fine

calebpollman (approved):

calebpollman (approved):

thaddmt (approved):

PR #5641, opened 2024-08-21
chore(storage-browser): add file extension as column type

timngyn:
#### Description of changes
- Created helper function `renderTableData` to help render the row items for the upload controls table
- Added `type` column to upload controls table
- Parse out file extension from file key and use it to display in the `type` column

image

image

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/UploadControls.tsx
```diff
@@ -75,6 +79,46 @@ export const ActionIcon = ({ status }: ActionIconProps): React.JSX.Element => {
...
+        );
+      case 'type': {
+        const indexOfDot = row.key.lastIndexOf('.');
+        return indexOfDot > -1 ? (
+          <TableDataText>
+            {row.key.slice(indexOfDot + 1).toUpperCase()}
+          </TableDataText>
+        ) : (
+          ''
+        );
```
calebpollman, line 102: Same feedback regarding usage of parsing based on `.` position. Just a general callout that if we have data that is only used for display that we should handle any processing/parsing previous to passing data to consumers to prevent additional work in the react render cycle

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -71,10 +72,11 @@ export function useHandleUpload({
 }): [tasks: CancelableTask[], handleUpload: () => void] {
   const getConfig = useGetLocationConfig();
   const [tasks, setTasks] = React.useState<CancelableTask[]>(() =>
-    (items ?? []).map(({ key, data, size }) => ({
+    (items ?? []).map(({ key, data, size, type }) => ({
       cancel: () => setTasks((prev) => removeTask(prev, key)),
       key,
       size,
+      type,
```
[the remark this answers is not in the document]
timngyn: oh that's a good question, I think we actually assign the `type` inside `ActionSelect` as `FILE` because of the `FileItem` type

```tsx
export interface FileItem {
  key: string;
  data?: File;
  lastModified: Date;
  size: number;
  type: 'FILE';
}
```

It does look like the actual type is exposed though. Do you think it makes sense to use the real type instead of `FILE`? Screenshot below of the type I'm able to see when adding a file

image

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -441,6 +441,15 @@ export const LocationDetailViewTable = (): JSX.Element | null => {
               case 'download' as keyof LocationItem: {
```
reesscot, line 441: non-blocking, but does this still need a cast if the ones below don't?

timngyn: I think so, I think if we don't want to cast it as `LocationItem`, then we'll have to add it to the `FileItem` type

```
export interface FileItem {
  key: string;
  data?: File;
  lastModified: Date;
  size: number;
  type: 'FILE';
}
```

calebpollman: Casting inside the input of a public callback as a code smell indicating potential friction points between the interface and its consumers that requires addressing.

In the context of React, a function that returns JSX  (a `render` function), providing `props` is idiomatic allowing for seamless passing of components as the callback parameter. 

Let's sync offline to track down where we need to modify our approach to address

hbuchel (dismissed):

calebpollman (approved):

PR #5659, opened 2024-08-22
fix(ai): fix excessive type depth error

thaddmt:
#### Description of changes
- Narrowing down the type of the V6 client seems to fix the issue with excessive type depth checking error

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-ai/src/types.ts
+++ b/packages/react-ai/src/types.ts
@@ -1,4 +1,4 @@
-import { V6Client } from '@aws-amplify/api-graphql';
+import { V6Client as SDKV6Client } from '@aws-amplify/api-graphql';
 import {
   Conversation as SDKConversation,
   ConversationMessage as SDKConversationMessage,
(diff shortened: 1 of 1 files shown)


packages/react-ai/src/types.ts
```diff
@@ -41,4 +41,8 @@ export interface SendMesageParameters {
 
 export type SendMessage = (input: SendMesageParameters) => void;
 
+type V6Client<T extends Record<any, any>> = Pick<
```
reesscot, line 44: Is the any type required?

reesscot (approved):

dbanksdesign (approved):

packages/react-ai/src/types.ts
calebpollman: Prefer not replicating a narrowed `V6Client` interface to prevent it leaking out of this file causing conflicts with the existing interface. Can we narrow the conditional inference of `Schema`?

PR #5675, opened 2024-08-26
chore(storage-browser): use DataTable for locations view table

timngyn:
#### Description of changes
- Use `DataTable` component to create table for locations view

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,185 @@
...
+                  : `sort-${direction}`
+              }
+            />
+          ) : (
+            <IconElement variant="sort-indeterminate" />
+          )}
+        </ButtonElement>
+      ),
+      key: 'th-scope',
+      className: `${TABLE_HEADER_CLASS_NAME} ${TABLE_HEADER_CLASS_NAME}--scope`,
```
timngyn, line 63: Adding this line from the previous `TableControl` implementation here: https://github.com/aws-amplify/amplify-ui/blob/8e34f98ac783c2964bf34e3da36d58581599dfd7/packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx#L90-L94

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+    () =>
+      getLocationsData({
+        data: data.result,
+        sortState,
+        onLocationClick: (payload) => {
+          handleUpdateState({
+            type: 'ACCESS_LOCATION',
+            location: payload,
+          });
+        },
```
calebpollman, line 157: Nit:
```suggestion
        onLocationClick: (location) => {
          handleUpdateState({ type: 'ACCESS_LOCATION', location });
        },
```

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+        onTableHeaderClick: (payload: keyof LocationAccess<Permission>) => {
```
calebpollman, line 158: Think `payload` can just be `string`?

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+          }));
+        },
+      }),
+    [data.result, handleUpdateState, sortState]
+  );
+
+  const hasLocations = !!data.result?.length;
+  const shouldRenderLocations = hasLocations && !isLoading;
+
+  return shouldRenderLocations ? <DataTable data={locationsData} /> : null;
```
calebpollman, line 172: `DataTableControl` should not be responsible for conditional rendering, can remove the logic here

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+      ),
+    },
+    { key: `td-type-${index}`, children: location.type },
+    { key: `td-permission-${index}`, children: location.permission },
+  ]);
+
+  return { columns, rows };
+};
+
+export function LocationsViewTableControl(): React.JSX.Element | null {
```
calebpollman, line 135: This can just be `DataTableControl` and `null` can be removed from the return type

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  if (direction === 'ascending') {
+    data.sort((a, b) => compareFn(a[selection], b[selection]));
+  } else {
+    data.sort((a, b) => compareFn(b[selection], a[selection]));
+  }
+
+  const rows = data.map((location, index) => [
+    {
+      key: `td-scope-${index}`,
```
calebpollman, line 117: Replace "scope" with "name"

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/LocationsView.tsx
```diff
@@ -61,6 +60,7 @@ export const LocationsMessage = (): React.JSX.Element | null => {
   ) : null;
 };
 
+// @ts-expect-error TODO: add Controls assignment
 const LocationsViewControls: LocationsViewControls = () => {
```
calebpollman, line 64: Just remove the type assignment

calebpollman:
Looks good, much cleaner approach!

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
timngyn: I think setting it as `string` causes some type issues when I try to use `selection` to grab the compare function `const compareFn = LocationsViewColumnSortMap[selection]`. It says:


Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ scope: (a: string, b: string) => number; type: (a: string, b: string) => number; permission: (a: string, b: string) => number; }'.

timngyn: This is what I have at the moment:


```tsx
const getCompareFn = (selection: string) => {
  switch (selection) {
    case 'scope':
    case 'type':
    case 'permission':
      return compareStrings;
  }
};

const compareFn = getCompareFn(selection);

if (compareFn) {
  // not sure how to get it without any TS errors without casting as keyof LocationAccess
  const castSelection = selection as keyof LocationAccess;

  if (direction === 'ascending') {
    data.sort((a, b) => compareFn(a[castSelection], b[castSelection]));
  } else {
    data.sort((a, b) => compareFn(b[castSelection], a[castSelection]));
  }
}


```

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+  selection: string;
+  direction: SortDirection;
+  onTableHeaderClick: (location: string) => void;
+}) => ({
+  children: (
+    <ButtonElement variant="sort" className={TABLE_HEADER_BUTTON_CLASS_NAME}>
+      {capitalize(key)}
+      <IconElement
+        variant={
+          selection === key && direction !== 'none'
```
calebpollman, line 48: In what scenario is the `direction` check relevant if the `selection === key` is `true`?

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+  TABLE_DATA_BUTTON_CLASS,
+  TABLE_HEADER_BUTTON_CLASS_NAME,
+  TABLE_HEADER_CLASS_NAME,
+} from '../../../components/DataTable';
+import { useControl } from '../../../context/controls';
+import { useLocationsData } from '../../../context/actions';
+import { LocationAccess } from '../../../context/types';
+import { compareStrings } from '../../../context/controls/Table';
+import { ButtonElement, IconElement } from '../../../context/elements';
+import { capitalize } from '@aws-amplify/ui';
```
calebpollman, line 14: External imports should be listed before local imports, e.g.:
```
import React from 'react';

import { capitalize } from '@aws-amplify/ui';

import { useControl } from '../../../context/controls';
```

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
timngyn: I'm not sure if the case could ever happen, but I added it because I was getting a typescript error that said `sort-none` wasn't a variant for `IconElement` and so this check helps with the typescript complaining

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+      />
+    </ButtonElement>
+  ),
+  key: `th=${key}`,
+  className: `${TABLE_HEADER_CLASS_NAME} ${TABLE_HEADER_CLASS_NAME}--${key}`,
+  onClick: () => onTableHeaderClick(key),
+  'aria-sort': selection === key ? direction : 'none',
+});
+
+const columnsArray = ['scope', 'type', 'permission'];
```
timngyn, line 61: still trying to figure this part out. At the moment need to have a custom display name

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+    onClick: () => onTableHeaderClick(key),
+    'aria-sort': selection === key ? direction : 'none',
+  };
+};
+
+const displayColumns: Record<string, string>[] = [
+  { scope: 'name' },
+  { type: 'type' },
+  { permission: 'permission' },
+];
```
timngyn, line 70: would anyone do this differently? Needed this since `scope` uses the `name` display name

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/LocationsView.tsx
timngyn: I tried removing it, but I think I'll have to update some of the other types too like for `LocationsView`. I think I can do this as a follow up though just so there isn't extra changes

hbuchel (dismissed):

reesscot (dismissed):

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <IconElement
+          variant={
+            selection === key && direction !== 'none'
+              ? `sort-${direction}`
+              : 'sort-indeterminate'
+          }
+        />
+      </ButtonElement>
+    ),
+    key: `th=${key}`,
```
calebpollman, line 59: Why `=` here instead of `_`? Can we align on a syntax?

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
calebpollman: This is fine for now but not what we want long term. We can update the structure of `LocationAccess` to help here

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
```diff
@@ -0,0 +1,156 @@
...
+  return { columns, rows };
+};
+
+export function DataTableControl(): React.JSX.Element {
+  const [{ data }] = useLocationsData();
+
+  const [, handleUpdateState] = useControl({ type: 'NAVIGATE' });
+
+  const [sortState, setSortState] = React.useState<SortState>({
+    selection: 'scope',
```
calebpollman, line 129: Can we please stop using `scope` for reference values?

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
timngyn: ah that was a typo, will update

packages/react-storage/src/components/StorageBrowser/Views/LocationsView/Controls/DataTable.tsx
timngyn: I think I chose to use `scope` as the selection because that's how I would access it inside the location object when doing the compare. `name` isn't technically a property on the `LocationAccess` object is it?

calebpollman: No it's not. For context here, exposing `scope` increases the need for conditional logic to be applied to convert it to `name`

calebpollman (approved):

PR #5683, opened 2024-08-26
Add missing class to text cells

hbuchel:
#### Description of changes

Adds back the missing wrapper for text cells in LocationsView table

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/components/DataTable.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/components/DataTable.tsx
@@ -16,6 +16,7 @@ export const TABLE_HEADER_CLASS_NAME = `${TABLE_CLASS_NAME}__header`;
 export const TABLE_HEADER_BUTTON_CLASS_NAME = `${TABLE_CLASS_NAME}__header__button`;
 export const TABLE_ROW_CLASS_NAME = `${TABLE_CLASS_NAME}__row`;
 export const TABLE_DATA_CLASS_NAME = `${TABLE_CLASS_NAME}__data`;
+export const TABLE_DATA_TEXT_CLASS_NAME = `${TABLE_CLASS_NAME}__data__text`;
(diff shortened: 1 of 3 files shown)


packages/react-storage/src/components/StorageBrowser/components/DataTable.tsx
```diff
@@ -16,6 +16,7 @@ export const TABLE_HEADER_CLASS_NAME = `${TABLE_CLASS_NAME}__header`;
 export const TABLE_HEADER_BUTTON_CLASS_NAME = `${TABLE_CLASS_NAME}__header__button`;
 export const TABLE_ROW_CLASS_NAME = `${TABLE_CLASS_NAME}__row`;
 export const TABLE_DATA_CLASS_NAME = `${TABLE_CLASS_NAME}__data`;
+export const TABLE_DATA_TEXT_CLASS_NAME = `${TABLE_CLASS_NAME}__data__text`;
```
calebpollman, line 19: Can we use the existing `TABLE_DATA_CLASS_NAME ` instead of adding an additional class name?

hbuchel: HMM let me see if I can drop the span. I think it makes sense if we keep this span element to have different classes on them since they are useful hooks for developers to style.

But, I'll see if we can forego the span element all together. We still need something to truncate text and make the items line up so they don't look goofy when in a table (like the LocationsDetailView) where some of the name columns have buttons, and some have padding.

calebpollman: Think your approach make sense actually, didn't realize `TABLE_DATA_CLASS_NAME` was being applied to the wrapping `TableDataElement` already

calebpollman (approved):

timngyn (dismissed):

PR #5703, opened 2024-08-28
fix(storage-browser): update useHandleUpload to return handleFileSelect function

timngyn:
#### Description of changes
- Update `useHandleUpload` to return a `handleFileSelect` function that helps create and manages previous tasks
- Pass the `handleFileSelect` function to `useFileSelect` so that `handleFileSelect` can accept the new files from the selection and manages the state of these files and tasks
- Disable "start", "add files" and "add folder" buttons after start action is pressed

To test:
1. Try uploading a folder (with multiple files) from the location items table
   - Verify that you see the folder name in the `folder` column
2. Remove one of the files from the folder you uploaded from step 1
   - Verify the file gets removed and nothing gets added to the tasks table
3. Add a new file with the "Add files" button
   - Verify the file gets added
4. Try uploading the same folder you uploaded from step 1 using the "Add folder" button
   - Verify that the folder gets added and the items that were added appear at the top of the table
   
Feel free to test out different combinations of adding files and deleting them and re-adding files

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -26,10 +26,18 @@ interface TaskUpdate extends Partial<CancelableTask> {
 
 const removeTask = <T extends Task | CancelableTask>(
   tasks: T[],
-  key: string
+  key: string,
+  updateFiles: React.Dispatch<React.SetStateAction<File[]>>
```
calebpollman, line 30: Running a side effect from a consumer can lead to subtle complexity as useHandleUpload is no longer a pure function, relying on the behavior of an external module. Ideally we should update useHandleUpload to expose a handler that can be provided to useFileSelect that will accept files and convert them to tasks/merge with existing tasks to ensure a single source of truth and simplify task lifecycle management

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -67,39 +73,70 @@ const updateTasks = <T extends Task | CancelableTask>(
...
-}): [tasks: CancelableTask[], handleUpload: () => void] {
+}): [
+  tasks: CancelableTask[],
+  handleUpload: () => void,
+  handleFileSelect: (files: File[]) => void,
+] {
   const getConfig = useGetLocationConfig();
 
   const [tasks, setTasks] = React.useState<CancelableTask[]>(() => []);
+  const [_, setFiles] = React.useState<File[]>([]);
```
timngyn, line 114: Added a `files` state here so that I could keep track of what files have been added and remove them when the task is removed.

Would this be okay, or is this coupling the files and tasks too closely?

packages/react-storage/src/components/StorageBrowser/Views/Controls/Table.tsx
```diff
@@ -17,7 +17,7 @@ import {
 export type SortDirection = 'ascending' | 'descending' | 'none';
 
 export type SortState<T> = {
-  selection: keyof T;
+  selection: keyof T | undefined;
```
calebpollman, line 20: Not sure i understand this change?

timngyn: It's so that I can set the upload table to have no default sorting so that when you try to re-add files, they appear at the top of the table

If we make it have a default sorting then it's not obvious that re-adding a file did anything because the sorting hides this update, if that makes sense

If a user does decide to sort the table and re-add an item then this UX issue is still there though

timngyn: Here's a video of what I'm talking about. I can remove these changes and just make the table have a default sort though, so no worries

In the video below, I'm re-adding the file `lighthouse-performance-results (2).zip` and can see it appear at the top of the table. Then I sort the column by name and try re-adding the same file, but it doesn't look like anything is happening because of the table sort making the file stay in it's sorted position:

https://github.com/user-attachments/assets/7b20dbca-6791-455f-967f-577ba1f9a4e1

calebpollman: Let's continue sorting based on the end user selection, it's deterministic

reesscot: I think it's ok if it's not obvious that re-adding a file did anything. If the table is sorted, it should always go wherever the sort says it should go.

timngyn: okay thank you both! I'll undo these changes then and have the tasks table sort by `name` by default again

calebpollman (approved):

reesscot (dismissed):

reesscot (dismissed):

hbuchel (dismissed):

calebpollman (dismissed):

calebpollman (approved):

reesscot (approved):

PR #5708, opened 2024-08-29
feat(storage-browser): process single task in useHandleUpload

timngyn:
#### Description of changes
- Updates to `useHandleUpload` 
   - Updated `useHandleUpload` to grab `batchSize` number of of tasks to initiate the upload
   - Return new `handleCancel` function to cancel all tasks
- Updates to `UploadControls`
   - Added helper function to print friendly names for the task status in the table
   - Added button to cancel all tasks
   - Print out progress as percentage

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +138,76 @@ export function useHandleUpload({
...
+      try {
+        await result;
+        setTasks((curr) => updateTasks(curr, { key, status: 'COMPLETE' }));
+      } catch (error) {
+        if (isCancelError(error)) {
           setTasks((curr) =>
-            // assign cancel to noop
-            updateTasks(curr, { key, cancel: () => null, status: 'CANCELED' })
+            updateTasks(curr, { key, status: 'CANCELED', cancel: undefined })
           );
```
timngyn, line 194: calling `result.cancel()` from `uploadData(input)` will cause the upload to throw an error and we have to check if it's a "cancel" error so that we can handle it as a cancel

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/UploadControls.tsx
```diff
@@ -112,6 +112,31 @@ const Destination = ({ children }: { children?: React.ReactNode }) => {
   );
 };
 
+const getFriendlyTaskStatus = (status: string): string => {
```
reesscot, line 115: nit: could probably go in a utils file so it can be unit tested separately

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/UploadControls.tsx
```diff
@@ -150,16 +175,24 @@ const renderRowItem: RenderRowItem<LocationActionViewColumns> = (
...
+        );
       case 'progress':
-        return <TableDataText>{row.progress}</TableDataText>;
-      case 'cancel':
         return (
-          <Cancel
-            onClick={row.cancel}
-            ariaLabel={`Cancel upload for ${row.key}`}
-          />
+          <TableDataText>{`${Math.round(row.progress * 100)}%`}</TableDataText>
```
reesscot, line 183: nit: pull this out into a util function

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -1,9 +1,16 @@
...
+  uploadData,
+  UploadDataWithPathInput,
+  isCancelError,
+} from 'aws-amplify/storage';
 
 import { useGetLocationConfig } from '../../context/config';
 import { TaskStatus } from '../../context/types';
 
+// 5MB for multipart upload
+export const MULTIPART_UPLOAD_THRESHOLD_BYTES = 5242880;
```
reesscot, line 12: does this exact number match the one used by the JS library? If so, can we include a link?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -95,17 +107,25 @@ const mergeSelectedTasks = (
...
+  handleCancel: () => void,
 ] {
   const getConfig = useGetLocationConfig();
 
   const [tasks, setTasks] = React.useState<CancelableTask[]>(() => []);
+  const tasksRef = React.useRef<CancelableTask[]>(tasks);
+
+  React.useEffect(() => {
+    tasksRef.current = tasks;
+  }, [tasks]);
```
timngyn, line 128: for the life of me, I could not find any other way to get the inline cancel to work without this ref to the `tasks` array. Since we need to be able to cancel inline and also cancel all the tasks, I needed the most up to date status for each of the tasks and so the ref here was able to help with that

any feedback on this is appreciated

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -95,17 +107,25 @@ const mergeSelectedTasks = (
 export function useHandleUpload({
   prefix,
   preventOverwrite,
+  batchSize = 6,
```
reesscot, line 110: nit: should we put this default in a const?

reesscot (commented):
Just some small nits / questions

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: I didn't get it from the JS team, but now that you mention it I should do that. I just looked it up on their repo and they make it a lot easier to understand: https://github.com/aws-amplify/amplify-js/blob/1a5366d113c9af4ce994168653df3aadb142c581/packages/storage/src/providers/s3/utils/constants.ts#L17

I'll add this as a const somewhere

reesscot: cool, and might as well link to the JS repo as well, since they are the source of truth for when we can cancel or not.

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -95,17 +110,25 @@ const mergeSelectedTasks = (
...
+  batchSize = DEFAULT_BATCH_SIZE,
 }: {
   prefix: string;
   preventOverwrite: boolean;
+  batchSize?: number;
 }): [
   tasks: CancelableTask[],
   handleUpload: () => void,
   handleFileSelect: (files: File[]) => void,
+  handleCancel: () => void,
```
calebpollman, line 122: This cancels all right?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
calebpollman: What does storing `tasks` in a ref provide that usage of `tasks` directly does not?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +138,76 @@ export function useHandleUpload({
...
+      } finally {
+        // Start the next upload in the queue
+        uploadNext();
+      }
+    };
+
+    // Start the initial set of uploads
+    const initialUploads = Array(batchSize)
+      .fill(null)
+      .map(() => uploadNext());
```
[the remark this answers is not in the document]
timngyn: we can set `batchSize` to be 1 and just hardcode it so that tasks upload one at a time

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +152,148 @@ export function useHandleUpload({
...
 
-          setTasks((curr) =>
-            // assign cancel to noop
-            updateTasks(curr, { key, cancel: () => null, status: 'CANCELED' })
-          );
-        };
+    await Promise.all(initialUploads);
```
calebpollman, line 269: There is no reason to use `Promise.all` unless it is for tracking whether all tasks have completed or an exception has been reached. There is some nuance with the concurrency behavior we are looking for in this hook as it should start the next task as soon as a task completes, not once a btach completes.

Additionally believe that `Promise.all` would introduce bugs since it resolves on any exception, meaning that if task 1 of 6 throws then the `Promise.all` would reject while the other 5 tasks are still running. If awaiting a batch was the desired behavior, `Promise.allSettled` would be what we are looking for here

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +152,148 @@ export function useHandleUpload({
...
+  };
 
+  const handleCancel = () => {
```
calebpollman, line 272: We should just be able to iterate over the current `tasks` here and call `cancel` for the cancelable tasks without the additional logic i think?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +152,148 @@ export function useHandleUpload({
...
+        await result;
+
+        setTasks((prevTasks) =>
+          updateTasks(prevTasks, { key, status: 'COMPLETE' })
+        );
+      } catch (error) {
+        setTasks((prevTasks) =>
+          updateTasks(prevTasks, {
+            key,
+            status: isCancelError(error) ? 'CANCELED' : 'FAILED',
```
calebpollman, line 255: Does a cancel failure guarantee an actual upload failure?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +152,148 @@ export function useHandleUpload({
...
+        );
+      } finally {
+        uploadNext();
+      }
+    };
 
-        const handleCancel = () => {
-          cancel();
+    // Start the initial set of uploads
+    const initialUploads = Array(batchSize)
```
calebpollman, line 265: As discussed offline, this PR should cover a single operation at a time to get the base approach correctly before expanding to multiple concurrent operations

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +152,148 @@ export function useHandleUpload({
...
+
+      const pendingTask = processUpload(queuedTask);
+
+      currentIndex++;
+
+      const { task, result } = pendingTask;
+      const { key } = task;
+
+      try {
+        await result;
```
calebpollman, line 246: `uploadData` is a sync operation with a nested async `result`. Believe on the our call last week we used `result.then` in `processUpload` to determine whether to run the next upload?

calebpollman (commented):
Making progress! Let's sync on the approach if necessary, understand there's some complexities to the use case that might be proving difficult here

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: yes, `handleCancel` will cancel all tasks that are able to be canceled

calebpollman: `handleCancelAll`?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -20,6 +30,11 @@ export interface CancelableTask extends Omit<Task, 'status'> {
   status: TaskStatus | 'CANCELED';
 }
 
+interface PendingTask {
+  task: CancelableTask;
+  result: Promise<any>;
```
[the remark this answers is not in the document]
timngyn: I think the `result` I'm using here is coming from `uploadData(input)`, so it's not the actual task or anything like that

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
```diff
@@ -129,60 +152,148 @@ export function useHandleUpload({
...
+      ...task,
+      status: 'PENDING',
+      cancel: cancelFn,
+    };
+
+    setTasks((prevTasks) => updateTasks(prevTasks, pendingTask));
+
+    return {
+      task: pendingTask,
+      result,
```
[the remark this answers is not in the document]
timngyn: I think from our chat last time, we had a `process` function that was processing the task one at a time by calling `result.then(() => ...)`


With how I tried to set it up this time this was just to help separate out the logic for setting up the task and then I call `await result` in `uploadNext`

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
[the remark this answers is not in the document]
calebpollman: Why do we need it though? Interface changes need to well defined and explained

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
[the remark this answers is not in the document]
calebpollman: IIRC there was no longer a need to include `uploadNext` in the solution we had, and not seeing any benefits here?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: that's true, I could simplify the logic in here. I think we just need to check if `cancel` is the cancel function, right?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: It helps gives access to the most up to date state of the tasks. This helps with knowing when to to skip over a task that was canceled

Previously when I tried to access `tasks` directly, it was always showing the initial state when we first start the uploads

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
calebpollman: Think yes? Something like `tasks.forEach(({ cancel }) => isFunction(cancel) ? cancel() : undefined)` maybe?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
calebpollman: Believe that is because of the `Promise.all`?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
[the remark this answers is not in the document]
timngyn: oh gotcha, I think I added this because I was originally getting some typescript error for the return of my `processUpload` function and left it there. I just removed it and it looks like it's okay to remove

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: sounds good, will make that change

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: We would still need to update the state to show the tasks got canceled too right?

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/UploadControls.tsx
```diff
@@ -345,6 +357,16 @@ export const UploadControls = (): JSX.Element => {
       >
         Start
       </Primary>
+      <ButtonElement
+        variant="cancel"
+        disabled={queuedTasks === 0}
```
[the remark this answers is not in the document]
timngyn: ah gotcha that makes sense. will make that change

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: I think I was running into this issue when trying out the original method we discussed too. Since `setTasks` is react state, is it supposed to happen instantaneous?

By instantaneous, I wanted to be able to access the updated state within the next recursive call

packages/react-storage/src/components/StorageBrowser/Views/LocationActionView/useHandleUpload.ts
timngyn: sorry, what do you mean by this? could you phrase it a different way?

timngyn: if you're wondering about why this is in the `catch`, this is how the `cancel` method works from JS. Calling `cancel` causes the current request to throw an error and we need to check if the error was because it was "canceled" or if it was something else

packages/react-storage/src/components/StorageBrowser/components/ActionsMenu.tsx
```diff
@@ -22,16 +22,21 @@ export interface ActionItemProps extends ButtonElementProps {}
...
+export function ActionItem(
+  {
+    className = BUTTON_CLASS_NAME,
+    key,
+    variant = ACTION_ITEM_VARIANT,
+    role = 'menuitem',
+    ...props
+  }: ActionItemProps,
+  index: number
+): React.JSX.Element {
```
calebpollman, line 34: Fixes a missing React `key` warning

packages/react-storage/src/components/StorageBrowser/Views/Controls/Overwrite.tsx
```diff
@@ -12,16 +12,19 @@ export const LABEL_TEXT = 'Overwrite existing files';
 
 interface OverwriteControlProps {
   defaultChecked?: boolean;
+  disabled?: boolean;
```
calebpollman, line 15: Allow disabling of input toggle once upload has started

reesscot (dismissed):
LGTM

dbanksdesign (dismissed):

esauerbo (approved):
LGTM, just a note to address this https://github.com/aws-amplify/amplify-ui/pull/5708/files#r1742795493 in a followup if it's something we still want

reesscot (approved):

dbanksdesign (approved):

PR #5728, opened 2024-09-03
fix(storage-browser): single page pagination

hbuchel:
#### Description of changes

Currently if there is only one item to show, it will always show a pagination next button.

**Before**

https://github.com/user-attachments/assets/4fa295c2-c25b-40de-9508-ac758d1f28ea

**After**
Screenshot 2024-09-03 at 11 13 35 AM

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/Views/utils.ts
+++ b/packages/react-storage/src/components/StorageBrowser/Views/utils.ts
@@ -16,11 +16,11 @@ export const listViewHelpers = ({
   disableRefresh: boolean;
   range: [start: number, end: number];
 } => {
-  const isLastPage =
-    // Checks if we have more results than pageSize and we're on the last page.
-    Math.round(resultCount / pageSize) === currentPage ||
-    // Checks if we have less results than pageSize (only one page)
-    Math.round(resultCount / pageSize) === 0;
+  // Use Math.ceil so we can round up. For example, if you have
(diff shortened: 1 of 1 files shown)


packages/react-storage/src/components/StorageBrowser/Views/utils.ts
```diff
@@ -16,7 +16,9 @@ export const listViewHelpers = ({
   disableRefresh: boolean;
   range: [start: number, end: number];
 } => {
-  const isLastPage = Math.round(resultCount / pageSize) === currentPage;
+  const isLastPage =
+    Math.round(resultCount / pageSize) === currentPage ||
+    Math.round(resultCount / pageSize) === 0;
```
calebpollman, line 21: Think this makes sense but having a hard time reading the additional condition here as relating to the current page, since `0` isn't an allowed value of `currentPage`. Can we add a small comment explaining the logic?

hbuchel: Yes, let me take another look. I just realized there are some other scenarios to test as well.

hbuchel: I think I was overthinking it. I updated the existing logic to use `Math.ceil` instead of .round. I think then it should cover all the scenarios.

calebpollman (approved):
🚢

timngyn (approved):

PR #5729, opened 2024-09-03
chore(storage-browser): use plain text for current breadcrumb item

hbuchel:
#### Description of changes

Rendered a span with plain text for current Navigate item instead of a button

**After**
Screenshot 2024-09-03 at 1 14 05 PM

Screenshot 2024-09-03 at 11 44 54 AM

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
@@ -17,9 +17,7 @@ import { CLASS_BASE } from '../constants';
 
 import { LocationData, useAction } from '../../context/actions';
 
-interface NavigateItemProps extends ButtonElementProps {
-  seperator?: React.ReactNode;
-}
+interface NavigateItemProps extends ButtonElementProps {}
 
 type RenderNavigateItem = (props: NavigateItemProps) => React.JSX.Element;
 
@@ -45,7 +43,6 @@ function Separator() {
 export const NavigateItem = ({
   className = `${BLOCK_NAME}__button`,
   children,
-  seperator = null,
   ...props
 }: NavigateItemProps): React.JSX.Element => {
   return (
@@ -53,7 +50,7 @@ export const NavigateItem = ({
       <ButtonElement {...props} className={className} variant="navigate">
         {children}
       </ButtonElement>
-      {seperator}
+      <Separator />
     </ListItemElement>
   );
 };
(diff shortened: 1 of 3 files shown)


packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
```diff
@@ -17,9 +17,7 @@ import { CLASS_BASE } from '../constants';
 
 import { LocationData, useAction } from '../../context/actions';
 
-interface NavigateItemProps extends ButtonElementProps {
-  seperator?: React.ReactNode;
```
hbuchel, line 21: I removed this prop since with this change we can always render the `<Seperator />`

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
```diff
@@ -114,15 +110,20 @@ export function NavigateControl(): React.JSX.Element {
...
 
-        return (
+        return isCurrent ? (
+          <ListItemElement
+            className={`${BLOCK_NAME}__item`}
+            aria-current="page"
+          >
+            <SpanElement variant="navigate-current">{displayValue}</SpanElement>
+          </ListItemElement>
+        ) : (
```
calebpollman, line 120: This complicates the logic needed for having a single render item, the conditional logic for rendering plain text needs to be encapsulated to the `NavigateItem`

packages/react-storage/src/components/StorageBrowser/context/elements/defaults.tsx
```diff
@@ -125,6 +126,14 @@ function Heading(props: HeadingElementProps): React.JSX.Element {
   );
 }
 
+function Span(props: SpanElementProps): React.JSX.Element {
+  const { variant } = props;
+  if (variant === 'navigate-current') {
+    return <_View {...props} as="span" padding="xs" color="font.secondary" />;
```
calebpollman, line 132: The font size of the "current" item in the Amplify screenshot looks a bit larger than the button text, do we need to add anything here?

hbuchel: Good catch, fixed that in latest.

packages/react-storage/src/components/StorageBrowser/Views/Controls/Navigate.tsx
hbuchel: That makes sense. In that case, I updated NavigateItem to just have the prop "isCurrent", the it uses that to render out the correct markup/Separator.

timngyn (approved):

reesscot (approved):

PR #5764, opened 2024-09-06
fix(ai): fix return type for useAIGeneration

thaddmt:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-ai/src/hooks/useAIGeneration.tsx
+++ b/packages/react-ai/src/hooks/useAIGeneration.tsx
@@ -55,6 +55,9 @@ export function createUseAIGeneration<
     ): Promise<Schema[Key]['returnType']> => {
       const result = await handleGenerate(input);
 
+      // handleGenerate returns a Promised wrapper around Schema[Key]['returnType'] which includes data, errors, and clientExtensions
+      // The type of data is Schema[Key]['returnType'] which useDataState also wraps in a data return
+      // TODO: follow up with how to type handleGenerate to properly return the promise wrapper shape
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
       const data = (result as any).data as Schema[Key]['returnType'];
       // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
--- a/packages/react-ai/src/hooks/useAIGeneration.tsx
+++ b/packages/react-ai/src/hooks/useAIGeneration.tsx
@@ -61,7 +61,7 @@ export function createUseAIGeneration<
       _prev: Schema[Key]['returnType'],
       input: Schema[Key]['args']
     ): Promise<
-      Schema[Key]['returnType'] & { graphqlErrors: GraphQLFormattedError[] }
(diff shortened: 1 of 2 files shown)


packages/react-ai/src/hooks/useAIGeneration.tsx
```diff
@@ -55,7 +55,10 @@ export function createUseAIGeneration<
     ): Promise<Schema[Key]['returnType']> => {
       const result = await handleGenerate(input);
 
-      return { ...result };
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      const data = (result as any).data as Schema[Key]['returnType'];
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
+      return { ...data, ...(result as any).errors };
```
reesscot, line 61: Can you add comment on why this is needed?

dbanksdesign (dismissed):

packages/react-ai/src/hooks/useAIGeneration.tsx
calebpollman: Kind of confused what we are doing here, isn't this forcing a return value that doesn't match the shape of `Schema[Key]['returnType']`?

thaddmt: Discussed a bit offline, but the return type of the handleGenerate call actually wraps the `Schema[Key]['returnType']` in a Promise with graphqlErrors and some other V6Client related info, the data key of that return type matches the `Schema[Key]['returnType']`.

In terms of adding the errors, the graphql errors do not throw so it would not be caught by the useDataState layer. There is still a chance that the appsync layer could throw an error that would be caught, but we need a way to surface graphql errors to the customer as well

dbanksdesign: the issue this is fixing is that the type didn't match what is actually returned, and the type was the right DX, so this is massaging the actual data to match the type. 

CleanShot 2024-09-06 at 13 53 15@2x

calebpollman: @dbanksdesign `anything.anything` makes me feel crazy. Think the fix there is to return `data` from `handleGenerate`:
```diff
- const result = await handleGenerate(input);
+ const result = await handleGenerate(input).data;
```

Synced with @thaddmt offline abt the `errors` piece since we are moving them up a layer think that change requires us to update the return shape of `updateAIGenerationStateAction`:

```diff
- Promise<Schema[Key]['returnType']>
+ Promise<Schema[Key]['returnType'] & { graphqlErrors: string[] }>
```

Does that make sense?

calebpollman: Kinda have concerns abt the underlying types we're consuming from `Schema` here as AFAICT they do not match the shape returned from `handleGenerate` (obviously something we don't own and maybe i'm misunderstanding them)

dbanksdesign (approved):

calebpollman (approved):
🚢

PR #5782, opened 2024-09-10
chore(ai): add timestamp function to displayText

thaddmt:
#### Description of changes
- Adds displayText providers and hooks
- Adds a getMessageTimestamp function to displayText

#### Checklist

- [ ] PR description included

### DIFF
--- a/examples/next/pages/ui/components/ai/ai-conversation/index.page.tsx
+++ b/examples/next/pages/ui/components/ai/ai-conversation/index.page.tsx
@@ -13,6 +13,13 @@ const { useAIConversation } = createAIHooks(client);
 
 Amplify.configure(outputs);
 
+const formatDate = (date) =>
+  `Argh the time be round ${date.toLocaleTimeString('en-US', {
+    hour: 'numeric',
+    minute: 'numeric',
+    hour12: true,
+  })}`;
+
 function Chat() {
   const [
     {
@@ -25,6 +32,7 @@ function Chat() {
   return (
     <Card variation="outlined" width="50%" height="300px" margin="0 auto">
       <AIConversation
+        displayText={{ getMessageTimestampText: formatDate }}
         messages={messages}
         handleSendMessage={sendMessage}
         isLoading={isLoading}
--- a/examples/next/pages/ui/components/ai/ai-conversation/index.page.tsx
+++ b/examples/next/pages/ui/components/ai/ai-conversation/index.page.tsx
@@ -13,7 +13,7 @@ const { useAIConversation } = createAIHooks(client);
 
 Amplify.configure(outputs);
 
-const formatDate = (date) =>
+const formatDate = (date: Date): string =>
   `Argh the time be round ${date.toLocaleTimeString('en-US', {
     hour: 'numeric',
     minute: 'numeric',
--- a/packages/react-ai/src/components/AIConversation/context/index.ts
+++ b/packages/react-ai/src/components/AIConversation/context/index.ts
@@ -1,17 +1,13 @@
-import { ActionsContext } from './ActionsContext';
-import { AvatarsContext } from './AvatarsContext';
-import { ConversationInputContext } from './ConversationInputContext';
-import { MessagesContext, RoleContext } from './MessagesContext';
-import { SuggestedPromptsContext } from './SuggestedPromptsContext';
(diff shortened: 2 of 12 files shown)


packages/react-ai/src/components/AIConversation/context/index.ts
```diff
@@ -1,17 +1,13 @@
...
+export * from './AvatarsContext';
+export * from './ConversationInputContext';
+export * from './MessagesContext';
+export * from './SuggestedPromptsContext';
+export * from './MessageVariantContext';
+export * from './DisplayTextContext';
+export * from './ControlsContext';
+export * from './LoadingContext';
+export * from './ResponseComponentsContext';
+export * from './SendMessageContext';
```
calebpollman, line 11: This is a lot to be exporting from this directory, do we actually need to do this?

packages/react-ai/src/components/AIConversation/createProvider.tsx
```diff
@@ -42,32 +48,40 @@ export default function createProvider({
...
-                      </MessageVariantProvider>
-                    </ActionsProvider>
-                  </AvatarsProvider>
-                </SendMessageContextProvider>
-              </ConversationInputContextProvider>
-            </ResponseComponentsProvider>
-          </SuggestedPromptProvider>
-        </ControlsProvider>
-      </ElementsProvider>
+      <ConversationDisplayTextProvider {..._displayText}>
```
calebpollman, line 56: The display text provider should be nested much lower here, ideally below any "static" providers (e.g. `ElementsProvider`

packages/react-ai/src/components/AIConversation/context/index.ts
thaddmt: We could keep it as explicit exports for each one, in most cases from each of the files we want to export both the `Context` and the `Provider` so I figured this was less verbose, what do you think?

packages/react-ai/src/components/AIConversation/createProvider.tsx
thaddmt: Hm can you explain why? I believe you but just curious what is the reasoning

packages/react-ai/src/components/AIConversation/context/index.ts
calebpollman: My main concern is that having "export all" statements can lead to internal exports bleeding out unintentionally (see all the deprecated UI primitive interfaces in ui-react for an example). Out of scope for this PR, but is there any way we can consolidate some of these in to scoped providers?

packages/react-ai/src/components/AIConversation/createProvider.tsx
calebpollman: Yeah, basically we want any provider that maintains a stable reference to be on the "outside" since they will never cause a re-render

packages/react-ai/src/components/AIConversation/context/index.ts
thaddmt: Sure, I can revert this change, mostly changed it out of making it look cleaner but that makes sense.

And to your second point, yeah I think that makes sense and can make the `createProvider` cleaner

packages/react-ai/src/components/AIConversation/createProvider.tsx
thaddmt: Ah yes that makes a lot of sense

examples/next/pages/ui/components/ai/ai-conversation/index.page.tsx
```diff
@@ -13,6 +13,13 @@ const { useAIConversation } = createAIHooks(client);
 
 Amplify.configure(outputs);
 
+const formatDate = (date) =>
```
dbanksdesign, line 16: Non-blocking, could we type this function to make sure the types are working as expected?

dbanksdesign (dismissed):

examples/next/pages/ui/components/ai/ai-conversation/index.page.tsx
thaddmt: yep, will add

calebpollman (approved):

dbanksdesign (approved):

PR #5793, opened 2024-09-11
fix(e2e): fix placeholder matching rn e2e android

thaddmt:
#### Description of changes
- Noticed when running e2e RN android tests locally this placeholder test would consistently fail
- Updated to match by testId
- Updated to check attribute on element
- Example failure screenshot below to see what throwing looking like

Screenshot 2024-09-11 at 12 08 22 PM

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/e2e/detox/integration/common/shared.ts
+++ b/packages/e2e/detox/integration/common/shared.ts
@@ -93,7 +93,18 @@ Then(
 );
 
 Then('I see placeholder {string}', async (placeholder: string) => {
-  await expect(element(by.text(placeholder))).toExist();
+  const emailField = element(
+    by.id('authenticator__text-field__input-username')
+  );
+  await expect(emailField).toExist();
+  const attributes = await emailField.getAttributes();
+  if ((attributes as Detox.ElementAttributes).placeholder) {
+    if ((attributes as any).placeholder !== placeholder) {
+      throw new Error(
+        `No element found with placeholder value: ${placeholder}`
+      );
+    }
+  }
 });
 
 Then(


calebpollman (approved):
Nice!

packages/e2e/detox/integration/common/shared.ts
```diff
@@ -93,7 +93,18 @@ Then(
 );
 
 Then('I see placeholder {string}', async (placeholder: string) => {
-  await expect(element(by.text(placeholder))).toExist();
+  const emailField = element(
+    by.id('authenticator__text-field__input-username')
+  );
+  await expect(emailField).toExist();
+  const attributes = await emailField.getAttributes();
+  if ((attributes as Detox.ElementAttributes).placeholder) {
```
jordanvn, line 101: Do we want to take any action if there is no `attributes.placeholder`?

dbanksdesign (approved):

packages/e2e/detox/integration/common/shared.ts
thaddmt: I think right now I'm not sure how we want to handle it because it seems like it isn't always `placeholder` based on the platform? I saw in some cases it could be `hint` or other options instead

jordanvn: Ah, ok. So is the idea then that we're just covering those expected cases of `placeholder` here, and we can possibly expand to the other platforms in the future?

calebpollman: The detox utils are specific to React Native

jordanvn (approved):

PR #5796, opened 2024-09-12
chore(deps): Fix transitive dependencies

esauerbo:
#### Description of changes
- Add eslint `import/no-extraneous-dependencies` rule which will error if an import is used that isn't included in the module's `package.json`
- Ignore testing files so only dependencies that will be shipped to customer code are checked
- Update dependencies for packages where transitive dependencies were being included

Manually went through each package and removed a dependency and verified that eslint threw an error

#### Checklist

- [x] PR description included


calebpollman (changes requested):
Let's add the linting rule to the base eslint config that the rest of the packages extend from and configure ignore patterns for devDeps in test files there

packages/react/package.json
```diff
@@ -55,6 +55,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@aws-amplify/core": "^6.2.1",
```
esauerbo, line 58:
```suggestion
    "@aws-amplify/core": "6.4.0",
```

packages/react-core-auth/package.json
```diff
@@ -33,10 +33,12 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@aws-amplify/core": "6.4.0",
```
esauerbo, line 36: Wasn't sure what version to use, went with 6.4.0 because it's the latest that would be used by the  `"aws-amplify": "^6.3.2"` peer dep

packages/angular/package.json
```diff
@@ -9,7 +9,8 @@
     "build": "ng build --configuration production",
     "test": "jest",
     "test:watch": "yarn test --watch",
-    "lint": "ng lint"
+    "lint": "yarn typecheck && eslint . --ext .js,.jsx,.ts,.tsx",
```
jordanvn, line 12: Are `lint` scripts currently used in our pipeline?

esauerbo: Yeah it's part of the unit testing which gets run when testing PRs and publishing to @next

jordanvn: Great! Makes sense

jordanvn (dismissed):
These changes look good to me. Verified that it works locally as expected as well

packages/configs/eslint/configs/jest.js
```diff
@@ -1,10 +1,11 @@
 module.exports = {
-  plugins: ['@typescript-eslint', 'jest'],
+  plugins: ['@typescript-eslint', 'import', 'jest'],
   extends: ['plugin:jest/recommended'],
   rules: {
     // turn off for test file linting
     '@typescript-eslint/no-unsafe-assignment': 'off',
-
+    // ignore dev dependencies used in test files
+    'import/no-extraneous-dependencies': 'off',
```
calebpollman, line 8: Let's configure the base rule to use a glob pattern ignore devDeps on matches against test files or use the `packageDir` option to specify the monorepo root _package.json_ as well as the current directory _package.json_, e.g.:
```
'import/no-extraneous-dependencies': ['error', { packageDir: ['.', '../..'] },
```

packages/angular/package.json
calebpollman, line 12: Good with adding `typecheck` here but let's continue using the built in `ng lint` script. Angular has a lot going on under the hood ("magic") that we don't have awareness in to

packages/angular/.eslintrc.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   root: true,
+  ignorePatterns: ['.eslintrc.js', 'coverage', 'dist'],
```
calebpollman, line 3: Why do we need this now?

packages/angular/.eslintrc.js
```diff
@@ -39,7 +40,12 @@ module.exports = {
     },
     {
       extends: ['@aws-amplify/amplify-ui/jest'],
-      files: ['**/__mocks__/**', '**/__tests__/**'],
+      files: [
+        '**/__mocks__/**',
+        '**/__tests__/**',
+        '**/*.spec.ts',
+        '**/*.jest.ts',
```
calebpollman, line 47: Let's reconsider the approach to avoid adding these two lines. `'**/*.spec.ts'` is covered by the existing glob patterns (and if not we should figure out why not), and `'**/*.jest.ts'` only would capture _setup.jest.ts_ which is a config file that should be linted the same as other config files

packages/angular/projects/ui-angular/ng-package.json
```diff
@@ -6,9 +6,11 @@
     "entryFile": "src/index.ts"
   },
   "allowedNonPeerDependencies": [
```
calebpollman, line 8: Can you explain what listing deps here does?

packages/angular/package.json
```diff
@@ -21,7 +22,6 @@
     "@angular-eslint/template-parser": "~14.4.0",
     "@angular/animations": "^14.3.0",
     "@angular/cli": "^14.2.13",
-    "@angular/common": "^14.3.0",
```
calebpollman, line 24: Removing `"@angular/common"` (and `"rxjs"` below) doesn't feel quite right to me. Can you explain why we would remove them?

packages/angular/projects/ui-angular/package.json
```diff
@@ -18,8 +18,10 @@
     "aws-amplify": "^6.3.2"
   },
   "dependencies": {
+    "@angular/common": "^14.3.0",
     "@aws-amplify/ui": "6.4.1",
     "nanoid": "3.1.31",
+    "rxjs": "^7.8.1",
```
calebpollman, line 24: Guessing this is why we are removing the devDeps for these two but are we sure that won't cause conflict with the consuming projects deps?

packages/configs/eslint/configs/index.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
     'generator-star-spacing': 'off',
     'global-require': 'off',
     'implicit-arrow-linebreak': 'off',
+    'import/no-extraneous-dependencies': 'error',
```
calebpollman, line 48: See comment [here](https://github.com/aws-amplify/amplify-ui/pull/5796/files#diff-e4bfb948c8aea5491665b8d85669f7ca531065cd0f370ad7dd817fe5a9d882adR8)

packages/react-liveness/package.json
```diff
@@ -55,6 +56,7 @@
     "@smithy/fetch-http-handler": "^2.1.3",
     "@smithy/protocol-http": "^3.0.3",
     "@smithy/signature-v4": "2.1.4",
+    "@smithy/types": "^3.3.0",
```
calebpollman, line 59: Type packages should not be included as direct (or peer) dependencies

packages/react-native/package.json
```diff
@@ -36,6 +36,7 @@
   },
   "peerDependencies": {
     "aws-amplify": "^6.3.2",
+    "react": "^16.14.0 || ^17.0 || ^18.0",
```
calebpollman, line 39: Have we validated that our supported versions of RN work with all of these React versions?

packages/ui/package.json
```diff
@@ -37,13 +37,14 @@
     "check:esm": "node --input-type=module --eval 'import \"@aws-amplify/ui\"'",
     "clean": "rimraf dist node_modules",
     "dev": "yarn build --watch",
-    "lint": "yarn typecheck",
+    "lint": "yarn typecheck && eslint src --ext .js,.ts,.tsx",
```
calebpollman, line 40: I don't believe the UI package uses the internal linting config, meaning it would not be picking up the `import` rule updates. Can you confirm?

packages/react/src/primitives/Card/__tests__/Card.test.tsx
calebpollman, line 1: Nice!

calebpollman (commented):
Great progress. Left some feedback on the rule configuration/angular changes and some general questions. 

Additionally have some notes on the `aws-amplify` related scoped `"@aws-amplify/*"`deps: 
- Should always be peer deps as installing them as direct deps may lead to conflicts with the consuming project's `aws-amplify` version, potentially resulting in multiple versions of the singleton getting installed in the consuming project
- The listed versions of `aws-amplify` related scoped `"@aws-amplify/*"` peer deps need to match the version specified in the peer dep version of `aws-amplify`. For example if we have a peer dep of `aws-amplify@4.2` and need to specify `@aws-amplify/core` as a peer dep, we need to look at what version of `@aws-amplify/core` is used by `aws-amplify@4.2` and set our peer dep accordingly

packages/angular/.eslintrc.js
esauerbo: https://github.com/aws-amplify/amplify-ui/blob/main/packages/angular/projects/ui-angular/src/lib/components/authenticator/components/setup-totp/setup-totp.component.spec.ts isn't covered by the existing glob pattern

packages/angular/projects/ui-angular/ng-package.json
esauerbo: "To mitigate the risk of accidentally publishing libraries with dependencies" they make you add dependencies here as well https://github.com/ng-packagr/ng-packagr/blob/main/docs/dependencies.md#allowing-in-the-dependencies-section

packages/react-liveness/package.json
esauerbo: Can we turn off this rule for that file then? this import is used outside of testing

packages/angular/projects/ui-angular/package.json
esauerbo: Just verified that these packages aren't dependencies of angular 14. In that case it's probably safe to just go with the version we were using as dev deps right? 

`@angular/common` isn't a dependency of any of the packages we're using actually, so it does seem weird that it wasn't causing problems given it's imported in client code.

packages/ui/package.json
esauerbo: Correct it's not using the internal linting config. I added the rule separately to the ui package's config.

packages/react-liveness/package.json
esauerbo: Discussed offline and added a types file for these imports with an eslint disable.

packages/configs/eslint/configs/jest.js
esauerbo: Updated. `yarn lint` works as expected but I'm seeing `The package.json file could not be found.` with VScode, not sure if that's just a problem with my machine though

packages/angular/.eslintrc.js
esauerbo: Removed

packages/react-native/package.json
esauerbo: Updated this to use `"*"`

packages/angular/.eslintrc.js
```diff
@@ -39,7 +39,22 @@ module.exports = {
     },
     {
       extends: ['@aws-amplify/amplify-ui/jest'],
-      files: ['**/__mocks__/**', '**/__tests__/**'],
+      files: [
+        '**/__mocks__/**',
+        '**/__tests__/**',
```
calebpollman, line 44: Since these lines don't do match any patterns can remove

packages/angular/.eslintrc.js
```diff
@@ -39,7 +39,22 @@ module.exports = {
...
+        './projects/ui-angular/src/lib/components/authenticator/components/setup-totp/setup-totp.component.spec.ts',
```
calebpollman, line 45: Small comment explaining this would be good for future context

packages/react-ai/package.json
```diff
@@ -42,11 +42,12 @@
     "typecheck": "tsc --noEmit"
   },
   "peerDependencies": {
-    "aws-amplify": "^6.3.2",
+    "aws-amplify": "^6.6.0",
     "react": "^16.14.0 || ^17.0 || ^18.0",
     "react-dom": "^16.14.0 || ^17.0 || ^18.0"
   },
   "dependencies": {
+    "@aws-amplify/api-graphql": "^4.1.11",
```
calebpollman, line 50: Should be a peer dep

packages/react-core/package.json
```diff
@@ -46,8 +46,11 @@
     "react-hook-form": "^7.43.5",
     "xstate": "^4.33.6"
   },
+  "devDependencies": {
+    "@aws-amplify/core": "^6.4.0"
+  },
```
calebpollman, line 51: Let's list in the top level devDeps with the rest of the `aws-amplify` related scoped packages

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/createStreamingClient/types.ts
```diff
@@ -0,0 +1,16 @@
...
+  RequestPresigningArguments,
+} from '@smithy/types';
+
+export {
+  HttpRequest,
+  Provider,
+  RequestHandler,
+  RequestHandlerMetadata,
+  RequestPresigningArguments,
+};
```
calebpollman, line 16:
```suggestion
export {
  HttpRequest,
  Provider,
  RequestHandler,
  RequestHandlerMetadata,
  RequestPresigningArguments,
} from '@smithy/types';
```

packages/react-native/package.json
```diff
@@ -35,7 +35,9 @@
     "style-dictionary": "3.9.1"
   },
   "peerDependencies": {
-    "aws-amplify": "^6.3.2",
+    "@react-native-async-storage/async-storage": "*",
```
calebpollman, line 38: This only used for a mock that is not bundled in _dist_, can it be listed as a dev dependency?

packages/react-storage/package.json
```diff
@@ -46,7 +46,8 @@
     "tslib": "^2.5.2"
   },
   "peerDependencies": {
-    "aws-amplify": "^6.3.2",
+    "@aws-amplify/core": "^6.4.0",
```
calebpollman, line 49: We're only using `@aws-amplify/core` in the storage package to reference a deprecated type (`StorageAccessLevel`), can we manually recreate the type and remove the peer dep? Can be done in a follow up

yarn.lock
```diff
@@ -482,45 +482,45 @@
...
+    "@aws-sdk/client-kinesis" "3.621.0"
+    "@aws-sdk/client-personalize-events" "3.621.0"
     "@smithy/util-utf8" "2.0.0"
     tslib "^2.5.0"
 
-"@aws-amplify/api-graphql@4.1.11":
-  version "4.1.11"
-  resolved "https://registry.yarnpkg.com/@aws-amplify/api-graphql/-/api-graphql-4.1.11.tgz#cf197cfc0105ac6f387dbfba2d0eed0ba88384dc"
-  integrity sha512-RAUG6y2VrTVjIQnOGgLS5e77XAtLDdf1d4Lj4oScN1dEp+38DyGgbyVew9AQpmD9f+KvpHkM/R0dHYFTSD8/JA==
+"@aws-amplify/api-graphql@4.3.0", "@aws-amplify/api-graphql@^4.1.11":
```
calebpollman, line 496: Let's make sure we only have one version specified in the repo

calebpollman (commented):
Almost there! Most of the feedback can be addressed in a follow up. only concern for getting merged is the `@aws-amplify/api-graphql` peer dep, left specific feedback in the code

packages/angular/.eslintrc.js
esauerbo: removed `__mocks__` but actually there are some `__tests__` in the angular package

packages/angular/.eslintrc.js
esauerbo: Added

packages/react-ai/package.json
esauerbo: Oh good catch, moved to peer dep

packages/react-core/package.json
esauerbo: Moved

packages/react-storage/package.json
esauerbo: Yeah will do this in a followup

packages/react-ai/package.json
```diff
@@ -42,7 +42,8 @@
     "typecheck": "tsc --noEmit"
   },
   "peerDependencies": {
-    "aws-amplify": "^6.3.2",
+    "@aws-amplify/api-graphql": "4.3.0",
```
calebpollman, line 45:
```suggestion
    "@aws-amplify/api-graphql": "^4.3.0",
```

yarn.lock
```diff
@@ -665,10 +665,24 @@
...
+    "@aws-sdk/types" "3.398.0"
+    "@smithy/util-hex-encoding" "2.0.0"
+    "@types/uuid" "^9.0.0"
+    js-cookie "^3.0.5"
+    rxjs "^7.8.1"
+    tslib "^2.5.0"
+    uuid "^9.0.0"
+
+"@aws-amplify/core@^6.4.0":
+  version "6.4.1"
```
calebpollman, line 683: Let's figure out what we need to do make sure there is only one version of `@aws-amplify/core` installed

esauerbo: `aws-amplify@6.6.0` uses the exact version `@aws-amplify/core@6.4.0`. 
Since our `aws-amplify` peerDep uses version ^6.6.0 we would need to either specify an exact version for both or an exact version for `@aws-amplify/core`. Not sure which is correct in this scenario

packages/react-native/package.json
esauerbo: It's listed as a devDependency for `@aws-amplify/react-native`, in that case is our documentation incorrect for telling people to install it separately?

packages/react-ai/package.json
esauerbo: Same question as https://github.com/aws-amplify/amplify-ui/pull/5796#discussion_r1763639952, in the main package.json we specify an exact version of `aws-amplify` which in turn uses exact versions of `@aws-amplify/api-graphql` and `@aws-amplify/api-core`, but our peer dep of `aws-amplify` uses a range of versions. So it seems like we would end up with multiple versions whenever `aws-amplify` updates. 

Should we update our devDep in the main package.json to use a range instead?

yarn.lock
esauerbo: Discussed offline and updated `aws-amplify` dep in top-level package.json to use a relative version

packages/react-ai/package.json
esauerbo: Discussed offline and updated `aws-amplify` dep in top-level package.json to use a relative version

jordanvn (approved):
This looks good to me. I believe I understand the change, but I would still like to see @calebpollman's remaining question answered

packages/angular/package.json
esauerbo: See https://github.com/aws-amplify/amplify-ui/pull/5796#discussion_r1759233733

reesscot (approved):

packages/vue/.eslintrc.cjs
```diff
@@ -19,6 +20,12 @@ module.exports = {
   },
   ignorePatterns: ['dist'],
   rules: {
+    'import/no-extraneous-dependencies': [
```
calebpollman, line 23: Nce catch!

calebpollman (approved):
Smoke tested locally and everything is working as expected, thanks for the hard work here @esauerbo

jordanvn (approved):

timngyn (approved):

PR #5798, opened 2024-09-12
fix(liveness): use video stream dimensions instead of videoEl dimensions

thaddmt:
#### Description of changes
- In some very specific cases when flipping the phones orientation many times before attempting a check there seems to have been a mismatch between the video dimensions and the dimensions we send to the rekognition service which would cause a distorted audit image
- Tested this by sending flipped dimensions to the rekognition service and got a distorted image
- Will need to test this extensively since I believe we initially used the layout width and height because some browser / device combinations would return flipped width and height from `getUserMedia`

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
@@ -60,7 +60,7 @@ export class LivenessStreamProvider {
   private _reader!: ReadableStreamDefaultReader;
   private videoEl: HTMLVideoElement;
   private _client!: RekognitionStreamingClient;
-  private _stream: MediaStream;
+  private stream: MediaStream;
   private initPromise: Promise<void>;
 
   constructor({
@@ -74,7 +74,7 @@ export class LivenessStreamProvider {
   }: StreamProviderArgs) {
     this.sessionId = sessionId;
     this.region = region;
-    this._stream = stream;
+    this.stream = stream;
     this.videoEl = videoEl;
     this.videoRecorder = new VideoRecorder(stream);
     this.credentialProvider = credentialProvider;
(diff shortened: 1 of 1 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
```diff
@@ -198,13 +198,15 @@ export class LivenessStreamProvider {
...
+
     const response = await this._client.send(
       new StartFaceLivenessSessionCommand({
         ChallengeVersions: 'FaceMovementAndLightChallenge_1.0.0',
         SessionId: this.sessionId,
         LivenessRequestStream: livenessRequestGenerator,
-        VideoWidth: this.videoEl.videoWidth.toString(),
-        VideoHeight: this.videoEl.videoHeight.toString(),
+        VideoWidth: (mediaSettings.width ?? this.videoEl.width).toString(),
+        VideoHeight: (mediaSettings.height ?? this.videoEl.height).toString(),
```
calebpollman, line 209: Would it make sense to resolve both height/width from one source or does it not matter? e.g
```ts
const getVideoDimensions = (mediaSettings, videoElement) => {
  if (!!mediaSettings.height && !!mediaSettings.width) {
    // return mediaSettings values
  }

  // return videoElement values
}
```

calebpollman (approved):

jordanvn (approved):

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/streamProvider.ts
thaddmt: Think it doesnt matter since it should usually have both but this is probably a safer change to make, I believe we're going to test this change on alpha first before proceding though

PR #5799, opened 2024-09-13
fix(ai): fix renders for empty toolUse messages

thaddmt:
#### Description of changes
- Developer can override the `sendMessage` utility to send messages with tools that are not renderable in the AIConversation component example below:
- Updates all of our response components to be prefixed with `UI_` to be easier to tell as response component tool uses rather than regular tool uses
- Filters out messages that are not renderable, i.e only render empty tool use blocks
```
handleSendMessage={(content) =>
          sendMessage({
            ...content,
            toolConfiguration: {
              tools: {
                generateRecipe: {
                  description: "List ingredients needed for a recipe",
                  inputSchema: {
                    json: {
                      type: "object",
                      properties: {
                        ingredients: {
                          type: "array",
                          items: {
                            type: "object",
                            properties: {
                              ingredientName: { type: "string" },
                              quantity: { type: "number" },
                              unit: { type: "string" },
                            }
                          },
                        },
                      },
                    },
                  },
                },
              }
            }
          })
        }
```

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
+++ b/packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
@@ -27,10 +27,13 @@ export const ResponseComponentsProvider = ({
   children?: React.ReactNode;
   responseComponents?: ResponseComponents;
 }): JSX.Element => {
+  const _responseComponents = React.useMemo(
+    () => prependResponseComponents(responseComponents),
+    [responseComponents]
+  );
+
   return (
-    <ResponseComponentsContext.Provider
-      value={prependResponseComponents(responseComponents)}
-    >
+    <ResponseComponentsContext.Provider value={_responseComponents}>
       {children}
     </ResponseComponentsContext.Provider>
   );
--- a/packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
+++ b/packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
@@ -4,7 +4,7 @@ import { ToolConfiguration } from '../../../types';
 
 type ResponseComponentsContextProps = ResponseComponents | undefined;
 
-export const RESPONSE_COMPONENT_PREFIX = 'UI_';
+export const RESPONSE_COMPONENT_PREFIX = 'AMPLIFY_UI_';
 
 export const ResponseComponentsContext =
   React.createContext<ResponseComponentsContextProps>(undefined);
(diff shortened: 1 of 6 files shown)


packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
```diff
@@ -4,9 +4,22 @@ import { ToolConfiguration } from '../../../types';
 
 type ResponseComponentsContextProps = ResponseComponents | undefined;
 
+export const RESPONSE_COMPONENT_PREFIX = 'UI_';
+
 export const ResponseComponentsContext =
   React.createContext<ResponseComponentsContextProps>(undefined);
 
+const prependResponseComponents = (responseComponents?: ResponseComponents) => {
```
thaddmt, line 12: This makes it so that we have to run this utility every time responseComponents changes, which responseComponents is set on the `createAIConversation` so it shouldnt be changing much if at all

dbanksdesign (approved):

packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
```diff
@@ -4,9 +4,22 @@ import { ToolConfiguration } from '../../../types';
 
 type ResponseComponentsContextProps = ResponseComponents | undefined;
 
+export const RESPONSE_COMPONENT_PREFIX = 'UI_';
```
calebpollman, line 7: Probably overkill and something i usually nitpick, but should we make this more specific by prefixing `AMPLIFY_` as well?

packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
```diff
@@ -15,7 +28,9 @@ export const ResponseComponentsProvider = ({
   responseComponents?: ResponseComponents;
 }): JSX.Element => {
   return (
-    <ResponseComponentsContext.Provider value={responseComponents}>
+    <ResponseComponentsContext.Provider
+      value={prependResponseComponents(responseComponents)}
```
calebpollman, line 32: This will create a new reference on each render, regardless of whether `responseComments` changes. Tbh kind of surprised that the React linting rules aren't complaining

packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
thaddmt: Sure, happy to add

packages/react-ai/src/components/AIConversation/context/ResponseComponentsContext.tsx
thaddmt: Yeah I guess I'm used to our linting rules complaining abou this if you send in an object like `{foo, bar}` so when this didn't complain i was a little unsure lol

calebpollman (approved):

reesscot (approved):

PR #5812, opened 2024-09-18
feat(theming) add custom component style rendering

dbanksdesign:
#### Description of changes

This is a new feature of the experimental theme v2 work. 

We have come across a use-case where a customer wants to create a custom component theme in a separate package, but cannot add that theme to the `createTheme` function due to a circular dependency. This functionality allows customers to just write the CSS that only their component needs, whenever they need to. 

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/src/components/ThemeProvider/ComponentStyle.tsx
+++ b/packages/react/src/components/ThemeProvider/ComponentStyle.tsx
@@ -1,0 +1,53 @@
+import * as React from 'react';
+import { WebTheme, createComponentCSS } from '@aws-amplify/ui';
+import {
+  BaseComponentProps,
+  ElementType,
+  ForwardRefPrimitive,
+  Primitive,
+  PrimitiveProps,
+} from '../../primitives/types';
+import { primitiveWithForwardRef } from '../../primitives/utils/primitiveWithForwardRef';
+import { BaseComponentTheme } from '@aws-amplify/ui';
+import { Style } from './Style';
+
+interface BaseComponentStyleProps extends BaseComponentProps {
+  /**
+   * Provide a server generated nonce which matches your CSP `style-src` rule.
+   * This will be attached to the generated <style> tag.
+   * @see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src
+   */
+  nonce?: string;
+  theme: Pick<WebTheme, 'name' | 'breakpoints' | 'tokens'>;
+  componentThemes: BaseComponentTheme[];
+}
+
+export type ComponentStyleProps<Element extends ElementType = 'style'> =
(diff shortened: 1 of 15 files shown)


packages/react/src/components/ThemeProvider/ComponentStyle.tsx
```diff
@@ -0,0 +1,53 @@
...
+  nonce?: string;
+  theme: Pick<WebTheme, 'name' | 'breakpoints' | 'tokens'>;
+  componentThemes: BaseComponentTheme[];
+}
+
+export type ComponentStyleProps<Element extends ElementType = 'style'> =
+  PrimitiveProps<BaseComponentStyleProps, Element>;
+
+const ComponentStylePrimitive: Primitive<ComponentStyleProps, 'style'> = (
+  { theme, componentThemes = [], nonce, ...rest },
```
calebpollman, line 29: If we aren't referencing `nonce` in the body below, it can just be kept inside `rest` i think?

packages/ui/src/theme/createTheme/createComponentCSS.ts
```diff
@@ -58,17 +58,21 @@ function recursiveComponentCSS(baseSelector: string, theme: BaseTheme) {
   return str;
 }
 
+interface CreateComponentCSSProps {
```
calebpollman, line 61: Nit: `Props` is a React specific naming construct and may lead to consumer confusion when using util functions that are not React components, for functions interfaces prefer `Input` or `Params`

packages/react/src/components/ThemeProvider/ComponentStyle.tsx
dbanksdesign: Right you are

packages/ui/src/theme/createTheme/createComponentCSS.ts
dbanksdesign: Good catch

packages/react/src/components/ThemeProvider/Style.tsx
```diff
@@ -0,0 +1,93 @@
...
+  } else {
+    return (
+      <style
+        {...rest}
+        ref={ref}
+        // eslint-disable-next-line react/no-danger
+        dangerouslySetInnerHTML={{ __html: cssText }}
+      />
+    );
+  }
```
calebpollman, line 83:
```suggestion
  if (cssText === undefined || /<\/style/i.test(cssText)) {
    return null;
  }
  
  return (
    <style
      {...rest}
      ref={ref}
      // eslint-disable-next-line react/no-danger
      dangerouslySetInnerHTML={{ __html: cssText }}
    />
  );
```

calebpollman (approved):

calebpollman (approved):

esauerbo (approved):

PR #5816, opened 2024-09-19
fix(ai): update useAIGeneration to add graphqlErrors before return

thaddmt:
#### Description of changes
- Danny noticed an error where simple string type returns from the client were returning malformed objects
- This is because we were assuming all return types from the generate client would be spreadable objects
- Updated called to useDataState to just return the data and massage the return type to add graphqlErrors after to avoid having to spread the return data

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-ai/src/hooks/useAIGeneration.tsx
+++ b/packages/react-ai/src/hooks/useAIGeneration.tsx
@@ -1,7 +1,6 @@
 import { DataState, useDataState } from '@aws-amplify/ui-react-core';
 import { V6Client } from '@aws-amplify/api-graphql';
 import { getSchema } from '../types';
-import React from 'react';
 
 export interface UseAIGenerationHookWrapper<
   Key extends keyof AIGenerationClient<Schema>['generations'],
@@ -59,10 +58,6 @@ export function createUseAIGeneration<
     state: GenerateState<Schema[Key]['returnType']>,
     handleAction: (input: Schema[Key]['args']) => void,
   ] => {
-    const [graphqlErrors, setGraphqlErrors] = React.useState<
-      GraphQLFormattedError[] | undefined
-    >(undefined);
-
     const handleGenerate = (
       client.generations as AIGenerationClient<Schema>['generations']
     )[routeName];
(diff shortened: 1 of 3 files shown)


dbanksdesign (approved):

calebpollman (commented):
The issue here reads as misuse of `useDataState` and not necessarily an issue with `useDataState` itself. Adding `graphqlErrors` to the `DataState` output itself can be done by massaging the output of `useDataState` before returning from `useAIGeneration`

thaddmt:
> The issue here reads as misuse of `useDataState` and not necessarily an issue with `useDataState` itself. Adding `graphqlErrors` to the `DataState` output itself can be done by massaging the output of `useDataState` before returning from `useAIGeneration`

Ended up going with your solution, great call out, thanks for catching

dbanksdesign (dismissed):

packages/react-ai/src/hooks/useAIGeneration.tsx
```diff
@@ -45,31 +55,37 @@ export function createUseAIGeneration<
...
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return data;
     };
 
-    return useDataState(updateAIGenerationStateAction, {});
+    const [data, handler] = useDataState(
+      updateAIGenerationStateAction,
+      undefined
+    );
+    return [{ ...data, graphqlErrors }, handler];
```
calebpollman, line 88: Can simplify a bit and remove the `graphqlErrors` state by returning `result` directly from `updateAIGenerationStateAction` and restructuring the hook return output, something like:
```ts
    const { data, errors } =
      (result?.data as SingularReturnValue<Schema[Key]['returnType']>) ?? {};

    const output: [
      state: GenerateState<Schema[Key]['returnType']>,
      handleAction: (input: Schema[Key]['args']) => void,
    ] = [{ ...result, data, graphqlErrors: errors }, handler];
```

thaddmt: Ah damn this is much cleaner, let me try that and ping you :)

dbanksdesign (approved):

calebpollman (approved):

PR #5867, opened 2024-10-04
chore: Recreate StorageAccessLevel type and remove amplify/core peerDep from…

esauerbo:
… storage package

#### Description of changes
Since StorageAccessLevel is a deprecated type we can just copy it and remove a dependency 

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-storage/src/components/FileUploader/utils/getInput.ts
+++ b/packages/react-storage/src/components/FileUploader/utils/getInput.ts
@@ -1,5 +1,4 @@
 import { fetchAuthSession } from 'aws-amplify/auth';
-import { StorageAccessLevel } from '@aws-amplify/core';
 import { UploadDataWithPathInput, UploadDataInput } from 'aws-amplify/storage';
 
 import { isString, isFunction } from '@aws-amplify/ui';
@@ -7,6 +6,8 @@ import { isString, isFunction } from '@aws-amplify/ui';
 import { ProcessFile } from '../types';
 import { resolveFile } from './resolveFile';
 import { PathCallback, PathInput } from './uploadFile';
+
+type StorageAccessLevel = 'guest' | 'protected' | 'private';
 
 export interface GetInputParams {
   accessLevel: StorageAccessLevel | undefined;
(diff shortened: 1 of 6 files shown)


packages/react-storage/src/components/FileUploader/utils/getInput.ts
```diff
@@ -8,6 +7,8 @@ import { ProcessFile } from '../types';
 import { resolveFile } from './resolveFile';
 import { PathCallback, PathInput } from './uploadFile';
 
+type StorageAccessLevel = 'guest' | 'protected' | 'private';
```
jordanvn, line 10: Nit: would it be better to declare StorageAccessLevel here or import it from `FileUploader/types`?

jordanvn (dismissed):
This makes sense to me. Confirmed that this is the only usage within ui-react-storage

packages/react-storage/src/components/FileUploader/utils/getInput.ts
esauerbo: Yeah I think it would make sense to import it from the types file, updated to do that.

packages/react-storage/src/components/StorageManager/types.ts
```diff
@@ -1,7 +1,5 @@
 import * as React from 'react';
 
-import type { StorageAccessLevel } from '@aws-amplify/core';
-
 import { FileStatus } from '../FileUploader/types';
```
calebpollman, line 3: Can we just import `StorageAccessLevel` here to dedupe?

esauerbo: Not sure what you mean, import from where?

calebpollman: From _ ../FileUploader/types_

esauerbo: Oh lol yes, updated

calebpollman (approved):

jordanvn (approved):

PR #5881, opened 2024-10-09
fix(storage-browser): invalid list ui state on error

pranavosu:
#### Description of changes

Use `hasError` to determine table visibility. This prevents being able to continually click through the same folder where an error occurred. 

nav-bug

In the future, we would want to reconsider combining error states with valid data in [useDataState](https://github.com/aws-amplify/amplify-ui/blob/9ac7b8d776bbb605a2e9283bb55ad5fdbbb59c09/packages/react-core/src/hooks/useDataState.ts#L43). 

#### Issue #, if available

https://app.asana.com/0/1208039352063378/1208383276287447

Create a folder with Deny perms in S3 bucket. Navigate to it in storage browser. Verify that you cannot keep clicking on the same folder.

#### Checklist

- [ ] PR description included


AllanZhengYP (dismissed):

packages/react-storage/src/components/StorageBrowser/views/LocationsView/__tests__/LocationsView.spec.tsx
```diff
@@ -125,6 +131,16 @@ describe('LocationsListView', () => {
     const messageText = screen.getByText(errorMessage);
     expect(message).toBeInTheDocument();
     expect(messageText).toBeInTheDocument();
+
+    // table doesn't render
+    const table = screen.queryByRole('table');
+    expect(table).not.toBeInTheDocument();
+
+    // pagination disabled
```
calebpollman, line 139: Appreciate the comments ⭐

calebpollman (approved):
Thanks for handling this! 

Approach makes sense but have some thoughts that can be addressed as follow ups - afaict in the error scenario addressed here the state of `LocationsView` and `LocationDetailView` becomes more or less unusable with the exception of the refresh button and breadcrumbs navigation (detail view only). Wondering if we should provide a descriptive error view inside the `Table` (rather than hiding it entirely) with a retry button to improve the UX? Thinking we need product input from @dbanksdesign here

pranavosu:
> Thanks for handling this!
> 
> Approach makes sense but have some thoughts that can be addressed as follow ups - afaict in the error scenario addressed here the state of `LocationsView` and `LocationDetailView` becomes more or less unusable with the exception of the refresh button and breadcrumbs navigation (detail view only). Wondering if we should provide a descriptive error view inside the `Table` (rather than hiding it entirely) with a retry button to improve the UX? Thinking we need product input from @dbanksdesign here

I agree! We aren't making the best of use of screen real estate. I'm not sure it has to be a part of the Table component, but we could use a more useful Error / Empty State view.

calebpollman (approved):

AllanZhengYP (approved):

reesscot:
@pranavosu Can you please push this to a branch in the amplify-ui repo rather than using a fork? Our workflows aren't working right for forks right now

pranavosu:
> @pranavosu Can you please push this to a branch in the amplify-ui repo rather than using a fork? Our workflows aren't working right for forks right now

I created a [new PR](https://github.com/aws-amplify/amplify-ui/pull/5889) and the workflows in this one started working. I think it's because the commit hash is identical i.e. 
9ef4ebb. I'm just going to merge this PR if the tests pass, and move the rest of my work to a branch from my fork for next time.

PR #5883, opened 2024-10-09
fix(ai): no more unnecessary re-renders in AIConversation

dbanksdesign:
#### Description of changes

Removing `createProvider` since it wasn't actually providing any technical value over just having a normal provider component. This fixes 2 issues dealing with re-rendering:
1. Response components re-rendering with every message
2. Scrollview scrolling from top to bottom instead of current position to bottom

#### Checklist

- [ ] PR description included


thaddmt (approved):

packages/react-ai/src/components/AIConversation/AIConversationProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+  controls,
+  displayText,
+  allowAttachments,
+  messages,
+  handleSendMessage,
+  avatars,
+  isLoading,
+  children,
+}: AIConversationInput &
+  AIConversationProps & { children?: React.ReactNode }): React.JSX.Element => {
```
calebpollman, line 37: Let's define a single interface for the `props`:
```
interface ProviderProps extends AIConversationInput, AIConversationProps {
  children?: React.ReactNode
}
```
Will help ensure that there are no unintentional type overrides through intersection

calebpollman (commented):
LGTM, just needs a small type update

calebpollman (approved):

PR #5900, opened 2024-10-14
chore(ai): add graphql errors to useAIGeneration

dbanksdesign:
#### Description of changes

Errors in GraphQL for AI generation routes are currently being sent to the client in the `errors` part of the response, but the `useAIGeneration` hook wasn't setting the `hasError`. I also moved the error to the `message` part of the hook. It seemed weird that there could be different places to check for the error message?

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-ai/src/hooks/useAIGeneration.tsx
+++ b/packages/react-ai/src/hooks/useAIGeneration.tsx
@@ -75,27 +75,30 @@ export function createUseAIGeneration<
       data: undefined,
     }));
 
-    const handleGeneration = async (input: Schema[Key]['args']) => {
-      setDataState(({ data }) => ({ ...LOADING_STATE, data }));
+    const handleGeneration = React.useCallback(
+      async (input: Schema[Key]['args']) => {
+        setDataState(({ data }) => ({ ...LOADING_STATE, data }));
 
-      const result = await (
-        client.generations as AIGenerationClient<Schema>['generations']
-      )[routeName](input);
+        const result = await (
+          client.generations as AIGenerationClient<Schema>['generations']
+        )[routeName](input);
 
-      const { data, errors } = result as SingularReturnValue<
-        Schema[Key]['returnType']
-      >;
+        const { data, errors } = result as SingularReturnValue<
+          Schema[Key]['returnType']
+        >;
 
-      if (errors) {
-        setDataState({
-          ...ERROR_STATE,
-          data,
-          messages: errors,
-        });
-      } else {
-        setDataState({ ...INITIAL_STATE, data });
-      }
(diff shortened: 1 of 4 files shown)


packages/react-ai/src/hooks/useAIGeneration.tsx
```diff
@@ -74,10 +74,17 @@ export function createUseAIGeneration<
...
     const { data, errors } =
       (result?.data as SingularReturnValue<Schema[Key]['returnType']>) ?? {};
 
-    return [{ ...result, data, graphqlErrors: errors }, handler];
+    if (errors) {
+      hasError = true;
+      message = errors.map((error) => error.message).join(' ');
+    }
+
+    return [{ ...result, data, hasError, message }, handler];
```
thaddmt, line 87: Can we ever get into a case where useDataState's result has an error `message` and we also have graphqlErrors? In that case graphqlErrors would override that `useDataState` message?

Though I think that would only happen in the case where our previous call had graphqlErrors, then our next generation resulted in an error

dbanksdesign: I think we should be fine because the amplify data client (using appsync) will always return errors in the same way. I don't think there is another way to get an error from the data client.

calebpollman: Thinking `useDataState` is the wrong approach here for a handful of reasons:
- `output.hasError` is `true`  when the API call itself fails, which is why it's `false` handled in this use case
- `output.message` exposes the error message of the failed API call
- afaict `handleGenerate` is setting a subscription in a fire and forget fashion rather than using  continuous calls to the `useDataState` handler to provide stateful `data`

calebpollman (approved):
🚢

thaddmt (approved):

thaddmt:
I forgot my approval counts for nothing :sweat_smile:

esauerbo (approved):

packages/react-ai/src/hooks/useAIGeneration.tsx
```diff
@@ -55,29 +65,42 @@ export function createUseAIGeneration<
...
-      (result?.data as SingularReturnValue<Schema[Key]['returnType']>) ?? {};
+        const { data, errors } = result as SingularReturnValue<
+          Schema[Key]['returnType']
+        >;
+
+        if (errors) {
+          setDataState({
+            ...ERROR_STATE,
+            data,
+            messages: errors,
```
calebpollman, line 94: Late question and asking mostly out of curiousity, how do consumers correlate `data` and `errors`?

PR #5908, opened 2024-10-16
feat: useLocationsView hook

pranavosu:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
@@ -1,23 +1,16 @@
 import React from 'react';
-
 import { CLASS_BASE } from '../constants';
 import { Controls } from '../Controls';
 import { useLocationsData } from '../../context/actions';
-
-import { usePaginate } from '../hooks/usePaginate';
-import { listViewHelpers, resolveClassName } from '../utils';
-
+import { resolveClassName } from '../utils';
 import { DataTableControl } from './Controls/DataTable';
+import { LocationAccess } from '../../context/types';
+import { useLocationsView } from './useLocationsView';
 
 export interface LocationsViewProps {
   className?: (defaultClassName: string) => string;
 }
 
-const DEFAULT_PAGE_SIZE = 100;
-export const DEFAULT_LIST_OPTIONS = {
-  exclude: 'WRITE' as const,
-  pageSize: DEFAULT_PAGE_SIZE,
-};
 export const DEFAULT_ERROR_MESSAGE = 'There was an error loading locations.';
 
 const {
@@ -64,40 +57,16 @@ const LocationsEmptyMessage = () => {
 export function LocationsView({
   className,
 }: LocationsViewProps): React.JSX.Element {
-  const [{ data, isLoading, hasError }, handleList] = useLocationsData();
+  const [{ data, isLoading, hasError }, handleAction] = useLocationsView();
 
-  const { result, nextToken } = data;
-  const resultCount = result.length;
-  const hasNextToken = !!nextToken;
+  const { items, hasMoreData, page } = data;
 
-  // initial load
-  React.useEffect(() => {
-    handleList({
-      options: { ...DEFAULT_LIST_OPTIONS, refresh: true },
-    });
-  }, [handleList]);
+  const handleLocationClick = (location: LocationAccess) => {
+    handleAction({ type: 'select-location', location });
+  };
 
-  const onPaginateNext = () =>
-    handleList({
-      options: { ...DEFAULT_LIST_OPTIONS, nextToken },
-    });
-
-  const {
-    currentPage,
-    handlePaginateNext,
-    handlePaginatePrevious,
-    handleReset,
-  } = usePaginate({ onPaginateNext, pageSize: DEFAULT_PAGE_SIZE });
-
-  const { disableNext, disablePrevious, disableRefresh, range } =
-    listViewHelpers({
-      currentPage,
-      hasNextToken,
-      isLoading,
-      pageSize: DEFAULT_PAGE_SIZE,
-      resultCount,
-      hasError,
-    });
+  const disableNext = !hasMoreData || isLoading || hasError;
+  const disablePrevious = page <= 1 || isLoading || hasError;
 
   return (
     <div
@@ -106,26 +75,30 @@ export function LocationsView({
     >
       <Title>Home</Title>
       <RefreshControl
-        disableRefresh={disableRefresh}
+        disableRefresh={isLoading}
         handleRefresh={() => {
-          handleReset();
-          handleList({
-            options: { ...DEFAULT_LIST_OPTIONS, refresh: true },
-          });
+          handleAction({ type: 'refresh' });
         }}
       />
       <Paginate
-        currentPage={currentPage}
+        currentPage={page}
         disableNext={disableNext}
         disablePrevious={disablePrevious}
         handleNext={() => {
-          handlePaginateNext({ resultCount, hasNextToken });
+          handleAction({ type: 'paginate-next' });
         }}
-        handlePrevious={handlePaginatePrevious}
+        handlePrevious={() => {
+          handleAction({ type: 'paginate-previous' });
+        }}
       />
       <LocationsMessage />
       <Loading />
-      <DataTableControl range={range} />
+      {hasError ? null : (
+        <DataTableControl
+          items={items}
+          handleLocationClick={handleLocationClick}
+        />
+      )}
       <LocationsEmptyMessage />
     </div>
   );
--- a/packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
+++ b/packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
@@ -1,0 +1,130 @@
+import { LocationAccess } from '../../context/types';
+import { ActionState } from '../../context/actions/createActionStateContext';
+import { useLocationsData } from '../../context/actions';
+import { useControl } from '../../context/control';
+import React from 'react';
+import { usePaginate } from '../hooks/usePaginate';
+
+interface UseLocationsView {
+  // items to display
+  items: LocationAccess[];
+
+  // current page
+  page: number;
+
+  // more items available to display
+  hasMoreData: boolean;
+
+  // escape hatch to access internal hook state
+  getState: () => LocationsViewState;
+}
+
+interface LocationsViewState {
+  // all items in memory
+  allItems: LocationAccess[];
+  // current page size
+  pageSize: number;
+}
+
+export type LocationsViewActionType =
+  // refresh data
+  | { type: 'refresh' }
+  // reset view to initial state
+  | { type: 'reset' }
+  // paginate
+  | { type: 'paginate-next' }
+  | { type: 'paginate-previous' }
+  // set location to be provided to LocationDetailView
+  | { type: 'select-location'; location: LocationAccess }
+  // query locations
+  | { type: 'search'; query: string; includeSubfolders?: boolean };
+
+const DEFAULT_PAGE_SIZE = 100;
+export const DEFAULT_LIST_OPTIONS = {
+  exclude: 'WRITE' as const,
+  pageSize: DEFAULT_PAGE_SIZE,
+};
+
+export function useLocationsView(): ActionState<
+  UseLocationsView,
+  LocationsViewActionType
+> {
+  const [state, handleList] = useLocationsData();
+  const [, handleUpdateState] = useControl('NAVIGATE');
+
+  const { data, message, hasError, isLoading } = state;
+  const { result, nextToken } = data;
+  const resultCount = result.length;
+  const hasNextToken = !!nextToken;
+  const pageSize = DEFAULT_PAGE_SIZE;
+
+  // initial load
+  React.useEffect(() => {
+    handleList({
+      options: { ...DEFAULT_LIST_OPTIONS, refresh: true },
+    });
+  }, [handleList]);
+
+  // set up pagination
+  const onPaginateNext = () =>
+    handleList({
+      options: { ...DEFAULT_LIST_OPTIONS, nextToken },
+    });
+  const {
+    currentPage,
+    handlePaginateNext,
+    handlePaginatePrevious,
+    handleReset,
+    range,
+  } = usePaginate({ onPaginateNext, pageSize });
+
+  const handleAction = (action: LocationsViewActionType) => {
+    const { type } = action;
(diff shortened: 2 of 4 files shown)


packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,130 @@
+import { LocationAccess } from '../../context/types';
+import { ActionState } from '../../context/actions/createActionStateContext';
+import { useLocationsData } from '../../context/actions';
+import { useControl } from '../../context/control';
+import React from 'react';
+import { usePaginate } from '../hooks/usePaginate';
+
+interface UseLocationsView {
+  // items to display
+  items: LocationAccess[];
```
pranavosu, line 10: @calebpollman this differs from the doc, but it was easier to consume processed items directly instead of calling a function.

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,130 @@
...
+    data: {
+      page: currentPage,
+      hasMoreData: hasNextToken,
+      items: processedItems,
+      getState() {
+        return {
+          allItems: result,
+          pageSize,
+        };
+      },
```
calebpollman, line 126: Let's sync on this with @cshfang to figure out what make sense for the `Table` to consume

packages/react-storage/src/components/StorageBrowser/views/hooks/usePaginate.ts
```diff
@@ -25,6 +26,10 @@ export const usePaginate = ({
     setCurrentPage(1);
   }).current;
 
+  const isInitialPage = currentPage === 1;
+  const start = isInitialPage ? 0 : (currentPage - 1) * pageSize;
+  const end = isInitialPage ? pageSize : currentPage * pageSize;
```
calebpollman, line 31: Since `start` and `end` are resolved with values already included in the memoized return dep array, might make sense to resolve them inside the return value?

packages/react-storage/src/components/StorageBrowser/views/LocationsView/__tests__/LocationsView.spec.tsx
```diff
@@ -290,7 +318,80 @@ describe('LocationsListView', () => {
...
+
+    // first page data matches input
+    expect(screen.queryByLabelText('Page 1')).toBeInTheDocument();
+    expect(screen.queryByText('test-bucket-0')).toBeInTheDocument();
+    expect(screen.queryByText('next-page-0')).not.toBeInTheDocument();
+
+    useLocationsDataSpy.mockReturnValue(nextPageState);
+
+    // go forward
+    await act(async () => {
```
pranavosu, line 353: we shouldn't have to add `act` here -- but we have [this issue](https://github.com/testing-library/user-event/discussions/906#discussioncomment-2522468) in the repo

packages/react-storage/src/components/StorageBrowser/views/LocationsView/__tests__/useLocationsView.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+];
+const PAGE_SIZE = 3;
+
+function mockUseLocationsData(
+  returnValue: DataState<{ result: LocationAccess[] }>
+) {
+  const handleList = jest.fn();
+  (useLocationsData as jest.Mock).mockReturnValue([returnValue, handleList]);
+  return handleList;
+}
```
calebpollman, line 28: What does this provide that using a `jest.spyOn` does not? Concerned abt creating test utilities that do not match the expected behavior of the API they are meant to mock

packages/react-storage/src/components/StorageBrowser/views/LocationsView/__tests__/useLocationsView.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+
+describe('useLocationsView', () => {
+  beforeEach(() => {
+    (useControl as jest.Mock).mockReturnValue([{}, jest.fn()]);
+    (usePaginate as jest.Mock).mockReturnValue({
```
calebpollman, line 33: Since `usePaginate` is a fairly lightweight hook that does not require external dependencies to define its behavior think we can avoid mocking it. Thoughts?

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,96 @@
...
+    handleList({
+      options: { ...DEFAULT_LIST_OPTIONS, nextToken },
+    });
+
+  const {
+    currentPage,
+    handlePaginateNext,
+    handlePaginatePrevious,
+    handleReset,
+    range,
```
pranavosu, line 64: I'll refactor `usePaginate` to avoid exposing `range` as a follow up

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export type LocationsViewActionType =
+  | { type: 'REFRESH' }
+  | { type: 'RESET' }
+  | { type: 'PAGINATE_NEXT' }
+  | { type: 'PAGINATE_PREVIOUS' }
+  | { type: 'PAGINATE'; page: number }
+  // set location to be provided to LocationDetailView
+  | { type: 'SELECT_LOCATION'; location: LocationAccess }
+  | { type: 'SEARCH'; query: string; includeSubfolders?: boolean };
```
calebpollman, line 28: Does LCAG allow for delimiting subfolders?

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
```diff
@@ -1,23 +1,16 @@
 import React from 'react';
-
```
calebpollman, line 2: Nit: Prefer to section imports with an empty line between external imports, "internal" amplify imports, and local imports

packages/react-storage/src/components/StorageBrowser/views/LocationsView/__tests__/useLocationsView.spec.ts
pranavosu: Yea, I think replacing it with a spy like the LocationsView test makes sense.

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
pranavosu: It doesn't, good catch i'll remove it.

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,118 @@
...
+    message,
+    isPaginateNextDisabled: !hasNextToken || isLoading || hasError,
+    isPaginatePreviousDisabled: currentPage <= 1 || isLoading || hasError,
+    page: currentPage,
+    hasNextPage: hasNextToken,
+    pageItems: processedItems,
+    onNavigate: (destination: LocationData) => {
+      if (typeof onNavigate === 'function') {
+        onNavigate(destination);
+      }
```
cshfang, line 104: Are we actually concerned that `onNavigate` is a non-function object? If so, do we need to be concerned about this everywhere? And if not, wouldn't just an optional chain achieve the same result?
```ts
onNavigate?.(destination);
```

packages/react-storage/src/components/StorageBrowser/views/hooks/usePaginate.ts
```diff
@@ -25,8 +26,11 @@ export const usePaginate = ({
     setCurrentPage(1);
   }).current;
 
-  return React.useMemo(
-    () => ({
+  return React.useMemo(() => {
+    const isInitialPage = currentPage === 1;
```
cshfang, line 30: Nit: Given that this is more or less "hardcoded" to mean page 1, would it be clearer to name this `isFirstPage`?

cshfang (approved):

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
pranavosu: Good point. I think we can use optional chaining

packages/react-storage/src/components/StorageBrowser/views/hooks/usePaginate.ts
```diff
@@ -8,6 +8,7 @@ interface UsePaginate {
   }) => void;
   handlePaginatePrevious: (input?: {}) => void;
   handleReset: () => void;
+  range: [start: number, end: number];
 }
 
 export const usePaginate = ({
```
calebpollman, line 14: Not sure if this was previously discussed but would make sense to return `pageItems` from `usePaginate` removing the need to expose `range`. Can be done in a follow up as we will need to update `usePaginate` to return the singular `onPaginate`

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
```diff
@@ -57,54 +50,27 @@ const LocationsEmptyMessage = () => {
...
+    isPaginateNextDisabled,
+    page,
+    isLoading,
+    onRefresh,
+    onPaginateNext,
+    onPaginatePrevious,
+    onNavigate,
+  } = useLocationsView({ onNavigate: onNavigateProp });
 
   // FIXME: Eventually comes from useView hook
```
reesscot, line 68: this can be removed now?

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,116 @@
...
+  | { type: 'REFRESH_DATA' }
+  | { type: 'RESET' }
+  | { type: 'PAGINATE_NEXT' }
+  | { type: 'PAGINATE_PREVIOUS' }
+  | { type: 'PAGINATE'; page: number }
+  | { type: 'NAVIGATE'; location: LocationData }
+  | { type: 'SEARCH'; query: string };
+
+export interface UseLocationsViewOptions {
+  initialValues?: InitialValues;
```
calebpollman, line 37: Should just be `defaultPageSize`, "initial" implies updates

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
```diff
@@ -118,17 +84,17 @@ export function LocationsView({
...
+          currentPage={page}
+          disableNext={isPaginateNextDisabled}
+          disablePrevious={isPaginatePreviousDisabled}
+          handleNext={onPaginateNext}
+          handlePrevious={onPaginatePrevious}
         />
         <LocationsMessage />
         <Loading />
-        <DataTableControl onNavigate={onNavigate} range={range} />
+        {hasError ? null : (
```
reesscot, line 95: shouldn't we show an empty message in the table rather than rendering null?

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const pageItems = React.useMemo(() => {
+    const [start, end] = range;
+    return result.slice(start, end);
+  }, [range, result]);
+
+  return {
+    isLoading,
+    hasError,
+    message,
```
reesscot, line 95: This doesn't appear to be getting used. Will that be part of the Message component refactor?

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
pranavosu: We'll show an error message but it looks like they're currently shown outside the table. I'm not sure if the new DataTable supports messages inside the table.

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -0,0 +1,116 @@
...
+    ...DEFAULT_LIST_OPTIONS,
+    ...initialValues,
+  });
+
+  // initial load
+  React.useEffect(() => {
+    handleList({
+      options: { ...listOptions, refresh: true },
+    });
+  }, [handleList, listOptions]);
```
calebpollman, line 71: `defaultPageSize` can be put inside a ref since there will be no changes to it that require re-rendering, e.g.
```ts
  const defaultListOptions = React.useRef({ ...DEFAULT_LIST_OPTIONS, pageSize: defaultPageSize });

  React.useEffect(() => {
    handleList({
      options: { ...defaultListOptions.current, refresh: true },
    });
  }, [handleList]);
```
Also if there is no use case for `DEFAULT_LIST_OPTIONS` besides this call we can move `refresh` inside of it as well

calebpollman (approved):
LGTM. Have some minor feedback, can be addressed in a follow up

packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
pranavosu: Yeah, the message component directly accesses `useLocationData` currently. I believe this `message` will be passed in to the ControlsContext for the Message composable.

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
pranavosu: I think we'll have some churn here as we finalize Composables and the ControlsContext interface. It's better if we handle this in a follow up

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
reesscot: I guess if you can't see the locations you're sort of out of luck for using the entire component. Maybe we could have a follow up to revisit the experience on this view, but not a blocker.

reesscot (approved):

packages/react-storage/src/components/StorageBrowser/views/hooks/usePaginate.ts
pranavosu: Yeah, we discussed it. I'd love to get rid of range, but it'll have to be a follow up

Issue #5911, opened 2024-10-16
Bug(Authenticator): Clicking the Sign in tab when already on the Sign In tab opens the 


hbuchel:
### Before creating a new issue, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [X] I have tried disabling all browser extensions or using a different browser
- [X] I have tried deleting the node_modules folder and reinstalling my dependencies
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Next.js

### Please describe your bug.

In the Authenticator component, if you click on the tab that is currently active, it switches the active tab.

### What's the expected behaviour?

Clicking the active tab button should not switch the active tab to the other.

### Help us reproduce the bug!

In the example on https://ui.docs.amplify.aws/react/connected-components/authenticator, with the "Sign in" tab open, click the "Sign in" tab again and it switches the active tab to Create Account.

### Code Snippet

```typescript
// Put your code below this line.

```

mjdepalma:
Some extra context in case it helps.  Selecting a particular page title should (1) switch to that page if user focus is currently on opposite page (which works), and (2) stay on that page if user is already on that page (which doesn't work).   That is the way all menus and spreadsheet pages work, and is the way that users will expect it to work as well.

(1) With focus on Sign In page, select "Sign In", and focus will incorrectly toggle to "Create Account"
(2) select "Sign In" again, and focus will toggle back to "Sign In"
(3) you can repeat steps 3 & 4 over and over to see the issue
(4) with focus on Create Account page, select "Create Account" page title, and focus will incorrectly toggle to "Sign In"
(5) select "Create Account" again, and focus will toggle back to "Create Account"
(6) you can repeat steps 6 & 7 over and over to see the issue

Hope this helps!

hbuchel:
Thanks for the extra context @mjdepalma

PR #5912, opened 2024-10-16
fix(Authenticator): clicking on current tab switches tabs

hbuchel:
#### Description of changes

Fixes an issue where clicking on the currently open Authenticator tab switches to the other tab. i.e. clicking on the "Sign in" tab while it is already active will no longer cause the "Create account" tab to open.

**before**

https://github.com/user-attachments/assets/286e5574-d991-4f4d-8a9e-67e0efb0477a

**after**

https://github.com/user-attachments/assets/b53efd48-d6fc-403a-86bc-42ec757f0762

#### Issue #, if available

Fixes: https://github.com/aws-amplify/amplify-ui/issues/5911

Verified in local docs examples

#### Checklist

- [ ] PR description included


packages/react/src/components/Authenticator/shared/SignInSignUpTabs.tsx
```diff
@@ -32,7 +32,11 @@ export const SignInSignUpTabs = ({
         <Tabs.Container
           value={route}
           isLazy
-          onValueChange={() => (route === 'signIn' ? toSignUp() : toSignIn())}
+          onValueChange={(prevRoute) => {
+            if (prevRoute !== route) {
+              route === 'signIn' ? toSignUp() : toSignIn();
```
reesscot, line 37: I don't know why, but this is hurting my brain. Is the route the tab I clicked on or the route I'm currently at? So if I click any tab then it will essentially toggle to the other one? Why is comparing the prevRoute to the current route necessary?

jordanvn: I think these variable names are switched. It looks like `route` is the active tab and `prevRoute` is the clicked tab.

So the old logic was just "if Sign In is the active tab go to Sign Up, otherwise go to Sign In", which results in always going to the other tab no matter what is clicked.

And the new logic is just adding the check of "if the clicked tab isn't the active tab...". It's still just flipping the value without regard for what was clicked, but as long as there are only the two tabs it will continue to work.

jordanvn: `newRoute` might be the more consistent name, since TabContainer's `setActiveTab()` is using the name `newValue` for the argument passed to `onValueChange()`

hbuchel: Updated in latest to use newRoute instead, that makes more sense for the actual logic happening.

jordanvn (approved):
LGTM

reesscot (approved):

jordanvn (approved):

reesscot (approved):
LGTM

PR #5913, opened 2024-10-16
chore: replace 'name' column with 'bucket' and 'folder' columns

jordanvn:
#### Description of changes
* Replace "Name" column with "Bucket" and "Folder" columns
  * "Bucket" contains the bucket name (without `s3://`), and is not clickable
  * "Folder" contains the folder prefix ("[bucket name]/" if it is a bucket), and is clickable
  * Folder is now the leftmost column and results are set to be sorted by folder name by default
* Remove "Type" column from LocationsView

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/views/LocationsView/Controls/DataTable.tsx
```diff
@@ -101,37 +110,49 @@ const getLocationsData = ({
...
+    }
+    return row;
+  });
 
   return { columns, rows };
 };
 
 export function DataTableControl({
   range,
+  hideType = false,
```
reesscot, line 145: do we need an option or should we just remove it by default? Do we anticipate adding it back?

jordanvn: Ah, I didn't recognize that this table was only used for the locations view. Yes, just removing type altogether makes the most sense.

jordanvn: Removed both `hideType` and usage of type in the table

reesscot (approved):

PR #5923, opened 2024-10-18
feat(react-storage): Introduce accountId option to createStorageBrowser

jimblanc:
#### Description of changes
Introduces an optional accountId parameter to the createStorageBrowser utilities and wires it in with the Amplify JS internal APIs used by the StorageBrowser.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/actions/types.ts
+++ b/packages/react-storage/src/components/StorageBrowser/actions/types.ts
@@ -3,7 +3,7 @@ import { LocationCredentialsProvider } from '../storage-internal';
 import { ActionState } from '../context/actions/createActionStateContext';
 
 export interface ActionInputConfig {
-  accountId: string;
+  accountId?: string;
   bucket: string;
   credentials: LocationCredentialsProvider;
   region: string;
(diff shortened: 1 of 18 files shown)


AllanZhengYP (commented):
We should also add this to the Amplify auth adapter when it's ready.

packages/react-storage/src/components/StorageBrowser/actions/types.ts
```diff
@@ -3,7 +3,7 @@ import { LocationCredentialsProvider } from '../storage-internal';
 import { ActionState } from '../context/actions/createActionStateContext';
 
 export interface ActionInputConfig {
-  accountId: string;
+  accountId?: string;
```
ashika112, line 6: why change to optional? i thought this is required in storage browser 🤔

packages/react-storage/src/components/StorageBrowser/context/config.tsx
```diff
@@ -12,6 +12,7 @@ import { LocationConfig } from './types';
 import { parseLocationAccess } from './navigate/utils';
 
 export interface LocationConfigProviderProps {
+  accountId?: string;
```
ashika112, line 15: diito as abve. why optional now?

packages/react-storage/src/components/StorageBrowser/actions/types.ts
jimblanc: This was done to avoid introducing a new prop & API change on the Amplify-auth StorageBrowser component. Discussed with Caleb, we only need the account ID for the control-plane APIs and will address that with the provider refactoring.

ashika112 (approved):

calebpollman (approved):

PR #5936, opened 2024-10-22
feat(storage-browser): add useLocationDetailView hook

pranavosu:
#### Description of changes

- Refactored LocationDetailViewControl to separate logic into a hook.
- Added unit tests

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/views/Controls/Table.tsx
```diff
@@ -189,23 +189,21 @@ const sortFns = {
...
   handleDroppedFiles,
-  onNavigate,
-  range,
+  handleLocationItemClick
 }: {
+  items: LocationItemData[];
   handleDroppedFiles: (files: File[]) => void;
-  onNavigate?: (location: LocationData) => void;
-  range: [start: number, end: number];
+  handleLocationItemClick: (location: LocationData) => void;
```
cshfang, line 198: Probably not really in scope or even relevant to your PR but we should probably stick to the `onX` convention for props to be consistent. But it shouldn't be a concern here as we will replace this table anyway

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/Controls.tsx
```diff
@@ -73,92 +66,20 @@ export const LocationDetailViewControls = ({
...
-  const handleRefresh = () => {
-    if (hasInvalidPrefix) return;
-    handlePaginateReset();
-    handleList({ prefix, options: DEFAULT_REFRESH_OPTIONS });
-    dispatchStoreAction({ type: 'RESET_LOCATION_ITEMS' });
-  };
+    page,
+    isPaginatePreviousDisabled,
+    isPaginateNextDisabled,
+  } = locationDetailView;
```
cshfang, line 76: Similar comment as another PR. Seems to me like we should be able to destructure from the hook call directly and just drop some of the `locationDetailView.` from below

packages/react-storage/src/components/StorageBrowser/views/hooks/usePaginate.ts
```diff
@@ -25,15 +26,17 @@ export const usePaginate = ({
     setCurrentPage(1);
   }).current;
 
-  return React.useMemo(
-    () => ({
+  return React.useMemo(() => {
+    const isInitialPage = currentPage === 1;
```
cshfang, line 30: This is the same change in two PRs, right?

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -0,0 +1,156 @@
...
+interface InitialValues {
+  pageSize?: number;
+  delimiter?: string;
+}
+
+export interface UseLocationDetailViewOptions {
+  initialValues?: InitialValues;
+  onDispatch?: React.Dispatch<LocationDetailViewActionType>;
+  onActionSelect?: (type: string) => void;
+  onExit?: () => void;
```
cshfang, line 45: I wonder if this should be something more akin to `onNavigateAway` - "exiting" a component is a little strange to me

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -0,0 +1,156 @@
...
+  delimiter: '/',
+  pageSize: DEFAULT_PAGE_SIZE,
+};
+
+export function useLocationDetailView(
+  options?: UseLocationDetailViewOptions
+): UseLocationDetailView {
+  const initialValues = options?.initialValues;
+  const onActionSelect = options?.onActionSelect;
+  const onNavigate = options?.onNavigate;
```
cshfang, line 60: Nit: Just to reduce some of the boilerplate for readability
```suggestion
  const { initialValues, onActionSelect, onNavigate } = options ?? {};
```

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -0,0 +1,156 @@
...
+  }, [
+    dispatchStoreAction,
+    handleList,
+    handleReset,
+    listOptions,
+    hasInvalidPrefix,
+    prefix,
+  ]);
+
+  const processedItems = React.useMemo(() => {
```
cshfang, line 122: Nit: Might as well just call this `pageItems` and use shorthand below since "processed" doesn't really add much context anyway.

cshfang (dismissed):

packages/react-storage/src/components/StorageBrowser/views/Controls/Table.tsx
pranavosu: Yeah, we have an existing `handleDroppedFiles` prop here, so I wanted to stick to that convention. But in general we should use `onX` 👍🏽

packages/react-storage/src/components/StorageBrowser/views/hooks/usePaginate.ts
pranavosu: yeah, i forgot about that. it's been a while since i opened the other PR. i'll rebase after `useLocationsView` is merged.

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/Controls.tsx
```diff
@@ -73,92 +66,20 @@ export const LocationDetailViewControls = ({
...
 
   const contextValue: ControlsContext = {
     data: {
-      isDataRefreshDisabled: disableRefresh,
+      isDataRefreshDisabled: isLoading,
     },
-    onRefresh: handleRefresh,
+    onRefresh: locationDetailView.onRefresh,
```
reesscot, line 82: This can just be destructured from the hook right?
```suggestion
    onRefresh,
```

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -0,0 +1,156 @@
...
+  pageSize: DEFAULT_PAGE_SIZE,
+};
+
+export function useLocationDetailView(
+  options?: UseLocationDetailViewOptions
+): UseLocationDetailView {
+  const initialValues = options?.initialValues;
+  const onActionSelect = options?.onActionSelect;
+  const onNavigate = options?.onNavigate;
+  const listOptions = React.useMemo(() => {
```
reesscot, line 61: Is use memo really necessary here? Doesn't seem like an expensive calculation.

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -0,0 +1,156 @@
...
+    handleList({ prefix, options: { ...listOptions, refresh: true } });
+    dispatchStoreAction({ type: 'RESET_LOCATION_ITEMS' });
+  };
+
+  React.useEffect(() => {
+    if (hasInvalidPrefix) return;
+    handleList({ prefix, options: { ...listOptions, refresh: true } });
+    handleReset();
+  }, [
+    dispatchStoreAction,
```
reesscot, line 114: is this a required dependency? Doesn't seem to be getting used.

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
pranavosu: it think the intention is to 'exit' a view, but I think we could consider changing that in a separate PR as `onExit` is used in a few different places.

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/__tests__/useLocationDetailView.spec.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+function mockListItemsAction(
+  returnValue: DataState<ListLocationItemsActionOutput>
+) {
+  const handleList = jest.fn();
+  jest
+    .spyOn(ActionsModule, 'useAction')
+    .mockReturnValue([returnValue, handleList]);
+  return handleList;
+}
```
calebpollman, line 59: Think this could just be:
```function mockListItemsAction(

  const handleListSpy =  jest
    .spyOn(ActionsModule, 'useAction')

  // usage
  handleListSpy.mockReturnValue([someMockData, someMockHandler])
```

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
reesscot: Don't disagree that it's a little weird, but I think we should keep onExit for now, as that's what we landed on as best option in the fit and finish review.

reesscot (approved):

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
calebpollman: We need to align all of the disparate names in these hooks. Can be done in a handled in a follow up since it shouldn't impact logic

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/__tests__/useLocationDetailView.spec.tsx
pranavosu: This function was more complicated in an earlier iteration of the tests, but i think we can keep it as is for now. I think it makes the tests a bit cleaner.

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -0,0 +1,149 @@
...
+    handleReset();
+    handleList({ prefix, options: { ...listOptions, refresh: true } });
+    dispatchStoreAction({ type: 'RESET_LOCATION_ITEMS' });
+  };
+
+  React.useEffect(() => {
+    if (hasInvalidPrefix) return;
+    handleList({ prefix, options: { ...listOptions, refresh: true } });
+    handleReset();
+  }, [handleList, handleReset, listOptions, hasInvalidPrefix, prefix]);
```
calebpollman, line 113: Nit: `listOptions` wouldn't need to be in the dep array if you use reference `listOptions.current` directly in the body of the effect

calebpollman (approved):
🚢

PR #5940, opened 2024-10-23
refactor(react-storage): Cancel to composable ActionCancel

Samaritan1011001:
#### Description of changes
Refactors the usage and implementation of Cancel button to ActionCancel button while making it composable.

- Exposes ActionCancelProps for a composable custom ActionCancel component
- Uses ControlContext to supply the data/information needed for the composable component to display it.
- Introduces a useActionCancel hook that converts the controlContext into props that is fed into the ResolvedComposable ActionCancel component.
- Adds tests

Next example app manual tests

#### Checklist

- [x] PR description included


packages/react-storage/src/components/StorageBrowser/controls/ActionCancelControl.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { ControlProps } from './types';
+import { useResolvedComposable } from './hooks/useResolvedComposable';
+import { ActionCancel } from '../composables/ActionCancel';
+import { useActionCancel } from './hooks/useActionCancel';
+import { ViewElement } from '../context/elements';
+
+export const ActionCancelControl = ({
+  className,
+}: ControlProps): React.JSX.Element | null => {
```
reesscot, line 11: I don't think it would even be null?

packages/react-storage/src/components/StorageBrowser/controls/__tests__/ActionCancelControl.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ActionCancelControl } from '../ActionCancelControl';
+import * as useActionCancelModule from '../hooks/useActionCancel';
+
+describe('ActionCancelControl', () => {
+  const useActionCancelSpy = jest.spyOn(
+    useActionCancelModule,
+    'useActionCancel'
+  );
+
+  afterEach(() => {
```
reesscot, line 12: nit: afterEach seems odd compared to using beforeEach for this. I don't know if there's really a difference here, so probably just personal pref

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/UploadControls.tsx
```diff
@@ -153,11 +154,24 @@ const renderRowItem: RenderRowItem<LocationActionViewColumns> = (
...
+            },
+            onActionCancel: row.cancel,
+          };
+
           return (
-            <Cancel
-              onClick={row.cancel}
-              ariaLabel={`Cancel upload for ${row.key}`}
-            />
+            <ControlsContextProvider {...contextValue}>
```
reesscot, line 170: It feels weird to wrap every cancel control in each row with a provider.

reesscot (commented):
Had some questions/comments

packages/react-storage/src/components/StorageBrowser/controls/ActionCancelControl.tsx
Samaritan1011001: True, previously it would return null if there are no props but now it will always be there so I can remove this.

packages/react-storage/src/components/StorageBrowser/controls/__tests__/ActionCancelControl.spec.tsx
Samaritan1011001: can go with beforeEach

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/UploadControls.tsx
Samaritan1011001: It is all the rows but only all the rows of the Cancel column. The data for ariaLabel and onActionCancel are specific to that column and row. So not sure how I can lift it up.

Samaritan1011001: Discussed offline and leaving it as is since row data for this column is only available here.

reesscot (dismissed):

packages/react-storage/src/components/StorageBrowser/composables/ActionCancel.tsx
```diff
@@ -0,0 +1,33 @@
...
+  ariaLabel?: string;
+  isDisabled?: boolean;
+  text?: string;
+}
+
+export const ActionCancel = ({
+  onCancel,
+  ariaLabel,
+  isDisabled,
+  text,
```
calebpollman, line 17: This should be `label` to align with `ActionStart` right?

Samaritan1011001: Won't it be confusing with ariaLabel?

Samaritan1011001: Reverted to label

calebpollman: We're using `label` for the `ActionStart` button, tbh not really sure why we have `ariaLabel` here, control props aren't meant to be that low level

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/UploadControls.tsx
reesscot: Instead of using this ActionCancelControl here, this could just be converted to using a ButtonElement. I don't think we are going to offer composability at this level.

packages/react-storage/src/components/StorageBrowser/composables/ActionCancel.tsx
```diff
@@ -0,0 +1,30 @@
...
+  label,
+}: ActionCancelProps): React.JSX.Element => (
+  <ButtonElement
+    variant="cancel"
+    className={`${CLASS_BASE}__action-cancel`}
+    onClick={onCancel}
+    disabled={isDisabled}
+  >
+    {label ?? (
+      <IconElement
```
reesscot, line 24: I don't think we're using the IconElement are we?

Samaritan1011001: True, since I removed the table's cancel button, don't need this Icon here.

reesscot (dismissed):

calebpollman (approved):

PR #5954, opened 2024-10-24
fix(liveness): Fix oval render when switching cameras 

esauerbo:
#### Description of changes
- Fix for an issue where the oval renders in the wrong location when switching cameras to unblock data collection
- Add an event listener to make sure video metadata (which includes `video.videoHeight` and `video.videoWidth`) is loaded before attempting to draw the oval
- Remove `drawStaticOval` from state machine because we already call it in the LivenessCameraModule. Calling from multiple places caused weird/unpredictable behavior 

- Deployment https://testing.d2pnskzbmchcgk.amplifyapp.com/
- Compare to https://alpha.d3v0f004886eky.amplifyapp.com/

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -163,35 +163,20 @@ export const LivenessCameraModule = (
     (!isMobileScreen || isFaceMovementChallenge);
 
   React.useEffect(() => {
-    if (
+    const shouldDrawOval =
       canvasRef?.current &&
       videoRef?.current &&
       videoStream &&
       isStartView &&
-      isMetadataLoaded
-    ) {
-      drawStaticOval(canvasRef.current, videoRef.current, videoStream);
-    }
-  }, [
-    canvasRef,
-    videoRef,
-    videoStream,
-    colorMode,
-    isStartView,
-    isMetadataLoaded,
-  ]);
+      isMetadataLoaded;
 
-  React.useEffect(() => {
+    if (shouldDrawOval) {
+      drawStaticOval(canvasRef.current, videoRef.current!, videoStream);
+    }
+
     const updateColorModeHandler = (e: MediaQueryListEvent) => {
-      if (
-        e.matches &&
-        canvasRef?.current &&
-        videoRef?.current &&
-        videoStream &&
-        isStartView &&
-        isMetadataLoaded
-      ) {
-        drawStaticOval(canvasRef.current, videoRef.current, videoStream);
+      if (e.matches && shouldDrawOval) {
+        drawStaticOval(canvasRef.current, videoRef.current!, videoStream);
       }
     };
 
@@ -209,7 +194,14 @@ export const LivenessCameraModule = (
       darkModePreference.removeEventListener('change', updateColorModeHandler);
       lightModePreference.addEventListener('change', updateColorModeHandler);
     };
-  }, [canvasRef, videoRef, videoStream, isStartView, isMetadataLoaded]);
+  }, [
+    canvasRef,
+    videoRef,
+    videoStream,
+    colorMode,
+    isStartView,
+    isMetadataLoaded,
+  ]);
 
   React.useLayoutEffect(() => {
     if (isCameraReady) {
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -194,14 +194,7 @@ export const LivenessCameraModule = (
       darkModePreference.removeEventListener('change', updateColorModeHandler);
       lightModePreference.addEventListener('change', updateColorModeHandler);
     };
-  }, [
-    canvasRef,
-    videoRef,
-    videoStream,
-    colorMode,
-    isStartView,
-    isMetadataLoaded,
-  ]);
+  }, [videoRef, videoStream, colorMode, isStartView, isMetadataLoaded]);
 
   React.useLayoutEffect(() => {
     if (isCameraReady) {
--- a/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
@@ -128,6 +128,7 @@ export const LivenessCameraModule = (
   const freshnessColorRef = useRef<HTMLCanvasElement | null>(null);
 
   const [isCameraReady, setIsCameraReady] = useState<boolean>(false);
+  const [isMetadataLoaded, setIsMetadataLoaded] = useState<boolean>(false);
   const isInitCamera = state.matches('initCamera');
   const isInitWebsocket = state.matches('initWebsocket');
   const isCheckingCamera = state.matches({ initCamera: 'cameraCheck' });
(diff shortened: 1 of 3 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -541,13 +541,10 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
       }),
       updateDeviceAndStream: assign({
         videoAssociatedParams: (context, event) => {
-          const { canvasEl, videoEl, videoMediaStream } =
-            context.videoAssociatedParams!;
           setLastSelectedCameraId(event.data?.newDeviceId as string);
           context.livenessStreamProvider?.setNewVideoStream(
             event.data?.newStream as MediaStream
           );
-          drawStaticOval(canvasEl!, videoEl!, videoMediaStream!);
```
jordanvn, line 550: Info: why is this being removed here?

jordanvn (dismissed):
This makes sense to me

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -253,21 +272,25 @@ export const LivenessCameraModule = (
...
-          type: 'UPDATE_DEVICE_AND_STREAM',
-          data: { newDeviceId, newStream },
-        });
+        // Only update the stream and draw oval once metadata is loaded
+        if (isMetadataLoaded) {
+          send({
+            type: 'UPDATE_DEVICE_AND_STREAM',
+            data: { newDeviceId, newStream },
+          });
+        }
```
calebpollman, line 289: How does run if `setIsMetaDataLoaded` is set to `false` in the same closure on line 275?

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -162,10 +163,23 @@ export const LivenessCameraModule = (
     (!isMobileScreen || isFaceMovementChallenge);
 
   React.useEffect(() => {
-    if (canvasRef?.current && videoRef?.current && videoStream && isStartView) {
+    if (
+      canvasRef?.current &&
+      videoRef?.current &&
+      videoStream &&
+      isStartView &&
+      isMetadataLoaded
```
calebpollman, line 171: Concerned abt fragility/maintainability with so many boolean values used for a condition

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -194,10 +209,10 @@ export const LivenessCameraModule = (
       darkModePreference.removeEventListener('change', updateColorModeHandler);
       lightModePreference.addEventListener('change', updateColorModeHandler);
     };
-  }, [canvasRef, videoRef, videoStream, isStartView]);
+  }, [canvasRef, videoRef, videoStream, isStartView, isMetadataLoaded]);
 
   React.useLayoutEffect(() => {
-    if (isCameraReady) {
+    if (isCameraReady && isMetadataLoaded) {
```
calebpollman, line 215: Using `isMetadataLoaded` to gate behavior in three different side effects creates complexity in readability and refactoring, can we isolate some of this behavior?

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
esauerbo: Yeah fair, maybe we can extract them into one variable for readability?

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
esauerbo: This is updated now so it just sets `IsMetaDataLoaded` to false.

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
esauerbo: I think that would be hard to do without a lot of refactoring. But I was actually able to remove this particular gate

reesscot (dismissed):

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -162,10 +163,23 @@ export const LivenessCameraModule = (
...
       drawStaticOval(canvasRef.current, videoRef.current, videoStream);
     }
-  }, [canvasRef, videoRef, videoStream, colorMode, isStartView]);
+  }, [
+    canvasRef,
+    videoRef,
+    videoStream,
+    colorMode,
+    isStartView,
+    isMetadataLoaded,
```
calebpollman, line 181: The fact the dependency array here and the dependency array on line 212 are basically the same leads me to believe that we should be able to combine the behavior of these effects in a single `useEffect` which can be abstracted in to a util hook and unit tested in isolation

esauerbo: Combined into a single useEffect and added a unit test to ensure the oval is drawn when metadata loads

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/__tests__/LivenessCameraModule.test.tsx
```diff
@@ -258,7 +269,9 @@ describe('LivenessCameraModule', () => {
       />
     );
     const videoEl = screen.getByTestId('video');
-    videoEl.dispatchEvent(new Event('canplay'));
+    await waitFor(() => {
```
reesscot, line 272: thanks for fixing these!

reesscot (dismissed):

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -162,21 +163,20 @@ export const LivenessCameraModule = (
...
-        e.matches &&
-        canvasRef?.current &&
-        videoRef?.current &&
-        videoStream &&
-        isStartView
-      ) {
-        drawStaticOval(canvasRef.current, videoRef.current, videoStream);
+      if (e.matches && shouldDrawOval) {
+        drawStaticOval(canvasRef.current, videoRef.current!, videoStream);
       }
```
calebpollman, line 180: Nice cleanup!

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
```diff
@@ -194,7 +194,14 @@ export const LivenessCameraModule = (
       darkModePreference.removeEventListener('change', updateColorModeHandler);
       lightModePreference.addEventListener('change', updateColorModeHandler);
     };
-  }, [canvasRef, videoRef, videoStream, isStartView]);
+  }, [
+    canvasRef,
+    videoRef,
```
calebpollman, line 199: We don't need to track the refs in the dependency array, their references maintain stability since we access `.current` inside the body of the hook
```suggestion
```

calebpollman (approved):

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
esauerbo: I'll remove `canvasRef` but react gets mad if we take out videoRef, I think because the ref itself might change since it comes from the `useMediaStreamInVideo` hook 
https://github.com/aws-amplify/amplify-ui/blob/4393c78a590499391090abe925acc834c792b815/packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx#L131

calebpollman: Ahh got it. For sure out of scope for this PR (and tbh think it will be a non-trivial change) but we should refactor to keep `videoRef` scoped to wherever it's used directly. Passing refs around leads to unexpected behaviors

calebpollman (approved):

reesscot (dismissed):

packages/react-liveness/src/components/FaceLivenessDetector/LivenessCheck/LivenessCameraModule.tsx
esauerbo: Yeah makes sense I'll add that to our backlog. It definitely made this like 10x harder to fix

reesscot (approved):

calebpollman (approved):

PR #5971, opened 2024-10-28
feat(storage-browser): implement useSearch hook

pranavosu:
#### Description of changes

Adds a hook that can filter an arbitrary list of items. This will be used in `use*View` hooks to add search functionality.

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
```diff
@@ -0,0 +1,77 @@
...
+  debounceDelay?: number;
+}
+
+interface UseSearchProps<T> {
+  initialValues: InitialSearchValues<T>;
+  items: T[];
+  onSearch: (term: string, includeSubfolders?: boolean) => void;
+}
+
+interface UseSearch<T> {
```
cshfang, line 17: Maybe nit but we having the type match the function name usually tells me the typing is for the entire function signature as opposed to just the return type.

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const debouncedHandler = React.useCallback(
+    debounce((term: string, includeSubfolders?: boolean) => {
+      setSearchTerm(term);
+      if (typeof onSearch === 'function') {
+        onSearch(term, includeSubfolders);
+      }
+    }, debounceDelay),
+    [onSearch, debounceDelay]
```
cshfang, line 62: Should `setSearchTerm` be included?

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
```diff
@@ -0,0 +1,77 @@
...
+interface UseSearchProps<T> {
+  initialValues: InitialSearchValues<T>;
+  items: T[];
+  onSearch: (term: string, includeSubfolders?: boolean) => void;
+}
+
+interface UseSearch<T> {
+  filteredItems: T[];
+  searchTerm: string;
+  handleSearch: (term: string, includeSubfolders?: boolean) => void;
```
cshfang, line 20: I understand the intent here of distinguishing between the onSearch from the props and the wrapped version coming out of this hook but I wonder if it's actually necessary to sacrifice the `onX` convention. Imo it wouldn't actually be that bad to just vend the debounced `onSearch` as also `onSearch` from this hook. Ultimately, the consumer of the hook doesn't need to know there's even a difference.

Alternatively, if that is really too unpalatable, something more pointed like `debouncedOnSearch` would be ok?

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @template T - The type of items in the search list.
+ * @param {UseSearchProps<T>} props
+ * @returns {UseSearch<T>} - The search results, current term, and handler.
+ */
+export function useSearch<T>({
+  items,
+  initialValues: initialValuesProp,
+  onSearch,
+}: UseSearchProps<T>): UseSearch<T> {
+  const [initialValues] = React.useState(initialValuesProp);
```
cshfang, line 34: Is this needed? Afaict, it seems like we can just have `initialValues` be `values` in the hook param and just always use below.

packages/react-storage/src/components/StorageBrowser/views/hooks/__tests__/useSearch.spec.ts
```diff
@@ -0,0 +1,113 @@
+import { renderHook, act } from '@testing-library/react-hooks';
```
cshfang, line 1: We've been trying to get away from using this package and using `'@testing-library/react'` instead (even though it will tell you `act` is deprecated 😅 )

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: `useState` returns stable setter functions, so we don't have to. In this case, I disabled `exhaustive-deps` because eslint is actually complaining about passing the `debounce` function directly. it wants `React.useCallback(() => debounce(...))` which is unnecessary IMO just to satisfy eslint

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: I'm not opposed to using `onSearch`, but you're right about the intention. I can change it to be consistent with other hooks

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: The intention with `initialValues` is to keep it stable for the lifecycle of a component as they're just initial values as the names suggest. e.g after the initial render, this hook is the source of truth for what the current `searchTerm` and it manages that internally (with debouncing for e.g) . We don't want to let the component change that after the first render.

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: makes sense. changed it to `UseSearchResult`

cshfang (dismissed):

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: @calebpollman suggested using useMemo, which works better.

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
```diff
@@ -0,0 +1,77 @@
...
+  const filteredItems = React.useMemo(
+    () =>
+      items.filter((item) => {
+        const test = item[searchKey];
+        if (typeof test === 'string') {
+          return test.includes(searchTerm);
+        }
+      }),
+    [items, searchKey, searchTerm]
+  );
```
calebpollman, line 52: Can we handle this logic in the list handler so we don't have to track `searchTerm` in state?

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const debouncedHandler = React.useMemo(
+    () =>
+      debounce((term: string, includeSubfolders?: boolean) => {
```
calebpollman, line 56: Know we chatted offline but still not quite sure we need to add debouncing here. Our default implementation only runs `onSearch` on button click, and the button should be disabled while a search is in progress

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import debounce from 'lodash/debounce';
+
+const DEFAULT_DELAY = 300;
+export interface InitialSearchValues<T> {
+  searchKey: keyof T;
```
calebpollman, line 6: Not quite sure what `searchKey` is used for or why it's required? Also why `keyof`, afaict from the unit tests, this could just be `string`?

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
calebpollman: Prefer `useRef` for these scenarios

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: `searchKey` is the key used for filtering. e.g LocationsView will provide `prefix` to key into a `LocationData` object to find a value to filter on, and LocationDetail will pass `key` as a searchKey.


`keyof T` is more accurate than using `string` because we want to restrict it to keys in the provided type i.e. LocationData or LocationItem.

example

```
useLocationsView.ts

// this will filter items using the `prefix` key in LocationData. 
// we can make sure the `searchKey` is not an arbitrary string by limiting it to using keyof T`
const { filteredItems } = useSearch<LocationData>({ initialValues: { searchKey: 'prefix' }})


useLocationDetailsView.ts

const { filteredItems } = useSearch<LocationItem>({ initialValues: { searchKey: 'key' }})

```

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: I think it's better to handle it in useSearch because that allows us to consolidate logic for regular search and "subfolder" search. We don't always call the list handler.

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: I'll remove it for now, just to simplify this given our current UI. It's easy enough to add later.

packages/react-storage/src/components/StorageBrowser/views/hooks/useSearch.ts
pranavosu: Another option is to always call list handlers on search, and use previous state to search without subfolders. But we'll couple search with data fetching.

pranavosu:
Closing in favor of https://github.com/aws-amplify/amplify-ui/pull/5995

PR #5975, opened 2024-10-29
feat(storage-browser): expose overwrite prevented result

HuiSF:
#### Description of changes

Added `OVERWRITE_PREVENTED` as a task result in addition to `'COMPLETE' | 'FAILED' | 'CANCELED'`, and corresponding for rendering the status in a table.

1. manual testing with the example app
2. unit tests

#### Checklist

- [x] PR description included


packages/react-storage/src/components/StorageBrowser/views/LocationActionView/constants.ts
```diff
@@ -14,6 +15,7 @@ export const STATUS_DISPLAY_VALUES = {
   CANCELED: 'Canceled',
   COMPLETE: 'Completed',
   FAILED: 'Failed',
+  OVERWRITE_PREVENTED: 'Overwrite Prevented',
```
calebpollman, line 18: Overwrite prevention is specific to uploads and should not be included in the generic statuses

HuiSF: But the status is generic to the status column, no?

calebpollman: If we add it to the status column then we need to add it to the status counts and allow for conditionally hiding that column as it is not pertinent to most actions. IMO we should just be returning an overwrite prevention specific message from the `upload` handler on failure

calebpollman (approved):

PR #5985, opened 2024-10-31
feat(storage-browser): add customEndpoint option to createStorageBrowser and managedAuthAdapter

ashwinkumar6:
#### Description of changes
- Adds advanced customEndpoint option to storage-browser component. 
DX (finalization pending!):
```
const { StorageBrowser } = createStorageBrowser({
  config: {
    ...createManagedAuthAdapter({
      credentialsProvider: <credentialsProvider>,
      region: <region>,
      accountId: <accountId>,
      registerAuthListener: <registerAuthListener>,
      customEndpoint:  's3-control.dualstack.us-east-2.amazonaws.com' // <== s3 control plane endpoint
    }),
    customEndpoint: 's3-accelerate.amazonaws.com', // <=== s3 data plane endpoint
  }
});
```

- managedAuthAdapter takes customEndpoint input for both data and control plane APIs and returns customEndpoint only for data plane APIs which consumed by createStorageBrowser in the DX above

Tested with [Amplify JS customEndpoint PR](https://github.com/aws-amplify/amplify-js/pull/13961). Validated the baseEndpoint of the following internal APIs can successfully be overridden
 - list
 - remove
 - uploadData
 - getUrl
 - copy
 - listLocation
- getDataAccess

(downloadData and getProperties APIs aren't used)

#### Checklist

- [x] PR description included


AllanZhengYP (commented):
As talked offline, the s3 data plane endpoint may make more sense to exist in the base config. something like this(just an idea)
```
const { StorageBrowser } = createStorageBrowser({
  config: {
    ...createManagedAuthAdapter({
      credentialsProvider: <credentialsProvider>,
      region: <region>,
      accountId: <accountId>,
      registerAuthListener: <registerAuthListener>,
      customEndpoint:  's3-control.dualstack.us-east-2.amazonaws.com' // <== s3 control plane endpoint
    }),
    customEndpoint: 's3-accelerate.amazonaws.com', // <=== s3 data plane endpoint
  }
});
```

packages/react-storage/src/components/StorageBrowser/adapters/createManagedAuthAdapter/createManagedAuthAdapter.ts
```diff
@@ -13,24 +13,28 @@ import {
...
 
   const getLocationCredentials = createLocationCredentialsHandler({
     credentialsProvider,
     accountId,
+    customEndpoint,
     region,
   });
 
   return {
+    accountId,
```
ashwinkumar6, line 37: required for bucket owner

AllanZhengYP (approved):
The change makes sense to me.

packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/listLocations.spec.ts
```diff
@@ -63,6 +64,14 @@ describe('listLocationsHandler', () => {
     );
     expect(result.nextToken).toBeUndefined();
     expect(mockListCallerAccessGrants).toHaveBeenCalledTimes(1);
+    expect(mockListCallerAccessGrants).toHaveBeenCalledWith({
```
ashika112, line 67: nit: i dont think test like this where we are call with an input config makes much sense unless we are manipulating some things along the way. like ex, im calling multiple pages, then on recursion `nexTToken` will change then we will want to check if it is using a newNextToken. Something like that IMO. Not gonna hurt but should avoid tests like this

packages/react-storage/src/components/StorageBrowser/do-not-import-from-here/types.ts
```diff
@@ -20,6 +20,7 @@ export interface LocationConfig {
   accountId?: string;
   bucket: string;
   credentialsProvider: LocationCredentialsProvider;
+  customEndpoint?: string;
```
ashika112, line 23: nit: is the type from here still used somewhere in ur PR?

ashika112 (approved):

packages/react-storage/src/components/StorageBrowser/do-not-import-from-here/types.ts
ashwinkumar6: Good question, yep,
`LocationConfig` is referenced in `TaskActionInput`, `ListActionInput` and `DownloadActionInput` types on this file. It's used on all the `do-not-import-from-here/actions/*`files ([example](https://github.com/aws-amplify/amplify-ui/pull/5985/files#diff-d90dfff69123a63827049eccb3911e149f1592a2ce02a775450de4bb9618b0c5R37))

Currently we introduced the `customEndpoint?` type in two places `StorageBrowser/do-not-import-from-here/actions/` and `StorageBrowser/actions/handlers/*`. Some views use the former while others use the latter.

We're in the process of migrating all former ones to latter hence introduced the type in both places to be on the safer side

packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/listLocations.spec.ts
ashwinkumar6: Discussed async, Adds value to check if underlying service call is made with the correct params if there's additional logic in between. In this case all action handlers are directly calling the service underneath.

We could potentially remove `toHaveBeenCalledWith` for all action handlers

ashwinkumar6: EDIT:
Discussed async, since we're moving things around and in future these handlers could hold additional logic, it makes sense to have this check

PR #5990, opened 2024-10-31
chore(storage-browser): refactor Navigation for composability

cshfang:
#### Description of changes

This PR refactors the breadcrumb navigation to for composability. Due to recent changes allowing the StorageBrowser to be initialized at a particular location and path, this necessitated sweeping changes to how navigation is handled. Namely, the useHistory hook has been renamed to useLocation and no longer keeps track of previously visited paths. Instead, breadcrumb navigation items are always generated based on the current location and path.

#### Checklist

- [x] PR description included


packages/react-storage/src/components/StorageBrowser/views/Controls/Table.tsx
```diff
@@ -316,6 +316,7 @@ export const LocationDetailViewTable = ({
   // @TODO: This should be it's own component instead of using `useCallback`
   const renderRowItem: RenderRowItem<LocationItemData> = React.useCallback(
     (row, index) => {
+      const { prefix = '' } = current ?? {};
```
calebpollman, line 319: If we are only using `current` in the body of `renderRowItem` think we can just destructure out `prefix` closer to line 201 and add a comment there explaining the need for the defaulting `prefix` and `path` to empty strings

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DeleteView/useDeleteView.ts
```diff
@@ -16,12 +16,13 @@ export const useDeleteView = ({
...
   const [
     {
-      history,
+      location,
       locationItems: { fileDataItems: selected },
     },
     dispatchStoreAction,
   ] = useStore();
-  const { current } = history;
+  const { current, path = '' } = location;
```
calebpollman, line 24: Might require deeper thought on the default values of `location.current`, but wondering if `path` should default to an empty string in the `LocationProvider` logic? Feels somewhat fragile from a maintenance perspective to fallback to a default value in consumers, would also reduce the amount of branches for unit testing

packages/react-storage/src/components/StorageBrowser/providers/store/location/context.tsx
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * current `location` metadata
+   */
+  current: LocationData | undefined;
+
+  /**
+   * current `location` subpath within a `location`
+   */
+  path: string | undefined;
+}
```
calebpollman, line 52: After going through the PR thinking we should add `key` to this interface along with `path` to avoid needing to concatenate `current.prefix` and `path` throughout the consumer code. `key` being the entire string following the `bucket`, e.g. `${current.prefix}${path}`

packages/react-storage/src/components/StorageBrowser/components/BreadcrumbNavigation.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export interface BreadcrumbProps {
+  isCurrent?: boolean;
+  name?: string;
+  onNavigate?: () => void;
+}
+
+interface BreadcrumbNavigationProps {
+  breadcrumbs: BreadcrumbProps[];
+}
```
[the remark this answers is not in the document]
cshfang: I originally had just Breadcrumbs but [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Breadcrumb_Navigation) referred to it as Breadcrumb navigation so I went with that

packages/react-storage/src/components/StorageBrowser/composables/Navigation.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface NavigationItem {
+  isCurrent?: boolean;
+  name?: string;
+  onNavigate?: () => void;
+}
+
+export interface NavigationProps {
+  items: NavigationItem[];
+}
```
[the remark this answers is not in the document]
cshfang: I can see it both ways. There are times where contexualized prop names intuitively makes sense and times like this where there's not really a better name abstraction.

packages/react-storage/src/components/StorageBrowser/controls/NavigationControl.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const NavigationControl = ({
+  className,
+}: ControlProps): React.JSX.Element | null => {
+  const props = useNavigation();
+
+  const ResolvedNavigation = useResolvedComposable(Navigation, 'Navigation');
+
+  if (!props) {
+    return null;
+  }
```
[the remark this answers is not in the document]
cshfang: Yeah I had the same thought when I was writing tests on this one. I think it's probably fair to just pass down empty props at this level and making the composables decide what to do with it. But I think this is another one we should do in a separate PR across multiple components

packages/react-storage/src/components/StorageBrowser/providers/store/location/context.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export const DEFAULT_STATE: LocationState = {
+  current: undefined,
+  path: undefined,
+};
+
+export type LocationActionType = {
+  type: 'NAVIGATE';
+  location?: LocationData;
+  path?: string;
```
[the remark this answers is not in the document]
cshfang: As part of the navigate action type, it kind of just unnecessarily requires an explicit `path: undefined` when just omitting it would do. I think making the state provided path default to `''` will be more helpful to consumers of the state

packages/react-storage/src/components/StorageBrowser/views/Controls/Table.tsx
cshfang: I think this was more to avoid having to track another item in the dependencies list when already tracking current. But I can make this change for now - this should all go away with table integration anyway

Samaritan1011001 (commented):
ActionStart related changes lgtm

packages/react-storage/src/components/StorageBrowser/providers/store/location/context.tsx
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * ```
+ * const state: LocationState = {
+ *   location: {
+ *     bucket: 'my-bucket',
+ *     permisson: 'READWRITE',
+ *     prefix: 'my/prefix/',
+ *     type: 'PREFIX',
+ *   },
+ *   path: 'a-path/a-nested-path/',
```
calebpollman, line 41:
```suggestion
 *   path: 'a-path/a-nested-path/',
 *   key: 'my/prefix/a-path/a-nested-path/',
```

packages/react-storage/src/components/StorageBrowser/controls/hooks/useNavigation.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const useNavigation = (): NavigationProps | null => {
+  const { data, onNavigate, onNavigateHome } = useControlsContext();
+  const { location } = data;
+
+  const { current, path = '' } = location ?? {};
+
+  return React.useMemo(() => {
+    if (!current) {
+      return null;
```
pranavosu, line 13: Can we return an empty breadcrumb or maybe `[{ name: 'Home', onNavigate: onNavigateHome }]`  as on line 35? consumers of the hook could then avoid null checks.

calebpollman (approved):

pranavosu (approved):

PR #5991, opened 2024-11-01
fix(react-storage): nested folder uploads

Samaritan1011001:
#### Description of changes
Fixes a bug where folder uploads got uploaded to the base location where upload was done rather than the folder structure path. 
We do so by using the `webKitRelativepath` value when available which gives us the complete folder structure path.

Manual tests and unit tests
- Upload a folder with nested structure.
- See that the uploaded files are in the same nested structure.

#### Checklist

- [x] PR description included


packages/react-storage/src/components/StorageBrowser/actions/handlers/upload.ts
```diff
@@ -46,8 +46,11 @@ export const uploadHandler: UploadHandler = ({
 
   const bucket = constructBucket(config);
 
+  const resolvedKey = isEmpty(data.webkitRelativePath)
+    ? key
+    : data.webkitRelativePath;
```
calebpollman, line 51: Correct approach but should be applied where the `key` is create and set in the store on file so consumers (override `uploadHandler` functions) don't need to introduce this logic

AllanZhengYP: TIL, I think a similar fix could be added to upload cache. Although unrelated to the StoBro work.

packages/react-storage/src/components/StorageBrowser/views/Controls/Navigate.tsx
```diff
@@ -115,8 +115,9 @@ export function NavigateControl({
       </NavigateItem>
       {previous?.map((destination, index) => {
         const { bucket, id, prefix: _prefix } = destination;
-        // remove trailing `/` from `prefix`
-        const prefix = _prefix?.endsWith('/') ? _prefix.slice(0, -1) : _prefix;
+
+        // remove trailing `/` from `prefix` and use only the last segment since we only want current location
+        const prefix = _prefix?.replace(/\/$/, '').split('/').pop() ?? '';
```
Samaritan1011001, line 120: This change is required to show the correct location when navigating to nested folders. It extracts only the last segment and removes the trailing `/`, if nothing is present falls back to empty string.

packages/react-storage/src/components/StorageBrowser/providers/store/files/utils.ts
```diff
@@ -21,7 +21,16 @@ export const resolveFiles = (
           name === file.name && webkitRelativePath === file.webkitRelativePath
       )
         ? items
-        : [...items, { key: file.name, id: crypto.randomUUID(), item: file }],
+        : [
+            ...items,
+            {
+              key: isEmpty(file.webkitRelativePath)
```
reesscot, line 27: Can we add a unit test for this?

packages/react-storage/src/components/StorageBrowser/views/Controls/Navigate.tsx
reesscot: How is this related to the folder uploads bug?

calebpollman: +1. Don't think we need this and also think it might be getting touched in https://github.com/aws-amplify/amplify-ui/pull/5990

Samaritan1011001: Got it,  I thought it was due to my change, if it exists regardless I won't update it.

packages/react-storage/src/components/StorageBrowser/providers/store/files/utils.ts
Samaritan1011001: Done

packages/react-storage/src/components/StorageBrowser/providers/store/files/utils.ts
```diff
@@ -21,7 +21,16 @@ export const resolveFiles = (
...
+                ? file.name
+                : file.webkitRelativePath,
```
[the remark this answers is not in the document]
Samaritan1011001: Yeah it does.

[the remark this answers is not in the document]
reesscot: I also confirmed this in the api docs

reesscot (approved):

calebpollman (approved):
🚢

PR #5993, opened 2024-11-01
chore: refactor search to add composability support

jordanvn:
#### Description of changes
* Add new Search composable and control based on existing search field
* Add relevant tests
* Update styling
Screenshot 2024-11-04 at 1 28 27 PM

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/context/elements/definitions.ts
```diff
@@ -182,6 +182,9 @@ export const InputElement = defineBaseElement<
   | 'onFocus'
   | 'checked'
   | 'defaultChecked'
+  | 'placeholder'
+  | 'value'
+  | 'ref'
```
jordanvn, line 187: I believe this change is causing tests to fail. Needs a second thought to consider if it is necessary and if so what other changes should be made

packages/react-storage/src/styles/storage-browser.css
jordanvn, line 1: These styles are not final. They get the search field close to how it should look, but should be replaced by styling more in line with defined patterns and more exact.

packages/react-storage/src/components/StorageBrowser/context/elements/definitions.ts
calebpollman: Yes that would be an issue because there is no `forwardRef` wrapped around the underlying `input`, why do need it?

jordanvn: I was including `ref` to use with the input field for the purpose of returning focus to the input after a user clicks the button to clear the current input

jordanvn: Will address this in a follow-up task

packages/react-storage/src/styles/storage-browser.css
```diff
@@ -483,3 +483,49 @@
...
+
+/* Search view styles */
+.storage-browser__search {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  min-width: 35rem;
+}
+
+.storage-browser__search__field-clear-button {
```
reesscot, line 495: This is fine for now, but I don't think we should have styling for search other than layout between the search field and the subfolders button, as this component will be replaced by the SearchField https://ui.docs.amplify.aws/react/components/searchfield

reesscot (dismissed):
Let's revisit this component after fit n finish, as we shouldn't be recreating the SearchField component.

reesscot (approved):

pranavosu (approved):

reesscot (approved):

PR #5982, opened 2024-10-30
feat: storage browser copy action view

reesscot:
#### Checklist

- [ ] Have read the [Pull Request Guidelines](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md)
- [ ] PR description included
- [ ] `yarn test` passes and tests are updated/added
- [ ] PR title and commit messages follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax
- [ ] _If this change should result in a version bump_, [changeset added](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) (This can be done after creating the PR.) This does not apply to changes made to `docs`, `e2e`, `examples`, or other private packages.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5994, opened 2024-11-01
feat: enhanced list actions integration

pranavosu:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/__tests__/createEnhancedHandler.spec.ts
+++ b/packages/react-storage/src/components/StorageBrowser/__tests__/createEnhancedHandler.spec.ts
@@ -1,0 +1,34 @@
+import { createEnhancedListHandler } from '../actions/createEnhancedHandler';
+
+const mockAction = jest.fn();
+
+describe('createEnhancedHandler', () => {
+  it('should call action', async () => {
+    const handler = createEnhancedListHandler(mockAction);
+    mockAction.mockReturnValue({
+      items: [],
+      nextToken: undefined,
+    });
+
+    const result = await handler(
+      {
+        nextToken: undefined,
+        items: [],
+      },
+      {
+        config: {
+          bucket: '',
+          credentials: jest.fn(),
+          region: '',
+        },
+        prefix: 'a_prefix',
+      }
+    );
+
+    expect(mockAction).toHaveBeenCalled();
+    expect(result).toEqual({
+      items: [],
+      nextToken: undefined,
+    });
+  });
+});
--- a/packages/react-storage/src/components/StorageBrowser/actions/createEnhancedHandler.ts
+++ b/packages/react-storage/src/components/StorageBrowser/actions/createEnhancedHandler.ts
@@ -1,0 +1,86 @@
+import { AsyncDataAction } from '@aws-amplify/ui-react-core';
+import {
+  ListHandler,
+  ListHandlerOptions,
+  ListHandlerInput,
+  ListHandlerOutput,
+} from './types';
+
+interface SearchOptions<K> {
+  query: string;
+  filterKey: keyof K;
+}
+
+export interface EnhancedListHandlerOptions<K> extends ListHandlerOptions {
+  refresh?: boolean;
+  reset?: boolean;
+  search?: SearchOptions<K>;
+}
+
+interface EnhancedListHandler<T, K>
+  extends AsyncDataAction<
+    ListHandlerOutput<K>,
+    ListHandlerInput<EnhancedListHandlerOptions<K> & T>
+  > {}
+
+function filter<K>(items: K[], key: keyof K, term: string) {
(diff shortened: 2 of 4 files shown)


packages/react-storage/src/components/StorageBrowser/__tests__/createEnhancedHandler.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { createEnhancedListHandler } from '../actions/createEnhancedHandler';
+
+const mockAction = jest.fn();
+
+describe('createEnhancedHandler', () => {
```
pranavosu, line 5: wip

packages/react-storage/src/components/StorageBrowser/actions/createEnhancedHandler.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  });
+}
+
+export const SEARCH_LIMIT = 10000;
+export const createEnhancedListHandler = <
+  T extends ListHandlerOptions<I>,
+  K,
+  I = never,
+>(
```
pranavosu, line 40: I'll try to clean up these types, `<T, K, I = never`  might be excessive.

packages/react-storage/src/components/StorageBrowser/actions/createEnhancedHandler.ts
```diff
@@ -0,0 +1,86 @@
...
+  search?: SearchOptions<K>;
+}
+
+interface EnhancedListHandler<T, K>
+  extends AsyncDataAction<
+    ListHandlerOutput<K>,
+    ListHandlerInput<EnhancedListHandlerOptions<K> & T>
+  > {}
+
+function filter<K>(items: K[], key: keyof K, term: string) {
```
calebpollman, line 26: Any reason we can't inline the _.filter_ call on line 69?

packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
```diff
@@ -42,4 +48,32 @@ export interface ListLocationItemsHandler
   > {}
 
 export const listLocationItemsHandler: ListLocationItemsHandler =
-  null as unknown as ListLocationItemsHandler;
+  async function listLocationItemsAction(
+    input: ListLocationItemsHandlerInput
+  ): Promise<ListLocationItemsHandlerOutput> {
```
calebpollman, line 53: Don't think we need to declare this function?

packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
```diff
@@ -42,4 +48,32 @@ export interface ListLocationItemsHandler
...
+      options: {
+        bucket,
+        expectedBucketOwner: accountId,
+        nextToken,
+        pageSize,
+        subpathStrategy,
+      },
+    };
+
+    const output = await list(listInput);
```
calebpollman, line 76: We are missing logic that is built in to the [current handler](https://github.com/aws-amplify/amplify-ui/blob/feat-storage-browser/main/packages/react-storage/src/components/StorageBrowser/do-not-import-from-here/actions/listLocationItemsAction.ts#L80) that we could extend for search recursion

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -102,20 +120,30 @@ export function useLocationDetailView(
...
+    handleList({ config, prefix, options: { ...listOptions, refresh: true } });
     dispatchStoreAction({ type: 'RESET_LOCATION_ITEMS' });
   };
 
+  const onSearch = (query: string, includeSubfolders: boolean) => {
+    if (hasInvalidPrefix) return;
+    const searchOptions = {
+      ...listOptions,
+      delimiter: includeSubfolders ? undefined : listOptions.delimiter,
+      search: { query, filterKey: 'key' as const },
```
calebpollman, line 132: Will `key` always be the `filterKey`? If so do we need to pass it an option at all?

pranavosu: It could be different for each view. for example: List Locations uses the filterKey `prefix`. I think we might actually want to make this a list to support multi column search, but we don't have that req currently.

packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
pranavosu: Plan to remove this handler from the PR. @reesscot 's [PR](https://github.com/aws-amplify/amplify-ui/pull/5982) will add the new handler.

PR #6003, opened 2024-11-04
chore(storage-browser): update task handling

calebpollman:
#### Description of changes
- fix cancel handling and progress support in `useProcessTasks`
- update handler types to allow stabilize `items` references passed to `useProcessTasks`
- update and align handler types

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/views/LocationActionView/UploadControls.tsx
```diff
@@ -232,9 +223,11 @@ export const UploadControls = ({
     };
   }, [dispatchStoreAction]);
 
-  const [tasks, handleProcess] = useProcessTasks(uploadHandler, files, {
-    concurrency: 4,
-  });
+  const [{ tasks, isProcessing }, handleProcess] = useProcessTasks(
+    uploadHandler,
+    files,
```
reesscot, line 228: Temporary workaround to make the order of the tasks match the sort order. Tested this out and it makes the upload progress order make sense.
```suggestion
    files?.sort(({ key: key1 }, { key: key2 }) => compareStrings(key1, key2)),
```

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DeleteView/useDeleteView.ts
```diff
@@ -17,21 +17,17 @@ export const useDeleteView = ({
...
   ] = useStore();
-  const { current, key } = location;
+  const { current } = location;
 
   const getInput = useGetActionInput();
 
-  const [tasks, handleProcess] = useProcessTasks(
+  const [{ tasks }, handleProcess] = useProcessTasks(
     deleteHandler,
-    // @ts-expect-error
```
reesscot, line 30: nice!

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/UploadControls.tsx
calebpollman: Can we do the sorting on the returned values? Would prefer not to mess with them before they get added as tasks

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DeleteView/useDeleteView.ts
```diff
@@ -40,10 +36,7 @@ export const useDeleteView = ({
 
   const onActionStart = () => {
     if (!current) return;
-    handleProcess({
-      config: getInput(),
-      prefix: key,
-    });
+    handleProcess({ config: getInput() });
```
reesscot, line 39: prefix is no longer needed because we have the full key, right?

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/utils.ts
```diff
@@ -74,62 +78,47 @@ export const getDeleteActionViewTableData = ({
...
+              content: { text: getFileTypeDisplayValue(item.data.key) },
             };
           }
           case 'size':
             return {
               key,
               type: 'text',
-              content: {
-                text: humanFileSize(
-                  parseInt((item as Task & { size: string }).size),
```
reesscot, line 121: So much great cleanup!

reesscot (approved):

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/UploadControls.tsx
reesscot: I don't think so because if you don't sort before creating the tasks, then the order they will execute in will not match the table order. That's why the UI upload behavior feels so random, because it's the order the tasks were originally created in.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DeleteView/useDeleteView.ts
calebpollman: Right, `key` is directly pulled from the list output items

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/UploadControls.tsx
calebpollman: Cool, we can add it as a follow up. MIght make sense to just handle the sorting in the reducer

pranavosu (approved):

PR #6005, opened 2024-11-05
feat: integrate new actions and search

pranavosu:
#### Description of changes

- Integrate new list location details action
- Integrate search with list locations and list location detail view
- Update unit tests

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/views/LocationsView/useLocationsView.ts
```diff
@@ -89,6 +91,10 @@ export function useLocationsView(
     return result.slice(start, end);
   }, [range, result]);
 
+  const filteredItems = React.useMemo(() => {
```
pranavosu, line 94: filtering in the view for now, i'll follow up with list locations integration once the adapters support the new interface.

packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
```diff
@@ -92,14 +92,13 @@ export const listLocationItemsHandler: ListLocationItemsHandler = async (
   const { bucket: _bucket, credentials, region } = config;
 
   const {
+    exclude: excludedType,
```
reesscot, line 95: We can just update the use of excludedType below
```suggestion
    exclude,
```

packages/react-storage/src/components/StorageBrowser/composables/__tests__/Search.spec.tsx
```diff
@@ -48,9 +54,9 @@ describe('Search', () => {
   it('searches for the term typed into the field', async () => {
     const onSearch = jest.fn();
 
-    render(<Search onSearch={onSearch} />);
+    render(<Search onSearch={onSearch} searchPlaceholder={'Placeholder'} />);
```
reesscot, line 57: nit:
```suggestion
    render(<Search onSearch={onSearch} searchPlaceholder="Placeholder" />);
```

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/__tests__/useLocationDetailView.spec.tsx
```diff
@@ -284,4 +300,33 @@ describe('useLocationsView', () => {
...
+    const handleListMock = jest.fn();
+    useActionSpy.mockReturnValue([mockDataState, handleListMock]);
+
+    const { result } = renderHook(() => useLocationDetailView());
+    act(() => {
+      const state = result.current;
+      state.onSearch('moo', true);
+    });
+
+    //
```
reesscot, line 321: nit: extra comment
```suggestion
```

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -163,12 +198,23 @@ export function useLocationDetailView(
...
+    onSearch: (query, includeSubfolders) => {
+      if (hasInvalidPrefix) return;
+      const searchOptions = {
+        ...listOptions,
+        delimiter: includeSubfolders ? undefined : listOptions.delimiter,
+        search: { query, filterKey: 'key' as const },
+      };
+      handleReset();
+      handleList({ config: getConfig(), prefix, options: searchOptions });
+    },
```
reesscot, line 218: We probably need to reset selection state as well in case they selected files we're going to replace.

```suggestion
      dispatchStoreAction({ type: 'RESET_LOCATION_ITEMS' });
    },
```

Not sure if ` dispatchStoreAction({ type: 'RESET_ACTION_TYPE' });` might also be necessary?

packages/react-storage/src/styles/storage-browser.css
```diff
@@ -529,3 +530,8 @@
 .storage-browser__search__submit-button:not(:focus) {
   border-inline-start-color: transparent;
 }
+
+.storage-browser__locations-view-search,
```
reesscot, line 534: Maybe just leave CSS out for now, Danny overhauled a lot of the CSS in my copy view pr.

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
```diff
@@ -18,9 +18,15 @@ const TOGGLE_BLOCK = 'toggle';
 
 export interface SearchProps {
   onSearch: (term: string, includeSubfolders: boolean) => void;
+  searchPlaceholder?: string;
+  showIncludeSubfolders?: boolean;
```
calebpollman, line 22: Rather than add an prop that is specific to a singular `View` that will be surfaced on the composable API for both list View components, can we provide the toggle `Checkbox` to `Search` through composition?

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
```diff
@@ -18,9 +18,15 @@ const TOGGLE_BLOCK = 'toggle';
 
 export interface SearchProps {
   onSearch: (term: string, includeSubfolders: boolean) => void;
+  searchPlaceholder?: string;
```
calebpollman, line 21: Noted elsewhere, the `searchPlaceholder` changes should be removed

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/Controls.tsx
```diff
@@ -40,22 +39,20 @@ function Loading({ show }: { show?: boolean }) {
...
-export const LocationDetailMessage = (): React.JSX.Element | null => {
-  const [{ hasError, message }] = useAction('LIST_LOCATION_ITEMS');
-
-  return hasError ? (
+export const LocationDetailMessage = ({
+  show,
+  message,
+}: {
+  show?: boolean;
+  message?: string;
```
calebpollman, line 47: Don't need `show`, can gate rendering based on the presence of `message`

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/Controls.tsx
```diff
@@ -40,22 +39,20 @@ function Loading({ show }: { show?: boolean }) {
...
-  const [{ data, hasError, isLoading }] = useAction('LIST_LOCATION_ITEMS');
-  const shouldShowEmptyMessage =
-    data.result.length === 0 && !isLoading && !hasError;
-
-  return shouldShowEmptyMessage ? (
-    <EmptyMessage>No items to show.</EmptyMessage>
-  ) : null;
+const LocationDetailEmptyMessage = ({ show }: { show?: boolean }) => {
+  return show ? <EmptyMessage>No items to show.</EmptyMessage> : null;
 };
```
calebpollman, line 56: We should only need one `Message` component

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/Controls.tsx
```diff
@@ -72,24 +69,34 @@ export const LocationDetailViewControls = ({
     page,
     isPaginatePreviousDisabled,
     isPaginateNextDisabled,
+    showIncludeSubfolders,
```
calebpollman, line 72: This is not a configurable property

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
```diff
@@ -22,26 +22,24 @@ const {
...
-  const [{ data, isLoading, hasError }] = useLocationsData();
-  const shouldShowEmptyMessage =
-    data.result.length === 0 && !isLoading && !hasError;
-
-  return shouldShowEmptyMessage ? (
-    <EmptyMessage>No locations to show.</EmptyMessage>
-  ) : null;
+const LocationsEmptyMessage = ({ show }: { show: boolean }) => {
+  return show ? <EmptyMessage>No locations to show.</EmptyMessage> : null;
 };
```
calebpollman, line 43: The empty message value should be provided to the `Message` component dynamically

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
```diff
@@ -51,21 +49,28 @@ export function LocationsView({
   const {
     pageItems,
     hasError,
+    message,
     isPaginatePreviousDisabled,
     isPaginateNextDisabled,
     page,
     isLoading,
+    searchPlaceholder,
```
calebpollman, line 57: Just noting that the same feedback related to `searchPlaceholder` in `LocationsView` applies here as well

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -55,6 +68,12 @@ export const DEFAULT_LIST_OPTIONS = {
   pageSize: DEFAULT_PAGE_SIZE,
 };
 
+const listLocationItemsAction = createEnhancedListHandler<
+  ListLocationItemsHandlerOptions,
+  LocationItemData,
+  LocationItemType
+>(listLocationItemsHandler);
```
calebpollman, line 75: Shouldn't need to provide these types as generics, if there is an issue with `createEnhancedListHandler` not being able to infer these types from `listLocationItemsHandler` we should address that in the util

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
pranavosu: We have different placeholders for LocationsView, LocationDetailView and CopyView. Is there a better way to handle that?

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/Controls.tsx
pranavosu: I thought about that but we have a `DEFAULT_ERROR_MESSAGE `

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
calebpollman: That's the use case for the controls context, it allows for providing low level props to the composables without impacting their top level APIs

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
pranavosu: Good call, i think we can treat it like a refresh, so `dispatchStoreAction({ type: 'RESET_ACTION_TYPE' });` might not be necessary unless we want to do that `onRefresh` as well.

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
pranavosu: I am providing the `searchPlaceholder` through ControlsContext in SearchControl [here](https://github.com/aws-amplify/amplify-ui/pull/6005/files#diff-c88f1b5bd17c3ba59ce3087ecf9abb609b23259d28b691374d65ebae22036a14R13), I don't think Search composable itself should use ControlContext ?

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/__tests__/useLocationDetailView.spec.tsx
```diff
@@ -327,5 +327,8 @@ describe('useLocationDetailView', () => {
       },
       prefix: 'item-b/',
     });
+    expect(handleStoreActionMock).toHaveBeenCalledWith({
+      type: 'RESET_LOCATION_ITEMS',
```
reesscot, line 331: Thank you!

reesscot (approved):

packages/react-storage/src/components/StorageBrowser/views/LocationsView/LocationsView.tsx
pranavosu: Can handle this in a follow up

packages/react-storage/package.json
```diff
@@ -71,7 +71,7 @@
       "name": "createStorageBrowser",
       "path": "dist/esm/browser.mjs",
       "import": "{ createStorageBrowser }",
-      "limit": "18.55 kB",
+      "limit": "19.10 kB",
```
reesscot, line 74: We are wasting so much time on this, should we just set it to a high number and revisit when we launch?

reesscot (dismissed):

packages/react-storage/package.json
pranavosu: Yea, not opposed to setting this to something reasonable. I think this is useful for catching big deviations, but we're constantly updating it.

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
calebpollman: We can follow up on this

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
calebpollman: Yep 👍🏼

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
```diff
@@ -119,16 +139,25 @@ export function useLocationDetailView(
...
+      config: getConfig(),
       prefix: key,
       options: { ...listOptions, refresh: true },
     });
     handleReset();
-  }, [handleList, handleReset, listOptions, hasInvalidPrefix, prefix, key]);
+  }, [
+    handleList,
+    handleReset,
+    listOptions,
```
calebpollman, line 150: `listOptions` is a ref so it shouldn't need to be in the deps array and `prefix` is not used in the effect afaict?

calebpollman (dismissed):

packages/react-storage/src/components/StorageBrowser/composables/Search.tsx
pranavosu: Thx, i'd like to understand this properly

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
calebpollman: `listOptions` changing would result in a rerun of the effect which is not the desired the behavior for an initial load?

reesscot (approved):

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
pranavosu: I agree, but we're never changing the ref in this hook. And eslint complains if i exclude `listOptions`

calebpollman (approved):

packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/useLocationDetailView.ts
calebpollman: Not needed to merge but that is due to assigning `listOptionsRef.current` to a variable. The dep can be removed by direct usage of the ref in the effect:
```
...listOptionsRef.current,
// other options
```

PR #6018, opened 2024-11-06
fix(storage-browser): Reintroduce folder conditional on drag and drop

cshfang:
#### Description of changes

This PR reintroduces a conditional which determines whether a drag and dropped items are files, folders or both. This determination subsequently triggers either routing to the upload file or folder actions. In the case of a mix of files/folders being dropped, the action will be treated as uploading files.

This should also fix the drag and drop e2e test which currently fails locally.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/__tests__/useLocationDetailView.spec.tsx
+++ b/packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/__tests__/useLocationDetailView.spec.tsx
@@ -342,7 +342,63 @@ describe('useLocationDetailView', () => {
       type: 'ADD_FILE_ITEMS',
       files: mockFiles,
     });
+    expect(handleStoreActionMock).toHaveBeenCalledWith({
+      type: 'SET_ACTION_TYPE',
+      actionType: 'UPLOAD_FILES',
+    });
   });
+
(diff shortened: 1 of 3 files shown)


packages/react-storage/src/components/StorageBrowser/views/LocationDetailView/__tests__/useLocationDetailView.spec.tsx
```diff
@@ -342,7 +342,63 @@ describe('useLocationDetailView', () => {
...
+      type: 'ADD_FILE_ITEMS',
+      files: [mockFile, mockFolder],
+    });
+    expect(handleStoreActionMock).toHaveBeenCalledWith({
+      type: 'SET_ACTION_TYPE',
+      actionType: 'UPLOAD_FILES',
+    });
+  });
+
+  it('should handle as files if adding files and folders', () => {
```
jordanvn, line 378: Are these tests named correctly? It looks like this test is handling as a folder and the previous test is handling as files

cshfang: Oops, good catch. Copy/pasted the test and updated the wrong one.

PR #6031, opened 2024-11-08
fix(textarea): autoresize textarea resets when value is empty

dbanksdesign:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/src/primitives/TextArea/useAutoresizeTextarea.ts
+++ b/packages/react/src/primitives/TextArea/useAutoresizeTextarea.ts
@@ -7,7 +7,7 @@ export const useAutoresizeTextArea = (
 ): void => {
   useEffect(() => {
     const resizeTextArea = () => {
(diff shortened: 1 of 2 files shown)


calebpollman (dismissed):

packages/react/src/primitives/TextArea/useAutoresizeTextarea.ts
```diff
@@ -7,7 +7,7 @@ export const useAutoresizeTextArea = (
 ): void => {
   useEffect(() => {
     const resizeTextArea = () => {
-      if (textAreaRef && value) {
+      if (textAreaRef) {
```
dbanksdesign, line 10: The bug was we were using value in the conditional here, value could be falsey and we still want to re-trigger the effect. So value is in the dependency array, but removing it from here  to fix the bug

calebpollman (approved):

PR #6032, opened 2024-11-08
chore(storage-browser): add styles in UI package

dbanksdesign:
#### Checklist

- [ ] PR description included


packages/ui/src/theme/components/index.ts
```diff
@@ -136,6 +138,7 @@ export type AllComponentThemes = {
   selectfield: SelectFieldTheme;
   sliderfield: SliderFieldTheme;
   stepperfield: StepperFieldTheme;
+  storagebrowser: StorageBrowserTheme;
```
calebpollman, line 141: Would prefer we don't couple StorageBrowser styles with the primitive themes, is that possible?

dbanksdesign: Let me think about it, we have other connected components in there too and it is only pure types

calebpollman: Just a separation of concerns thing, ideally we scope the theme types to the components in their respective directories

calebpollman (approved):

PR #6038, opened 2024-11-08
feat: add create folder, copy and delete views display text

dindjarinjs:
#### Description of changes
Add display text for create folder, delete and copy views.


packages/react-storage/src/components/StorageBrowser/displayText/types.ts
```diff
@@ -52,6 +52,26 @@ export interface DefaultLocationDetailViewDisplayText
...
 }
 
+/**
+ * Action view display text values
+ */
+
+export interface DefaultCreateFolderViewDisplayText
+  extends DefaultActionViewDisplayText {
+  getValidationMessage: (error: Error) => string;
+  getActionFailedMessage: (error: Error) => string;
```
calebpollman, line 62: Should be covered by `getActionCompleteMessage`?

packages/react-storage/src/components/StorageBrowser/displayText/types.ts
```diff
@@ -52,6 +52,26 @@ export interface DefaultLocationDetailViewDisplayText
   tableColumnTypeHeader: string;
 }
 
+/**
+ * Action view display text values
+ */
+
+export interface DefaultCreateFolderViewDisplayText
+  extends DefaultActionViewDisplayText {
+  getValidationMessage: (error: Error) => string;
```
calebpollman, line 61: We don't throw an error on validation failures, but do think it makes sense to use a callback here. Maybe we can just provide the folder name string?

packages/react-storage/src/components/StorageBrowser/displayText/libraries/en/createFolderView.ts
```diff
@@ -0,0 +1,19 @@
...
+import { DefaultCreateFolderViewDisplayText } from '../../types';
+
+export const DEFAULT_CREATE_FOLDER_VIEW_DISPLAY_TEXT: DefaultCreateFolderViewDisplayText =
+  {
+    ...DEFAULT_ACTION_VIEW_DISPLAY_TEXT,
+    title: 'Create folder',
+    actionStartLabel: 'Create folder',
+    getValidationMessage: (_error) => {
+      // @TODO: maybe add logic to check which character caused the failure
+      return `Folder name must be at least one character and cannot contain a "/", ".", or ".."`;
```
calebpollman, line 11: Think we can drop the piece regarding a single character since we disable the start button for an empty string, does that make sense?

calebpollman (approved):

dbanksdesign (approved):

PR #6051, opened 2024-11-11
fix(react-storage): filter out only dot or slash items

Samaritan1011001:
#### Description of changes
Filter out display items that have only "/" or "." or ".." as their key so it cannot be navigated into.

Manual tests and unit tests

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
+++ b/packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
@@ -62,13 +62,28 @@ const parseItems = (
 const parseExcludedPaths = (paths: string[] | undefined): LocationItemData[] =>
   paths?.map((key) => ({ key, id: crypto.randomUUID(), type: 'FOLDER' })) ?? [];
 
+export const filterDotItems = (
+  items: LocationItemData[],
+  prefix: string
+): LocationItemData[] =>
+  items.filter(
+    // matches strings that are exactly either a single forward slash ("/") or one to two dots ("." or ".."), nothing else.
+    (item) =>
+      !(
+        item.key.startsWith(prefix)
+          ? item.key.substring(prefix.length)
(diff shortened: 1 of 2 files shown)


packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
```diff
@@ -62,13 +62,28 @@ const parseItems = (
 const parseExcludedPaths = (paths: string[] | undefined): LocationItemData[] =>
   paths?.map((key) => ({ key, id: crypto.randomUUID(), type: 'FOLDER' })) ?? [];
 
+export const filterDotItems = (
```
calebpollman, line 65: `parseItems` already filters out items with `path` values equivalent to `/` online 45 of this file, we either need to replace that filtering or extend it to include the additional values in this util

packages/react-storage/src/components/StorageBrowser/actions/handlers/listLocationItems.ts
```diff
@@ -62,13 +62,28 @@ const parseItems = (
...
   prefix: string
-): LocationItemData[] => [
-  ...parseExcludedPaths(excludedSubpaths),
-  ...parseItems(items, prefix),
-];
+): LocationItemData[] =>
+  filterDotItems(
+    [...parseExcludedPaths(excludedSubpaths), ...parseItems(items, prefix)],
+    prefix
+  );
```
calebpollman, line 86: Do we already have recursive logic in place that ensures the returned `items` is equal to the expected `pageSize`?

Samaritan1011001: Discussed offline and confirmed that the new items do get fetched until the pageSize is reached when items get filtered out due to this logic.

calebpollman (approved):

PR #6062, opened 2024-11-12
feat(dropzone): add dropzone support for files within dropped directories

jordanvn:
#### Description of changes
* Add logic to recursively retrieve files from within directories dropped into dropzone

#### Checklist

- [ ] PR description included


packages/react-core/src/hooks/useDropZone.ts
```diff
@@ -81,11 +81,54 @@ export default function useDropZone({
...
+
+    const files: File[] = [];
+
+    // Add any directly dropped files
+    if (event.dataTransfer.files.length > 0) {
+      files.push(...Array.from(event.dataTransfer.files));
+    }
+
+    // Handle items for directory support
+    if (event.dataTransfer.items) {
```
pranavosu, line 97: are `items` and `files` mutually exclusive? i wonder if we could include something twice

packages/react-core/src/hooks/useDropZone.ts
```diff
@@ -86,17 +87,35 @@ export default function useDropZone({
...
-    );
 
     if (isFunction(_onDrop)) {
       _onDrop(event);
     }
-    if (isFunction(onDropComplete)) {
-      onDropComplete({ acceptedFiles, rejectedFiles });
+
+    // If no DataTransfer items interface, handle as simple files
+    if (!event.dataTransfer.items) {
```
cshfang, line 96: Nit: Can we destructure dataTransfer or items/files earlier to cut down on some of this chaining verbosity?

packages/react-core/src/hooks/useDropZone.ts
```diff
@@ -86,17 +87,35 @@ export default function useDropZone({
...
+        acceptedFileTypes
+      );
+      if (isFunction(onDropComplete)) {
+        onDropComplete({ acceptedFiles, rejectedFiles });
+      }
+      return;
     }
+
+    // Process items using util
+    processDroppedEntries(Array.from(event.dataTransfer.items)).then(
```
cshfang, line 108: Nit: `processDroppedItems` to be more specific that we're dealing with `items`

packages/react-core/src/hooks/useDropZone.ts
cshfang, line 96: Not as nit: Is there any reason that we would not just exclusively deal with items in the presence of them? E.g.
```ts
if (isFunction(_onDrop)) {
   _onDrop(event);
}

// replacing line 89
const { items, files } = event.dataTransfer;

// 
!items ? processDroppedItems(items) : completeDrop(files);
```

Where `completeDrop` would encapsulate the duplicated functionality of line 97-103 and be called by processDroppedItems instead of chaining then here. (I think the current processing function even already creates this flattened list of `Files`, right?)

packages/react-core/src/utils/processDroppedEntries.ts
```diff
@@ -0,0 +1,59 @@
...
+    } else if (entry.isDirectory) {
+      const dirReader = (entry as FileSystemDirectoryEntry).createReader();
+      const dirEntries = await readAllDirectoryEntries(dirReader);
+      await Promise.all(dirEntries.map(processFileSystemEntry));
+    }
+  };
+
+  // Process all items
+  await Promise.all(
+    Array.from(dataTransferItems)
```
cshfang, line 51: Isn't dataTransferItems already an array?

packages/react-core/src/utils/processDroppedEntries.ts
```diff
@@ -0,0 +1,59 @@
...
+      .filter((dataTransferItem) => dataTransferItem.kind === 'file')
+      .map((dataTransferItem) => dataTransferItem.webkitGetAsEntry())
+      .filter((entry): entry is FileSystemEntry => entry !== null)
+      .map(processFileSystemEntry)
```
cshfang, line 55: Nit: Maybe possible to save some of these chained loops using reduce?

packages/react-core/src/hooks/useDropZone.ts
jordanvn: Updated

packages/react-core/src/hooks/useDropZone.ts
jordanvn: Good call out, updated

jordanvn: I'm not sure if I understand. Maybe I'm misinterpreting your first statement, but we were already only dealing with items if they existed, because conditionally called logic for handling files included an early return.

That said, I agree with your approach for simplifying the logic here. I moved the repeated ending logic into `completeDrop`. However, rather than passing `completeDrop` into `processDroppedItems`, I thought it might be clearer to just chain it as `processDroppedItems(items).then(completeDrop)`. What do you think?

packages/react-core/src/utils/processDroppedEntries.ts
jordanvn: Yep, updated to remove the `Array.from()`

packages/react-core/src/utils/processDroppedEntries.ts
jordanvn: Updated to use `.reduce().map()` instead

packages/react-core/src/hooks/useDropZone.ts
cshfang: I think my confusion was actually over the shadowed `files` variable due to it being declared in this parent function but then the same variable name was used as the parameter of the `then`. I initially thought it was the same files being used in both places

packages/react-core/src/utils/processDroppedItems.ts
```diff
@@ -46,12 +45,16 @@ export async function processDroppedEntries(
...
-  // Process all items
+  // Filter out and process files from the data transfer items
   await Promise.all(
-    Array.from(dataTransferItems)
-      .filter((dataTransferItem) => dataTransferItem.kind === 'file')
-      .map((dataTransferItem) => dataTransferItem.webkitGetAsEntry())
-      .filter((entry): entry is FileSystemEntry => entry !== null)
+    dataTransferItems
+      .reduce<FileSystemEntry[]>(
+        (acc, item) =>
```
cshfang, line 52: Super nit: Could destructure `item` but it's arguable whether that's more or less readable

packages/react-core/src/hooks/useDropZone.ts
```diff
@@ -102,20 +102,11 @@ export default function useDropZone({
         onDropComplete({ acceptedFiles, rejectedFiles });
       }
       return;
```
cshfang, line 104: Nit: I think this `return` is now unneeded?

cshfang (dismissed):

packages/react-core/src/hooks/useDropZone.ts
calebpollman: +1

packages/react-core/src/hooks/useDropZone.ts
```diff
@@ -85,18 +86,27 @@ export default function useDropZone({
...
+      );
+      if (isFunction(onDropComplete)) {
+        onDropComplete({ acceptedFiles, rejectedFiles });
+      }
+      return;
+    };
+
+    !items
+      ? completeDrop(Array.from(files))
+      : processDroppedItems(Array.from(items)).then(completeDrop);
```
calebpollman, line 108: Ternaries that aren't returning anything read strangely imo

packages/react-core/src/hooks/useDropZone.ts
jordanvn: Ah, yep! Nice catch. Removed

packages/react-core/src/utils/processDroppedItems.ts
calebpollman: I'd destructure since we use reference `webkitGetAsEntry` multiple times

packages/react-core/src/hooks/useDropZone.ts
jordanvn: Do you think it would read better as an if else?

jordanvn: Updated to if else

packages/react-core/src/utils/processDroppedItems.ts
jordanvn: Agreed. Updated to destructure

calebpollman (approved):

cshfang (approved):

PR #6069, opened 2024-11-13
fix(storageBrowser): cancel and upload same file

ashwinkumar6:
#### Description of changes
Fixes a bug in `react -> components -> FileSelect` where canceling and attaching the same file again doesn't trigged the onChange event for a file input.

This seems like an existing bug see, StackOverflow reference: [HTML input file selection event not firing upon selecting the same file](https://stackoverflow.com/questions/12030686/html-input-file-selection-event-not-firing-upon-selecting-the-same-file) 

- In ActionView, add an item with `Add files`
- remove the item by hitting `X`
- again add the same item
- upload is successful 

#### Checklist

- [x] PR description included


packages/react/src/components/FileSelect/FileSelect.tsx
```diff
@@ -129,6 +129,11 @@ export const useFileSelect = (
           onSelect?.([...event.target.files]);
         }
 
+        // Reset the input value to allow re-selecting the same file
```
ashwinkumar6, line 132: another approach is to add a `key` prop to `FileSelect` component and dynamically update it when onChange is called to make sure it re-renders

packages/react/src/components/FileSelect/__tests__/FileSelect.spec.tsx
```diff
@@ -54,8 +54,6 @@ describe('useFileSelect', () => {
         await user.upload(input!, file);
       });
 
-      expect(input?.files?.[0]).toStrictEqual(file);
```
calebpollman, line 57:
```suggestion
      expect(input?.files?.[0]).not.toStrictEqual(file);
```

ashwinkumar6: makes sense, updated

calebpollman (approved):

PR #6087, opened 2024-11-14
[fix] search suffix  bug

ashika112:
#### Description of changes

- Earlier use of split -> split the array on empty string and also duplicate folder name. Current fix searches in suffix after prefix is sliced from given string.

#### Checklist

- [ ] PR description included


packages/react-storage/src/components/StorageBrowser/actions/__tests__/createEnhancedListHandler.spec.ts
```diff
@@ -73,6 +73,60 @@ describe('createEnhancedListHandler', () => {
...
+  it('should collect and filter results when search and empty prefix is provided', async () => {
+    mockAction
+      .mockResolvedValueOnce({
+        items: [{ name: 'a_prefix/apple' }, { name: 'a_prefix/banana' }],
+        nextToken: 'next',
+      })
+      .mockResolvedValueOnce({
+        items: [{ name: 'a_prefix/cherry' }, { name: 'a_prefix/date' }],
+        nextToken: null,
+      });
```
pranavosu, line 85: i would change the names to remove `a_prefix` from them to avoid confusion, since prefix is "" for this case.

packages/react-storage/src/components/StorageBrowser/actions/__tests__/createEnhancedListHandler.spec.ts
```diff
@@ -73,6 +73,60 @@ describe('createEnhancedListHandler', () => {
...
+      search: { query: 'foo', filterKey: 'name' as const },
+    };
+
+    const result = await handler(prevState, {
+      config,
+      prefix: 'foo/',
+      options,
+    });
+
+    expect(result.items).toEqual([{ name: 'foo/date/foo' }]);
```
pranavosu, line 126: nice!

packages/react-storage/src/components/StorageBrowser/actions/__tests__/createEnhancedListHandler.spec.ts
pranavosu: i think the test works better if we match everything except `cherry` with `query = "a"`. we exercise the filtering logic that way too. if i'm not mistaken, names without `a_prefix/` would fail with the existing logic.

ashika112: There is a separate test for that, this is in addition, so no prefix is given. So like root ones, still need to search down stream path/

pranavosu (approved):
:ship:

PR #6135, opened 2024-11-18
feat: add date formatting to displayText API

dindjarinjs:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/displayText/libraries/en/__tests__/__snapshots__/uploadView.spec.ts.snap
+++ b/packages/react-storage/src/components/StorageBrowser/displayText/libraries/en/__tests__/__snapshots__/uploadView.spec.ts.snap
@@ -1,55 +1,122 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`CopyView display text values \`getActionCompleteMessage\` returns the expected values in the all failed scenario 1`] = `
+exports[`UploadView display text values \`getActionCompleteMessage\` returns the expected values in the all failed scenario 1`] = `
 {
   "content": "All files failed to upload.",
   "type": "error",
 }
 `;
 
-exports[`CopyView display text values \`getActionCompleteMessage\` returns the expected values in the all prevented overwrites scenario 1`] = `
+exports[`UploadView display text values \`getActionCompleteMessage\` returns the expected values in the all overwrite prevented or failed scenario 1`] = `
+{
+  "content": "Overwrite prevented for 6 files, 5 files failed to upload.",
+  "type": "error",
+}
+`;
+
+exports[`UploadView display text values \`getActionCompleteMessage\` returns the expected values in the all overwrite prevented scenario 1`] = `
 {
   "content": "Overwrite prevented for all files.",
   "type": "warning",
 }
 `;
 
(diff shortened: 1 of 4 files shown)


dbanksdesign (dismissed):

packages/react-storage/src/components/StorageBrowser/composables/DataTable/dataCells/DateDataCell.tsx
```diff
@@ -6,18 +6,19 @@ import { STORAGE_BROWSER_BLOCK_TO_BE_UPDATED } from '../../../constants';
 export interface DateDataCellProps {
   content: {
     date?: Date;
+    text?: string;
```
cshfang, line 9: Nit: Could this be `displayText` to distinguish the purpose of the prop a bit more? (Similar to the value and displayValue props for number cells)

packages/react-storage/src/components/StorageBrowser/displayText/types.ts
```diff
@@ -22,6 +22,7 @@ export interface DefaultListViewDisplayText {
   searchPlaceholder: string;
   searchSubmitLabel: string;
   searchClearLabel: string;
+  getDateText: (date: Date) => string;
```
calebpollman, line 25: `getDateLabel` for consistency?

packages/react-storage/src/components/StorageBrowser/composables/DataTable/dataCells/DateDataCell.tsx
```diff
@@ -6,18 +6,19 @@ import { STORAGE_BROWSER_BLOCK_TO_BE_UPDATED } from '../../../constants';
...
   content,
 }: DateDataCellProps): React.JSX.Element => {
-  const { date } = content;
+  const { text } = content;
   return (
     <ViewElement
       className={`${STORAGE_BROWSER_BLOCK_TO_BE_UPDATED}__table-date-data-cell`}
     >
-      {date?.toLocaleString()}
+      {text}
```
cshfang, line 21: Since text is optional, would the toLocaleString be a reasonable fallback value?

packages/react-storage/src/components/StorageBrowser/displayText/types.ts
cshfang: Per my other comment - maybe getDateDisplayText? Esp since it's not quite a label for a date either?

calebpollman: Had thought abt suggesting `getDateValue` here but iirc we already have a `getDownloadLabel` key, would prefer we keep consistency with appending either `value` or `label` for these

packages/react-storage/src/components/StorageBrowser/composables/DataTable/dataCells/DateDataCell.tsx
cshfang: Sure I think that's ok

packages/react-storage/src/components/StorageBrowser/composables/DataTable/dataCells/DateDataCell.tsx
cshfang: Yeah but at that point there is nothing more we can do for this cell. 🪦

calebpollman (approved):

cshfang (approved):

PR #6141, opened 2024-11-18
feat: add action list text to displayText API

dindjarinjs:
#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/actions/useAction/__tests__/createEnhancedListHandler.spec.ts
+++ b/packages/react-storage/src/components/StorageBrowser/actions/useAction/__tests__/createEnhancedListHandler.spec.ts
@@ -18,6 +18,7 @@ const config: ActionInputConfig = {
 };
 type Output = ListHandlerOutput<{
   name: string;
+  alt: string;
   id: number;
 }>;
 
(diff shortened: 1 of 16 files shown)


packages/react-storage/src/components/StorageBrowser/actions/configs/types.ts
```diff
@@ -69,7 +69,12 @@ export interface ActionListItemConfig {
   /**
    * list item label
    */
-  label: string;
+  label?: string;
+
+  /**
+   * list item key (used for displayText)
+   */
+  key: string;
```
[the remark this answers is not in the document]
calebpollman: No key at all, since we wouldn't be able to provide typing for it on the displayText API without complicated type inference. 

in the displayText API we add `getActionListItemLabel` in the location detail view interface, and then run the getter for each action list item with the `label` value provided from the configuration, our default implementation will look like:
```
getActionListItemLabel: (label) => label
```

packages/react-storage/src/components/StorageBrowser/displayText/libraries/en/locationDetailView.ts
```diff
@@ -52,6 +52,20 @@ export const DEFAULT_LOCATION_DETAIL_VIEW_DISPLAY_TEXT: DefaultLocationDetailVie
     tableColumnTypeHeader: 'Type',
     selectFileLabel: 'Select file',
     selectAllFilesLabel: 'Select all files',
+    getActionListLabel: (key: string | undefined) => {
```
cshfang, line 55: Nit: Could use a default value of `''` and get rid of the nullish coalesce

cshfang (approved):

packages/react-storage/src/components/StorageBrowser/actions/configs/types.ts
```diff
@@ -69,7 +69,7 @@ export interface ActionListItemConfig {
   /**
    * list item label
    */
-  label: string;
+  label?: string;
```
calebpollman, line 72: This should remain required for consistency

packages/react-storage/src/components/StorageBrowser/displayText/types.ts
```diff
@@ -67,6 +67,7 @@ export interface DefaultLocationDetailViewDisplayText
   tableColumnNameHeader: string;
   tableColumnSizeHeader: string;
   tableColumnTypeHeader: string;
+  getActionListLabel: (key: string | undefined) => string;
```
calebpollman, line 70: For discoverability think adding "Item" will help consumers to better understand that this is used for resolving list item labels:
```
getActionListItemLabel
```

calebpollman (commented):
Minor feedback but overall lgtm

cshfang (dismissed):

calebpollman (approved):

PR #6144, opened 2024-11-18
chore: move storage browser icon data from icon element to react icon primitive

jordanvn:
#### Description of changes
* Move icon paths and variant types to react package icon primitive, to resolve type issue

#### Checklist

- [ ] PR description included


cshfang (dismissed):

packages/react/src/primitives/Icon/internal.ts
```diff
@@ -1,2 +1,6 @@
 export * from './icons';
 export { useIcons } from './context/useIcons';
+export {
+  STORAGE_BROWSER_ICON_PATHS,
+  StorageBrowserIconType,
+} from './context/StorageBrowserIcons';
```
calebpollman, line 6: Think this line is creating the public export

packages/react/__tests__/__snapshots__/exports.ts.snap
```diff
@@ -111,6 +111,7 @@ exports[`@aws-amplify/ui-react/internal exports should match snapshot 1`] = `
   "IconVisibilityOff",
   "IconWarning",
   "PrimitiveCatalog",
+  "STORAGE_BROWSER_ICON_PATHS",
```
calebpollman, line 114: Should not be a public export, needs to be exported from _packages/react/src/internal.ts_

calebpollman, line 114: Should not be a public export

jordanvn: Removed from snapshot after changing export location

packages/react/src/primitives/Icon/internal.ts
jordanvn: Added export at suggested file

packages/react/__tests__/__snapshots__/exports.ts.snap
jordanvn: I had thought that I saw this test pass, but it is failing again. It seems like this test is confirming internal exports though, so I think it being in this snapshot should be correct? Or am I missing something?

calebpollman (approved):

PR #6147, opened 2024-11-18
test(react-storage): add upload copy and delete e2e tests

ashwinkumar6:
#### Description of changes
e2e tests for upload, it tests uploading a file and folder on LocationActionView and listing it on the LocationDetailsView to see if it was successful.

Tested in cypress locally

#### Checklist

- [x] PR description included


packages/e2e/features/ui/components/storage/storage-browser/upload.feature
```diff
@@ -0,0 +1,63 @@
+Feature: Upload tests within Storage Browser
```
ashwinkumar6, line 1: `drag-and-drop.feature` renamed to `upload.feature` with two additional tests added

packages/react-storage/src/components/StorageBrowser/composables/OverwriteToggle.tsx
```diff
@@ -25,6 +25,7 @@ export const OverwriteToggle = ({
       id={OVERWRITE_TOGGLE_ID}
       onChange={onToggle}
       type="checkbox"
+      aria-label="overwrite-toggle"
```
cshfang, line 28: Align with others i.e. `Overwrite Toggle`

ashwinkumar6: Updated approach to find the overwrite checkbox. 

Looks like 
- managedAuth is using the default OverwriteToggle composable. This contains input checkBox
- defaultAuth is overriding default composable with it's own. It's using label + span + icon instead.

Updated to a more generic approach, "look for certain text inside a label"

packages/e2e/features/ui/components/storage/storage-browser/upload.feature
```diff
@@ -0,0 +1,83 @@
...
+    When I click the first button containing "public"
+    Then I see the "Actions" button
+    When I click the "Actions" button
+    # LocationActionView -> Upload files
+    Then I see the "Upload files" menuitem
+    When I click the "Upload files" menuitem
+    Then the "Upload" button is disabled
+    Then I upload a text file with name "e2eTest"
+    Then I see "Queued"
+    Then I click the "overwrite-toggle" checkbox
```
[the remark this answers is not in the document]
ashwinkumar6: do you mean encapsulating multiple steps into one step. 
In this case we're hitting the upload button in the end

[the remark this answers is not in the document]
ashika112: Sorry i meant `i click the overwrite checkbox` as its own step if it is something we plan on testing across. I see [this](https://github.com/aws-amplify/amplify-ui/pull/6147#discussion_r1847352587) now, so this should be ok.

ashika112 (dismissed):

packages/react-storage/src/components/StorageBrowser/composables/OverwriteToggle.tsx
ashwinkumar6: rebased ignore comment

packages/e2e/features/ui/components/storage/storage-browser/upload.feature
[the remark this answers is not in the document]
ashwinkumar6: rebased ignore comment

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -270,6 +276,10 @@ When('I click the {string}', (id: string) => {
   cy.findByTestId(id).click();
 });
 
+When('I click the element with id attribute {string}', (id: string) => {
```
ashwinkumar6, line 279: An alternative to this is `I click the {string}` that uses `cy.findByTestId`. This requires the element to have `data-testid` prop.

Adding to prop to dynamic components like table might be complex hence introduced this as an alternative.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -581,6 +596,10 @@ Then('I click the submit button', () => {
   }).click();
 });
 
+Then('I click the label containing text {string}', (labelText: string) => {
```
ashwinkumar6, line 599: Ideally we would click on an input/button but since the `Overwrite existing files` checkbox compossible contains label+span+svg, we're using this approach

packages/e2e/features/ui/components/storage/storage-browser/toggle-menu.feature
```diff
@@ -0,0 +1,140 @@
...
+    When I click the "Create Folder" menuitem
+    Then I see "Folder name"
+    Then I see the "Create Folder" button
+    Then the "Create Folder" button is disabled 
+    When I type a new "Folder name" with value "Blackberry/"
+    When I lose focus on "Folder name" input
+    Then I see 'Folder name cannot contain \"/\", nor end or start with \".\"'
+
+  @react
+  Scenario: upload, list, copy and delete a file using toggle menu
```
ashwinkumar6, line 66: - On the Location Action View each of these operations also have an error state
  - copy fails if copying to source object's location
  - upload fails if overwrite is disabled and obj exists...etc
- `Menu Toggle -> upload` could be it's own e2e test, testing all the clickable items on page

This PR mainly tests CRUD operations. We can discuss the approach ^ and implement as a follow-up

packages/e2e/cypress/integration/common/shared.ts
jordanvn: Nit: should click always have `force` set to true when we want to select by id? And would it make sense to allow any selector rather than an id? Or would that be too generic?

packages/e2e/features/ui/components/storage/storage-browser/toggle-menu.feature
jordanvn, line 1: Nit: wouldn't this menu be referred to as `action-menu`?

jordanvn (dismissed):
Looks good to me. Not sure if there are some ways some changes could be improved, but that might be bumping up against gherkin best practices.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -8,7 +8,13 @@ import { get, escapeRegExp } from 'lodash';
 let language = 'en-US';
 let window = null;
 let stub = null;
-const randomFileName = `fileName${Math.random() * 10000}`;
+export const randomFileName = `fileName${Math.random() * 10000}`;
+
+const doesDocumentContainText = (text: string) => {
```
ashika112, line 13: nit: best practice for cypress would be create a cy command if it is used across

packages/e2e/features/ui/components/storage/storage-browser/toggle-menu.feature
ashika112: +1 makes more sense to be action menu or real component name

packages/e2e/features/ui/components/storage/storage-browser/toggle-menu.feature
```diff
@@ -0,0 +1,140 @@
...
+    Then I see the two files with random names
+    # delete created file
+    Then I click the element with id attribute "header-checkbox"
+    When I click the "Menu Toggle" button
+    Then I click the "Delete" menuitem
+    Then I click the "Delete" button
+    Then I see "All files deleted"
+    When I click the "Exit" button
+    # verify all files are deleted
+    Then I see "No files"
```
ashika112, line 140: Goal of UI test should be just to test the UI e2e. In this case this seems like its already covered in above test. List and delete. maybe uneccesarry unless you are covering a different UI scenarios here.

I would just check if there is a folder `e2eTemp`

packages/e2e/features/ui/components/storage/storage-browser/toggle-menu.feature
ashwinkumar6: thx updated

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -645,3 +664,33 @@ Then('I see the {string} radio button checked', (label: string) => {
...
+      },
+      {
+        contents: Cypress.Buffer.from('File 2 content'),
+        fileName: `${folderName}/${randomFileName}2`,
+        mimeType: 'text/plain',
+      },
+    ];
+    cy.get('input[type="file"]').selectFile(folderFiles, { force: true });
+  }
+);
```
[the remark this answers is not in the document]
ashwinkumar6: sounds good, updated

packages/e2e/cypress/integration/common/shared.ts
ashwinkumar6: >Nit: should click always have force set to true when we want to select by id?

We're using this to select the table header checkbox input which is hidden so force is required in this scenario. 

>would it make sense to allow any selector rather than an id? Or would that be too generic?

Yeah i think it might too generic, we could add one later if required

jordanvn: > We're using this to select the table header checkbox input which is hidden so force is required in this scenario.

I was thinking if this could be an optional choice for situations where we want to select by an id and _not_ force the click. But now that I think of it, I can't think of a scenario where forcing the click in this circumstance would be unwanted.

As you say, it can be updated later if additional test cases require more flexible behavior.

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -362,10 +388,12 @@ Then('I see tab {string}', (search: string) => {
...
 });
 
-Then('I see {string}', (message: string) => {
-  cy.findByRole('document')
-    .contains(new RegExp(escapeRegExp(message), 'i'))
-    .should('exist');
+Then('I see {string}', doesDocumentContainText);
+
+Then('I see {string} files with random names', (count: string) => {
+  for (let i = 1; i <= parseInt(count); i++) {
```
jordanvn, line 394: Does it make sense to consider cases where `count` could resolve to NaN, since it's a string? I guess if tests have an error they're (hopefully) not getting merged. What do you think?

packages/e2e/features/ui/components/storage/storage-browser/action-menu.feature
```diff
@@ -0,0 +1,141 @@
...
+    Then I type my password
+    Then I click the "Sign in" button
+    When I click the first button containing "public"
+    Then I see the "Menu Toggle" button
+    When I click the "Menu Toggle" button
+    # upload file
+    Then I see the "Upload" menuitem
+    When I click the "Upload" menuitem
+    Then the "Upload" button is disabled
+    Then I upload "1" files with random names
```
jordanvn, line 77: I knew we'd run into the grammar issue of "1 files", but I'm fine ignoring it. Not worth having two separate gherkin steps imo

packages/e2e/cypress/integration/common/shared.ts
ashwinkumar6: If it results in NaN say we input "two" instead of "2", the file simply won't get uploaded.
So in the following step we won't see `All files successfully uploaded` status and the test would fail.

We could do something like this but i was thinking it might be a bit overkill
```
  if (isNaN(numberOfFiles)) {
    throw new Error(`Invalid count: "${count}" is not a valid number`);
  }
```

packages/e2e/features/ui/components/storage/storage-browser/action-menu.feature
ashwinkumar6: yeah i was on the fence in naming it this or ` I upload "2" file with random name` 😂

packages/e2e/cypress/integration/common/shared.ts
Guywiththelaugh: You love to kill birds right? You can not kill a 🥔 potato , just kill it with two stones🤣🍕

Guywiththelaugh: More emotion 🐟🐟🐟

Guywiththelaugh: You want jon snow them Ashwin 😉

Guywiththelaugh: May be a bottle of soju will fix this♥️🐟😉

packages/e2e/features/ui/components/storage/storage-browser/action-menu.feature
Guywiththelaugh: ALWAYS 2 is greater than 1Ashein

ashika112 (approved):

packages/e2e/cypress/integration/common/shared.ts
ashwinkumar6: updating in a follow-up PR

PR #6194, opened 2024-11-21
chore(storage-browser): Update control unit tests to be more isolated

cshfang:
#### Description of changes

This PR streamlines the control unit tests to isolate its mocks a bit better so that it is not implicitly testing upstream dependent components

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/controls/hooks/__tests__/useDataRefresh.spec.ts
+++ b/packages/react-storage/src/components/StorageBrowser/controls/hooks/__tests__/useDataRefresh.spec.ts
@@ -1,0 +1,37 @@
+import { renderHook } from '@testing-library/react';
+import { DataRefreshProps } from '../../../composables/DataRefresh';
+import { useControlsContext } from '../../../controls/context';
+import { useDataRefresh } from '../useDataRefresh';
+
+jest.mock('../../../controls/context');
+
+describe('useDataRefresh', () => {
+  const data = {
+    dataRefreshLabel: 'add-folder-label',
+    isDataRefreshDisabled: false,
+  };
+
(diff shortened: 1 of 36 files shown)


jordanvn (approved):
These changes mstm. Confirmed no regressions locally, and for this improves overall coverage as well.

packages/react-storage/src/components/StorageBrowser/controls/hooks/__tests__/useDataRefresh.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  afterEach(() => {
+    mockUseControlsContext.mockReset();
+  });
+
+  it('returns DataRefresh props', () => {
+    const { result } = renderHook(() => useDataRefresh());
+
+    const expected: DataRefreshProps = {
+      isDisabled: data.isDataRefreshDisabled,
+      onRefresh: expect.any(Function),
```
pranavosu, line 32: we could consider using mock `onRefresh` from line 19, possibly expect `onRefresh: mockOnRefresh` instead of any function. this applies to a lot of the other tests, but i don't think it's critical.

pranavosu (approved):

packages/react-storage/src/components/StorageBrowser/controls/hooks/__tests__/useDataRefresh.spec.ts
cshfang: Yeah, I had considered doing this but we'd have to pull them out into `const mockOnRefresh = jest.fn()` just to assert something that's more or less trivial in the end. So just opted to keep it marginally more concise

PR #6206, opened 2024-11-25
test(e2e): add storage-browser offline tests

ashwinkumar6:
#### Description of changes
Add offline e2e tests for action menu items

Tested in cypress locally
Screenshot 2025-01-08 at 1 22 00 PM

#### Checklist

- [x] PR description included


packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -687,3 +687,9 @@ When(
   (folderName: string, count: string) =>
     fileInputUpload(parseInt(count), folderName)
 );
+
+When('I simulate network disconnection', () => {
```
jordanvn, line 691: This is written as a `When`, which I think is correct, but it is written with `Then` in all scenarios

packages/e2e/features/ui/components/storage/storage-browser/action-menu.feature
```diff
@@ -139,3 +139,71 @@ Feature: Create folder with Storage Browser
     # verify all files are deleted
     Then I see "No files"
 
+  @react
+  Scenario: Upload file shows a Network error if offline
+      When I type my "email" with status "CONFIRMED"
+      Then I type my password
+      Then I click the "Sign in" button
```
jordanvn, line 146: Nit: this is entirely non-blocking of this PR, but would it fit for there to be step specifically for signing into the storage browser which would just be equivalent to these three steps? Having every test start with 3 - 4 steps to get to the right starting place can really cause these files to get large and less straightforward to parse

packages/e2e/cypress/integration/common/shared.ts
ashwinkumar6: Good question, if we define the step as `when` would you suggest using it in the test as `when`

Depending on the previous step, this may vary, example
```
When I see the first button containing "public"
Then I simulate network disconnection
```
or we could use a when after a when

ashwinkumar6: good catch, updated to `when` in all places.

ashika112 (approved):

packages/e2e/features/ui/components/storage/storage-browser/drag-and-drop.feature
```diff
@@ -38,4 +31,3 @@ Feature: Drag and drop files within Storage Browser
     Then I press the "{esc}" key
     When I drag and drop a file into the storage browser with file name "test.txt"
     Then I see "test.txt"
-
```
jordanvn, line 41: Should there be an EOF newline?

packages/e2e/features/ui/components/storage/storage-browser/action-menu.feature
```diff
@@ -2,12 +2,12 @@ Feature: Create folder with Storage Browser
 
   Background:
     Given I'm running the example "ui/components/storage/storage-browser/default-auth"
+    And I type my "email" with status "CONFIRMED"
+    And I type my password
+    And I click the "Sign in" button
```
jordanvn, line 7: Nit: would it make sense to add the line `And I click the first button containing "public"`? It's also repeated at the start of every scenario, but I could see it going either way as to whether it could be considered "putting the system in a known state" as Givens are supposed to be used.

ashwinkumar6: That's a valid point. The `And I click the first button containing "public"` step is common in most files so this could also be moved to `Background`

The thought process is:
It's mandatory to signIn to view the StorageBrowser component only after which we can perform any integ tests on the component. So this is a required step for all integ tests now and in future.

But however, clicking the public folder isn't a mandatory step for all integ tests, we could have tests in future that click the private folder for example. Hence only these 3 steps related to signIn are added to Background in all tests

packages/e2e/features/ui/components/storage/storage-browser/drag-and-drop.feature
ashwinkumar6: looks like it show this because EOF was fixed

packages/e2e/features/ui/components/storage/storage-browser/action-menu.feature
ashika112: I think it is better to keep `And I click the first button containing "public"` because it happens on the screen once you start. I would consider moving to background only things like signIn but LocationView in itself is the first place you are in. Unless the test itself is focused on lets say another view like `LocationDetailView` i wouldn't move it to background IMO

ashika112 (dismissed):

packages/e2e/cypress/integration/common/shared.ts
```diff
@@ -655,3 +655,9 @@ When(
   (folderName: string, count: string) =>
     cy.fileInputUpload(`${folderName}/${randomFileName}`, parseInt(count))
 );
+
+When('I simulate network disconnection', () => {
```
calebpollman, line 659: Since a network connection failure is not something that a user would intentionally cause (in most scenarios at least) think the verbiage of this util would read more straightforward as:
```suggestion
When('A network failure occurs', () => {
```

calebpollman (commented):
Thanks for adding these!

packages/e2e/features/ui/components/storage/storage-browser/action-menu.feature
jordanvn: Returning to this, MSTM, I think that's a good reason not to. I think this point is resolved

ashika112 (approved):

jordanvn (approved):

PR #6211, opened 2024-11-26
feat(ai) add attachment validations

dbanksdesign:
#### Checklist

- [ ] PR description included


packages/react-ai/src/components/AIConversation/context/AttachmentContext.tsx
```diff
@@ -1,16 +1,32 @@
...
-  children?: React.ReactNode;
-  allowAttachments?: boolean;
-}): JSX.Element => {
+  allowAttachments = false,
+  // We save attachments as base64 strings into dynamodb for conversation history
+  // DynamoDB has a max size of 400kb for records
+  // This can be overridden so cutsomers could provide a lower number
+  // or a higher number if in the future we support larger sizes.
+  maxAttachmentSize = 400_000,
+  maxAttachments = 20,
```
calebpollman, line 22: Are these safe to override today?

packages/react-ai/src/components/AIConversation/context/ControlsContext.tsx
```diff
@@ -1,20 +1,24 @@
...
+import { ConversationDisplayText } from '../displayText';
 
 export interface ControlsContextProps {
   Form?: React.ComponentType<
     {
       handleSubmit: (e: React.FormEvent<HTMLFormElement>) => void;
       allowAttachments?: boolean;
       isLoading?: boolean;
-    } & Required<ConversationInputContext>
+      displayText: Required<ConversationDisplayText>;
```
calebpollman, line 14: Will making `displayText` required break composability?

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
```diff
@@ -62,8 +66,46 @@ export const Form: Required<ControlsContextProps>['Form'] = ({
...
+        let error = '';
+        if (hasMaxError) {
+          error += displayText.getMaxAttachmentErrorText(maxAttachments);
+        }
+        if (hasMaxSizeError) {
+          error += displayText.getAttachmentSizeErrorText(
+            // base64 size is about 137% that of the file size
+            // https://en.wikipedia.org/wiki/Base64#MIME
+            humanFileSize((maxAttachmentSize - 814) / 1.37, true)
+          );
```
calebpollman, line 89: Looks like we could avoid passing displayText as a prop if we handle the validation logic in the parent and just pass an `onValidate` callback to the `Form`?

packages/react-ai/src/components/AIConversation/context/AttachmentContext.tsx
dbanksdesign: They are safe to override to be lower, but if they are higher the customer will probably get an error further down in the stack (in lambda)

packages/react-ai/src/components/AIConversation/context/ControlsContext.tsx
dbanksdesign: Nope, because the composable Form control doesn't take any props. This is to override the form component (the Form control grabs all the data and stuff from contexts and passes it to this component).

packages/react-ai/src/components/AIConversation/context/AttachmentContext.tsx
calebpollman: Got it. Lmk if I missed it, but are we validating that the provided values are not larger than the known thresholds?

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
calebpollman: Just noting that if displayText is not being provided to override components not concerned abt the logic living here fir the immediate term

dbanksdesign: Good call, moved this up and passed an onValidate callback

packages/react-ai/src/components/AIConversation/context/ControlsContext.tsx
dbanksdesign: But it did make it kinda wonky, changing to just take an onValidate callback that handles the validation and state update

calebpollman (dismissed):
🚢 

May want to consider adding max threshold validation for the maxAttachments/maxAttachmentSize props to ensure override values don't exceed the dynamodb limits in the future

packages/react-ai/src/components/AIConversation/context/AttachmentContext.tsx
dbanksdesign: We aren't validating if the provided values are larger than the known thresholds. I did this because the thresholds will go up at a later point, the 400kb size limit per file is a limitation of our backend implementation and not Bedrock or the AI model's, and we do plan to fix this long term. I can change it so we enforce the limits though if we feel strongly about it.

jordanvn (dismissed):

packages/react-ai/src/components/AIConversation/views/Controls/FormControl.tsx
```diff
@@ -213,14 +221,52 @@ export const FormControl: FormControl = () => {
...
+          maxAttachmentSize,
+        });
+
+      if (hasMaxError ?? hasMaxSizeError) {
+        let error = '';
+        if (hasMaxError) {
+          error += displayText.getMaxAttachmentErrorText(maxAttachments);
+        }
+        if (hasMaxSizeError) {
+          error += displayText.getAttachmentSizeErrorText(
```
cshfang, line 240: If both errors are present, is the string just joined together without a space in between? Would it be better to do something like
```ts
const errors = [];
errors.push(maxAttachmentError);
errors.push(sizeError);
setError(errors.join(' '));
```

cshfang (approved):

cshfang (approved):

jordanvn (approved):

PR #6231, opened 2024-12-02
chore: replace defaultChecked with checked

jordanvn:
#### Description of changes
* Fix bug causing no radio buttons to be checked by default in Authenticator "verify user" state

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
+++ b/packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
@@ -39,7 +39,7 @@ const generateRadioGroup = (
           name="unverifiedAttr"
           value={key}
           key={key}
-          defaultChecked={index === 0}
+          checked={index === 0}
         >
           {translate(verificationType)}:{' '}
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Authenticator/VerifyUser/VerifyUser.tsx
```diff
@@ -39,7 +39,7 @@ const generateRadioGroup = (
           name="unverifiedAttr"
           value={key}
           key={key}
-          defaultChecked={index === 0}
```
jordanvn, line 42: `defaultChecked` is not used to pass in an initial boolean value of a radio button, [it is to return the initial boolean value of the radio button](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/defaultChecked). Setting it here will not actually set the value of the radio button.

jordanvn: This actually looks like it isn't the case. It appears that in the Amplify Radio component, `defaultChecked` does indeed set the default state of the checkbox

PR #6239, opened 2024-12-06
fix: resolve React Native build test failure

jordanvn:
#### Background
* The mega app used in cli React Native canary build tests uses the dependency `react-native-safe-area-context`
  * Currently, the workflow to run the build test is set to use the latest version of `react-native-safe-area-context`
  * The release of version 5.0.0 increased the minimum required version of React Native to 0.75.0+, while support for lower versions is still being included in tests
  * This caused an incompatibility on tests using versions of React Native below 0.75.0, resulting in test failure.
* Build test was confirmed to pass with this change

#### Description of changes
* Set version of `react-native-safe-area-context` used in pipeline build tests to latest minor change within v4, as v5 includes breaking changes to React Native versions required in testing process.

#### Checklist

- [ ] PR description included

### DIFF
--- a/.github/workflows/build-system-test-react-native.yml
+++ b/.github/workflows/build-system-test-react-native.yml
@@ -6,8 +6,9 @@ permissions:
   id-token: write # This is required for aws-actions/configure-aws-credentials
 
 on:
-  schedule:
-    - cron: '0 * * * *' # Run at the first minute of every hour
+  pull_request:
+    branches: [main, hotfix]
+    types: [opened, synchronize, labeled]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


HuiSF (approved):

scanlonp (approved):

jordanvn:
> Thanks for handling this. Have some questions and general feedback on the approach: Couple things:
> 
> * have we validated that React Native >0.75 works with the pinned version of react-native-safe-area-context?
> * feels off to be hard coding a version in the install script. Would expect the approach here to align with the approach taken with other build tests that rely on specific versions to prevent confusion in the future
> * our minimum allowed peer dep is `^4.2.5`, imo the pin should should match just even if `^4.14.0` is what ends up being installed to reduce the cognitive load for future maintainers
> * snowflake-y workarounds should have code comments that provide context for future maintainers

* Yes, I have validated that React Native >0.75 works with the pinned version.
* I agree with the additional comments here and will put together a follow-up to improve the quality

PR #6257, opened 2024-12-17
feat(authenticator): email mfa

scanlonp:
Moving fork branch into feature branch.

#### Description of changes

PR to add email MFA. User flows will have MFA selection if multiple MFA methods are available, setup for email MFA if need be, and confirmation of sign in with a code sent to user's email.

#### Checklist

- [ ] PR description included


packages/react/src/components/Authenticator/SelectMfa/SelectMfa.tsx
```diff
@@ -0,0 +1,178 @@
...
+          {value}
+        </Radio>
+      );
+    }
+  );
+  */
+};
+
+
+export const SelectMfa = ({
```
jordanvn, line 80: How are instances of a user only having one MFA option handled?

packages/react/src/components/Authenticator/SelectMfa/SelectMfa.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const opt: MfaOptions = {
+  email: 'EMAIL',
+  username: 'Totp Prop',
+};
+
+type MfaType = 'EMAIL' | 'wow';
+*/
+
+const generateRadioGroup = (
```
jordanvn, line 41: Should we have one of the radio options selected by default?

packages/react/src/components/Authenticator/SelectMfa/SelectMfa.tsx
scanlonp: The JS library should route directly to the confirm sign in step associated with that MFA option. I will make sure that behavior is tested as well.

jordanvn: Perfect, thanks for confirming that that case is already handled

packages/react/src/components/Authenticator/SelectMfa/SelectMfa.tsx
jjarvisp: Or make sure the continue button is disabled until one is selected.

packages/react/src/components/Authenticator/SelectMfa/SelectMfa.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { ConfirmSignInFooter } from '../shared/ConfirmSignInFooter';
+import { RemoteErrorMessage } from '../shared/RemoteErrorMessage';
+import { RouteContainer, RouteProps } from '../RouteContainer';
+import { toLower } from 'lodash';
+
+const { getChallengeText } = authenticatorTextUtil;
+
+function parseMfaTypes(types?: MfaType[]): string {
+  if(types) {
+    return types.join();
```
jjarvisp, line 21: If you're displaying this result to users (which may be redundant given they'll see them in the radio options), probably want to use `.join(', ')` instead.

PR #6301, opened 2025-01-16
fix(slider): remounted controlled value not updating

dbanksdesign:
#### Description of changes

Fixes an issue where the label value for the slider was not updating on external update. TBH there might be a cleaner way to do this, but I'm rusty. 

#### Issue #, if available

To repro:
1. set a value via props
2. unmount the component
3. mount the component again with a different value
4. Observe the slider bar has repositioned while the label still reports the original value

Added a test and built a test page to validate with my own 👁️ 

(the top slider is controlled, bottom one uncontrolled, the button just sets a random value)

Previous:

CleanShot 2025-01-16 at 13 27 03

With the fix:

CleanShot 2025-01-16 at 13 25 53

#### Checklist

- [ ] PR description included

### DIFF
--- a/packages/react/src/primitives/SliderField/SliderField.tsx
+++ b/packages/react/src/primitives/SliderField/SliderField.tsx
@@ -90,16 +90,17 @@ const SliderFieldPrimitive: Primitive<SliderFieldProps, 'span'> = (
     [onChange]
   );
 
+  const realValue = isControlled ? value : currentValue;
   const renderedValue = React.useMemo(() => {
     const formattedValue = isFunction(formatValue)
-      ? formatValue(currentValue)
-      : currentValue;
+      ? formatValue(realValue)
+      : realValue;
     return typeof formatValue === 'string' ? (
       <View as="span">{formattedValue}</View>
     ) : (
       formattedValue
     );
-  }, [currentValue, formatValue]);
+  }, [realValue, formatValue]);
 
   const isVertical = orientation === 'vertical';
   const componentClasses = classNames(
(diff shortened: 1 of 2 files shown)


packages/react/src/primitives/SliderField/SliderField.tsx
```diff
@@ -90,16 +90,17 @@ const SliderFieldPrimitive: Primitive<SliderFieldProps, 'span'> = (
...
-      : currentValue;
+      ? formatValue(realValue)
+      : realValue;
     return typeof formatValue === 'string' ? (
       <View as="span">{formattedValue}</View>
     ) : (
       formattedValue
     );
-  }, [currentValue, formatValue]);
+  }, [realValue, formatValue]);
```
calebpollman, line 103: Minor callout: think it probably makes sense to remove the `useMemo` here:
- if the provided `formatValue`  isn't memoized it negates the memoization here
- moving the ternary starting on line 98 inside line 141 would gate the creation of the element node when `isValueHidden` is truthy

packages/react/src/primitives/SliderField/__tests__/SliderField.test.tsx
```diff
@@ -207,6 +227,8 @@ describe('SliderField:', () => {
       );
     });
 
+    it('should work as controlled', async () => {});
```
calebpollman, line 230: Spooky missing test body!

calebpollman (approved):
🚢

jordanvn (approved):

Issue #4800, opened 2023-12-01
Add a service override to React Authenticator for the Resend Code button.


joealynn:
### On which framework/platform would you like to see this feature implemented?

React

### Which UI component is this feature-request for?

Authenticator

### Please describe your feature-request in detail.

My web application has always upper-cased the user's entry of username: We store it in Cognito as all uppercase. Under V5, I was able to do this by overriding the handleSignUp, handleSignIn, handleForgotPassword, and handleForgotPasswordSubmit services to grab the username entered, convert it to uppercase and invoke the corresponding Auth service.

Under V6, I found that I also had to override handleConfirmSignUp. I have everything working, except when the Resend Code button is clicked: the username is sent as entered and not found by Cognito if it was not all uppercase. I see that there is a 'resendSignUpCode' function in aws-amplify/auth, but no 'handleResetSignUpCode' in the defaultServices file. Can this be added, please?

### Please describe a solution you'd like.

Support a new entry in the services passed to the Authenticator component: services.handleResendSignUpCode, which if present should be called when the Resend Code button is clicked.

### We love contributors! Is this something you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request.
- [ ] ⚠️ This feature might incur a breaking change.

calebpollman:
@joealynn This makes sense to me, will look in to adding it to our roadmap!

zerbusdetroy:
Hello,

I have a similar requirement for Angular. Should I create a separate issue for this? Is this feature on the roadmap?

Thank you.

stevexm:
Hello from Seattle, I had the very same requirement for a React project and (finally) gave up on the documentation pointing to any specific override function. Figured that no one has gotten around to working on comprehensively expanding the overrides yet.

If you are in the same situation, here is a hacky bit of a workaround until something more formal gets posted:

```
document.querySelector("button[type='button']").onclick = (event) => { 
     event.stopPropagation(); //do not bubble
     event.stopImmediatePropagation(); //stop the other Authenticator handlers from getting a hold of the onclick; take out if you want to run 
     event.preventDefault(); //stop default button click
     console.log('do whatever you need to do for a code resend, cutting out the default/or not as the case may be')
}

```

The particular selector employed works because the submit button is of type='submit', whereas the "Resend Code" button we are interested in is of type="button". Yes I know, hacky. But for my use case all I needed to do was trigger my server to tell SendGrid to reissue an email with a fresh code.

Cheers

[...]

PR #6312, opened 2025-01-23
feat(ui): allow override of resendSignUpCode function call

kevin860:
#### Description of changes

This adds the `resendSignUpCode` function to the list of function calls that can be overridden using a `handleResendSignUpCode` option on the services object. This is important for anyone who is modifying the username, which is actually what the current example in the documentation suggests.

This PR also updates the documentation to match.

#### Issue #, if available

https://github.com/aws-amplify/amplify-ui/issues/4800

I have reviewed the existing tests for this component and found that they are currently skipped, making it hard to truly verify that these changes are functional. However, this change is quite simple and exactly follows the pattern of other function call overrides.

#### Checklist

- [x] PR description included


packages/ui/src/machines/authenticator/actors/signIn.ts
```diff
@@ -297,7 +296,7 @@ export function signInActor({ services }: SignInMachineOptions) {
           return resetPassword({ username });
         },
         handleResendSignUpCode({ username }) {
-          return resendSignUpCode({ username });
+          return services.handleResendSignUpCode({ username });
```
jordanvn, line 299: Will this change also have to be made for the usage of `resendSignUpCode()` in /actors/signUp.ts?

kevin860: @jordanvn yes you're right - I was focused on the signIn use case we've been working through but the same change should be done there. I'll push an update here.

jordanvn (approved):
These changes look good to me

calebpollman:
@jordanvn Have we verified the `Authenticator` interfaces include the new service override signature?

kevin860:
@calebpollman I'm not sure if this is what you're referencing, but the `AuthContext` interface (which I did not change) is based on `Partial<typeof defaultServices` (see https://github.com/kevin860/amplify-ui/blob/fa89be92eed0cd3dbf58f4bdfaebac2322334677/packages/ui/src/machines/authenticator/types.ts#L133).

Since the `defaultServices` now includes the new function it shows up on that type.

jordanvn:
We were able to confirm that the interfaces include the new service override signature, however in smoke testing we discovered a small issue related to `resendSignUpCode` separate from these changes. We're going to address that in a different PR before merging these changes in

jordanvn:
Upon further testing it was determined that the behavior identified within `resendSignUpCode` was not a break from expected behavior and will not require a fix. As such, this PR is clear to move forward once it receives sufficient reviews.

joao-squillace-razr:
Any updates on this?

calebpollman (approved):

PR #6333, opened 2025-02-03
fix(angular): stop auth actor when destroying `AuthenticatorService`

QuentinFchx:
#### Description of changes

The xstate machine started during AuthenticatorService initialization was not properly closed.
When starting a machine, it adds a bunch of listeners and register the actor in a global registry ([XState interpreter documentation](https://stately.ai/docs/xstate-v4/xstate/running-machines/intro#interpret) and [source on the v4 branch](https://github.com/statelyai/xstate/blob/xstate%404.38.3/packages/core/src/interpreter.ts#L537)).
Started machine (aka actors) have to be stopped using `.stop()`. Otherwise it seems the created subscriptions and registrations can leak when used in a long-lived running session such as server-side rendering.

I modified my local copy of [the memory-leak reproduction](https://github.com/QuentinFchx/amplify-leak) to include the present fix.
I then ran 2 memory snapshots (one after 2 requests, and one after 202 requests) to see if it was making any differences : it did, the leak disappeared.

#### Additional notes

~~It seems that the Vue implementation has a similar issue : the actor is not stopped when disposing the `useAuth` composable ([as seen here](https://github.com/aws-amplify/amplify-ui/blob/cd508ee3334e217fd8ff8e500e1eb6b7074a12c5/packages/vue/src/composables/useAuth.ts#L53C3-L53C17)).
I didn't have the time to validate the issue for Vue-based SSR environments so the fix here only apply to Angular. Feel free to investigate further.~~

EDIT: the use of `useActor` seems to stop the actor on teardown, please ignore the above note

#### Checklist

- [x] PR description included


jordanvn (dismissed):

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
```diff
@@ -192,6 +192,7 @@ export class AuthenticatorService implements OnDestroy {
   ngOnDestroy(): void {
     if (this._machineSubscription) this._machineSubscription.unsubscribe();
     if (this._unsubscribeHub) this._unsubscribeHub();
+    this._authService.stop();
```
HuiSF, line 195: nit: should we add a unit test to ensure `stop()` is always being called?

cshfang (dismissed):

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
HuiSF: Accessing private members seems an existing pattern in the codebase, we can improve this on our side. I'm good with the current test. Thank you for spending time on this!

HuiSF (approved):

packages/angular/projects/ui-angular/src/lib/services/authenticator.service.ts
calebpollman: I'm good with this as well. We can clean up the typing on our end

calebpollman (approved):
Thanks for the contribution @QuentinFchx, great job!

jordanvn (approved):

ashika112 (approved):

PR #6383, opened 2025-02-25
chore(react-native): update react-native-safe-area-context peer dep

jjarvisp:
#### Description of changes
The purpose of this pull request is to expand the range of supported peer dependency `react-native-safe-area-context` in `@aws-amplify/ui-react-native` from `^4.2.5` to `*` to allow customers to install a version compatible with their version of `react-native`.

- `react-native` is currently 0.78
- `react-native@0.78` requires `react-native-safe-area-context@^5.2.0`
- `react-native` < 0.74 requires `react-native-safe-area-context@^4`
- `@aws-amplify/ui-react-native` currently has a peer dependency with `react-native-safe-area-context@^4.2.5`

[See `react-native-safe-area-context` supported versions here.](https://appandflow.github.io/react-native-safe-area-context/#supported-react-native-versions)

RN version compatibility callout SS
image

Build test with changes to align with recommendation in our docs
https://github.com/aws-amplify/amplify-ui/actions/runs/13706540260

- manual tests
- e2e tests
- build test

#### Checklist

- [X] PR description included
- [X] PR title and commit messages follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax
- [X] _If this change should result in a version bump_, [changeset added](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) (This can be done after creating the PR.) This does not apply to changes made to `docs`, `e2e`, `examples`, or other private packages.

### DIFF
--- a/docs/src/pages/[platform]/getting-started/installation/dependencies.mdx
+++ b/docs/src/pages/[platform]/getting-started/installation/dependencies.mdx
@@ -82,6 +82,22 @@ import { InstallScripts, TerminalCommand } from '@/components/InstallScripts.tsx
 
   <InstallScripts framework="react-native" />
 
+  <Accordion.Container>
+    <Accordion.Item value="Accordion-item">
+      <Accordion.Trigger>
+        Important note for integration with React Native projects using version `0.74` or below
+        <Accordion.Icon />
+      </Accordion.Trigger>
+      <Accordion.Content>
+        Different major versions of peer dependency `react-native-safe-area-context` are designed to be compatible with specific React Native versions.
+        <br />
+        For React Native version `0.74` or below, you must install version `4.x` of `react-native-safe-area-context`.
+        <br />
+        <TerminalCommand command="npm install react-native-safe-area-context@^4.2.5" />
+      </Accordion.Content>
+    </Accordion.Item>
+  </Accordion.Container>
+
   If your project will support Federated Sign In using the React Native `Authenticator` the `@aws-amplify/rtn-web-browser` package is also required:
 
   <TerminalCommand command="npm install @aws-amplify/rtn-web-browser" />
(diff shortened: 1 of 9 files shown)


packages/react-native/package.json
```diff
@@ -38,7 +38,7 @@
     "aws-amplify": "^6.9.0",
     "react": "*",
     "react-native": ">=0.70",
-    "react-native-safe-area-context": "^4.2.5"
+    "react-native-safe-area-context": ">=4.2.5"
```
calebpollman, line 41: Leaning towards loosening the version to `"*"` here unless we have a strong argument not to. Since we have a direct dep on React Native, the compatible version of `react-native-safe-area-context` is dictated by RN itself and not our package's responsibility.

Assuming we go that route, we'll need to update the RN UI build canary install scripts (and ui/primary docs) to specify the `react-native-safe-area-context` version based on the RN versikn.

What you think?

jjarvisp: Yeah, I'm on board. Original thought was to suggest a lower bound compatible with our RN lower bound, but given this needs to be spelled out in our docs anyways during installation it makes sense to leave open.

I'll add docs updates to this PR.

jjarvisp: Updated! Will resolve this convo for now.

docs/src/pages/[platform]/getting-started/installation/dependencies.mdx
```diff
@@ -82,6 +82,22 @@ import { InstallScripts, TerminalCommand } from '@/components/InstallScripts.tsx
...
+      </Accordion.Trigger>
+      <Accordion.Content>
+        Different major versions of peer dependency `react-native-safe-area-context` are designed to be compatible with specific React Native versions.
+        <br />
+        For React Native version `0.74` or below, you must install version `4.x` of `react-native-safe-area-context`.
+        <br />
+        <TerminalCommand command="npm install react-native-safe-area-context@^4.2.5" />
+      </Accordion.Content>
+    </Accordion.Item>
+  </Accordion.Container>
```
Comment (calebpollman) [inline docs/src/pages/[platform]/getting-started/installation/dependencies.mdx:99]: Might make sense to put this in a component and share between the two places where we are using it in this file, but will defer to you since there's probably a lot of duplicated content in this file

packages/react-native/package.json
```diff
@@ -27,7 +27,7 @@
     "@react-native-async-storage/async-storage": "*",
     "@testing-library/react-native": "^12.9.0",
     "metro-react-native-babel-preset": "^0.77.0",
-    "react-native-safe-area-context": "^4.7.3"
+    "react-native-safe-area-context": "^5.2.0"
```
calebpollman, line 30: Why do we need to bump the dev dep?

jjarvisp: I suppose we don't have to... but v5 has larger support overlap with our supported versions.

calebpollman: Does it cause issues with our example app since it's pointing to `0.71.x`? Think `react-native-safe-area-context` is in the `noHoist` list but just want to confirm

jjarvisp: Yes, exactly. No issues with example app, `react-native-safe-area-context` isn't hoisted.

docs/src/pages/[platform]/getting-started/installation/dependencies.mdx
Comment (jjarvisp) [reply docs/src/pages/[platform]/getting-started/installation/dependencies.mdx]: Initially avoided due to rendering conflicts between MDX / TSX files, but I found a workaround and updated.

calebpollman (approved):
🚢

ashika112 (approved):

PR #6392, opened 2025-03-01
fix(authenticator): apply default checked to select mfa type

jjarvisp:
#### Description of changes
The purpose of this PR is apply default checked behavior to the radio groups on the select mfa type screen. We automatically select the first radio option.

#### Checklist

- [X] PR description included
- [X] PR title and commit messages follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax

### DIFF
--- a/packages/vue/src/components/select-mfa-type.vue
+++ b/packages/vue/src/components/select-mfa-type.vue
@@ -77,7 +77,7 @@ const onBackToSignInClicked = () => {
                             :aria-labelledby="`amplify-field-${random}`"
                         >
                             <template
-                                v-for="mfaType in allowedMfaTypes"
+                                v-for="(mfaType, index) in allowedMfaTypes"
                                 :key="mfaType"
                             >
                                 <base-label
@@ -94,6 +94,7 @@ const onBackToSignInClicked = () => {
                                         name="mfa_type"
                                         type="radio"
                                         :value="mfaType"
+                                        :checked="index === 0"
                                     >
                                     </base-input>
                                     <base-text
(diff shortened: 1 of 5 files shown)


packages/react-native/src/Authenticator/hooks/useFieldValues/useFieldValues.ts
```diff
@@ -23,14 +28,26 @@ export default function useFieldValues<FieldType extends TypedField>({
...
+  // select mfa type screen should auto select first radio option
+  const [values, setValues] = useState(() => {
+    const result: Record<string, string> = {};
+    if (isSelectMfaTypeRoute) {
+      const initialValue = fields[0]?.value;
+      if (initialValue) {
+        result.mfa_type = initialValue;
+      }
+    }
+    return result;
```
calebpollman, line 48: Can we update this logic to include a default checked value for `verifyUser`?

packages/vue/src/components/select-mfa-type.vue
```diff
@@ -94,6 +94,7 @@ const onBackToSignInClicked = () => {
                                         name="mfa_type"
                                         type="radio"
                                         :value="mfaType"
+                                        :checked="index === 0"
```
calebpollman, line 97: Curious, does the leading colon of the `checked` attribute indicate a default/initial value?

calebpollman (approved):
LGTM, just left a Vue related question for my own understanding. Thanks for taking care of this!

packages/vue/src/components/select-mfa-type.vue
jjarvisp: Yeah, this is effectively the same as `defaultChecked` in React. There is no automatic two way data binding in Vue's reactivity system so this component remains uncontrolled and we can modify the checked attribute outside the bound JS state.

PR #6426, opened 2025-03-12
fix(ui-react-storage): StorageImage showing fallback text while the image is still fetching

aadimch:
#### Description of changes

Currently, when using `StorageImage`, the user would see the fallback text while the image is still being fetched. The issue is more prominent when the network is throttled. 

This PR uses the loading state exposed by `useGetUrl` and returns a `View` with the provided styles while the image is being fetched. 

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageImage/StorageImage.tsx
+++ b/packages/react-storage/src/components/StorageImage/StorageImage.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import { classNames, ComponentClassName } from '@aws-amplify/ui';
-import { Image } from '@aws-amplify/ui-react';
+import { Image, View } from '@aws-amplify/ui-react';
 import { useDeprecationWarning } from '@aws-amplify/ui-react/internal';
 import { useGetUrl, useSetUserAgent } from '@aws-amplify/ui-react-core';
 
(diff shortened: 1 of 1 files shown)


packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -99,11 +100,16 @@ export const StorageImage = ({
...
+  const { url, isLoading } = useGetUrl(input);
+
+  if (isLoading) {
+    return <View style={{ ...style }}></View>;
+  }
 
   return (
     <Image
       {...rest}
+      style={{ ...style }}
```
calebpollman, line 112: For future reference, no need to spread `style` and create a new object here

packages/react-storage/src/components/StorageImage/StorageImage.tsx
```diff
@@ -99,11 +100,16 @@ export const StorageImage = ({
     ]
   );
 
-  const { url } = useGetUrl(input);
+  const { url, isLoading } = useGetUrl(input);
+
+  if (isLoading) {
+    return <View style={{ ...style }}></View>;
+  }
```
calebpollman, line 107: Rather than destructure `style`, let's add a `loadingElement` key on `StorageImageProps` and render it conditionally while `isLoading` is `true` if provided.

```ts
interface StorageImageProps {
  loadingElement?: React.ReactNode;
}
```

Issue #6378, opened 2025-02-22
Can not input Chinese characters in the textarea of AIConversation in Firefox


zxkane:
### Before creating a new issue, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [x] I have tried disabling all browser extensions or using a different browser
- [x] I have tried deleting the node_modules folder and reinstalling my dependencies
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Other

### How is your app built?

Next.js

### What browsers are you seeing the problem on?

Firefox

### Please describe your bug.

Can't input Chinese characters in the textarea of `AIConversation` component in Firefox browser. It works well in Chrome and Safari.

I can reproduce this problem in the documentation page, https://docs.amplify.aws/react/ai/conversation/ai-conversation/

### What's the expected behaviour?

Input non-English characters for my conversation

### Help us reproduce the bug!

type Chinese or non-English characters into the textarea of AI Conversation

### Code Snippet

```typescript
// Put your code below this line.

```

osama-rizk:
@zxkane ,
This bug is now fixed and released can you please upgrade, can you please upgrade [@aws-amplify/ui-react-ai](https://www.npmjs.com/package/@aws-amplify/ui-react-ai)
 to the to version `1.5.1` or latest. 

Please try it out and let us know if you encounter any issues.

PR #6389, opened 2025-02-28
feat(ai): adding document support

dbanksdesign:
#### Description of changes

This is the last of the changes to support document attachments in the AIConversation component. Changes to other packages are:
* Backend: https://github.com/aws-amplify/amplify-backend/pull/2566
* Data schema: https://github.com/aws-amplify/amplify-data/pull/548

#### Checklist

- [ ] Have read the [Pull Request Guidelines](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md)
- [ ] PR description included
- [ ] `yarn test` passes and tests are updated/added
- [ ] PR title and commit messages follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax
- [ ] _If this change should result in a version bump_, [changeset added](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) (This can be done after creating the PR.) This does not apply to changes made to `docs`, `e2e`, `examples`, or other private packages.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6429, opened 2025-03-13
fix(ui-react-ai): Addressing cross-browser inconsistencies in AIConversation IME input handling

joon-won:
#### Description of changes

- Address the issue AIConversation component was not able to properly update Chinese pinyin IME suggestion

#### Issue #, if available

- https://github.com/aws-amplify/amplify-ui/issues/6378

- Manual testing with local copy of AIConversation

#### Checklist

- [ ] PR description included


packages/react-ai/src/components/AIConversation/views/default/Form.tsx
```diff
@@ -107,8 +107,16 @@ export const Form: Required<ControlsContextProps>['Form'] = ({
...
-          onCompositionEnd={() => setComposing(false)}
+          onCompositionStart={() => {
+            setComposing(true);
+          }}
+          onCompositionEnd={(e) => {
+            setComposing(false);
+            setInput?.((prevValue) => ({
+              ...prevValue,
+              text: (e.target as HTMLTextAreaElement).value,
+            }));
```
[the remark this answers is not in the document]
calebpollman: Casting is a code smell and should be avoided. It looks like the issue here might be the usage of `e.target` instead of `e.currentTarget` in accessing the `value`?

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
```diff
@@ -120,7 +128,7 @@ export const Form: Required<ControlsContextProps>['Form'] = ({
           onChange={(e) => {
             setInput?.((prevValue) => ({
               ...prevValue,
-              text: e.target.value,
+              text: (e.target as HTMLTextAreaElement).value,
```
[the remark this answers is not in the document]
calebpollman: Yes the preference is to never cast. As noted above, casting is an anti-pattern that leads to subtle issues in TS code

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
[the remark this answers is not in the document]
joon-won: Thanks for the suggestion, `e.currentTarget` worked gracefully without a need to add casting, just had to set a default value as it was sometimes `None`. Hopefully this approach is okay

packages/react-ai/src/components/AIConversation/views/default/__tests__/Form.spec.tsx
```diff
@@ -57,4 +58,57 @@ describe('Form', () => {
...
+    await waitFor(() => {
+      fireEvent.compositionStart(textInput);
+      fireEvent.change(textInput, preMatureInput);
+    });
+
+    await waitFor(() => {
+      fireEvent.compositionEnd(textInput, completeInput);
+    });
+
+    expect(textInput.value).not.toBeNull();
```
calebpollman, line 111: When does `textInput.value` resolve to `null`?

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
```diff
@@ -107,8 +107,17 @@ export const Form: Required<ControlsContextProps>['Form'] = ({
...
+          onCompositionStart={() => {
+            setComposing(true);
+          }}
+          onCompositionEnd={(e) => {
+            const composedText = e?.currentTarget?.value || '';
+            setComposing(false);
+            setInput?.((prevValue) => ({
+              ...prevValue,
+              text: composedText,
+            }));
```
calebpollman, line 119: What is the shape of `prevValue`? Tryng to understand the need for spreading the previous state

calebpollman (commented):
Thanks for addressing the feedback. Left a couple small questions, also can we verify whether the fix will impact the [document support feature PR](https://github.com/aws-amplify/amplify-ui/pull/6389)?

packages/react-ai/src/components/AIConversation/views/default/__tests__/Form.spec.tsx
joon-won: Ah realized it can only be an empty value and can't resolve to `null`.

joon-won:
> Thanks for addressing the feedback. Left a couple small questions, also can we verify whether the fix will impact the [document support feature PR](https://github.com/aws-amplify/amplify-ui/pull/6389)?

Yup sounds good, I'll pull from Danny's feature branch and apply the changes to see if nothing's conflicting. When I took a look at it briefly the only change that might cause something is behavioral change in setComposition state. I might need to make a change to match the new behavior

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
joon-won: My understanding is that the interface used for `prevValue` here is [ConversationInput](https://github.com/aws-amplify/amplify-ui/blob/326dfd6928d3273ac95ef5d088946f57afe6f5de/packages/react-ai/src/components/AIConversation/context/ConversationInputContext.tsx#L3) which only have `text` or `files` used to keep in track the data user previously composed.

calebpollman: Got it, thanks!

joon-won:
Cherry-picked AIConversation form changes associated with `onCompositionEnd`, `onCompositionStart`, and `onKeyDown` from the PR and confirmed working with test. Also added a change to AI docs PR to resolve failing unit test on that PR let me know how you think

packages/react-ai/src/components/AIConversation/views/default/__tests__/Form.spec.tsx
```diff
@@ -57,4 +64,57 @@ describe('Form', () => {
...
+
+    const preMatureInput = Object.defineProperty(textInput, 'value', {
+      configurable: true,
+      value: 'しあわせ',
+    });
+
+    const completeInput = Object.defineProperty(textInput, 'value', {
+      configurable: true,
+      value: '幸せならおkです',
+    });
```
HuiSF, line 106: Q: Aren't the `preMatureInput` and `completeInput` both have `value: '幸せならおkです'` after line 106? What are we trying to verify then if they are the same?

calebpollman, line 106: Bit confused on why we are directly mutating the test input `value` property?

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
```diff
@@ -107,20 +105,14 @@ export const Form: Required<ControlsContextProps>['Form'] = ({
...
-          }}
+          onCompositionStart={onCompositionStart}
+          onCompositionEnd={onCompositionEnd}
+          onKeyDown={onKeyDown}
           onChange={(e) => {
+            const composedText = e?.currentTarget?.value || '';
             setInput?.((prevValue) => ({
               ...prevValue,
-              text: e.target.value,
+              text: composedText,
```
calebpollman, line 115: These changes seem unrelated to the issue we are trying to address in this PR? If they are not part of the fix then this increases risk of regression

packages/react-ai/src/components/AIConversation/views/Controls/FormControl.tsx
```diff
@@ -272,6 +283,9 @@ export const FormControl: FormControl = () => {
         isLoading={isLoading}
         error={error}
         setError={setError}
+        onKeyDown={handleOnKeyDown}
+        onCompositionStart={() => setComposing(true)}
+        onCompositionEnd={handleOnCompositionEnd}
```
calebpollman, line 288: Why are these event callbacks typed for a `textarea` but being added to a `Form` interface?

packages/react-ai/src/components/AIConversation/views/default/Form.tsx
joon-won: This change is intended to be a part of the fix, where we decided to go with `e.currentTarget.value` to address eslint's complaint on `e.target.value` not being determined. The issue with `currentTarget.value` is that sometimes it could go undefined, defaulting to an empty string was needed to address the issue.

packages/react-ai/src/components/AIConversation/views/Controls/FormControl.tsx
joon-won: This part was built on Danny's approach with the AI docs PR [here](https://github.com/aws-amplify/amplify-ui/blob/f5a6a16c10d735df7dddf8d639877e9a57c185d0/packages/react-ai/src/components/AIConversation/views/Controls/FormControl.tsx#L344). Should we discuss with Danny on moving the handle to a different file?

packages/react-ai/src/components/AIConversation/views/default/__tests__/Form.spec.tsx
joon-won: Sorry for it being confusing, seems Hui also be concerned about this part. my understanding is that `value` property of `textInput` will not be updated in the actual `textInput` until they're actually passed with `.change` or `onCompositionEnd` which we are testing. Maybe do you think it could be more clear if I add an `expect` line on `textInput.value` before first `change()` is called to ensure the actual object's value is empty after the definition of `preMatureInput` and `completeInput`?

calebpollman (approved):

HuiSF (approved):

PR #6455, opened 2025-03-25
chore(dependency-review): Remove untracked licenses 

joon-won:
#### Description of changes
Removing untracked licenses for test purpose. 
DO NOT MERGE

#### Checklist

- [ ] PR description included


joon-won:
As we were not able to identify dependencies needing the licenses in concern, setting PR ready for review.

.github/dependency-review/config.yml
```diff
@@ -2,7 +2,6 @@ allow-licenses:
   - '0BSD'
   - 'Apache-2.0'
   - 'Apache-2.0 AND MIT'
-  - 'BlueOak-1.0.0'
```
calebpollman, line 5: Let's keep BlueOak-1.0.0 since it is tracked

.github/dependency-review/config.yml
```diff
@@ -20,9 +19,7 @@ allow-licenses:
   - 'JSON'
   - 'MIT'
   - 'MIT AND Zlib'
-  - 'MPL-2.0'
```
ashika112, line 23: did u run the workflow to see if it will pass without these two?

.github/dependency-review/config.yml
joon-won: Got it, will be updating soon

.github/dependency-review/config.yml
joon-won: Yes I believe [this run](https://github.com/aws-amplify/amplify-ui/actions/runs/14069695878/job/39400747967?pr=6455) was testing the repo with the updated allowlist

joon-won: Could it be possible that nested dependencies have been changed since then and it's not longer being used?

joon-won: I've locally verified we still have 5 dependencies using `BlueOak-1.0.0` and 2 for `MPL-1.0.0`

PR #6597, opened 2025-06-20
fix(liveness): cypress/e2e liveness test issue

osama-rizk:
#### Description of changes
Currently multiple liveness e2e test fails due file definition lacking (error details below). This PR will fix that by adding a step definition to handle the selection element. 

```
     Error: 
Step implementation missing for "I click the "FaceMovementAndLightChallenge" selectfield and select the "FaceMovementChallenge" option".

We tried searching for files containing step definitions using the following search pattern templates:

```

- By running the liveness e2e test successfully `yarn test:liveness `

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/e2e/cypress/integration/common/liveness.ts
+++ b/packages/e2e/cypress/integration/common/liveness.ts
@@ -12,3 +12,10 @@ Then('I see the {string} timeout error', (message: string) => {
     })
     .should('exist');
 });
+
+Then(
+  'I click the {string} selectfield and select the {string} option',
+  (selectFieldName: string, optionValue: string) => {
+    cy.get('select').scrollIntoView().select(optionValue);
+  }
+);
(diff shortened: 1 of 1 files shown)


packages/e2e/cypress/integration/common/liveness.ts
```diff
@@ -12,3 +12,10 @@ Then('I see the {string} timeout error', (message: string) => {
     })
     .should('exist');
 });
+
+Then(
+  'I click the {string} selectfield and select the {string} option',
+  (selectFieldName: string, optionValue: string) => {
+    cy.get('select').scrollIntoView().select(optionValue);
```
adrianjoshua-strutt: question: Is there a reason we are not using the `selectFieldName` and instead we match any select element? Also, should we add the `timeout: LIVENESS_TIMEOUT` to cy.get, as seen in the code above?

osama-rizk: I updated the code to locate the select element by its display name using `selectFieldName`. Using the display name is a better approach as it prevents test failures when additional select elements are added to the page in the future.

osama-rizk: > should we add the timeout: LIVENESS_TIMEOUT to cy.get, as seen in the code above?

Adding the timeout isn't necessary here. This select element is static and rendered immediately with the component, so it should be available right away. The LIVENESS_TIMEOUT is specifically for elements that appear after async operations complete, like the error message in the step above

adrianjoshua-strutt (approved):

PR #6611, opened 2025-07-01
fix: null-check for getSettings

adrianjoshua-strutt:
#### Description of changes

We blindly call getSettings on the the stream.getTracks(). This can cause NPEs. This PR adds null-safety to avoid this.

yarn test
e2e tests: 

#### Checklist

- [x ] Have read the [Pull Request Guidelines](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md)
- [x ] PR description included
- [x ] `yarn test` passes and tests are updated/added
- [x ] PR title and commit messages follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/hooks/useMediaStreamInVideo.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/hooks/useMediaStreamInVideo.ts
@@ -25,12 +25,12 @@ export function useMediaStreamInVideo(
       if (isObject(videoRef.current)) {
         videoRef.current.srcObject = stream;
       }
-      const { height: streamHeight, width: streamWidth } = stream
-        .getTracks()[0]
-        .getSettings();
 
-      setVideoHeight(streamHeight);
-      setVideoWidth(streamWidth);
+      const settings = stream.getTracks()?.[0]?.getSettings();
+      if (settings) {
+        setVideoHeight(settings.height);
+        setVideoWidth(settings.width);
+      }
     }
 
     return () => {


pranavosu (commented):
we should add a unit test for this

packages/react-liveness/src/components/FaceLivenessDetector/hooks/useMediaStreamInVideo.ts
```diff
@@ -25,12 +25,12 @@ export function useMediaStreamInVideo(
       if (isObject(videoRef.current)) {
         videoRef.current.srcObject = stream;
       }
-      const { height: streamHeight, width: streamWidth } = stream
-        .getTracks()[0]
-        .getSettings();
 
-      setVideoHeight(streamHeight);
-      setVideoWidth(streamWidth);
+      const settings = stream.getTracks()?.[0]?.getSettings();
```
osama-rizk, line 29: Should we add "?." on calling `getSettings?.()`. if getTracks return empty array then `[0]` will be undefined and if we call it `undefined.getSettings()` It will throw an error `Uncaught TypeError: [0]?.getSettings is not a function`. Suggested change 

```
const settings = stream?.getTracks?.()?.[0]?.getSettings?.();
```

adrianjoshua-strutt: Thanks for the feedback. However, I do not think that this is necessary. 

1. Stream (`MediaStream`) is not nullable and we can expect it to be present. I would argue that throwing an error here, instead of safely casting it to an undefined, would be appropriate. 
2. `[0]` can indeed be undefined, yes. That is why the `.[0]?` was added. The check that you added to `getTracks`, as far as I understand, checks if the method `getTracks` is present on the object, which should always be the case for a `MediaStream` object as it's part of the standard API.
3. The check for `getSettings?.()` seems a bit redundant. While it would prevent a runtime error if the method doesn't exist, `getSettings` is a standard method on `MediaStreamTrack` objects. Additionally, our existing `.[0]?` already prevents calling it on undefined, and we check if the result is undefined right after this line with `if (settings) {`.

This is also reflected by the passing unit tests.
Please, let me know if I am misunderstanding something here :)

osama-rizk (approved):

PR #6746, opened 2025-11-11
feat(storage-browser): add multi-file zipdownload

bobbor:
#### Description of changes

This change introduces the possibility to download multiple files in form of a zip-file.
ZipFiles are supported by all operating systems.

This change adds a new argument to the handlers, which contains all files the handler will be called on.

#### Issue #, if available
_- none -_

- I downloaded multiple files locally.
  - small files and large files
- Unit-tests have been added / adjusted
- e2e tests

#### Checklist

- [x] PR description included

### DIFF
--- a/.changeset/smooth-horses-peel.md
+++ b/.changeset/smooth-horses-peel.md
@@ -1,0 +1,5 @@
+---
+"@aws-amplify/ui-react-storage": minor
+---
+
+feat(storage-manager): add multi-file zipdownload


bobbor:
~~there is no changeset by design.~~
~~the implementation can be used in next (unstable) releases, but should not be part of a version~~

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -0,0 +1,174 @@
...
+                if (blobURL) {
+                  const anchor = document.createElement('a');
+                  const clickEvent = new MouseEvent('click');
+                  anchor.href = blobURL;
+                  anchor.download = `${folder || 'archive'}.zip`;
+                  anchor.dispatchEvent(clickEvent);
+                }
+              });
+            }
+          }, 250);
```
ahmedhamouda78: question: what's the reason behind 250ms specifically?

bobbor: so the issue is, that the handler gets invoked for every file individually.
The handler itself manages a queue of files to download or which are downloading.

Since the download-handler, is only invoked on 4 files at a time, it can happen, that before the next file is being queued by the storage-manager, the download-handler "thinks" the queue is empty and assumes it has processed all files.

this small delay, gives the storage-manager some time push more files into the download-handler, and thus continue downloading.

It would be much more beneficial, if the action-handlers would know how many files are being processed in total e.g. like `Array.map`, where the signature is `<T, R>(item: T, index: number, all: T[]) => R[]`

but I'll do a follow-up on this one in another PR

osama-rizk:
> there is no changeset by design. the implementation can be used in next (unstable) releases, but should not be part of a version

Can you please elaborate more?

osama-rizk (commented):
Can you please update the PR description explaining the changes on high level, also  test and verification.

packages/react-storage/src/components/StorageBrowser/actions/handlers/download.ts
```diff
@@ -53,14 +53,14 @@ export const downloadHandler: DownloadHandler = ({ config, data }) => {
...
+  .then(({ url }) => {
+    downloadFromUrl(key, url.toString());
+    return { status: 'COMPLETE' as const, value: { url } };
+  })
+  .catch((error: Error) => {
+    const { message } = error;
+    return { error, message, status: 'FAILED' as const };
+  });
 
   return { result };
```
osama-rizk, line 65: Do we need to define the cancel callback here? 


```
  return {
    result,
    cancel: () => {
      //cancel download
    },
  };
```

bobbor: this is the old download handler.

This creates 5 million attachment-links and clicks on them.

there is no canceling that.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
bobbor: I have removed the timeout, as I introduced an `all` property to the taskHandler.

basically I did what I suggested in my previous comment:
> It would be much more beneficial, if the action-handlers would know how many files are being processed in total e.g. like Array.map, where the signature is <T, R>(item: T, index: number, all: T[]) => R[]

PR #6801, opened 2026-01-13
fix: resolve @smithy/config-resolver security vulnerabilities

adrianjoshua-strutt:
- *Increases package sizes*
- Update AWS SDK packages to latest versions (3.967.0)
- Update aws-amplify to 6.15.9
- Add resolution for @smithy/config-resolver ^4.4.5 to ensure all transitive dependencies use safe version
- Keep environment packages using 'latest' for automatic security updates
- Fixes CVE with CVSS 3.7 (LOW) - region validation vulnerability

All @smithy/config-resolver versions now >= 4.4.0 (safe) Resolves Dependabot alerts:

https://github.com/aws-amplify/amplify-ui/security/dependabot/284

https://github.com/aws-amplify/amplify-ui/security/dependabot/283

https://github.com/aws-amplify/amplify-ui/security/dependabot/280

https://github.com/aws-amplify/amplify-ui/security/dependabot/276

https://github.com/aws-amplify/amplify-ui/security/dependabot/275

#### Checklist

- [ ] PR description included

### DIFF
--- a/environments/liveness/liveness-environment/amplify/backend/function/livenessenvironment09873267/src/package-lock.json
+++ b/environments/liveness/liveness-environment/amplify/backend/function/livenessenvironment09873267/src/package-lock.json
@@ -9,7 +9,8 @@
       "version": "2.0.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@aws-sdk/client-rekognition": "latest"
+        "@aws-sdk/client-rekognition": "latest",
+        "bowser": "2.11.0"
       },
       "devDependencies": {
         "@types/aws-lambda": "^8.10.92"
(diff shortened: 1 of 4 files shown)


osama-rizk (dismissed):

sarayev (dismissed):

osama-rizk (approved):

.github/dependency-review/config.yml
```diff
@@ -18,6 +18,7 @@ allow-licenses:
   - 'ISC'
   - 'JSON'
   - 'MIT'
+  - 'MITNFA'
```
sarayev: Why do we need the new license?

adrianjoshua-strutt: This is required because bowser (a transitive dependency) uses MITNFA instead of the standard MIT license that's already allowlisted. MITNFA is essentially MIT with one additional clause: if you modify the software and redistribute it, you must update or remove the original author's contact info to avoid misattributing bugs you introduced. Since we're just consuming bowser as-is without modifications, this clause should not affect us. But I already created a thread to discuss this internally.

bobbor: I don't think that introducing MITNFA is a problem here, but:

1. it is weird that you need to introduce `yarn.lock` files for environments, where there haven't been any before. (I see `package-lock.json` files in some)
2. if adding lock files is necessary, could you edit them and pin `bowser` to `2.11.0` like we do in the root `yarn.lock`? since this version seems ok - license wise.
    ```
    bowser@^2.11.0:
    version "2.11.0"
    resolved "https://registry.npmjs.org/bowser/-/bowser-2.11.0.tgz"
    integrity sha512-AlcaJBi/pqqJBIQ8U9Mcpc9i8Aqxn88Skv5d+xBX006BY5u8N3mGLHa5Lgppa7L/HfwgwLgZ6NYs+Ag6uUmJRA==
    ```

That should remove the need for the license change.

adrianjoshua-strutt: You are correct in pointing that out. I removed the unnecessary yarn.lock file and pinned the bowser version. This allowed me to remove MITFNA from the allowlist

PR #6816, opened 2026-01-23
chore(deps): update aws-amplify version.

osama-rizk:
#### Description of changes

Updated `aws-amplify` peer dependency from `^6.14.3` to `^6.16.0` across all packages in the monorepo to ensure version consistency and prevent dependency conflicts for end users. Also updated the `aws-amplify` version in the root `package.json`.

**Files updated:**
- Root `package.json` (both peer dependency and regular dependency)
- All package-specific `package.json` files (React, Vue, Angular, React Native, and related packages)
- Updated `yarn.lock` to reflect dependency changes

This change aligns all packages to use the same aws-amplify version, following monorepo best practices.

#### Issue #, if available

N/A - Maintenance update to align aws-amplify versions

- Verified all `package.json` files now reference `aws-amplify: "^6.16.0"`
- Confirmed root `package.json` aws-amplify version updated to `^6.16.0`
- Confirmed `yarn.lock` was properly updated
- No breaking changes expected as this is a minor version bump within the same major version range

#### Checklist

- [x] PR description included

### DIFF
--- a/.changeset/sixty-bats-thank.md
+++ b/.changeset/sixty-bats-thank.md
@@ -1,7 +1,5 @@
 ---
-"@aws-amplify/ui-docs": patch
-"@aws-amplify/ui-next-app-router-example": patch
-"@aws-amplify/ui-next-pages-router-example": patch
+"@aws-amplify/ui-react-liveness": patch
 ---
 
 feat: Add nextjs 16 compatibility.


package.json
```diff
@@ -152,7 +152,7 @@
     "@types/react-dom": "^18.3.0",
     "@types/react-test-renderer": "^18.3.0",
     "@vitejs/plugin-vue": "^2.3.4",
-    "aws-amplify": "6.15.9",
+    "aws-amplify": "^6.16.0",
```
ahmedhamouda78: Is it safe to use caret versioning here? The previous version was specifically pinned.

packages/angular/projects/ui-angular/package.json
```diff
@@ -16,7 +16,7 @@
   "peerDependencies": {
     "@angular/core": ">= 19.0.0",
     "@angular/common": ">= 19.0.0",
-    "aws-amplify": "^6.14.3",
+    "aws-amplify": "^6.16.0",
```
ahmedhamouda78: Any reason we need to change this (i.e. hot-fix, etc.)? I am asking because the caret versioning we had before should install the latest minor version by default while being backwards compatible with versions all the way to 6.14.3.

This question also applies to the other package.json files.

package.json
osama-rizk: I used the pinned version.

bobbor:
short question:

have you tried simply removing all `yarn.lock` files and then reinstalling all dependencies?
that should unpin the resolved versions and achieve the same as without updating every single amplify dependency.

packages/angular/projects/ui-angular/package.json
bobbor: @ahmedhamouda78 this is due to version resolution in the `yarn.lock`

if it resolves to (let's say) 6.14.5 it fulfills `^6.14.3` and would never bump the version (unless major release as you say).

deleting the yarn.lock (or at least the aws-amplify segments) would do the trick

osama-rizk: > deleting the yarn.lock

I deleted yarn.lock and re-install.

PR #6853, opened 2026-02-16
fix(authenticator): exclude username from user attributes in force change password flow

sarayev:
#### Description of changes

This PR fixes a bug in the Authenticator's force change password flow where the `username` field was being incorrectly sent as a user attribute to AWS Cognito's `confirmSignIn` API.

**Changes:**
- Modified `handleForceChangePassword` service in `signIn.ts` to explicitly exclude `username` from the destructured `userAttributes` object
- Added comprehensive unit tests to verify that `username`, `confirm_password`, and `country_code` are properly excluded from user attributes
- Tests also verify proper phone number formatting with country codes

**Impact:**
This prevents potential API errors when username is present in form values during the force change password flow. The username field is not a valid user attribute for the `confirmSignIn` operation and should be filtered out along with other UI-only fields.

1. Verified in a sample app that reproduces the issue. 

2. **Unit Tests**: Added 3 new test cases in `signIn.test.ts` that verify:
   - Username is excluded from userAttributes when calling confirmSignIn
   - Phone number formatting works correctly and country_code is excluded
   - Empty userAttributes are handled when only password fields are provided

2. **Test Execution**: All tests pass successfully
   ```bash
   npm test -- --testPathPattern=signIn.test.ts --testNamePattern="handleForceChangePassword"

### DIFF
--- a/.changeset/nice-buses-train.md
+++ b/.changeset/nice-buses-train.md
@@ -1,0 +1,6 @@
+---
+'@aws-amplify/ui': patch
+'@aws-amplify/ui-react': patch
+---
+
+fix(authenticator): exclude username from user attributes in force change password flow
--- a/packages/ui/src/machines/authenticator/actors/__tests__/signIn.test.ts
+++ b/packages/ui/src/machines/authenticator/actors/__tests__/signIn.test.ts
@@ -571,4 +571,127 @@ describe('signInActor', () => {
     await flushPromises();
     expect(service.getSnapshot().value).toStrictEqual('rejected');
   });
+
+  describe('handleForceChangePassword service', () => {
+    let confirmSignInSpy: jest.SpyInstance;
+    let testService: any;
+
+    beforeEach(() => {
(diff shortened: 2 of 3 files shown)


.changeset/nice-buses-train.md
```diff
@@ -0,0 +1,6 @@
+---
+'@aws-amplify/ui': patch
```
osama-rizk, line 2: Why we included only these two packages? All ui packages depend on `@aws-amplify/ui` so perform any changes in this package should pump all the packages. Also , please keep in mind that authenticator is supported for other frameworks e.g Angular, vue ...etc

soberm: All of our packages use exact versions. So bumping @aws-amplify/ui will automatically do a patch bump for all packages that depend on it. @sarayev You can also verify this by executing `changeset version` locally.

sarayev: Verified that, it works exactly as Michael described.

PR #6926, opened 2026-03-25
feat(storage-browser): improve download

bobbor:
#### Description of changes
this PR addresses two issues, which have been reported
1. do not zip single file download. (for this a facade has been introduced as downloadHandler)
2. switch from `jszip` to `@zip.js/zip.js` as its memory and fs performance is much better
    a. this comes at the cost of an increased bundle size, but allows the download of >10GB zips

adjust bundle-size for storage-browser.

#### Issue #, if available
_- none -_

the following flow:
1. create 10 files ~1GB in size using `urandom` (overall ~10GB)
2. upload them through the storage-browser
3. download them using storage-browser
4. Verify that:
  1. the zip is not corrupted
  2. all files are in the zip
  3. all files have the same hash (sha256 and md5) than the original files. 

also: 
- `yarn lint`
- `yarn test`
- `yarn size`
#### Checklist

- [x] PR description included

### DIFF
--- a/build-system-tests/e2e/package.json
+++ b/build-system-tests/e2e/package.json
@@ -31,7 +31,7 @@
     "undici": "6.24.0",
     "@actions/http-client/undici": "6.24.0",
     "diff": "8.0.3",
-    "serialize-javascript": "^7.0.3",
+    "serialize-javascript": "^7.0.5",
     "**/find-cypress-specs/minimatch": "3.1.4",
     "**/mocha/minimatch": "5.1.8",
     "**/@typescript-eslint/typescript-estree/minimatch": "9.0.7",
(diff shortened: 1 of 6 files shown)


packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -215,7 +162,7 @@ const downloadHandler = (() => {
         .then((): DownloadTaskResult => {
           fileDownloadQueue.set(key, true);
           return {
-            status: 'LOADED',
+            status: 'COMPLETE' as const,
```
ahmedhamouda78, line 165: Is the zipdownload handler exposed for customization by customers? If yes, could some of the changes in these file cause any breaking changes?

bobbor: I don't think so:

| Flow | before | after |
| --- | --- | --- |
| for 1 file | PENDING → LOADED → FINISHING → COMPLETE | PENDING → COMPLETE |
| for 2+ files | for each file: PENDING → LOADED _waiting*_ → FINISHING _waiting*_ → COMPLETE | for each file: PENDING → LOADED → FINISHING → COMPLETE |

_*) waiting means: waiting for all other files to reach the same state. e.g. all files need to be downloaded, before they can be zipped_

and the reason why I don't think that it will break flows, is that all `onProgress` callbacks get `_data`, which is the file downloaded, `progress` and `status` and their signature + moment of invocation has not changed. just the order.

Issue #6966, opened 2026-04-23
Wrong, hardcoded punctuation in Japanese


jiw-mh:
### Before creating a new issue, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [x] I have tried disabling all browser extensions or using a different browser
- [x] I have tried deleting the node_modules folder and reinstalling my dependencies
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

### On which framework/platform are you having an issue?

React

### Which UI component?

Authenticator

### How is your app built?

Vite

### Please describe your bug.

In some translation texts, the regular '.' is hard coded. For example:

https://github.com/aws-amplify/amplify-ui/blob/a6470c163288a6dcd928959d41b4cf74e71ddd27/packages/ui/src/helpers/authenticator/textUtil.ts#L40

In japanese the ends of sentences use `。` instead. It would be good not assume any punctuation. Not sure it wouldn't be a good idea to also open the sentence order to reduce the length of a sentence in some language packs.

### What's the expected behaviour?

To have a japanese punctuation consistently.

### Help us reproduce the bug!

Load the Login component in a japanese display.

bobbor:
hey @jiw-mh 

thanks for letting us know. I think you are right, we shouldn't do this.

just out of curiosity, the `.` is a different one to `。` by size and followed by a space, right. very interesting.

I'll handle this as a bug and will start working on it.

Cheers Philipp

jiw-mh:
`。` is a full-width character. In Japanese/Chinese all characters are mono-spaced. The `。` is left-aligned and thus it looks like it has some space added to it.

bobbor:
TIL.

I've worked on the bug today and have created a PR:
https://github.com/aws-amplify/amplify-ui/pull/6968

all punctuation should now be / can be  part of the translation. If I haven't missed any, there is no more period `.` (or other punctuation for that matter) in the code, it's all in the translation. 

Also, it has some extra functionality added to it, so that some strings can be easier constructed. translation strings allow for placeholders like `{foo}` which can then be resolved via 2nd parameter to the translate function

e.g.
```
// in translations
const texts = {
  MOTORCYCLE: "I once owned a {brand}."
}

[...]

PR #6968, opened 2026-04-24
Fix/bobbor/i18n texts

bobbor:
#### Description of changes

This change introduces a more fine-grained control over translation texts in the Authenticator.

1.) **strings do accept placeholders and values now.**

translations with different positions of values is allowed. no more fragile construction in the source code.

```
// in english
an email has been sent to {emailAddress}.

// in german
Wir haben eine Email an {emailAddress} gesendet.
```

2.) **punctuation is part of translation**

Since the japanese language uses different period signs for delimiting sentences and in thai it is common to not use any at all. the translation controls punctuation.
This allows for a more natural way of communcation. 

3.) **backwards compatibility**
a small number of keys has been introduced to the translation files for backwards compatibility and to even out cognito inconsistencies.

#### Issue #, if available
https://github.com/aws-amplify/amplify-ui/issues/6966

- the textUtils class contains unit-tests. these still pass.
- the Authenticator still renders.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/ui/src/i18n/translations.ts
+++ b/packages/ui/src/i18n/translations.ts
@@ -68,8 +68,17 @@ export type Dict = Record<string, string>;
  *
  * You can also use translate<string> to handle custom strings or dynamic content.
  */
-export function translate<T extends string>(phrase: T): string {
-  return I18n.get(phrase);
+export function translate<T extends string>(
+  phrase: T,
+  values?: Record<string, string>
+): string {
+  let result = I18n.get(phrase);
+  if (values) {
+    for (const [key, val] of Object.entries(values)) {
+      result = result.replaceAll(`{${key}}`, val);
+    }
+  }
+  return result;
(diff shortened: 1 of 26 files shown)


packages/ui/src/i18n/translations.ts
```diff
@@ -68,8 +68,17 @@ export type Dict = Record<string, string>;
  *
  * You can also use translate<string> to handle custom strings or dynamic content.
  */
-export function translate<T extends string>(phrase: T): string {
-  return I18n.get(phrase);
+export function translate<T extends string>(
```
jiw-mh, line 71: It feels like this should maybe be part of: https://github.com/aws-amplify/amplify-js/blob/93487ff0967b8b4f752aca4aacd341052b343177/packages/core/src/I18n/I18n.ts#L61-L82

PR #7005, opened 2026-06-04
fix(react-liveness): add onError handlers to prevent stuck state on model load failure

pranavosu:
## Description

Adds missing `onError` handlers to face detection and stream initialization invoke states in the liveness state machine.

## Problem

When the BlazeFace WASM backend or model fails to load (network issues, CDN unreachable, WASM unsupported on device), `detectFaces()` throws a `TypeError` because `this._model` is undefined. The `detectFaceBeforeStart`, `detectFaceDistanceBeforeRecording`, and `initializeLivenessStream` states had no `onError` transition defined. In XState 4, when an invoked promise rejects without an `onError` handler, the machine silently stays stuck in that state forever — no error callback fires, no timeout triggers.

This causes the component to appear permanently frozen with the camera active but no oval drawn and no error surfaced to the integrator.

## Changes

- Add `onError` transitions to `detectFaceBeforeStart`, `detectFaceDistanceBeforeRecording`, and `initializeLivenessStream` states, all routing to the `error` state with `updateErrorStateForRuntime`
- Remove the silent `try/catch` around `modelLoadingPromise` in `detectFace` and `detectInitialFaceAndDrawOval` services so model load failures propagate as rejections instead of causing a downstream TypeError on `this._model.estimateFaces()`

## How it was tested

- All 36 existing test suites pass (265 tests)
- TypeScript compiles cleanly (`tsc --noEmit`)
- Verified with a local XState test that a rejected invoke without `onError` stays stuck, and with `onError` transitions to the error state correctly

### DIFF
--- a/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
+++ b/packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
@@ -260,6 +260,10 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
                   'spawnResponseStreamActor',
                 ],
               },
+              onError: {
+                target: '#livenessMachine.error',
+                actions: 'updateErrorStateForRuntime',
+              },
             },
           },
           waitForSessionInfo: {
@@ -288,6 +292,10 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
             target: 'checkFaceDetectedBeforeStart',
             actions: 'updateFaceMatchBeforeStartDetails',
           },
+          onError: {
+            target: 'error',
+            actions: 'updateErrorStateForRuntime',
+          },
         },
       },
       checkFaceDetectedBeforeStart: {
@@ -305,6 +313,10 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
           onDone: {
             target: 'checkFaceDistanceBeforeRecording',
             actions: 'updateFaceDistanceBeforeRecording',
+          },
+          onError: {
+            target: 'error',
+            actions: 'updateErrorStateForRuntime',
           },
         },
       },
@@ -1093,12 +1105,7 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
         const { faceDetector } = context.ovalAssociatedParams!;
 
         // initialize models
-        try {
-          await faceDetector!.modelLoadingPromise;
-        } catch (err) {
-          // eslint-disable-next-line no-console
-          console.log({ err });
-        }
+        await faceDetector!.modelLoadingPromise;
 
         // detect face
         const faceMatchState = await getFaceMatchState(faceDetector!, videoEl!);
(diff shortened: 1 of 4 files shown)


packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
```diff
@@ -260,6 +260,10 @@ export const livenessMachine = createMachine<LivenessContext, LivenessEvent>(
                   'spawnResponseStreamActor',
                 ],
               },
+              onError: {
+                target: '#livenessMachine.error',
```
soberm, line 264: The other two states only use `'error'`

packages/react-liveness/src/components/FaceLivenessDetector/service/types/faceDetection.ts
```diff
@@ -28,9 +30,28 @@ export abstract class FaceDetection {
...
    * Triggers the `loadModels` method and stores
    * the corresponding promise to be awaited later.
+   * Applies a timeout to prevent indefinite hangs if the
+   * network request for model assets stalls.
    */
   triggerModelLoading(): void {
-    this.modelLoadingPromise = this.loadModels();
+    this.modelLoadingPromise = Promise.race([
+      this.loadModels(),
+      new Promise<void>((_, reject) =>
```
soberm: When `loadModels()` wins the race and resolves first, the `setTimeout` callback remains scheduled in the event loop. It will fire 10s later, call `reject()` on an already-settled promise (no-op), but the timer reference itself stays alive until then. Each call leaks a timer.

packages/react-liveness/src/components/FaceLivenessDetector/service/types/faceDetection.ts
```diff
@@ -28,9 +30,28 @@ export abstract class FaceDetection {
   /**
    * Triggers the `loadModels` method and stores
    * the corresponding promise to be awaited later.
+   * Applies a timeout to prevent indefinite hangs if the
+   * network request for model assets stalls.
    */
   triggerModelLoading(): void {
-    this.modelLoadingPromise = this.loadModels();
+    this.modelLoadingPromise = Promise.race([
```
soberm, line 45: We could add a guard to ensure that there is only one load.

packages/react-liveness/src/components/FaceLivenessDetector/service/utils/__tests__/blazefaceFaceDetection.test.ts
```diff
@@ -98,4 +103,38 @@ describe('blazefaceFaceDetection', () => {
 
     expect(model.modelBackend).toBe('cpu');
   });
+
+  describe('triggerModelLoading timeout', () => {
```
soberm, line 107: The test never exercises the race condition. We could advance time here to validate it.

packages/react-liveness/src/components/FaceLivenessDetector/service/machine/machine.ts
pranavosu: `#livenessMachine.error` is correct for the nested state (needs full path to escape initWebsocket) e.g `recording.ovalDrawing`

soberm (approved):

PR #7044, opened 2026-07-06
feat(storage): service worker streaming zip download

bobbor:
Replace the in-memory blob-based zip download with a service worker streaming architecture for StorageBrowser multi-file downloads. Falls back to blob collection when the SW is unavailable.

Key changes:
- Add download-sw.ts service worker with skipWaiting + clients.claim
- Add zipdownload handler with WeakMap-scoped batch state (no singleton)
- Stream files sequentially into zip via zip.js (concurrency=1)
- MessageChannel handshake for SW stream transfer + keepalive pings
- Blob fallback with inline cleanup (no fire-and-forget race)
- composedDownloadHandler routes single-file vs multi-file downloads
- useServiceWorkerRegistration hook for automatic SW registration
- copy-serviceworker bin script for npm consumers
- Documentation for SW setup across frameworks
- Comprehensive test suite with deterministic async flushing

#### Checklist

- [ ] PR description included

### DIFF
--- a/.github/workflows/reusable-build-system-test-react-native.yml
+++ b/.github/workflows/reusable-build-system-test-react-native.yml
@@ -10,7 +10,16 @@ on:
 
 jobs:
   build:
-    runs-on: macos-15-intel
+    # Android builds run on ubuntu-latest so the emulator can use hardware (KVM)
+    # acceleration. The previous macos-15-intel runner regularly wedged/segfaulted
+    # the emulator boot under `-gpu host` (see #7031).
+    #
+    # NOTE: this repo restricts GitHub Actions to an allow-list (GitHub-authored,
+    # aws-actions/*, aws-amplify/* and changesets/action), so the emulator is
+    # launched with the Android SDK CLI directly rather than a third-party action.
+    runs-on: ubuntu-latest
+    # Bound the whole job so a wedged emulator boot can't run until GitHub's 6h hard limit.
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
@@ -47,49 +56,12 @@ jobs:
         env:
           SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
 
-      - name: Restore CocoaPods cache
-        if: ${{ matrix.platform == 'ios' }}
-        id: restore-cocoapods-cache
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: ./examples/react-native/ios/Pods
-          key: ${{ runner.os }}-cocoapods
-          restore-keys: pods-${{ hashFiles('examples/react-native/ios/Podfile.lock') }}
-        env:
-          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
-
-      - name: Restore node_modules cache
-        if: ${{ matrix.platform == 'ios' }}
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        id: restore-cache
-        with:
-          path: |
-            ./node_modules
-            **/node_modules
-          key: ${{ runner.os }}-node24-nodemodules
-        env:
-          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 3
-
       - name: Install Java 17
         if: ${{ matrix.platform == 'android' }}
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           distribution: 'corretto' # Amazon Corretto Build of OpenJDK
           java-version: '17'
-
-      - name: Cache Android SDK and emulator
-        if: ${{ matrix.platform == 'android' }}
-        id: android-cache
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: |
-            ~/Library/Android/sdk/system-images
-            ~/Library/Android/sdk/build-tools
-            ~/Library/Android/sdk/platform-tools
-            ~/.android/avd
-          key: ${{ runner.os }}-android-sdk-27-x86_64-snapshot-v1
-        env:
-          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 5
 
       - name: Cache Gradle dependencies
         if: ${{ matrix.platform == 'android' }}
@@ -104,36 +76,50 @@ jobs:
         env:
           SEGMENT_DOWNLOAD_TIMEOUT_MINS: 3
 
-      - name: Install iOS simulator
-        if: ${{ matrix.platform == 'ios' }}
+      # KVM enables hardware acceleration for the Android emulator on Linux runners.
+      - name: Enable KVM
+        if: ${{ matrix.platform == 'android' }}
         run: |
-          brew tap wix/brew
-          brew install applesimutils
-          brew install watchman
-          brew link --overwrite python@3.11
-          echo "ruby --version"
-          ruby --version
-        continue-on-error: true # brew overwrite step addresses a python install issue: https://github.com/actions/runner-images/issues/8500
-
-      - name: Update CocoaPods
-        if: ${{ matrix.platform == 'ios' }}
-        run: |
-          gem update cocoapods xcodeproj
-          yarn react-native-example ios:pod-install
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
 
       - name: Install Android emulator
         if: ${{ matrix.platform == 'android' }}
         run: |
-          echo -e "echo \$ANDROID_HOME"
-          echo $ANDROID_HOME
-          $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --install "build-tools;33.0.2" "platform-tools" "system-images;android-27;default;x86_64"
-          $ANDROID_HOME/cmdline-tools/latest/bin/avdmanager create avd --name Pixel_5_API_27 --force --device "pixel_5" --abi x86_64 --package "system-images;android-27;default;x86_64"
+          echo "ANDROID_HOME=$ANDROID_HOME"
+          # The new avdmanager writes AVDs to $HOME/.config/.android/avd, but the
+          # emulator looks in $ANDROID_AVD_HOME (default $HOME/.android/avd). Pin both
+          # tools to the same directory so the created AVD is actually found, and
+          # persist it via $GITHUB_ENV so the "Start Android emulator" step agrees.
+          export ANDROID_AVD_HOME="$HOME/.android/avd"
+          echo "ANDROID_AVD_HOME=$ANDROID_AVD_HOME" >> "$GITHUB_ENV"
+          mkdir -p "$ANDROID_AVD_HOME"
+          yes | $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --licenses > /dev/null || true
+          $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --install "platform-tools" "emulator" "build-tools;33.0.2" "platforms;android-27" "system-images;android-27;default;x86_64"
+          # `--tag` and `--abi` must be passed explicitly: without them avdmanager
+          # prints "Auto-selecting single ABI" and exits 0 WITHOUT creating the AVD.
+          echo "no" | $ANDROID_HOME/cmdline-tools/latest/bin/avdmanager --verbose create avd --force --name Pixel_5_API_27 --package "system-images;android-27;default;x86_64" --tag "default" --abi "x86_64" --device "pixel_5"
+          # Fail fast (instead of waiting 420s for a boot that can't happen) if the AVD was not created.
+          echo "Available AVDs:"
+          $ANDROID_HOME/emulator/emulator -list-avds
+          if ! $ANDROID_HOME/emulator/emulator -list-avds | grep -qx "Pixel_5_API_27"; then
+            echo "::error::AVD Pixel_5_API_27 was not created by avdmanager"
+            $ANDROID_HOME/cmdline-tools/latest/bin/avdmanager list avd || true
+            exit 1
+          fi
 
-      - name: Start Android emulator (cold boot - first run)
-        if: ${{ matrix.platform == 'android' && steps.android-cache.outputs.cache-hit != 'true' }}
+      - name: Start Android emulator
+        if: ${{ matrix.platform == 'android' }}
         run: |
-          $ANDROID_HOME/emulator/emulator -avd Pixel_5_API_27 -port ${{ env.EMULATOR_PORT }} -no-boot-anim -no-audio -no-snapshot-load -gpu host -accel on &
-          $ANDROID_HOME/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'
+          # Headless launch with software GPU (swiftshader) + KVM; this is the
+          # reliable CI combination on Linux runners.
+          $ANDROID_HOME/emulator/emulator -avd Pixel_5_API_27 -port ${{ env.EMULATOR_PORT }} -no-window -no-boot-anim -no-audio -no-snapshot -gpu swiftshader_indirect -accel on -camera-back none &
+          if ! timeout 420 $ANDROID_HOME/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'; then
+            echo "::error::Android emulator did not reach sys.boot_completed within 420s; failing fast instead of hanging until the 6h limit"
+            $ANDROID_HOME/platform-tools/adb devices
+            exit 1
+          fi
           $ANDROID_HOME/platform-tools/adb devices
           # disable spell checker
           $ANDROID_HOME/platform-tools/adb shell settings put secure spell_checker_enabled 0
@@ -141,15 +127,6 @@ jobs:
           $ANDROID_HOME/platform-tools/adb shell settings put global window_animation_scale 0.0
           $ANDROID_HOME/platform-tools/adb shell settings put global transition_animation_scale 0.0
           $ANDROID_HOME/platform-tools/adb shell settings put global animator_duration_scale 0.0
-          # Save snapshot for future runs
-          $ANDROID_HOME/platform-tools/adb emu avd snapshot save default_boot
-
-      - name: Start Android emulator (from snapshot - cached)
-        if: ${{ matrix.platform == 'android' && steps.android-cache.outputs.cache-hit == 'true' }}
-        run: |
-          $ANDROID_HOME/emulator/emulator -avd Pixel_5_API_27 -port ${{ env.EMULATOR_PORT }} -no-boot-anim -no-audio -snapshot default_boot -gpu host -accel on &
-          $ANDROID_HOME/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'
-          $ANDROID_HOME/platform-tools/adb devices
 
       - name: Create MegaApp ${{ env.MEGA_APP_NAME }} and run build on NodeJS ${{ matrix.node-version }}
         run: npm run setup:${{matrix.framework}}:${{matrix.build-tool}} -- --name ${{ env.MEGA_APP_NAME }} --platform ${{matrix.platform}} --tag ${{inputs.dist-tag}} --framework-version ${{matrix.framework-version.value}} --build-tool-version ${{matrix.build-tool-version}}
--- a/.github/workflows/reusable-build-system-test-react-native.yml
+++ b/.github/workflows/reusable-build-system-test-react-native.yml
@@ -10,16 +10,7 @@ on:
 
 jobs:
   build:
-    # Android builds run on ubuntu-latest so the emulator can use hardware (KVM)
-    # acceleration. The previous macos-15-intel runner regularly wedged/segfaulted
-    # the emulator boot under `-gpu host` (see #7031).
-    #
-    # NOTE: this repo restricts GitHub Actions to an allow-list (GitHub-authored,
-    # aws-actions/*, aws-amplify/* and changesets/action), so the emulator is
-    # launched with the Android SDK CLI directly rather than a third-party action.
-    runs-on: ubuntu-latest
-    # Bound the whole job so a wedged emulator boot can't run until GitHub's 6h hard limit.
-    timeout-minutes: 45
+    runs-on: macos-15-intel
     strategy:
       fail-fast: false
       matrix:
@@ -56,12 +47,49 @@ jobs:
         env:
           SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
 
+      - name: Restore CocoaPods cache
+        if: ${{ matrix.platform == 'ios' }}
+        id: restore-cocoapods-cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ./examples/react-native/ios/Pods
+          key: ${{ runner.os }}-cocoapods
+          restore-keys: pods-${{ hashFiles('examples/react-native/ios/Podfile.lock') }}
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
+
+      - name: Restore node_modules cache
+        if: ${{ matrix.platform == 'ios' }}
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        id: restore-cache
+        with:
+          path: |
+            ./node_modules
+            **/node_modules
+          key: ${{ runner.os }}-node24-nodemodules
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 3
+
       - name: Install Java 17
         if: ${{ matrix.platform == 'android' }}
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           distribution: 'corretto' # Amazon Corretto Build of OpenJDK
           java-version: '17'
+
+      - name: Cache Android SDK and emulator
+        if: ${{ matrix.platform == 'android' }}
+        id: android-cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: |
+            ~/Library/Android/sdk/system-images
+            ~/Library/Android/sdk/build-tools
+            ~/Library/Android/sdk/platform-tools
+            ~/.android/avd
+          key: ${{ runner.os }}-android-sdk-27-x86_64-snapshot-v1
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 5
 
       - name: Cache Gradle dependencies
         if: ${{ matrix.platform == 'android' }}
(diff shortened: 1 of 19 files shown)


packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -125,82 +197,224 @@ const download = async (
...
+      }
+      const { value, done } = await reader.read();
+      if (done) break;
+      transferred += value.length;
+      streamController.enqueue(value);
+
+      if (size > 0 && isFunction(options?.onProgress)) {
+        options!.onProgress(
+          data,
+          getProgress({ totalBytes: size, transferredBytes: transferred }),
```
osama-rizk, line 284: **Progress now reflects fetch bytes only, not zip finalization.**

The previous handler fed `onProgress` from two sources: the fetch read loop (`readBody`) *and* zip.js's own `onprogress` callback during `ZipWriter.add`. This new loop drives progress solely from bytes read off `response.body` (`transferred / size`), and the zip.js `onprogress` hook is gone — so progress reports download-side only.

With `level: 0` (store, no compression) the write phase is near-instant, so the practical drift is small and this is likely a deliberate simplification. But it's coupled to the removal of the `'FINISHING'`/"Zipping" status: together they mean the UI now jumps `PENDING → COMPLETE` with no finalization phase. Could you add a one-line note to the changeset that per-file progress is now download-only by design? That keeps it from reading as an accidental regression to future readers.

bobbor: Good catch — this is deliberate. With `level: 0` (store, no compression) the zip finalization phase is near-instant, so driving progress purely off the S3 fetch read loop keeps the reporting simple without meaningful drift. I've added a note to the changeset (commit 4455d41a4) documenting that per-file progress is download-only by design and that there's intentionally no separate "zipping"/finishing phase, so it won't read as an accidental regression to future readers.

packages/react-storage/src/components/StorageBrowser/service-worker/download-sw.ts
```diff
@@ -0,0 +1,79 @@
...
+});
+
+// Intercept fetch requests matching the download URL pattern
+self.addEventListener('fetch', (event) => {
+  const url = new URL(event.request.url);
+  if (!url.pathname.startsWith('/amplify-storage-download/')) return;
+
+  const downloadId = url.pathname.split('/amplify-storage-download/')[1];
+  const stream = pendingStreams.get(downloadId);
+
```
soberm, line 82: [blocker] `url.pathname` preserves percent-encoding. A folder name containing a space (common in S3 prefixes, e.g. `vacation photos`) produces `downloadId = 'vacation photos-1234567890.zip'` which is stored as the `pendingStreams` key. When the `<a>` href is set to `/amplify-storage-download/vacation photos-1234567890.zip`, the browser percent-encodes the space before dispatching the request. The SW splits the encoded pathname and calls `pendingStreams.get('vacation%20photos-1234567890.zip')`, which returns `undefined`. `respondWith` is never called, the browser gets a 404, and the download silently fails for any folder name with a space or other URL-unsafe character. All tests use plain-ASCII IDs so this is not caught.

Fix: decode the pathname segment before the lookup, and drop the now-redundant `decodeURIComponent` on the filename since it is already decoded:

```ts
const rawId = url.pathname.split('/amplify-storage-download/')[1];
const downloadId = decodeURIComponent(rawId);
const stream = pendingStreams.get(downloadId);
if (!stream) return;
pendingStreams.delete(downloadId);
const filename = downloadId.split('/').pop()!;
```

Also add a test case in the `getFolderName logic` describe block for a key like `my photos/vacation 2026/beach.jpg` to prevent regressions.

packages/react-storage/src/components/StorageBrowser/service-worker/download-sw.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (!stream) return;
+
+  pendingStreams.delete(downloadId);
+
+  const filename = decodeURIComponent(downloadId.split('/').pop()!);
+  event.respondWith(
+    new Response(stream, {
+      headers: {
+        'Content-Disposition': `attachment; filename="${filename}"`,
+        'Content-Type': 'application/octet-stream',
```
soberm, line 98: [minor] `Content-Disposition: attachment; filename="${filename}"` does not escape `"` or `\` inside the quoted string. S3 key components can legally contain `"`. RFC 5987 extended notation handles arbitrary UTF-8 without escaping and is broadly supported:

```ts
`attachment; filename*=UTF-8''${encodeURIComponent(filename)}`
```

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -1,116 +1,188 @@
...
+
+      // Cancel detection works via addPromise.catch: when the user dismisses the
+      // download dialog, the SW response stream closes → TransformStream writable
+      // errors → zipWriter.add() rejects → cancelled flag is set.
+
+      const { port1, port2 } = new MessageChannel();
+      port1.onmessage = () => {
+        const a = document.createElement('a');
+        a.href = `/amplify-storage-download/${state.downloadId}`;
+        a.download = `${state.folder}.zip`;
```
soberm, line 173: [blocker] This is the origin of the URL-encoding mismatch caught in `download-sw.ts`. `state.downloadId` (e.g. `vacation 2026-1234567890.zip`) is placed raw into the href string. The DOM normalises it to percent-encoded form when the navigation fires, so the SW receives `vacation%202026-1234567890.zip`. The `pendingStreams` map was keyed with the unencoded string, so the lookup returns `undefined` and the download silently fails.

The fix belongs in the SW fetch handler (decode the pathname segment before calling `pendingStreams.get`), but this line is the root cause worth noting during review.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -1,116 +1,188 @@
...
+      reg.active.postMessage(
+        { downloadId: state.downloadId, stream: state.zipReadable },
+        [state.zipReadable as unknown as Transferable, port2]
+      );
+      state.zipReadable = null;
+
+      // Keepalive pings run for the entire batch duration to prevent Firefox's
+      // 30s SW idle timeout from terminating the worker mid-stream.
+      state.keepaliveInterval = setInterval(() => {
+        reg.active?.postMessage({ type: 'keepalive' });
```
soberm, line 195: [major] Keepalive interval leak when `cancelBatch` fires before `state.swReady` resolves.

Sequence: (1) `initServiceWorkerStream` kicks off an async `getRegistration` call; (2) the user cancels immediately; (3) `cancelBatch` runs, `state.keepaliveInterval` is still `null`, so `clearInterval(null)` is a no-op and `reset()` removes the entry from `batchMap`; (4) later the `.then()` callback fires, sees an active registration, and sets `state.keepaliveInterval = setInterval(...)`. At that point no code path holds a reference to this state via `batchMap`, so `clearInterval` is never called. The interval pings the SW every 10 s indefinitely, preventing it from going idle.

Fix: add a cancelled guard at the top of the `.then()` callback:

```ts
.then((reg) => {
  if (state.cancelled) return;
  if (!reg?.active) { ... }
  // rest unchanged
})
```

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -125,82 +197,224 @@ const download = async (
...
+  // Distinguish stream-closure errors (cancel) from other zip.js errors (real failures).
+  addPromise.catch((error: unknown) => {
+    const err = error instanceof Error ? error : new Error(String(error));
+    // zip.js throws AbortError or errors the writable when the output stream
+    // is terminated. These are caused by the user dismissing the download dialog
+    // (SW path) or by our explicit abort(). Any other error is a genuine failure.
+    if (
+      err.name === 'AbortError' ||
+      err.message.includes('abort') ||
+      err.message.includes('close') ||
```
soberm: [nit] `err.message.includes('abort')` and `err.message.includes('close')` are substring matches on zip.js internal error messages that are not part of a public API contract and may silently break across library updates. The `state.batchAbort.signal.aborted` check already covers the explicit-cancel case. Consider relying solely on `signal.aborted || state.cancelled` as the cancellation signal and treating all other `addPromise` rejections as genuine failures.

packages/react-storage/src/components/StorageBrowser/service-worker/useServiceWorkerRegistration.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const SW_URL = '/amplify-storage-download/download-sw.js';
+
+export function useServiceWorkerRegistration(): void {
+  useEffect(() => {
+    if ('serviceWorker' in navigator) {
+      navigator.serviceWorker
+        .register(SW_URL, { scope: SW_DOWNLOAD_SCOPE })
+        .catch(() => {
+          // Registration failure is non-critical; blob fallback handles downloads
```
soberm: [minor] Silent catch hides registration failures entirely. A developer who skips the `copy-serviceworker` step (the most common setup error) gets no feedback; the component falls back to blob downloads with no indication that the SW path is broken.

```ts
.catch((err) => {
  if (process.env.NODE_ENV !== 'production') {
    console.warn('[StorageBrowser] Service worker registration failed:', err);
  }
});
```

packages/react-storage/bin/copy-serviceworker.js
```diff
@@ -0,0 +1,36 @@
...
+    `Error: Service worker source not found at ${SW_SOURCE}\n` +
+      'Make sure the package is properly installed.'
+  );
+  process.exit(1);
+}
+
+const resolvedTarget = path.resolve(targetDir, 'amplify-storage-download');
+
+if (!fs.existsSync(resolvedTarget)) {
+  fs.mkdirSync(resolvedTarget, { recursive: true });
```
soberm: [nit] `fs.existsSync(resolvedTarget)` before `mkdirSync(..., { recursive: true })` is redundant: `mkdirSync` with `recursive: true` already silently succeeds when the directory exists and does not throw. The extra check also introduces a TOCTOU window. Call `fs.mkdirSync(resolvedTarget, { recursive: true })` unconditionally.

.github/workflows/reusable-build-system-test-react-native.yml
```diff
@@ -10,16 +10,7 @@ on:
...
-    # the emulator boot under `-gpu host` (see #7031).
-    #
-    # NOTE: this repo restricts GitHub Actions to an allow-list (GitHub-authored,
-    # aws-actions/*, aws-amplify/* and changesets/action), so the emulator is
-    # launched with the Android SDK CLI directly rather than a third-party action.
-    runs-on: ubuntu-latest
-    # Bound the whole job so a wedged emulator boot can't run until GitHub's 6h hard limit.
-    timeout-minutes: 45
+    runs-on: macos-15-intel
     strategy:
```
soberm, line 23: [minor] Two related concerns:

1. The now-deleted comment documented that `macos-15-intel` with `-gpu host` "regularly wedged/segfaulted the emulator boot (see #7031)" and was the reason for switching to `ubuntu-latest`. This PR returns to `macos-15-intel` with `-gpu host` in both the cold-boot and cached-snapshot steps. If the root cause was the macOS-Intel + Metal GPU combination, this may reintroduce the same flakiness.

2. The job-level `timeout-minutes: 45` was removed at the same time. On a cache miss with a hanging emulator the job now runs until GitHub's 6-hour hard limit. Worth restoring a ceiling (e.g. 60 min for macOS) regardless of whether iOS testing is added.

.github/workflows/reusable-build-system-test-react-native.yml
```diff
@@ -56,13 +47,50 @@ jobs:
...
+      - name: Restore CocoaPods cache
+        if: ${{ matrix.platform == 'ios' }}
+        id: restore-cocoapods-cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ./examples/react-native/ios/Pods
+          key: ${{ runner.os }}-cocoapods
+          restore-keys: pods-${{ hashFiles('examples/react-native/ios/Podfile.lock') }}
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
```
soberm: [minor] The `restore-keys` prefix (`pods-...`) does not match the primary key prefix (`${{ runner.os }}-cocoapods`). Partial-hit matching is prefix-based, so this `restore-keys` entry can never produce a hit against an existing cache entry. Either align the prefix with the primary key (e.g. `${{ runner.os }}-cocoapods`) or drop `restore-keys` entirely.

.github/workflows/reusable-build-system-test-react-native.yml
```diff
@@ -76,57 +104,52 @@ jobs:
...
-          # Fail fast (instead of waiting 420s for a boot that can't happen) if the AVD was not created.
-          echo "Available AVDs:"
-          $ANDROID_HOME/emulator/emulator -list-avds
-          if ! $ANDROID_HOME/emulator/emulator -list-avds | grep -qx "Pixel_5_API_27"; then
-            echo "::error::AVD Pixel_5_API_27 was not created by avdmanager"
-            $ANDROID_HOME/cmdline-tools/latest/bin/avdmanager list avd || true
-            exit 1
-          fi
+          echo -e "echo \$ANDROID_HOME"
+          echo $ANDROID_HOME
```
soberm: [minor] `echo -e "echo \$ANDROID_HOME"` prints the literal string `echo $ANDROID_HOME` to stdout. This looks like a debugging leftover; the next line already logs the actual value via `echo $ANDROID_HOME`.

.github/workflows/reusable-build-system-test-react-native.yml
```diff
@@ -76,57 +104,52 @@ jobs:
...
+          $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --install "build-tools;33.0.2" "platform-tools" "system-images;android-27;default;x86_64"
+          $ANDROID_HOME/cmdline-tools/latest/bin/avdmanager create avd --name Pixel_5_API_27 --force --device "pixel_5" --abi x86_64 --package "system-images;android-27;default;x86_64"
```
soberm: [major] The original `sdkmanager --install` command included `"emulator"` and `"platforms;android-27"` which are both absent here:

Before: `"platform-tools" "emulator" "build-tools;33.0.2" "platforms;android-27" "system-images;android-27;default;x86_64"`

After: `"build-tools;33.0.2" "platform-tools" "system-images;android-27;default;x86_64"`

On a fresh `macos-15-intel` runner without a cache hit the `emulator` binary may not exist and `avdmanager create avd` may fail without the API-27 platform headers. Please confirm both are pre-installed on the macOS runner image, or add them back.

.github/workflows/reusable-build-system-test-react-native.yml
```diff
@@ -76,57 +104,52 @@ jobs:
...
-          # reliable CI combination on Linux runners.
-          $ANDROID_HOME/emulator/emulator -avd Pixel_5_API_27 -port ${{ env.EMULATOR_PORT }} -no-window -no-boot-anim -no-audio -no-snapshot -gpu swiftshader_indirect -accel on -camera-back none &
-          if ! timeout 420 $ANDROID_HOME/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'; then
-            echo "::error::Android emulator did not reach sys.boot_completed within 420s; failing fast instead of hanging until the 6h limit"
-            $ANDROID_HOME/platform-tools/adb devices
-            exit 1
-          fi
+          $ANDROID_HOME/emulator/emulator -avd Pixel_5_API_27 -port ${{ env.EMULATOR_PORT }} -no-boot-anim -no-audio -no-snapshot-load -gpu host -accel on &
+          $ANDROID_HOME/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'
           $ANDROID_HOME/platform-tools/adb devices
```
soberm, line 123: [major] `adb wait-for-device shell '...'` has no timeout. The original command wrapped this with `timeout 420 ...` and emitted `::error::` before exiting, preventing the job from hanging indefinitely. Without it a wedged emulator holds the job until GitHub's 6-hour hard limit. The same issue applies to the identical `adb wait-for-device` call in the cached-snapshot step below. Consider restoring `timeout 420` on both.

packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/zipdownload.spec.ts
```diff
@@ -46,93 +119,232 @@ const baseInput: DownloadHandlerInput = {
...
+    const input = createBaseInput();
+    const { result } = zipDownloadHandler(input);
+    // result now awaits cleanup completion before resolving
+    await result;
+
+    expect(mockRevokeObjectURL).toHaveBeenCalledWith('blob:revoke-test');
+  });
+
+  it('returns cancel function in early-exit cancelled path', () => {
+    // Simulate a cancelled batch by first starting a download, then cancelling,
```
soberm: [nit] The test name promises to exercise the `existingBatch?.cancelled` early-exit branch, but the assertions only verify that `first.cancel` is defined and is a function. `first.cancel!()` calls `cancelBatch` which calls `reset()`, removing the batch from `batchMap`; a subsequent call with the same `all` starts a fresh batch rather than hitting the cancelled path. The test body even documents this ("Since batch is reset on cancel, calling with same `all` starts fresh") but the assertions do not validate the branch. Either rename the test to match what it actually checks, or add assertions that genuinely reach the `existingBatch.cancelled` branch.

packages/react-storage/src/components/StorageBrowser/service-worker/download-sw.ts
bobbor: Fixed in 9f488b1d2. The fetch handler now decodes the pathname segment before the `pendingStreams` lookup (`const downloadId = decodeURIComponent(rawId)`), and the redundant `decodeURIComponent` on the filename is gone. Added a SW test that stores the stream under the unencoded id and requests the percent-encoded URL (round-trip), plus a `my photos/vacation 2026/beach.jpg` case in the `getFolderName logic` block.

packages/react-storage/src/components/StorageBrowser/service-worker/download-sw.ts
bobbor: Fixed in 9f488b1d2 — switched to RFC 5987 extended notation: `attachment; filename*=UTF-8''${encodeURIComponent(filename)}`. SW tests updated to assert the new form.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
bobbor: Acknowledged — the fix lives in the SW fetch handler (decode before lookup) as you suggested, in 9f488b1d2. Thanks for pinpointing the root cause.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
bobbor: Fixed in 9f488b1d2 — added `if (state.cancelled) return;` at the top of the `getRegistration().then()` callback, so no keepalive interval is created once the batch has been cancelled/reset. Prevents the orphaned-interval leak you described.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
bobbor: Addressed in 9f488b1d2, with one small deviation worth flagging. I dropped the fragile `.includes('abort')`/`.includes('close')` substring matches, but kept the standard `err.name === 'AbortError'` check alongside `signal.aborted`. Rationale: only the UI cancel button aborts `batchAbort`; the browser-dialog-cancel path terminates the SW response stream and surfaces as an `AbortError` from zip.js without aborting the signal. Relying solely on `signal.aborted` would stop flagging that case early, so the read loop would keep pulling bytes from S3 until the fetch completes. `AbortError` is a DOM-standard error name (not a zip.js-internal message string), so it isn't subject to the library-version fragility you called out. Happy to switch to strictly `signal.aborted` if you'd rather accept the late-stop tradeoff.

packages/react-storage/src/components/StorageBrowser/service-worker/useServiceWorkerRegistration.ts
bobbor: Fixed in 9f488b1d2 — the `.catch` now logs a `console.warn` guarded by `process.env.NODE_ENV !== 'production'`, pointing at the copy-serviceworker setup step so a skipped setup surfaces in dev.

packages/react-storage/bin/copy-serviceworker.js
bobbor: Fixed in 9f488b1d2 — now calls `fs.mkdirSync(resolvedTarget, { recursive: true })` unconditionally; the `existsSync` check (and its TOCTOU window) is gone.

packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/zipdownload.spec.ts
bobbor: Fixed in 9f488b1d2 — the test now genuinely reaches the `existingBatch.cancelled` early-exit branch: a 2-file batch where file 1's fetch rejects with an `AbortError` leaves the batch flagged-but-not-reset (batchDone 1 < 2), then the file-2 call hits the early-exit branch. It asserts the CANCELED result, that a noop cancel fn is returned and doesn't throw, and that file 2 was never fetched.

.github/workflows/reusable-build-system-test-react-native.yml
bobbor: Good catch — this workflow change was unintentional. It was pulled in by a bad conflict resolution during a rebase; this file has nothing to do with the SW streaming feature. I've reverted `reusable-build-system-test-react-native.yml` to match `main` in 9f488b1d2, so this diff (and the macos-15-intel/timeout/sdkmanager/restore-keys/echo concerns across this file) is gone from the PR entirely.

.github/workflows/reusable-build-system-test-react-native.yml
bobbor: Good catch — this workflow change was unintentional. It was pulled in by a bad conflict resolution during a rebase; this file has nothing to do with the SW streaming feature. I've reverted `reusable-build-system-test-react-native.yml` to match `main` in 9f488b1d2, so this diff (and the macos-15-intel/timeout/sdkmanager/restore-keys/echo concerns across this file) is gone from the PR entirely.

.github/workflows/reusable-build-system-test-react-native.yml
bobbor: Good catch — this workflow change was unintentional. It was pulled in by a bad conflict resolution during a rebase; this file has nothing to do with the SW streaming feature. I've reverted `reusable-build-system-test-react-native.yml` to match `main` in 9f488b1d2, so this diff (and the macos-15-intel/timeout/sdkmanager/restore-keys/echo concerns across this file) is gone from the PR entirely.

.github/workflows/reusable-build-system-test-react-native.yml
bobbor: Good catch — this workflow change was unintentional. It was pulled in by a bad conflict resolution during a rebase; this file has nothing to do with the SW streaming feature. I've reverted `reusable-build-system-test-react-native.yml` to match `main` in 9f488b1d2, so this diff (and the macos-15-intel/timeout/sdkmanager/restore-keys/echo concerns across this file) is gone from the PR entirely.

.github/workflows/reusable-build-system-test-react-native.yml
bobbor: Good catch — this workflow change was unintentional. It was pulled in by a bad conflict resolution during a rebase; this file has nothing to do with the SW streaming feature. I've reverted `reusable-build-system-test-react-native.yml` to match `main` in 9f488b1d2, so this diff (and the macos-15-intel/timeout/sdkmanager/restore-keys/echo concerns across this file) is gone from the PR entirely.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -125,82 +197,224 @@ const download = async (
...
+  const folder = getFolderName(firstKey);
+  const batchKey = getBatchKey(all);
+
+  const existingBatch = batchMap.get(batchKey);
+
+  // ─── STEP 1: Handle cancelled state ───
+  if (existingBatch?.cancelled) {
+    existingBatch.batchDone++;
+    const result = Promise.resolve<DownloadTaskResult>({
+      status: 'CANCELED',
```
osama-rizk, line 387: **UI cancel resurrects the download instead of stopping it.**

The cancelled-drain path guarded here is dead code for the UI-cancel case, so cancelling actually delivers a zip. The chain:

1. `cancel()` (line ~405) calls `cancelBatch(batchKey)`, which sets `state.cancelled = true` **and synchronously** calls `reset()` → `batchMap.delete(batchKey)` (line 64) in the same tick.
2. This view runs the batch with `concurrency: 1` (`useDownloadView`), so `useProcessTasks` processes files one at a time and re-dispatches the next `QUEUED` file to `zipDownloadHandler` on settle — with no batch-level cancelled guard.
3. The next file calls `batchMap.get(batchKey)` → `undefined` (entry was deleted) → `existingBatch?.cancelled` here is falsy → this early-exit is skipped → STEP 2 builds a **brand-new** `TransformStream`/`ZipWriter`, re-inits the SW, and downloads the remaining files into a fresh zip that gets delivered.

**Repro:** select 5 files, click download, cancel after file 1 → files 2–5 still download as a zip.

Root cause is that the `cancelled` sentinel and the map entry's lifetime share one key, so tearing down the entry erases the very flag the drain path relies on. Suggest decoupling them: keep the batch entry alive (mark it tombstoned) until all `batchTotal` files have drained, and only `reset()`/`batchMap.delete` once `batchDone === batchTotal` — for both the completion and the cancel paths.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -125,82 +197,224 @@ const download = async (
...
+      Uint8Array
+    >();
+    currentBatch = {
+      zipWriter: new ZipWriter(writable),
+      zipWritable: writable,
+      zipReadable: readable,
+      downloadId: `${folder}-${Date.now()}.zip`,
+      blobPromise: null,
+      swReady: Promise.resolve(),
+      cancelled: false,
```
osama-rizk, line 415: **SW-path filename leaks a timestamp and disagrees with the blob fallback.**

`downloadId` is used for two different purposes: (a) the internal key that uniquely identifies the stream in the SW's `pendingStreams` map, and (b) — transitively — the user-facing filename. In `download-sw.ts` the response is built as `Content-Disposition: attachment; filename*=UTF-8''${encodeURIComponent(filename)}` where `filename = downloadId.split('/').pop()` = `${folder}-${Date.now()}.zip`.

For a **same-origin** response (the SW serves a relative `/amplify-storage-download/...` URL), the WHATWG 'suggested filename' algorithm gives `Content-Disposition` precedence over the anchor's `download` attribute. So the SW path ignores `a.download = `${state.folder}.zip`` and saves e.g. `photos-1720080000000.zip` — a leaked epoch timestamp — while the blob-fallback path (which has no server header) honors `a.download` and saves `photos.zip`.

Result: the same feature produces two different filenames depending on SW availability, and the SW one exposes an internal timestamp. The `Date.now()` is only needed to keep the map key unique across concurrent batches — it shouldn't reach the filename. Suggest sourcing the Content-Disposition filename from `state.folder` (pass `folder` alongside `downloadId` in the postMessage, or send the desired filename explicitly), keeping the timestamped id purely as the internal stream key.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
bobbor: Great catch, and thank you for the precise repro — fixed in 896a5af88. Root cause was exactly as you described: `cancelBatch` called `reset()` synchronously, deleting the map entry in the same tick, so the next queued file (concurrency: 1) missed the `existingBatch.cancelled` early-exit and built a brand-new batch. I decoupled the sentinel from the entry lifetime: `cancelBatch` now sets `cancelled`, aborts the fetch + writable, and stops the keepalive — but keeps the (tombstoned) entry. `reset()`/`batchMap.delete` runs only once `batchDone === batchTotal`, in the STEP 1 early-exit or `onFileSettled`, for both the cancel and completion paths. Added a regression test that cancels a 2-file batch via the returned `cancel()` and asserts exactly one `ZipWriter` is constructed across both files (i.e. file 2 takes the early-exit and does not resurrect the download).

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
bobbor: Fixed in 896a5af88. You're right that the timestamped `downloadId` was doubling as the user-facing filename, and that same-origin Content-Disposition wins over the anchor's `download` attribute — so the SW path saved `photos-<epoch>.zip` while the blob fallback saved `photos.zip`. The page now sends the filename explicitly (`filename: `${state.folder}.zip``) in the postMessage; the SW stores `{ stream, filename }` and builds Content-Disposition from that filename, keeping the timestamped id purely as the internal `pendingStreams` key. Both paths now save `${folder}.zip`. Added a SW test asserting the explicit filename is used and the timestamp never reaches Content-Disposition.

osama-rizk (approved):

osama-rizk:
**Testing: is an e2e test feasible for this?**

The added coverage here is solid at the unit level, but by construction it mocks out every boundary the feature actually depends on — `ReadableStream`/`TransformStream` are polyfilled from `node:stream/web`, `MessageChannel` and `navigator.serviceWorker` are hand-stubbed, and `zip.js` is faked. So the specs verify the handler's bookkeeping (postMessage called, filename chosen, blob URL revoked) but can't exercise the parts that make this change risky in a real browser:

- a real SW intercepting the `/amplify-storage-download/` fetch and streaming to disk
- the MessageChannel handshake + transferred `ReadableStream` (`Transferable`)
- the Firefox 30s idle-timeout keepalive actually keeping the worker alive mid-stream
- cancel via the browser's own download-dialog dismissal
- SW registration timing and the blob fallback across Chrome/Firefox/Safari

**Question for the author:** is an e2e test feasible for this feature given the current harness? A service worker needs a secure context (https/localhost) and a real registration, plus a way to assert a file landed on disk — so I'm not sure how well it fits the existing e2e setup, and I don't want to ask for something the harness can't reasonably support.

If it *is* feasible, I don't think it needs to block this PR — it could be reasonable **follow-up work** (a tracking issue is fine), given the streaming path is the whole value proposition and the mocks give limited confidence there. If it's *not* feasible, could you fill in the PR's "Description of how you validated changes" with the manual cross-browser validation you ran (Chrome/Firefox/Safari, large batch, cancel path)? That gives reviewers the confidence the unit tests can't.

soberm (approved):

PR #6834, opened 2026-02-03
feat(storage-browser): Folder deletion.

osama-rizk:
#### Description of changes

This PR implements folder deletion functionality for the StorageBrowser component. The changes include:

- **Enhanced delete handler**: Updated the delete action handler to support both file and folder deletion with recursive folder deletion capabilities
- **New ActionConfirmationModal component**: Added a reusable confirmation modal for destructive actions like deletion
- **Folder count tracking**: Implemented functionality to track and display the number of items within folders before deletion
- **Updated delete view**: Enhanced the DeleteView to show folder contents and confirmation dialogs
- **Display text updates**: Added comprehensive text strings for folder deletion scenarios including confirmation messages and progress indicators
- **Table resolver improvements**: Updated delete resolvers to handle folder-specific actions and display appropriate UI elements
- **Test coverage**: Added comprehensive test coverage for new folder deletion functionality

The implementation ensures users can safely delete folders with proper confirmation dialogs showing the contents that will be deleted.

#### Issue #, if available

- #6830 
- #6615 

#### Description of how you validated changes

- **Unit Tests**: All existing tests pass, and new comprehensive test coverage has been added for folder deletion functionality
- **Manual Testing**: Tested folder deletion scenarios including:
  - Empty folder deletion
  - Folder deletion with nested files and subfolders
  - Confirmation modal behavior and user interactions
  - Error handling for failed deletion operations

#### Notes 
A PR will follow to add 
 - docs https://github.com/aws-amplify/amplify-ui/pull/6837

#### Checklist

- [x] Have read the [Pull Request Guidelines](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md)
- [x] PR description included
- [x] `yarn test` passes and tests are updated/added
- [x] PR title and commit messages follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax
- [x] _If this change should result in a version bump_, [changeset added](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) (This can be done after creating the PR.) This does not apply to changes made to `docs`, `e2e`, `examples`, or other private packages.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7046, opened 2026-07-10
feat(storage): folder download support in StorageBrowser

bobbor:
## Description

Phase 2 of the Storage Browser download improvements: **folder download support**. Previously the download button was disabled whenever the selection contained a folder (see Transfer Family webapp feedback). This enables downloading folders and mixed file/folder selections, building on the Phase 1 service-worker streaming download (already merged).

### What changed

- **Folder expansion at the view level.** When the `DownloadView` opens, any selected folders are recursively enumerated (`expandFolderToFiles`, paginated S3 `list`, unbounded, abortable) so their files render as rows. Enumeration runs on mount/selection-change — the download is dispatched only when the user clicks Start (`onActionStart`) and only once **every** selected folder is fully enumerated (`allFoldersReady` gate), so a partial/incomplete archive can never be dispatched.
- **Relative paths preserved in the zip.** Each expanded file carries a folder-relative `relativePath` (`key.slice(locationPrefix.length)`) used as its zip entry name, so nested directory structure is preserved and cross-folder collisions are avoided.
- **Archive naming.** The zip is named after the **longest common ancestor directory** of all downloaded files (`getArchiveName`), computed in the view and passed to the handler — e.g. `public/nested/one/pic.jpg` + `public/nested/two/pic.jpg` → `nested.zip`; adding files under `public/` → `public.zip`; falls back to `download`.
- **Disable gate removed.** `hasSelectedFolders` no longer disables the download action.
- **UX states.** "Listing folder contents", enumeration error (with retry via Start), and "no files" messages surfaced via display text. Enumeration is cancellable and retryable; individual expanded files can be removed from the pending download.

### Testing

- Unit/integration: **239 tests pass** (folder expansion, relative-path rule, multi-folder collisions, archive-name common-ancestor cases incl. edge inputs, mount auto-enumeration, no-auto-start, cancel/error → no partial dispatch + retry, per-row removal, message precedence).
- `yarn typecheck` clean, `eslint` clean, `yarn build` (rollup) clean (`download-sw.js` emitted).
- Manually validated: simple, nested, and deeply-nested folder downloads; zip naming; empty-view and empty-list handling.

### Notes

- No new dependencies; reuses existing `list`/`getUrl` and `@zip.js/zip.js`.
- Downloads remain sequential (`concurrency: 1`); no-partial-zip policy preserved (any file failure discards the archive).


packages/react-storage/src/components/StorageBrowser/actions/configs/__tests__/defaults.spec.ts
```diff
@@ -111,9 +111,9 @@ describe('defaultActionConfigs', () => {
       expect(disable?.([file])).toBe(false);
     });
 
-    it('is disabled when folders are selected', () => {
-      expect(disable?.([folder])).toBe(true);
-      expect(disable?.([file, folder])).toBe(true);
+    it('allows folder download (Phase 2)', () => {
```
soberm: [nit] The test name bakes a phased-delivery label into a permanent test. Once this merges, "Phase 2" is just the current behavior. Rename to something stable like `'is not disabled when folders are selected'`.

packages/react-storage/src/components/StorageBrowser/actions/configs/defaults.tsx
```diff
@@ -60,9 +60,11 @@ export const uploadActionConfig: UploadActionConfig = {
 export const downloadActionConfig: DownloadActionConfig = {
   viewName: 'DownloadView',
   actionListItem: {
+    // Folder selections are now supported (Phase 2 — folder download expands
```
soberm: [nit] Changelog annotation baked into source. Once merged, this comment should just explain the current invariant without referencing a delivery phase. Something like: "Only an empty selection disables Download; folder selections are expanded into files at the view level."

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
```diff
@@ -0,0 +1,153 @@
+import type {
+  ActionInputConfig,
+  DownloadHandlerData,
+  FileData,
+} from '../../../actions';
+import {
+  constructBucket,
```
soberm: [nit] `constructBucket` is imported directly from the internal handler module (`actions/handlers/utils`) rather than a public surface. `createDownloadItem` was just promoted to the `actions` index, so that import is fine, but `constructBucket` isn't. If view-level code legitimately needs it, promote it to the public `actions` export too, or get the resolved bucket from `getConfig()`. This creates a view-into-handler-internals dependency.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+const LIST_PAGE_SIZE = 1000;
+
+/**
+ * Soft threshold (file count) above which the UI should confirm a "large
+ * download" with the user before dispatching. Unlike DeleteView's hard 5000
+ * cap (which only affects a display *count*), download must fetch every file —
+ * a silently truncated zip would be data loss — so expansion itself is
+ * UNBOUNDED and this value is used purely as a confirmation guard.
+ */
+export const LARGE_DOWNLOAD_FILE_COUNT = 5000;
```
soberm, line 24: [minor] `LARGE_DOWNLOAD_FILE_COUNT` is exported but never enforced anywhere. The live code does `void LARGE_DOWNLOAD_FILE_COUNT` to satisfy the lint rule, which is a smell. Shipping an exported constant as a pure TODO placeholder adds noise to the public API surface and creates false confidence (the test that asserts it equals `5000` doesn't test any behavior). Remove it until the confirmation gate is actually wired up.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+ * - is cancellable via `signal`, checked between `list()` pages.
+ *
+ * @param folderKey - S3 key of the folder to expand (ends in `'/'`).
+ * @param config - Storage action config.
+ * @param locationPrefix - Current browse-location prefix `P`; each file's
+ *   `relativePath` is computed as `key.slice(P.length)`.
+ * @param signal - Optional abort signal; when aborted between pages the loop
+ *   throws an `AbortError` and no items are returned.
+ */
+/* eslint-disable max-params -- positional (folderKey, config, prefix, signal) signature is intentional and mirrors the abortable list-pagination pattern */
```
soberm: [nit] The `eslint-disable max-params` block comment is placed before the function signature and then re-enabled on the first line of the body. The conventional pattern is to pair both outside the function, or use `// eslint-disable-next-line max-params` directly above the `export const` line. The current placement works but is a bit unexpected.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+      // Skip directory markers (zero-byte keys ending in '/') — only real files
+      // become zip entries.
+      if (item.path.endsWith('/')) {
+        continue;
+      }
+
+      const fileData: FileData = {
+        key: item.path,
+        id: crypto.randomUUID(),
+        size: item.size!,
```
soberm: [minor] `item.size!` and `item.lastModified!` use non-null assertions. S3's `list()` can omit these fields for certain object types (e.g., incomplete multipart uploads). The `!` suppresses the type error but still passes `undefined` at runtime, silently corrupting the `FileData` object. Consider a defensive fallback (`item.size ?? 0`, `item.lastModified ?? new Date(0)`) or skip the item with a warning if these fields are absent.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -20,10 +73,245 @@ export const useDownloadView = (
...
+  const getConfig = useGetActionInput();
+
+  // Cache of folder id -> expanded download items, mirroring DeleteView's
+  // `folderCountsRef`. Survives re-renders so re-expanding is avoided.
+  // Intentional per-session cache: entries are keyed by the (stable, per-mount)
+  // folder id and are never individually invalidated — the whole ref is
+  // discarded when the view unmounts on exit (RESET_LOCATION_ITEMS), which is
+  // the only path that changes the underlying location. This matches
+  // DeleteView's `folderCountsRef` semantics.
+  const folderExpansionRef = React.useRef<Map<string, DownloadHandlerData[]>>(
```
soberm, line 122: [major] All expanded `DownloadHandlerData` objects for every selected folder are accumulated in `folderExpansionRef` and held for the entire view session. For a deep S3 hierarchy this can be a significant memory allocation kept alive until unmount. The PR description notes expansion is intentionally unbounded, but there's no cap and no eviction. Even a rough size check before calling `setResolvedItems` (wiring in `LARGE_DOWNLOAD_FILE_COUNT` as a confirmation gate) would prevent surprising tab crashes on large hierarchies.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -20,10 +73,245 @@ export const useDownloadView = (
...
+          setIsEnumerating(false);
+          return;
+        }
+
+        // TODO(oracle-phase2-design Q4): large-download confirmation. Design
+        // leaves the exact confirmation UX soft ("e.g. 5000, show a
+        // confirmation"); expansion is intentionally unbounded for
+        // completeness. Wire a confirmation gate here once product specifies
+        // the UX. For now we proceed to make the full set available.
+        void LARGE_DOWNLOAD_FILE_COUNT;
```
soberm: [minor] `void LARGE_DOWNLOAD_FILE_COUNT` is a no-op inserted purely to silence `no-unused-vars`. This is the wrong fix. If the constant isn't being used yet, remove the import or add an `// eslint-disable-next-line` with a TODO ticket reference. A void expression just confuses the next reader into wondering if there's supposed to be a side effect here.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -20,10 +73,245 @@ export const useDownloadView = (
...
+
+        // Apply any per-row removals the user made before enumeration settled
```
soberm: [question] After enumeration completes and calls `setResolvedItems(filterRemoved(resolved, removedItemIdsRef.current))`, the sync effect won't re-run unless `dataItems`, `current`, or `removedItemIds` changes. Is it possible for a removal that lands between the `buildDownloadItems` call and the `setResolvedItems` call inside `runEnumeration` to be silently dropped? The ref read (`removedItemIdsRef.current`) captures the latest value, but React's batching means the sync effect's `setResolvedItems` and the enumeration effect's `setResolvedItems` could both be committed in the same batch with the sync effect's write winning if it runs after. A test exercising removal-during-enumeration would give confidence the ordering is safe.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/DownloadViewProvider.tsx
```diff
@@ -37,10 +46,49 @@ export function DownloadViewProvider({
...
+  // Surface the no-files message in BOTH empty states, not just enumeration:
+  //   - `hasNoFilesToDownload`: enumeration found only empty folders.
+  //   - `allFoldersReady && !hasFilesToDownload`: the ready set went empty
+  //     because the user manually removed every row (mirrors the Start-disable
+  //     gate below). Guarded against enumeration/error/active-processing/
+  //     completion so those states keep ownership of the message via the
+  //     precedence order. Note this can also match a bare mount with no
+  //     selection (empty resolvedItems, vacuously-ready) — that matches the
+  //     pre-existing empty-set semantics and is not a regression.
+  const showNoFiles =
```
soberm, line 62: [minor] `showNoFiles` can be true on bare mount when there is no selection at all: an empty `dataItems` array makes `allFoldersReady` vacuously true and `hasFilesToDownload` false, matching this condition. The comment acknowledges it, but "The selected folders contain no files to download" reads as actively misleading when no folders were ever selected. Consider scoping this to cases where `hasFolders` was true during the current session, or using a more neutral message for the manually-emptied-rows case.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+ * - ['public/nested/one/pic.jpg', 'public/nested/two/pic.jpg'] -> 'nested'
+ * - ['public/nested/a.jpg', 'public/images/b.jpg', 'public/c.jpg'] -> 'public'
+ * - ['photos/a.jpg', 'photos/b.jpg'] -> 'photos'
+ * - ['a.jpg', 'b.jpg'] -> 'download'
+ * - ['public/x.jpg', 'other/y.jpg'] -> 'download'
+ * - [] -> 'download'
+ *
+ * Pure/synchronous; the base name only — the handler appends `.zip`.
+ */
+export const getArchiveName = (fileKeys: string[]): string => {
```
osama-rizk, line 73: Select folder `photos/` whose files all happen to live under `photos/vacation/`, and `getArchiveName` returns the longest common ancestor _of the files_ → `vacation`, so the user gets **`vacation.zip`** after selecting `photos/`. The test *"uses the deepest shared folder for a nested selection" → `vacation`* locks this in.

This looks intended per the changeset, so I'm raising it as a design question rather than a bug: when the selection _is_ a folder, the user's mental model is "I downloaded `photos/`", and LCA-of-files silently drills past the selected boundary. Consider naming the zip after the selected folder (`getFolderName(folder.key)`) when exactly one folder is selected, and reserving LCA for multi-item/mixed selections where there's no single obvious name.

Not a blocker — worth a one-line product confirmation on the intended single-folder name.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
osama-rizk: The changeset says *"arbitrarily large folders download with constant memory."* That's true of the zip **streaming** (Phase 1's service worker) but **not** of the new **enumeration**: `expandFolderToFiles` accumulates every `DownloadHandlerData` into an in-memory array, `folderExpansionRef` retains every folder's full list for the session, `resolvedItems` holds the flattened set, and one table row renders per file. A folder with ~500k objects builds a 500k-element array (and 500k rows) before a single byte is zipped. The `LARGE_DOWNLOAD_FILE_COUNT` guard that would bound this is imported, `void`-ed, and deferred to a TODO here.

Recommend either scoping the changeset wording to the _transfer_, or noting the enumeration memory cost as a known Phase-2 limitation. Quick wording fix, not a blocker.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -20,10 +73,245 @@ export const useDownloadView = (
...
+    };
+
+    runEnumeration();
+
+    // Abort the in-flight enumeration when the selection changes or the view
+    // unmounts, so its `list()` loop stops and no setState-after-unmount occurs.
+    return () => {
+      controller.abort();
+    };
+  }, [dataItems, current, hasFolders, getConfig, enumAttempt]);
```
osama-rizk, line 388: This effect keys on `[dataItems, current, hasFolders, getConfig, enumAttempt]`. Removing a **loose file** dispatches `REMOVE_LOCATION_ITEM`, which changes `dataItems` identity → cleanup fires `controller.abort()` → the effect re-runs. If a selected folder was still mid-enumeration (not yet cached), the abort discards its in-flight `list()` progress and the re-run re-expands it from page 1. For a mixed selection over a huge folder, a stray "remove one file" restarts a long recursive listing.

It converges correctly (the readiness gate holds, no data loss), so this is efficiency, not correctness — flagging because the inline comment calls removing a loose file "harmless", which is true for correctness but not for work done. Already-completed folders survive via the id-keyed cache; only the in-flight run is wasted. Low priority.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -20,10 +73,245 @@ export const useDownloadView = (
...
+    const { prefix } = current;
+    const controller = new AbortController();
+    enumAbortRef.current = controller;
+    setHasNoFilesToDownload(false);
+    setIsEnumerationError(false);
+    setIsEnumerating(true);
+
+    const runEnumeration = async () => {
+      try {
+        await Promise.all(
```
osama-rizk, line 317: `Promise.all(foldersToExpand.map(...))` launches one unbounded pagination loop **per selected folder simultaneously**. Select 40 folders and you fire 40 concurrent `list()` chains against S3 — a throttling/burst risk. `DeleteView` shares this shape but at lower stakes (it caps at 5000 and only counts); the download path is the one that most wants a small concurrency pool here.

Minor — note for a follow-up.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
osama-rizk: `size: item.size!` / `lastModified: item.lastModified!` run _after_ the `endsWith('/')` filter, so only real files reach them and real S3 objects carry both — safe in practice. But the `!` silences the one case that would matter (a malformed/edge list item) rather than handling it.

Cosmetic; matches the "trust the SDK shape" convention elsewhere — wouldn't block on it.

osama-rizk:
Really strong PR — the unit/integration coverage on the abort/retry/no-partial-dispatch paths is excellent, and the inline documentation of the state machine made this easy to reason about. 🙏

One request before merge, if you're open to it: could we add an **end-to-end / integration test** that exercises the real folder-download flow rather than the mocked hook? The core invariant here — "a partial/incomplete archive can never be dispatched" — spans the view, the enumeration, `useAction`, and the zip handler + service worker, and the current suite verifies each layer in isolation with `expandFolderToFiles` mocked. A test that selects a real nested folder, waits for enumeration, clicks Download, and asserts the resulting zip contains every expected entry (with correct `relativePath` structure) would guard the seams the unit tests can't see — especially the `list()` recursion assumption and the SW streaming path.

Totally fine to scope it as a fast-follow if that fits the release better. Thanks!

packages/react-storage/src/components/StorageBrowser/actions/configs/__tests__/defaults.spec.ts
bobbor: @soberm good catch, renamed to 'is not disabled when folders are selected' in 1e63a859a 👍

packages/react-storage/src/components/StorageBrowser/actions/configs/defaults.tsx
bobbor: @soberm done, the comment now just states the invariant: only an empty selection disables Download; folder selections are expanded into files at the view level. (1e63a859a)

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
bobbor: @soberm agreed on the view-into-handler-internals smell. Promoted `constructBucket` through the public `actions` index (same spot `createDownloadItem` already lives) and pointed the DownloadView utils at that. FWIW DeleteView imports from the `actions/handlers` barrel, so there wasn't a single established path; left DeleteView as-is. (1e63a859a)

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
bobbor: @soberm went one better: `expandFolderToFiles` now takes a single options object (a limit/counter param landed in this round anyway), so the `eslint-disable max-params` is gone entirely. (1e63a859a)

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
bobbor: @soberm switched to defensive fallbacks: `item.size ?? 0` and `item.lastModified ?? new Date(0)`, non-null assertions removed. (1e63a859a)

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @soberm fair, this was the real issue behind the placeholder. Wired the cap for real in 1e63a859a: a shared file counter across the whole selection (seeded with loose files + already-cached expansions) makes `expandFolderToFiles` bail once the total would exceed `LARGE_DOWNLOAD_FILE_COUNT` (5000). Over-limit surfaces a new `tooManyFilesMessage` (error variant) and sets `isOverFileLimit`, which blocks dispatch the same way `allFoldersReady` does, so no truncated archive can ever start. No silent truncation, no unbounded accumulation.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @soberm yep, the `void` was a smell. It's gone; the constant is now genuinely consumed by the cap logic (see the thread on the ref accumulation above). (1e63a859a)

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @soberm good question, dug into it: no removal can be dropped. `buildDownloadItems` and `setResolvedItems` are consecutive synchronous statements inside the enumeration continuation, so nothing interleaves between them. If the removal commits before the continuation runs, `removedItemIdsRef.current` is already updated and the completion filter includes it. If it commits after, the sync effect keys on `removedItemIds`, re-runs after the enumeration commit, and rebuilds from the now-populated `folderExpansionRef` with the latest removal set, so the removal wins on the rebuild. Added the test you asked for ('applies a removal that lands while enumeration is in flight', expandFolderToFiles held open on a controllable promise) in 1e63a859a.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/DownloadViewProvider.tsx
bobbor: @soberm agreed, that message on a bare mount would be confusing. The hook now exposes a sticky `hasSelection` flag (selection is or was non-empty this session) and `showNoFiles` is scoped to it, so an empty mount shows nothing and the manually-emptied case still gets the message. Test added for the bare-mount case. (1e63a859a)

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
bobbor: @osama-rizk you read the product intent right, and the intent is actually the opposite of what the LCA gave here: the zip should be named after the folder the download was initiated from. Fixed in 1e63a859a with `resolveArchiveName`: when the selection is exactly one folder, the archive is named after that folder (so `photos/` gives `photos.zip` even when all files live under `photos/vacation/`); LCA stays for file-only, multi-folder and mixed selections where there is no single obvious name. The old 'vacation' test was updated and the canonical file-selection examples ([public/nested/one/pic.jpg, public/nested/two/pic.jpg] -> nested.zip) still hold. 👍

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @osama-rizk fair point, and this round made it partly moot: enumeration is now capped at 5000 files (see the thread with soberm above), with over-limit blocking the download rather than truncating. Also reworded the changeset to scope the constant-memory claim to the transfer and call out that enumeration holds the listing in memory with the 5000-file cap. (1e63a859a)

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @osama-rizk nice catch on the wasted work. You're right that it's efficiency only: completed folders survive via the id-keyed cache and the readiness gate holds, so a stray loose-file removal costs at most the one in-flight folder's listing restarting from page 1. Keeping the current behavior for this PR and tracking a smarter cleanup (e.g. only abort when the removed item affects a folder) as a follow-up. The new 5000-file cap also bounds the worst-case restart cost.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @osama-rizk agreed. With the new 5000-file cap the worst-case burst is bounded, but a small concurrency pool (3-4 folders at a time) is the right shape. Noted as a follow-up alongside the enumeration-restart optimization above rather than growing this PR further. 👍

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
bobbor: @osama-rizk resolved together with soberm's comment on the same lines: swapped the assertions for `item.size ?? 0` / `item.lastModified ?? new Date(0)`. (1e63a859a)

bobbor:
@osama-rizk thanks for the kind words! 🙏 Agreed on the e2e value: the seams between enumeration, useAction and the SW streaming path are exactly what the unit suite can't see (this round's removal-during-enumeration race question from @soberm is a good example of why). Taking you up on the fast-follow scoping: I'll add a storage-browser e2e feature that selects a real nested folder, waits for enumeration, downloads, and asserts the zip entries and their relative paths, in a separate PR so this one can ship the feature. Meanwhile this round did add an integration-style test for the in-flight-removal ordering.

Summary of the review round (1e63a859a): 5000-file enumeration cap wired for real (over-limit blocks dispatch, no truncation), single-folder selections now name the zip after the selected folder, constructBucket promoted to the public actions surface, defensive fallbacks for list() fields, no-files message scoped to actual selections, phase labels scrubbed, changeset wording corrected. 272 tests green, typecheck/lint/build/size-limit all pass.

osama-rizk (dismissed):

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -20,10 +78,276 @@ export const useDownloadView = (
...
+      prevSelectionRef.current.dataItems !== dataItems ||
+      prevSelectionRef.current.current !== current;
+    prevSelectionRef.current = { dataItems, current };
+
+    // On a selection change no prior removals carry over; otherwise apply the
+    // current removal set. (Removing a LOOSE file also mutates dataItems, so it
+    // reads as a selection change here — harmless, since the reducer has already
+    // pruned that file from dataItems. Removing a folder-EXPANDED file no-ops in
+    // the reducer, so dataItems is unchanged and this filter is what removes it.)
+    const effectiveRemovedIds = selectionChanged ? EMPTY_SET : removedItemIds;
```
osama-rizk, line 232: Removing a loose file can resurrect a previously-removed expanded row.

Removals are recorded two different ways: a loose-file removal prunes `dataItems` (so its identity changes), while an expanded folder-file removal no-ops in the reducer and lives *only* in `removedItemIds`. Because this effect treats any `dataItems` identity change as a selection change and resets `removedItemIds` to empty:

1. Remove an expanded row `X` → `removedItemIds = {X}`, `dataItems` unchanged.
2. Remove a loose file `Y` → `dataItems` identity changes → `selectionChanged === true` → `removedItemIds` reset to `∅`.
3. `X` reappears in the download set even though the user explicitly removed it.

The comment notes a loose removal reading as a selection change is "harmless" — true for the loose file itself, but not for expanded removals whose only record is the set being reset. Could we distinguish a genuine selection change from a within-selection removal (e.g. compare by id-set, or skip the reset when the new selection is a subset of the previous one)? A regression test for the remove-expanded-then-remove-loose interleaving would lock it in.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -1,16 +1,74 @@
...
+        items.push(...expanded);
+      }
+    }
+  }
+  // Name the zip from the SELECTION (computed here, in the view — NOT in the
+  // handler) and stamp it uniformly onto every item so the zip handler can
+  // read it from `all[0].archiveName`. A single-folder selection is named
+  // after that folder; every other shape uses the common ancestor directory
+  // of the collected files (see resolveArchiveName).
+  const archiveName = resolveArchiveName(
```
osama-rizk, line 97: Minor: `archiveName` is computed over the pre-removal set. `buildDownloadItems` resolves the name from the full `items` list, then `filterRemoved` drops removed rows afterward. If a user removes rows such that the common ancestor shifts, the archive keeps the stale name. Cosmetic (no data loss) — just noting it in case recomputing post-filter is cheap.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
```diff
@@ -20,10 +78,276 @@ export const useDownloadView = (
...
+      controller.abort();
+    };
+  }, [dataItems, current, hasFolders, getConfig, enumAttempt]);
+
+  // Readiness gate: Start may only dispatch once EVERY selected folder has been
+  // expanded into the cache. Files are always ready; a FOLDER is ready only when
+  // folderExpansionRef has cached its expanded items. Recomputed each render —
+  // state changes from enumeration (setResolvedItems/setIsEnumerating) trigger the
+  // re-render that flips this true after the ref is populated.
+  const allFoldersReady = dataItems.every(
```
osama-rizk, line 395: FYI (no change needed): `allFoldersReady` reads a ref during render, and `hadSelectionRef.current = true` writes one during render. Both work because a `setState` always follows to trigger the observing re-render, and the comments acknowledge it — just flagging the load-bearing assumption since it's the kind of thing concurrent-rendering / StrictMode changes can disturb later.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+      // Enforce the cap per file so the loop stops as soon as the combined
+      // selection exceeds it, mid-page included.
+      fileCounter.count += 1;
+      if (fileCounter.count > LARGE_DOWNLOAD_FILE_COUNT) {
+        throw new FileLimitError();
+      }
+
+      const fileData: FileData = {
+        key: item.path,
+        id: crypto.randomUUID(),
```
osama-rizk, line 220: FYI (not a new risk): `crypto.randomUUID()` is `undefined` outside a secure context. Folder download already depends on a Service Worker, which itself only runs in secure contexts, so this path is already gated — noting it only so it's a conscious dependency rather than an accidental one.

osama-rizk (commented):
Thanks for the thorough write-up and the extensive tests — the no-partial-zip invariant discipline, the abort/cleanup handling, and the edge-case coverage all read really well. A few findings below, mostly around the state seams in `useDownloadView`. The first is the one I'd want addressed before merge; the rest are lower priority / FYI.

packages/react-storage/src/components/StorageBrowser/displayText/types.ts
```diff
@@ -183,6 +183,27 @@ export interface DefaultDeleteViewDisplayText
...
+   * Shown when enumeration completed but produced zero downloadable files
+   * (e.g. only empty folders were selected).
+   */
+  noFilesMessage: string;
+  /**
+   * Shown when the selection exceeds the maximum file count for a single
+   * download. The download is blocked in this state; no truncated zip is
+   * produced.
+   */
+  tooManyFilesMessage: string;
```
soberm, line 206: [question] These 4 new fields are required and this interface is reachable from the public `./browser` entry via `DefaultStorageBrowserDisplayText` (where `DownloadView: DefaultDownloadViewDisplayText` is a required member). Any consumer who constructs or extends an object typed as `DefaultDownloadViewDisplayText` will now get a TypeScript compile error. Strictly that is a semver-major type change, yet the changeset marks this `minor`. Consumers using the shipped default value or the optional `StorageBrowserDisplayText`/`DownloadViewDisplayText` customization interface are unaffected, so practical impact is low. Was `minor` a deliberate call, or should these fields be optional (or the change be `major`) to be safe?

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/types.ts
```diff
@@ -6,7 +6,53 @@ import type {
...
+   * download folders in smaller batches instead.
+   */
+  isOverFileLimit: boolean;
+  /**
+   * `true` when every selected folder has been expanded into its files (the
+   * download-readiness gate). Files are always ready; folders become ready only
+   * once enumeration has cached their contents. Start must not dispatch a
+   * partial set while this is false — the guard lives in `onActionStart`.
+   */
+  allFoldersReady: boolean;
```
soberm: [question] These 7 new required fields extend a type that is reachable to consumers via `UseView`/`useView('download')`. Anyone who has typed a custom download-view component against `DownloadViewState` will now fail to compile until they satisfy all 7. Same semver-vs-`minor` question as the display-text interface. Intentional, or should these be considered a breaking change?

packages/react-storage/src/components/StorageBrowser/displayText/types.ts
bobbor: @soberm deliberate call, and the practical surface is narrower than it looks: every consumer-facing path into these strings is `Partial`, so the required fields only exist on the library-owned defaults object.

Proof by code lines (all in `displayText/types.ts` unless noted):
- The customization interface consumers actually use wraps the default in `Partial`: `export interface DownloadViewDisplayText extends Partial<DefaultDownloadViewDisplayText> {}` (L240-241). Same pattern for every view (L232-252).
- That interface hangs off `StorageBrowserDisplayText` as an optional member: `DownloadView?: DownloadViewDisplayText` (L257-264).
- And the component prop itself is optional: `displayText?: StorageBrowserDisplayText` (`createStorageBrowser/types.ts:317`).

So anyone passing display text through the documented prop chain provides any subset and is unaffected; `DefaultDownloadViewDisplayText` describes the fully-resolved defaults the library ships and merges internally, and consumers don't construct it. The only theoretical break is someone typing their own object as `Default*` directly, which the naming signals as internal-default territory.

Precedent for exactly this shape shipping as minor: folder deletion (#6834) added 5 required keys to `DefaultDeleteViewDisplayText` (confirmationModalTitle etc.) with a minor changeset, same structure, no fallout. Happy to add a doc comment on `DefaultDownloadViewDisplayText` pointing consumers at `DownloadViewDisplayText` if you think the naming isn't signal enough. 👍

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/types.ts
bobbor: @soberm intentional, keeping them required with the minor. Reasoning:

- `DownloadViewState` is a read surface in practice: `useView('download')` consumers receive the object, and extra fields on a received object don't break reads or subset destructuring. The only compile break hits code that constructs a value typed as `DownloadViewState` (test fixtures, hand-rolled reimplementations of the hook), which is accepting that coupling.
- Making the 7 fields optional would be a type-level lie: the hook always returns all of them at runtime, and optionality would force undefined-handling into every internal read site including the Provider's dispatch gating, where non-optional booleans are exactly what we want.
- Precedent: folder deletion (#6834) added 3 required fields to `DeleteViewState` (`onConfirmDelete`, `onCancelConfirmation`, `confirmationModal`) and shipped as minor, so this follows the repo's established pattern for view-state growth.

Good scrutiny though, thanks for pushing on it! 👍

bobbor: @soberm update: we took the size of this surface to heart and shrank it. The 7 fields are now 3: `enumerationStatus: 'NOT_STARTED' | 'PENDING' | 'SUCCEEDED' | 'ERROR' | 'OVER_LIMIT'` (derived per-render, single source of truth), `hasFilesToDownload`, and `hasSelection`. `hasNoFilesToDownload` was derivable (`SUCCEEDED && !hasFilesToDownload && hasSelection`) and is gone; the over-limit case folded into the union as a terminal value that blocks both dispatch and retry. Cancel maps to `NOT_STARTED` which makes the Start-as-retry affordance explicit in the type. All behavioral tests migrated (272 green), bundle shrank ~1 kB. The semver reasoning from my previous reply stands, just with a much smaller footprint. 👍

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @osama-rizk great catch, this was a real resurrection bug. Fixed in b58696892 with your id-set suggestion: the sync effect now stores the previous dataItems id Set and only treats the change as a selection change when a genuinely NEW id appears (or the location changes). A pure removal (new set is a subset of the previous) carries removedItemIds over, so removing loose file Y no longer resurrects expanded row X. The old 'harmless' comment is rewritten, and the exact interleaving you described is locked in as a regression test ('removing a loose file does not resurrect a previously-removed expanded row'). Bonus: an identical-id re-select (the .concat new-array case) also no longer resets removals. 👍

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @osama-rizk cheap indeed, fixed in b58696892: archiveName is now stamped AFTER filterRemoved at both write sites (new resolveDownloadItems helper: build -> filter -> resolve name -> stamp), so the name always reflects the actual download set. The single-folder naming rule still reads the selection (removal-independent by design); only the LCA input is post-filter. Test added: removing rows that shift the common ancestor renames the archive accordingly.

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/useDownloadView.ts
bobbor: @osama-rizk fair flag. Both are the documented ref-mirror pattern (a setState always follows to schedule the observing render), and StrictMode double-render is idempotent for them since they're pure reads/latching writes. Agreed it's load-bearing; the comments call it out so a future concurrent-rendering migration trips over the docs rather than the behavior. Keeping as-is. 👍

packages/react-storage/src/components/StorageBrowser/views/LocationActionView/DownloadView/utils.ts
bobbor: @osama-rizk yes, conscious dependency: folder download rides the SW streaming path which already requires a secure context, and the blob fallback path in non-secure contexts predates this PR. Happy to leave it implicit-but-documented; noting it in the JSDoc if it ever moves out of the SW-gated flow.

osama-rizk (approved):

soberm (approved):

PR #6767, opened 2025-12-02
feat(angular): update minimal Angular version to v19

osama-rizk:
#### Description of changes

This PR completes the Angular 14 to 19 migration for Amplify UI library by cleaning up remaining Angular 14 references and updating build system configurations.

Changes made:
• Updated Jest setup to use setupZoneTestEnv() from jest-preset-angular/setup-env/zone (replaces deprecated setup-jest import)
• Upgraded @testing-library/angular from v12.3.0 to v18.1.1 for Angular 19 compatibility
• Created MockFormFieldComponent stub to replace ng-mocks usage for Angular 19 validation requirements
• Removed Angular 14 specific logic from build system scripts (mega-app-copy-files.sh)
• Simplified USE_STANDALONE conditional logic for Angular 19+ minimum support
• Fixed angular.json buildTarget configuration for Angular 19 CLI compatibility
• Updated build system tests documentation to remove Angular 14 test matrix
• Updated troubleshooting examples to use current Angular versions
• Cleaned up repository-wide Angular 14 references

Note: Passing standalone: false in component decorators does not change the behavior and maintains backward compatibility. This keeps the existing behavior while ensuring Angular 19+ compatibility.

#### Issue #, if available

N/A - Migration cleanup task

#### Description of how you validated changes

• Built and ran tests locally - all pass as expected
• Updated Jest snapshots to match new testing library DOM structure
• Verified Angular 19 compatibility with proper component decorators
• For build system changes, verified by running the following commands:
     • npm run setup:angular:angular-cli -- --framework-version latest --build-tool-version 19 --pkg-manager npm --tag latest
     • npm run setup:angular:angular-cli -- --framework-version latest --build-tool-version latest --pkg-manager npm --tag latest

#### skipLibCheck Added

Added `"skipLibCheck": true` to resolve type incompatibilities with two dev dependencies:

- `@testing-library/angular` - expects `Binding` export removed in Angular 19
- `@types/node` vs `@types/webpack-env` - conflicting `require` variable declarations

These are dev-only type conflicts that don't affect runtime functionality. `skipLibCheck` bypasses library type checking while preserving type safety for our source code.

#### Checklist

• [x] Have read the [Pull Request Guidelines](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md)
• [x] PR description included
• [x] yarn test passes and tests are updated/added
• [x] PR title and commit messages follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax
[...]

PR #4117, opened 2023-06-15
chore!: bump minimum Angular version to 14

sreeramsama:
#### Description of changes

This PR bumps the minimum supported Angular version from 12 to 14.

#### Commits

No commits other than below were made without a PR.

- #3861 
- #4035
- #4087
- #4116 
- #4128
- 6e9479a36a9fb3a48f068b90c277c18841220b5 (eslint fix in #4128)
- d2d12f9d730b1675ea7afd3ca904f6d6aebfd48f (code highlight fix in this PR)

#### Description of how you validated changes

Manual smoke tests + build system tests + existing CI.

#### Checklist

- [x] Have read the [Pull Request Guidelines](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md)
- [x] PR description included
- [x] Relevant documentation is changed or added (and PR referenced)
- [x] `yarn test` passes and tests are updated/added
- [x] No side effects or [`sideEffects`](https://github.com/aws-amplify/amplify-ui/blob/main/packages/react/CONTRIBUTING.md#code-standards) field updated

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7047, opened 2026-07-14
chore(angular): upgrade @aws-amplify/ui-angular to Angular 20

bobbor:
Supersedes the dependency-only bump in #7019.

Upgrades the Angular library and example app from Angular 19 to 20.3, making the Angular 20 move real and consistent (supersedes the effectively-cosmetic `@angular/common`-only bump in #7019).

## Dependencies
- All `@angular/*` → 20.3.26 (runtime + compiler-cli); `@angular/cli` & `@angular-devkit/build-angular` → 20.3.x
- `ng-packagr` → ^20.3; `@angular-eslint/*` → ^20.7; `@testing-library/angular` → ^18.1; `ng-mocks` → ^14.15
- `typescript` → ~5.8 (Angular 20 requires >=5.8 <5.9)
- Root `package.json` resolutions bumped (these pinned all `@angular/*` to 19.2.20 and were the reason the prior bump never took effect)

## Code
- Migrated constructor DI to field `inject()` across 19 component/directive files via Angular's official `inject()` schematic (required by `@angular-eslint/prefer-inject`, newly enforced in angular-eslint 20)

## Testing
- Library build: ✅ PASS
- Unit tests (jest): ✅ PASS (15 suites / 43 tests)
- Lint (tsc + eslint): ✅ PASS
- The example-app `@environments/*/aws-exports` build failure is pre-existing/environmental (missing e2e-generated config), not caused by this upgrade


osama-rizk:
Could we add a changeset for this one? 🙏

Even though the bulk of the diff is `yarn.lock` / the example app / root `resolutions` (all private or in the changeset `ignore` list), the 19 `inject()` migrations under `projects/ui-angular/src/lib/**` are source of the **published** `@aws-amplify/ui-angular` package, which isn't ignored. Without a changeset it won't be versioned/released, and consumers won't get a changelog entry for what is effectively an Angular-20 support move.

I'd suggest a `minor` on `@aws-amplify/ui-angular` — it makes the library build against and support Angular 20, which is consumer-relevant, not purely internal. (`patch` if the team considers it internal-only, but I'd not go lower.)

One reason the changelog entry matters here specifically: the published peer range is still `@angular/core: ">= 19.0.0"`, but the library is now built with the Angular 20 toolchain (`ng-packagr` 20, `compiler-cli` 20, TS 5.8). Worth a line so consumers know the build surface moved even though the peer range text didn't.

osama-rizk:
Question on scope for the `inject()` migration 🤔

The library migration cites `@angular-eslint/prefer-inject` (now `error` in the angular-eslint 20 `recommended` preset). The example app also bumped to angular-eslint 20 and extends `plugin:@angular-eslint/recommended`, but three of its components still use constructor DI:

- `examples/angular/src/pages/ui/components/authenticator/sign-up-with-email/sign-up-with-email.component.ts:46`
- `examples/angular/src/pages/ui/components/authenticator/auth-status/auth-status.component.ts:15`
- `examples/angular/src/pages/ui/components/authenticator/custom-slots/custom-slots.component.ts:76`

Should we also migrate these to field `inject()` for consistency (they'd be flagged by the same rule, and `eslint .` inside `examples/angular` would report them — though it isn't wired into the CI `lint` filter, so nothing goes red)?

Or is leaving them on constructor DI intentional — so the example app deliberately exercises **both** DI styles against Angular 20 and validates that constructor-parameter injection still works alongside the migrated `inject()` path? If that's the intent, it might be worth a quick comment so a future reader doesn't "tidy" them away.

Either way is defensible — mostly want to make sure the split is a choice rather than a miss. No strong opinion; these compile fine under Angular 20 regardless.

bobbor:
@osama-rizk thanks for flagging the changeset! 🙏 One landed shortly after your comment — but as **major** rather than minor, together with bumping the published peerDeps to `@angular/core / @angular/common >= 20.0.0`.

Reasoning: the library is now compiled with the Angular 20 toolchain, and partial-Ivy output is only forward-compatible — so Angular 19 consumers are outside the supported matrix. Keeping `>= 19.0.0` would promise something the build can no longer guarantee, and dropping a supported peer major is a semver-breaking change. On top of that, Angular 19 is EOL with unpatched high-sev advisories — this PR closes all 9 open `@angular/*` Dependabot alerts. 🎉

For what it's worth, precedent cuts both ways: the 14→19 jump shipped as minor in #6767, but the 2023 peer-floor raise went out as major in #4117. We went with the semver-correct option here — happy to discuss if the team prefers minor!

bobbor:
@osama-rizk good catch — that was a miss, not a choice 😅 The `inject()` schematic only ran against the library workspace, so the example app got skipped.

The three components you listed are migrated now (22cd5602c), and while at it we also caught `useAuthenticator` / `useAuthenticatorHome` (ba2f29a1b) — those were the last two with constructor DI, so the example app is now consistently on `inject()`. No intentional dual-DI-style showcase, promise 🙂

osama-rizk (dismissed):

osama-rizk (dismissed):

soberm (approved):

osama-rizk (approved):

PR #7055, opened 2026-07-16
fix(ui): make Authenticator delivery message punctuation translation-owned (i18n)

### LINKED ISSUE #6966: Wrong, hardcoded punctuation in Japanese

ahmedhamouda78:
## Summary

Fixes #6966. The Authenticator's `getDeliveryMessageText()` appended **hardcoded ASCII periods** to already-translated delivery messages, so non-Latin locales rendered incorrect punctuation — e.g. Japanese, which uses the ideographic full stop `。`, and Thai, which uses no period at all.

### Root cause

`packages/ui/src/helpers/authenticator/textUtil.ts`:

```ts
return `${translate(DefaultTexts.CODE_SENT)}. ${arrivalMessage}.`;        // unknown medium
return `${instructionMessage} ${Destination}. ${arrivalMessage}.`;        // email / sms
```

The `.` after the instruction, after `${Destination}`, and the trailing `.` were all hardcoded ASCII and could never be localized.

### Approach — punctuation is now owned by each translation

- `translate()` accepts an optional `values` argument that interpolates `{placeholder}` tokens via `replaceAll` (fully backward compatible with existing callers).
- New composite `DefaultTexts` keys `DELIVERY_MESSAGE_UNKNOWN` / `DELIVERY_MESSAGE_EMAIL` / `DELIVERY_MESSAGE_PHONE` carry `{destination}` / `{arrivalMessage}` placeholders; `CODE_ARRIVAL` carries its own trailing period.
- `getDeliveryMessageText()` builds the message from those keys instead of concatenating literal `.` characters.
- All 19 locale dictionaries gain the composite delivery keys (and the period-inclusive arrival key where it was missing) with locale-appropriate punctuation (`。` for Japanese/Chinese, none for Thai, `.` for Latin scripts).

### English output is byte-identical

`DELIVERY_MESSAGE_UNKNOWN` keeps the full English wording, so English rendering is unchanged — **no `@aws-amplify/ui`, `ui-vue`, or `ui-react-native` snapshots change**.

### Locale coverage

A data-driven test iterates all 18 non-English locales and asserts that the arrival phrase and the email delivery message resolve **in-locale** (never falling back to English). Japanese-specific tests additionally assert the ideographic full stop is used and that no ASCII period follows the destination.

### Verification

- `tsc --noEmit` ✅
- `jest` (`textUtil` + `authenticatorTextUtil`) ✅ — 75 tests / 14 snapshots pass (36 pre-existing assertions pass unchanged, confirming byte-identical English)
- `prettier --check` ✅
- changeset added (patch bump for `@aws-amplify/ui`)

### Credit

The composite-key + `translate()` interpolation approach (and the reviewed per-locale translations) was originally proposed by @bobbor in #6968 — thank you. This PR is an independent, completed take that additionally closes gaps in that PR:

- **Vue CI** — keeps English byte-identical, so Vue/React Native snapshots don't need regenerating (the Vue snapshot job on #6968 is red).
- **English regression** — #6968's `DELIVERY_MESSAGE_UNKNOWN` dropped "To log in, enter the code we sent you"; this PR preserves the original English wording.
- **`ja` / `de` / `hu` / `id` / `th` arrival fallback** — #6968 added a trailing period to `CODE_ARRIVAL` but omitted the period-inclusive arrival key in several locales, so the arrival phrase fell back to English; this PR ensures every locale resolves it in-locale (covered by the data-driven test).

Fixes #6966


osama-rizk:
Is it feasible to add e2e tests for the newly added functionality ?

osama-rizk (dismissed):

ahmedhamouda78:
Thanks @osama-rizk — added an e2e scenario for the delivery-message punctuation.

It extends the existing `authenticator/i18n` example (already wired for all four frameworks under `I18n.setLanguage('ja')`), mocks `SignUp` with the `sign-up-with-email` fixture, and drives sign-up → Confirm Sign Up (which renders `getDeliveryMessageText`). It asserts the Japanese line uses the ideographic full stop, not the old ASCII period:
- sees `送信先: a***@e***.com。` and `コードを受信するまで数分かかる場合があります。`
- does **not** see `a***@e***.com.`

Deterministic (fixture-mocked, no live backend). Files:
- `packages/e2e/features/ui/components/authenticator/i18n.feature` — new `@angular @react @vue @svelte` scenario
- `packages/e2e/cypress/integration/ui/components/authenticator/i18n/i18n.steps.ts` — one small locale-aware input step; assertions reuse shared `I see` / `I don't see`

Complements the unit coverage in `textUtil.test.ts` (locale×medium matrix, no-placeholder-leak, English byte-identity). Note: e2e is skipped on fork PRs and only runs after a maintainer approves — I can't trigger it myself.

soberm (dismissed):

osama-rizk (dismissed):

packages/ui/src/helpers/authenticator/textUtil.ts
```diff
@@ -24,6 +24,36 @@ const getChallengeText = (challengeName?: ChallengeName): string => {
...
+    return sentence;
+  }
+  if (CJK_CHARACTER.test(sentence)) {
+    return `${sentence}。`;
+  }
+  if (THAI_CHARACTER.test(sentence)) {
+    return sentence;
+  }
+  return `${sentence}.`;
+};
```
bobbor: 2 things here:

no.1
these early-return branches might cause problems.
`CJK_CHARACTER.test(sentence)` matches anywhere so ANY sentence in any language containing a single "CJK" adds a "。" - unless there is punctuation already. the same is true for `THAI_CHARACTER`.

this is safe in most cases, but e.g. korean phrases might contain CJK chars, and then utilize wrong periods or spaces.

IMO a safer approach would be to check if "the majority of the sentence" contains CJK or THAI (using exec or match).

no.2
the TERMINAL_PUNCTUATION, checks for punctuation at the end only.
meaning: `/[.!?。！？]$/` misses ellipsis `…`, full-width period `．`, and colon `:`. so any customers-override might add unwanted periods. (e.g. arrivalMessage is a custom "Might take a minute to two…"

packages/ui/src/helpers/authenticator/textUtil.ts
```diff
@@ -24,6 +24,36 @@ const getChallengeText = (challengeName?: ChallengeName): string => {
...
+// translated copy, so punctuation stays locale-correct without adding any
+// translation key, interpolation, or public API.
+const CJK_CHARACTER =
+  /[\u3040-\u30ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uff66-\uff9f]/;
+const THAI_CHARACTER = /[\u0e00-\u0e7f]/;
+const TERMINAL_PUNCTUATION = /[.!?。！？]$/;
+
+// Appends the script-appropriate sentence terminator to `sentence`, unless it
+// already ends with terminal punctuation (e.g. a customer vocabulary override).
+const terminateSentence = (sentence: string): string => {
```
bobbor: have you considered moving this fn in its own i18n-helper util/file?

packages/ui/src/helpers/authenticator/textUtil.ts
```diff
@@ -35,16 +65,22 @@ const getDeliveryMessageText = (
...
   }
 
   const instructionMessage = isEmailMessage
     ? translate(DefaultTexts.CODE_EMAILED)
     : translate(DefaultTexts.CODE_TEXTED);
 
-  return `${instructionMessage} ${Destination}. ${arrivalMessage}.`;
+  return `${terminateSentence(
+    `${instructionMessage} ${Destination}`
+  )}${spacer}${terminateSentence(arrivalMessage)}`;
```
bobbor, line 57: this is independent of the PR, but oh boy, how I don't like this.

but there is no way for a user to say:
"enter the code we emailed to ${Destination} to log in."

extend the translate function to interpolate, with sane defaults would be much nicer.
(check my [age-old PR here](https://github.com/aws-amplify/amplify-ui/pull/6968/changes#diff-333359b718024f11abaa428a5846d29bcc2f5295739b0f7ffa88fdec39bd28d8))

osama-rizk (approved):

packages/ui/src/helpers/authenticator/textUtil.ts
ahmedhamouda78: Good catches, fixed both in `4e1eb7e`. For no.1 I swapped the "contains any CJK/Thai char" check for a majority one: it counts CJK and Thai characters and only picks that script when it's more than half the non-whitespace text, otherwise it falls back to a plain `.`. So a Latin sentence with a lone CJK proper noun, or a mostly-Hangul phrase with a stray CJK codepoint, stays `.` now, while ja/zh keep `。` and th stays terminator-free. For no.2, the "already terminated" guard now also treats `…`, `．`, `｡`, `:` and `：` as terminal, so an override ending in `…` or `:` won't get an extra period tacked on. Covered by new tests in `sentencePunctuation.test.ts` plus a few integration cases through `getDeliveryMessageText`.

packages/ui/src/helpers/authenticator/textUtil.ts
ahmedhamouda78: Done. Moved the punctuation helpers into their own file at `packages/ui/src/i18n/sentencePunctuation.ts`, next to the other i18n modules. It's internal only, not exported from `i18n/index.ts` or the public entry, so the public API is unchanged. `textUtil.ts` just imports the two helpers now.

packages/ui/src/helpers/authenticator/textUtil.ts
ahmedhamouda78: Yeah, fair point. Agreed the interpolation approach reads nicer, and honestly it's the only way to let a locale move `{destination}` around like in your example. I kept this PR to a zero-API-change fix on purpose though: I didn't want to touch `translate()`'s signature or add keys just for a punctuation bug, especially with existing overrides in play. Interpolation really wants its own PR, and #6968 already lays the groundwork. Happy to pick that up as a follow-up once this one's in.

bobbor (approved):

osama-rizk (approved):

PR #7083, opened 2026-07-27
fix(authenticator): apply UI config when signing out before init

### LINKED ISSUE #6967: Bug: Signing out before the Authenticator is shown will break services, and probably more.

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-ui/issues?q=is%3Aissue+) and [discussions](https://github.com/aws-amplify/amplify-ui/discussions).
- [x] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-ui/blob/main/CONTRIBUTING.md#bug-reports).

React

### Which UI component?

Authenticator

### How is your app built?

Vite

Chrome

### Please describe your bug.

If you `signOut()` before showing the `<Authenticator/>` any `services` _(and probably also other configuration)_  will be ignored.

Showing the Authenticator after the logout should still use the configuration as expected.

I prepared an example repository that allows to easily reproduce the bug:

https://github.com/jiw-mh/amplify-ui-issue-6967

Clone the repository ↑ and follow the [Readme](https://github.com/jiw-mh/amplify-auth-problem/blob/main/Readme.md#amplify-issue-example-project) to replicate the problem.

It seems the problem is in the state machine.

https://github.com/aws-amplify/amplify-ui/blob/a6470c163288a6dcd928959d41b4cf74e71ddd27/packages/react-core/src/Authenticator/hooks/useAuthenticatorInitMachine/useAuthenticatorInitMachine.tsx#L17

After the `signOut` (_from what I can gather_) the state machine _should_ (?) be going to the `setup` route:

https://github.com/aws-amplify/amplify-ui/blob/a6470c163288a6dcd928959d41b4cf74e71ddd27/packages/ui/src/machines/authenticator/index.ts#L303-L308

But going through it with a debugger, it seems to land at `signIn` when the Authenticator is first rendered.

vishwakt:
#### Description of changes

Signing out before the `Authenticator` renders caused every prop passed to it (`services`, `formFields`, `initialState`, ...) to be ignored.

The machine waits in `setup.initConfig` for the UI to send `INIT`, which carries those props. `initConfig` also handles `SIGN_OUT`, and `done.invoke.signOutActor` targeted `setup.getConfig` unconditionally, skipping `initConfig` and the `configure` action. The machine settled on `signIn`, and since `useAuthenticatorInitMachine` only initializes when `route === 'setup'`, `INIT` was never sent.

Now returns to `setup` when setup has not completed, reusing the existing but unreferenced `shouldSetup` guard. `hasSetup` is only set after `getConfig` resolves, so the ordinary sign out path is unchanged.

#### Issue #, if available

Fixes #6967

Repro page passes `initialState="signUp"`, a custom `formFields` label, and a `services.validateCustomSignUp` override, then signs out before rendering the `Authenticator`.

Initial render, before and after. `initialState` and `formFields` ignored, then applied:

after-1-initial
before-1-initial

After submit, before and after. The default `handleSignIn` reaches Cognito, confirming `services` was dropped, then the override runs with no network call:

after-2-submitted
before-2-submitted

Tests in `packages/ui/src/machines/authenticator/__tests__/index.test.ts`:

- Added `should wait for INIT after signing out before the UI has initialized the machine`. Fails on `main` with `setup.getConfig` instead of `setup.initConfig`.
- Restored `should spawn the signOutActor`, `it.skip`'d with a `@todo-migration` note. It passed `getCurrentUser` where the machine invokes `handleGetCurrentUser` and never sent `INIT`, so it could not reach `authenticated`. Now asserts the existing `setup.getConfig` target is preserved once set up.

`yarn test` and `yarn lint` pass in `packages/ui`, `yarn test` passes in `packages/react-core`.

<details>
<summary>Repro page, no backend required</summary>

Drop into `examples/next/pages/ui/components/authenticator/signout-before-init/index.page.tsx`.

The sign out is gated on `route === 'setup'` because it otherwise races the `idle` state's `getCurrentUser`. `idle` has no `SIGN_OUT` handler, so an early event is dropped and the bug does not reproduce.

```tsx
import React from 'react';
import { Amplify } from 'aws-amplify';
import { signOut } from 'aws-amplify/auth';

import { Authenticator, useAuthenticator } from '@aws-amplify/ui-react';

// no backend calls are made by this page, the values only need to be well formed
Amplify.configure({
  Auth: {
    Cognito: {
      userPoolId: 'us-east-1_000000000',
      userPoolClientId: '00000000000000000000000000',
    },
  },
});

const formFields = {
  signUp: {
    username: { label: 'Custom Sign Up Label' },
  },
};

const services = {
  async validateCustomSignUp(formData: Record<string, string>) {
    if (!formData.username) {
      return;
    }
    return { username: 'Custom services validation ran' };
  },
};

function SignOutBeforeAuthenticator() {
  const { route } = useAuthenticator(({ route }) => [route]);
  const [hasSignedOut, setHasSignedOut] = React.useState(false);
  const hasSignedOutRef = React.useRef(false);

  React.useEffect(() => {
    if (route === 'setup' && !hasSignedOutRef.current) {
      hasSignedOutRef.current = true;
      signOut().finally(() => setHasSignedOut(true));
    }
  }, [route]);

  if (!hasSignedOut) {
    return <p>signing out</p>;
  }

  return (
    <Authenticator
      initialState="signUp"
      formFields={formFields}
      services={services}
    />
  );
}

  return (
    <Authenticator.Provider>
      <SignOutBeforeAuthenticator />
    </Authenticator.Provider>
  );
}
```

</details>

#### Checklist

- [x] PR description included
- [x] _If this change should result in a version bump_, [changeset added](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)


soberm (dismissed):

vishwakt:
@soberm , i saw the e2e test fail when the CI was run.

The svelte `Headless Usage` failure was a regression from my first approach and i've pushed a different fix.

`examples/svelte/.../useAuthenticator/home/+page.svelte` renders `Authenticator.Provider` with no `Authenticator` and navigates off the route:

```js
$effect(() => {
  if (authenticator.route === 'signIn') goto('.');
});
```

Nothing sends `INIT` there, so returning to `setup.initConfig` after the sign out left it at `route === 'setup'`. Navigation never fired. React and Vue passed because their pages call `router.push` in the click handler instead of reacting to the route.

Headless usage depends on reaching `signIn` without `INIT`, so the sign out transition is now identical to `main`. Fixed the initialization path instead:

- `configure` records that the UI configured the machine, and `INIT` is accepted at the root until it has. The guard also requires `hasSetup` so a late `INIT` cannot interrupt `idle` before `handleGetCurrentUser` resolves.
- `useAuthenticatorInitMachine` sends `INIT` on any route other than `idle`, so an `Authenticator` rendered after setup still delivers its props.

Verified by mirroring the svelte page in a React example: reproduced the CI failure locally, then confirmed it reaches `signIn` and navigates. Original repro still applies its `config` and `services`.

Tests: replaced the assertion that the machine stops at `setup.initConfig` with one that it reaches `signInActor` without `INIT`, which nothing covered before. Added late `INIT` applies `config`, `INIT` ignored once configured, and hook coverage for both routes. Changeset now bumps `ui-react-core` as well.

soberm (approved):

bobbor (approved):

soberm:
Thanks for this PR and your contribution!

PR #624, opened 2021-11-03
feat(angular): support sign up customization

wlee221:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #616, opened 2021-11-03
Add transformOrigin style property

hvergara:
*Description of changes:*
Add a `transformOrigin` style property to <View> derived components

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #615, opened 2021-11-03
Apply `box-sizing: border-box` to all Amplify theme children

reesscot:
*Description of changes:*
Applies styling rule `box-sizing: border-box` to all children of Amplify theme element.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #7101, opened 2026-08-04
fix(deps): bump undici to 6.28.0 to address CRLF/cookie-injection & response-desync advisories

osama-rizk:
#### Description of changes

Resolves six medium-severity Dependabot alerts for the transitive `undici` dependency (three in the root `yarn.lock`, three in `build-system-tests/e2e/yarn.lock`), all fixed in `undici@6.28.0`:

- **#622 / #614** — downstream response desynchronization
- **#623 / #615** — cookie attribute injection via unsanitized values
- **#624 / #616** — CRLF injection via blob-like body

`undici` was pinned to `6.27.0` in both lockfiles and is purely transitive. Both existing ranges (`^6.23.0`, `^6.25.0`) already allow `6.28.0`, so this is a lockfile-only re-resolution — surgically editing just the `undici` entry in each lockfile (version, `resolved`, `integrity`). `undici@6.28.0` has no runtime dependencies, so nothing else in the tree shifts.

#### Issue #, if available

Dependabot alerts #614, #615, #616, #622, #623, #624.

- Confirmed `6.28.0` is the first patched version for all three advisories and satisfies the existing `^6.x` ranges.
- Verified the pinned `version` / `resolved` tarball / `integrity` hash against the npm registry for `undici@6.28.0`.
- Diff is limited to the single `undici` block in each of the two lockfiles (no dependency lines, since `undici` has none).

#### Checklist

- [x] Have read the Pull Request Guidelines
- [x] PR description included
- [x] PR title and commit messages follow conventional commit syntax
- [ ] _If this change should result in a version bump_, changeset added

> Note: dependency-only security fix in lockfiles (transitive, purely in private `e2e` and non-published trees). No changeset needed; no source or test changes.

### DIFF
--- a/build-system-tests/e2e/yarn.lock
+++ b/build-system-tests/e2e/yarn.lock
@@ -3618,9 +3618,9 @@ typescript@^5.1.6, typescript@^5.9.3:
   integrity sha512-WRNW+sJgj5OBN4/0JpHFqtqzhpbnV0GuB+OozA9gCL7a993SmU+1JBZCzLNxYsbMfIeDL+lTsphD5jN5N+n0zg==
 
 undici@^6.23.0:
-  version "6.27.0"
-  resolved "https://registry.yarnpkg.com/undici/-/undici-6.27.0.tgz#41f9e48f7c5a40d27376caaead8c9a9fc7bca9c4"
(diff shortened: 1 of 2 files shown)


osama-rizk:
Superseded by #7102, which consolidates all open Dependabot alert fixes (including the e2e-lockfile undici bump this PR covered) into a single change. Closing in favor of that.

PR #7111, opened 2026-08-13
fix(authenticator): handle SIGN_OUT while resolving the current user

### LINKED ISSUE #7110: Bug: SIGN_OUT is silently dropped while the Authenticator machine is in `idle`, and can leave the machine `authenticated` after a successful sign out

- [x] I have searched for duplicate or closed issues and discussions.
- [x] I have read the guide for submitting bug reports.

React

_The root cause is in `@aws-amplify/ui`, so this applies to every framework that uses the authenticator machine._

### Which UI component?

Authenticator

### How is your app built?

Vite

Chrome

### Please describe your bug.

The machine's `idle` state invokes `handleGetCurrentUser` and declares no event handlers, and `SIGN_OUT` is not handled at the machine root:

https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/machines/authenticator/index.ts#L107-L114

`defaultAuthHubHandler` forwards both the `signedOut` and `tokenRefresh_failure` Hub events to the machine as `SIGN_OUT`. If either arrives while the machine is still in `idle`, xstate silently discards it. Two consequences:

1. **The machine can end up `authenticated` after a successful sign out.** Sequence: a signed in user loads a page that calls `signOut()` on mount (a `/logout` route, session cleanup, etc.). `idle` has already invoked `handleGetCurrentUser`. `signOut()` clears tokens and dispatches `signedOut`, the forwarded `SIGN_OUT` is dropped, and if the in-flight `getCurrentUser` resolves with the cached user (async token storage widens this window, e.g. React Native AsyncStorage or custom storage adapters), the machine runs `setUser` -> `setup` -> `getConfig`, the `hasUser` guard passes, and the machine settles in `authenticated`. `route` reports `authenticated` with a stale `user` while the session is gone. In React, `authStatus` flips to `unauthenticated` through the provider's own Hub listener, so `route` and `authStatus` now contradict each other.

2. **`useAuthenticator().signOut()` is a silent no-op during `idle`.** The facade's `signOut` is `sendToMachine('SIGN_OUT')`, so calling it before the machine leaves `idle` does nothing and gives the caller no feedback. Observed directly while building the repro for #6967: with the machine at `route === 'idle'`, clicking a button wired to `signOut()` produced no transition.

The same gap exists in the `getCurrentUser` state, which also only handles the invoke result.

`SIGN_OUT` is honored in every state where a sign out is meaningful. Signing out during startup should land the machine in `setup`/`signIn`, never `authenticated`.

Deterministic at the machine level, no backend needed. Drop this into `packages/ui/src/machines/authenticator/__tests__/` and run it, the final state is `{ authenticated: 'idle' }` with the stale user in context:

```ts
import { interpret } from 'xstate';
import { setImmediate } from 'timers';

import { createAuthenticatorMachine } from '..';

jest.mock('aws-amplify');
const flushPromises = () => new Promise(setImmediate);

it('SIGN_OUT during idle is dropped and the machine ends authenticated', async () => {
  let resolveUser: (u: unknown) => void;
  const services = {
    getCurrentUser: () =>
      new Promise((resolve) => {
        resolveUser = resolve;
      }),
    getAmplifyConfig: () => Promise.resolve({} as any),
  };

  const service = interpret(
    createAuthenticatorMachine().withContext({ config: {}, services })
  ).start();

  expect(service.getSnapshot().value).toStrictEqual('idle');

  // user signs out now, the Hub forwards it as SIGN_OUT
  service.send({ type: 'SIGN_OUT' });
  await flushPromises();
  expect(service.getSnapshot().value).toStrictEqual('idle'); // dropped

  // the in-flight getCurrentUser resolves with the cached user
  resolveUser!({ username: 'stale-user' });
  await flushPromises();
  service.send({ type: 'INIT', data: { services } });
  await flushPromises();

  // machine is authenticated even though the user signed out
  expect(service.getSnapshot().value).toStrictEqual({
    authenticated: 'idle',
  });
  expect(service.getSnapshot().context.user).toStrictEqual({
    username: 'stale-user',
  });
  service.stop();
});
```

Browser demonstration of the no-op: render `Authenticator.Provider`, read `route` from `useAuthenticator`, and call `signOut()` while `route === 'idle'`. Nothing happens. This is why the repro page in #6967 had to gate its `signOut()` call on `route === 'setup'`.

Found while working on #6967 / #7083.

Suggested fix: handle `SIGN_OUT` on `idle` (and `getCurrentUser`) targeting `#authenticator.signOut`. Exiting the state cancels the in-flight invoke, so the stale `getCurrentUser` result can no longer be applied. After the sign out actor completes the machine goes to `setup.getConfig`, and since #7083 a late `INIT` still applies the UI's `config` and `services`, so the early transition is safe for a UI rendered afterwards.

Fix incoming.

vishwakt:
The `idle` and `getCurrentUser` states only declare handlers for their invoke result, and `SIGN_OUT` is not handled at the machine root, so a sign out occurring while `handleGetCurrentUser` is in flight is silently dropped. If the in-flight call then resolves with the cached user, the machine settles in `authenticated` with a stale user after a successful sign out. `useAuthenticator().signOut()` is also a no-op until the machine leaves `idle`.

Handle `SIGN_OUT` in both states by transitioning to `signOut`. Exiting the state cancels the in-flight invoke, so the stale result can no longer be applied.

Fixes #7110

#### Checklist

- [ ] PR description included


bobbor (approved):
the changes look good 

thanks @vishwakt

soberm (approved):

PR #7115, opened 2026-08-25
fix(storage): handle multi-file download errors (e.g. 403 Glacier)

mwellmn:
#### Description of changes

Adds a response-status check to the multi-file (zip) download handler so that non-OK HTTP responses are surfaced as errors instead of being silently written into the downloaded archive.

Previously, download() in zipdownload.ts passed the fetch response body straight to the zip writer without checking `response.ok`. When S3 returned an error for an object — for example a 403 for an object in Glacier / Glacier Deep Archive storage that hasn't been restored — the error response body (XML error payload) was zipped into the archive as if it were the file's contents. The user got a "successful" download containing a corrupt/garbage file rather than a clear failure.

1. In a Storage Browser multi-file download, include an object that returns a non-OK response (e.g. a 403 from an unrestored Glacier object).
2. Before: the archive downloads "successfully" but the affected file contains the S3 error XML.
3. After: the download for that key throws failed to download; the rest of the batch is unaffected.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/zipdownload.spec.ts
+++ b/packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/zipdownload.spec.ts
@@ -137,6 +137,9 @@ describe('zipDownloadHandler', () => {
     mockGetUrl.mockResolvedValue({ expiresAt, url });
 
     globalThis.fetch = jest.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      statusText: 'OK',
       headers: { get: (h: string) => (h === 'content-length' ? '100' : null) },
       body: new ReadableStream({
         start(ctrl) {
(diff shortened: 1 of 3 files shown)


packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/zipdownload.spec.ts
```diff
@@ -229,6 +232,55 @@ describe('zipDownloadHandler', () => {
     });
   });
 
+  it('marks a non-ok response (e.g. 403 Glacier) FAILED without blocking the rest of the batch', async () => {
```
osama-rizk, line 235: This test is good — it asserts the FAILED result shape and batch isolation — but it verifies the *result status*, not the behavior the fix actually exists to prevent. A few gaps worth closing:

1. **The core symptom isn't directly asserted (the one I'd add).** The bug is that the error-response body gets written into the zip as file content. This test infers "not zipped" from `status: 'FAILED'`, but never asserts that no zip entry was created for the failed key. Since the `!response.ok` throw happens *before* the `ZipWriter.add`, a direct assertion — the zip writer's `add` was not called for `prefix/archived` (or the resulting archive contains only the successful key) — would pin the actual defect rather than a proxy for it.

2. **Sequential, not concurrent.** `file1` and `file2` are invoked one after the other (`await r1`, then `r2`), so this doesn't exercise a real in-flight batch with a live abort signal. Adequate today because the error path never aborts `batchAbort`, but it wouldn't catch a future regression that made a per-file failure cancel the batch.

3. **Not covered:** the empty-`statusText` path (HTTP/2 responses carry no reason phrase, so the message renders `Failed to download prefix/archived: 403 ` with a trailing space), and whether the successful sibling's bytes actually land in the archive (`r2` asserts `COMPLETE`, not content).

None block the fix — it's verified correct — but (1) in particular would make the test guard the thing the PR is about.

osama-rizk (commented):
Reviewed the fix end to end — correct and well-placed: the `response.ok` throw lands before any zip entry is created (so no error body is written into the archive), and I confirmed the error path never touches `batchAbort`, so sibling downloads are genuinely unaffected as the description claims. Recommend approve. One note on test coverage inline.

osama-rizk (dismissed):
Approving. The fix is correct and minimal — the `response.ok` throw fires before any zip entry is created, and the error path never touches `batchAbort`, so sibling downloads are unaffected. The follow-up commit addresses the test-coverage note directly: the `mockAddedFilenames` assertions now pin the no-error-body-in-zip behavior itself (not just the FAILED status), and the added concurrent-dispatch test covers batch isolation. Thanks for the thorough follow-up.

packages/react-storage/src/components/StorageBrowser/actions/handlers/__tests__/zipdownload.spec.ts
mwellmn: Added/updated tests to address most of these gaps.

The only one I didn't address is the trailing white space with empty `statusText`. We're not currently surfacing these messages directly to the customer so I don't think it adds much value. If we do want to show it to the customer we may want to use the error `message` instead which is more likely to have actionable guidance, but that can be a separate discussion for future consideration.

osama-rizk (dismissed):

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
```diff
@@ -236,6 +236,11 @@ const download = async (
     mode: 'cors',
     signal: state.batchAbort.signal,
   });
+  if (!response.ok) {
+    throw new Error(
+      `Failed to download ${key}: ${response.status} ${response.statusText}`
+    );
+  }
```
bobbor, line 245: nit (non-blocking): the non-OK response carries a body (S3 error XML) that is never consumed — throwing leaves it unread, which can delay connection reuse under HTTP/1.1 keep-alive. Cheap to drain before throwing:

```suggestion
  if (!response.ok) {
    // Release the connection — the error body (S3 error XML) is never consumed.
    response.body?.cancel().catch(() => {});
    throw new Error(
      `Failed to download ${key}: ${response.status} ${response.statusText}`
    );
  }
```

bobbor (dismissed):
Approving — the guard is correctly placed (fires before `zipWriter.add`, so no dangling entry and batch finalization accounting in `onFileSettled` is unaffected) and the error path never touches `batchAbort`, so sibling downloads continue. Approval is contingent on @osama-rizk's review feedback being addressed, which the follow-up commit does.

One non-blocking nit inline. We also looked at the all-files-fail scenario (empty zip still downloads) — that's pre-existing behavior in the settlement logic, out of scope here; will be tracked as a follow-up.

packages/react-storage/src/components/StorageBrowser/actions/handlers/zipdownload.ts
mwellmn: Tested locally. It's a good suggestion, applying the change.

osama-rizk (approved):

bobbor (approved):

